PR #244, opened 2020-07-04
Suggested changes in API review

JinmingHu-MSFT:


sdk/storage/inc/common/concurrent_transfer.hpp
```diff
@@ -0,0 +1,70 @@
...
+        int chunkId = nextChunkId.fetch_add(1);
+        if (chunkId >= numChunks || failed)
+        {
+          break;
+        }
+        int64_t chunkOffset = offset + chunkSize * chunkId;
+        int64_t chunkLength = std::min(length - chunkSize * chunkId, chunkSize);
+        try
+        {
+          transferFunc(chunkOffset, chunkLength, chunkId, numChunks);
```
katmsft, line 39: Put block will be one of the transferFunc, but the chunk ID cannot be passed out, meaning that the caller does not know the chunk ID being passed into the Put block operation if chunk ID is not a number starting from 0.
Adding a chunk ID prefix parameter to the `ConcurrentTransfer` and make sure chunkId is a string instead of an int will resolve this issue.

sdk/storage/inc/common/concurrent_transfer.hpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+    threadFunc();
+    for (auto& handle : threadHandles)
+    {
+      handle.get();
+    }
+
+    if (numWorkingThreads != 0)
+    {
+      std::abort();
```
katmsft, line 66: I believe even in a thread leak situation like this, we still should not call std::abort(). Please throw an exception instead.

sdk/storage/src/blobs/block_blob_client.cpp
```diff
@@ -75,6 +77,64 @@ namespace Azure { namespace Storage { namespace Blobs {
         options.Context, *m_pipeline, m_blobUrl.ToString(), content, protocolLayerOptions);
   }
 
+  BlobContentInfo BlockBlobClient::UploadFromBuffer(
+      const uint8_t* buffer,
+      std::size_t bufferSize,
+      const UploadBlobOptions& options) const
+  {
+    constexpr int64_t c_defaultBlockSize = 8 * 1024 * 1024;
```
katmsft, line 85: `c_defaultBlockSize ` should be a non-nullable optional parameter with a default value of 8MB instead.

sdk/storage/src/blobs/block_blob_client.cpp
```diff
@@ -75,6 +77,64 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    constexpr int64_t c_maximumNumberBlocks = 50000;
+    constexpr int64_t c_grainSize = 4 * 1024;
```
katmsft, line 87: `c_grainSize ` should also be optional parameter. If HTTP stack decides to read some other value than 4kb per reading, we can optimize with changing this value.
However, not a priority. Only need to doc this if you don't want to change it.

sdk/storage/inc/common/concurrent_transfer.hpp
katmsft: Hmm, saw your solution, ignore this comment then.

sdk/storage/inc/common/concurrent_transfer.hpp
JinmingHu-MSFT: It's expressly stated in our guidelines that, in this situation, we should crash.

https://azure.github.io/azure-sdk/cpp_design.html#post-conditions

sdk/storage/src/blobs/block_blob_client.cpp
```diff
@@ -75,6 +77,64 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+      int64_t minBlockSize = (bufferSize + c_maximumNumberBlocks - 1) / c_maximumNumberBlocks;
+      chunkSize = std::max(chunkSize, minBlockSize);
+      chunkSize = (chunkSize + c_grainSize - 1) / c_grainSize * c_grainSize;
+    }
+
+    std::vector<std::pair<BlockType, std::string>> blockIds;
+    auto getBlockId = [](int64_t id) {
+      constexpr std::size_t c_blockIdLength = 64;
+      std::string blockId = std::to_string(id);
+      blockId = std::string(c_blockIdLength - blockId.length(), '0') + blockId;
```
katmsft, line 105: Please reference track 1 or other languages for block ID format.

sdk/storage/src/blobs/block_blob_client.cpp
JinmingHu-MSFT: In `options` we already have a nullable `ChunkSize`, which will override this default value if provided.

What's the point of having them both in `options`?

sdk/storage/src/blobs/block_blob_client.cpp
JinmingHu-MSFT: I know they are different, but which is better than which?

sdk/storage/inc/common/concurrent_transfer.hpp
katmsft: This is not a post-condition according to the definition of post-condition:
> Post-Conditions
Post-Condition violations happen when some function didn’t do the correct thing, these are always bugs in the function itself, and users shouldn’t be expected to handle them.

I am not sure when this will happen. I would not even examine this value, to be honest.

sdk/storage/src/blobs/block_blob_client.cpp
JinmingHu-MSFT: This is not for HTTP stack optimization or something. It's designed for two purposes

1. For page blob, this value must be a multiple of 512.
2. For other kinds of blob, like block blob, I just feel it's kind of strange if the block size is not aligned to some power-of-2 boundaries. For example, uploading a 700KB blob with 6 threads, 120KB block size looks better than 119467 bytes.

JinmingHu-MSFT: And necessity of making it customizable by users?

sdk/storage/inc/common/concurrent_transfer.hpp
JinmingHu-MSFT: This is never supposed to happen. If it did happen, it means this function didn't do the correct thing, there was a bug in it. So by definition it's a post-condition check.

JinmingHu-MSFT: > I would not even examine this value, to be honest.

Yes, it's totally reasonable if you don't want to check this value.

But that doesn't mean post-condition check is unnecessary. Since our guidelines mention how we should do post-condition check, I think that implies that post-condition check is not prohibited. So I suppose it's okay to do it here.

sdk/storage/src/blobs/block_blob_client.cpp
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/246

sdk/storage/src/blobs/block_blob_client.cpp
katmsft: Then I don't think it's neccessary to have it here. We should have `ChunkSize` default to 8MB.

sdk/storage/src/blobs/block_blob_client.cpp
katmsft: Customer might chose their own HTTP stack, the grain size can be used to align with the HTTP stack's reading size per read for a better performance.

JinmingHu-MSFT: I didn't see how it's relevant to performance.

PR #247, opened 2020-07-07
Concurrent upload block blob from file, concurrent download blob to file

JinmingHu-MSFT:


sdk/storage/inc/common/file_io.hpp
```diff
@@ -0,0 +1,50 @@
...
+#else
+  using FileHandle = int;
+#endif
+
+  class FileReader {
+  public:
+    FileReader(const std::string& filename);
+
+    ~FileReader();
+
```
katmsft, line 26: FileReader does not provide `Read` is somewhat strange. However, I don't find a place to call this `Read`. It is your choice to or not to add the API.

JinmingHu-MSFT: okay

sdk/storage/src/blobs/blob_client.cpp
```diff
@@ -222,29 +223,161 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+        remainingOffset, remainingSize, chunkSize, options.Concurrency, downloadChunkFunc);
+    ret.ContentLength = blobRangeSize;
+    return ret;
+  }
+
+  BlobDownloadInfo BlobClient::DownloadToFile(
+      const std::string& file,
+      const DownloadBlobToFileOptions& options) const
+  {
+    constexpr int64_t c_defaultChunkSize = 8 * 1024 * 1024;
```
katmsft, line 274: Why 8MB?

JinmingHu-MSFT: I'm fine with any value, but we have to pick a value.

katmsft: I think it's normal to make it 4MB, any other SDK making it 8MB?
Also, I still think it should be the default value of the optional parameter in options, not here.

JinmingHu-MSFT: Okay, i'll change it to 4MB.

JinmingHu-MSFT: About default value or optional parameter, here's the deal. I want the customer to be able to fully control the block size if they want to, and if they don't, we pick the optimal value for them.

For example, if they're trying to download 4.2MiB blob or blob range with 2 threads. If they explicitly specify the block size is 4MiB, we're going to use 4MiB. If they don't, we'll use 2.1MiB to achieve the best overall performance.

JinmingHu-MSFT:
@damirault @danieljurek @Azure/azure-sdk-eng Can you please review [this](https://github.com/Azure/azure-sdk-for-cpp/pull/247/commits/22a5d2374c71008de35996e3a7b3f096f338526c) commit? I made some change to `eng/` to make CI pass.

JinmingHu-MSFT:
@RickWinter for your awareness

JinmingHu-MSFT:
> I saw this earlier today and have a similar fix in another PR for it. But this one is more likely to go in first so I can rebase. Thanks for making this change!
> 
> This is the right change to make for now but in the future it'd be nice if we didn't have to install dependencies from vcpkg just to generate doc artifacts. I've already got an issue tracking this here: #196

Thank you very much for your quick approval.

PR #252, opened 2020-07-08
Storage Blob Service Access Conditions, Bitwise enum class

JinmingHu-MSFT:

PR #257, opened 2020-07-09
Datalake full feature support PR.

katmsft:
The PR contains below changes:
1. Adapting the breaking changes introduced in new bodystream.
2. Add support for Nullable in Azure Storage DataLake.
3. Refine interface according to Jeffrey's review.
4. Adding E2E validation.
5. Resolve product code bugs found in E2E validation.


sdk/storage/inc/common/shared_request_options.hpp
```diff
@@ -14,7 +15,7 @@ namespace Azure { namespace Storage {
      *        when the request is received by the service. If the timeout value
      *        elapses before the operation completes, the operation fails.
      */
-    int32_t Timeout = 0; // TODO: This is going to be changed into Nullable.
+    Azure::Core::Nullable<int32_t> Timeout;
```
JeffreyRichter, line 18: Customers should be controlling timeout via the context (which will work for ALL client libraries); not via an explicit field/parameter.

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -217,65 +228,41 @@ namespace Azure { namespace Storage { namespace DataLake {
...
-    std::string IfModifiedSince;
+    Azure::Core::Nullable<std::string> IfModifiedSince;
 
     /**
      * @brief A date and time value. Specify this header to perform the
      *        operation only if the resource has not been modified since
      *        the specified date and time.
      */
-    std::string IfUnmodifiedSince;
+    Azure::Core::Nullable<std::string> IfUnmodifiedSince;
```
JeffreyRichter, line 265: This is different than how done for blobs since we added structures to blobs. Maybe we should be consistent across client libraries? Thoughts?

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -871,18 +854,18 @@ namespace Azure { namespace Storage { namespace DataLake {
...
-        int32_t Timeout
-            = int32_t(); // The timeout parameter is expressed in seconds. For more
-                         // information, see <a
-                         // href="https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/setting-timeouts-for-blob-service-operations">Setting
-                         // Timeouts for Blob Service Operations.</a>
+        Azure::Core::Nullable<std::string>
+            ClientRequestId; // Provides a client-generated, opaque value with a 1 KB character
+                             // limit that is recorded in the analytics logs when storage analytics
+                             // logging is enabled.
+        Azure::Core::Nullable<int32_t>
```
JeffreyRichter, line 861: Done via context

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -900,24 +883,41 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+            Timeout; // The timeout parameter is expressed in seconds. For more information, see <a
```
JeffreyRichter, line 914: context

sdk/storage/inc/common/shared_request_options.hpp
katmsft: This is a server-end timeout, which tells when the service should throw a timeout error. This is left as an optional parameter for other SDKs, is it a good approach to remove them?
However, I do agree that we can make the name to be a more obvious hint for a server-end timeout.

sdk/storage/inc/datalake/datalake_options.hpp
katmsft: Yes, waiting for that PR to be merged to abstract access conditions.

sdk/storage/inc/common/shared_request_options.hpp
JeffreyRichter: In the Go SDK I had thee retry policy add this query parameter automatically based on the next cancellation time remaining.

sdk/storage/test/test_base.cpp
```diff
@@ -92,22 +92,34 @@ namespace Azure { namespace Storage { namespace Test {
...
   }
 
+  std::map<std::string, std::string> RandomMetadata(size_t size)
+  {
+    std::map<std::string, std::string> result;
+    for (unsigned i = 0; i < size; ++i)
+    {
+      // TODO: Use mixed casing after Azure::Core lower cases the headers.
+      // Metadata keys cannot start with a number.
+      result.insert("m" + LowercaseRandomString(5), LowercaseRandomString(5));
```
JinmingHu-MSFT, line 118: Value should not be all-lowercase

sdk/storage/test/test_base.cpp
```diff
@@ -92,22 +92,34 @@ namespace Azure { namespace Storage { namespace Test {
...
+    auto str = RandomString(size);
     std::transform(
         str.begin(), str.end(), str.begin(), [](unsigned char c) { return char(std::tolower(c)); });
     return str;
   }
 
+  std::map<std::string, std::string> RandomMetadata(size_t size)
+  {
+    std::map<std::string, std::string> result;
+    for (unsigned i = 0; i < size; ++i)
```
JinmingHu-MSFT, line 114: `std::size_t`

sdk/storage/test/test_base.cpp
```diff
@@ -92,22 +92,34 @@ namespace Azure { namespace Storage { namespace Test {
...
+  std::string LowercaseRandomString(size_t size)
   {
-    auto str = RandomString();
+    auto str = RandomString(size);
     std::transform(
         str.begin(), str.end(), str.begin(), [](unsigned char c) { return char(std::tolower(c)); });
     return str;
   }
 
+  std::map<std::string, std::string> RandomMetadata(size_t size)
```
JinmingHu-MSFT, line 111: std::size_t

sdk/storage/test/test_base.cpp
katmsft: This is changed to call `RandomString` instead for metadata value.

sdk/storage/test/CMakeLists.txt
```diff
@@ -7,17 +7,25 @@ add_executable (
...
+     datalake/service_client_test.cpp
+     datalake/file_system_client_test.hpp
+     datalake/file_system_client_test.cpp
+     datalake/path_client_test.hpp
+     datalake/path_client_test.cpp
      main.cpp
-)
+    )
+
+target_include_directories(azure-storage-test PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>)
```
JinmingHu-MSFT, line 28: Why do we need `BUILD_INTERFACE` here?

sdk/storage/test/test_base.cpp
katmsft: `size_t` is more concise. We can bring this up in the next meeting.

sdk/storage/test/test_base.cpp
```diff
@@ -151,4 +163,12 @@ namespace Azure { namespace Storage { namespace Test {
 
   void DeleteFile(const std::string& filename) { std::remove(filename.data()); }
 
+  std::vector<uint8_t> RandomBuffer(std::size_t length)
+  {
+    std::vector<uint8_t> result(length);
+    uint8_t* dataPtr = const_cast<uint8_t*>(result.data());
```
JinmingHu-MSFT, line 169: `const_cast` is not something we want

katmsft: It's in test code and I don't want to spend all day reimplementing test utilities.

sdk/storage/test/test_base.cpp
```diff
@@ -151,4 +163,12 @@ namespace Azure { namespace Storage { namespace Test {
...
+    RandomBuffer(reinterpret_cast<char*>(dataPtr), length);
```
JinmingHu-MSFT, line 170:
```suggestion
    uint8_t* dataPtr = reinterpret_cast<char*>(&result[0]);
    RandomBuffer(dataPtr, length);
```

JinmingHu-MSFT: come on, it's not that complex.

sdk/storage/test/CMakeLists.txt
katmsft: Removed.

sdk/storage/inc/common/shared_request_options.hpp
katmsft: This indeed is more ideal. Is there a default value for this timeout in GO?

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
katmsft: This is not exposed to customers as this is protocol layer options.

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
katmsft: This is not exposed to customers as this is protocol layer options.

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -75,7 +79,7 @@ namespace Azure { namespace Storage { namespace DataLake {
      *        Note that the string may only contain ASCII characters in the
      *        ISO-8859-1 character set.
      */
-    std::map<std::string, std::string> Metadata;
+    Azure::Core::Nullable<std::map<std::string, std::string>> Metadata;
```
JinmingHu-MSFT, line 82: We have a decision that this isn't nullable

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -88,14 +92,14 @@ namespace Azure { namespace Storage { namespace DataLake {
      *        operation only if the resource has been modified since the
      (        specified date and time.
     */
-    std::string IfModifiedSince;
+    Azure::Core::Nullable<std::string> IfModifiedSince;
```
JinmingHu-MSFT, line 95: Will adapt to AccessConditions struct when that PR is merged.

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -105,6 +109,13 @@ namespace Azure { namespace Storage { namespace DataLake {
   {
   };
 
+  /**
+   * @brief Optional parameters for DataLakeFileSystemClient::GetProperties
+   */
+  struct FileSystemPropertiesOptions : public SharedRequestOptions
```
JinmingHu-MSFT, line 115: Maybe `Get` is left out?

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -173,12 +184,12 @@ namespace Azure { namespace Storage { namespace DataLake {
     /**
      * @brief Specify the transactional md5 for the body, to be validated by the service.
      */
-    std::string ContentMD5;
+    Azure::Core::Nullable<std::string> ContentMD5;
```
JinmingHu-MSFT, line 187: Can you double-check CRC64 is not supported?

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -99,6 +101,14 @@ namespace Azure { namespace Storage { namespace DataLake {
     constexpr static const char* c_HeaderXMsAcl = "x-ms-acl";
     constexpr static const char* c_HeaderXMsClientRequestId = "x-ms-client-request-id";
   } // namespace Details
+  struct DataLakeHttpHeaders
+  {
+    Azure::Core::Nullable<std::string> CacheControl;
+    Azure::Core::Nullable<std::string> ContentDisposition;
+    Azure::Core::Nullable<std::string> ContentEncoding;
+    Azure::Core::Nullable<std::string> ContentLanguage;
+    Azure::Core::Nullable<std::string> ContentType;
```
JinmingHu-MSFT, line 110: Should these be nullable?

sdk/storage/inc/datalake/datalake_utilities.hpp
```diff
@@ -12,9 +13,10 @@
 namespace Azure { namespace Storage { namespace DataLake { namespace Details {
   UrlBuilder GetBlobUriFromDfsUri(const UrlBuilder& dfsUri);
 
-  std::map<std::string, std::string> DeserializeMetadata(
-      const std::string& dataLakePropertiesString);
+  Azure::Core::Nullable<std::map<std::string, std::string>> DeserializeMetadata(
+      const Azure::Core::Nullable<std::string>& dataLakePropertiesString);
```
JinmingHu-MSFT, line 17:
```suggestion
  std::map<std::string, std::string> DeserializeMetadata(std::string& dataLakePropertiesString);
```

sdk/storage/inc/datalake/file_system_client.hpp
```diff
@@ -117,6 +109,22 @@ namespace Azure { namespace Storage { namespace DataLake {
         const std::map<std::string, std::string>& metadata,
         const FileSystemSetMetadataOptions& options = FileSystemSetMetadataOptions()) const;
 
+    /**
+     * @brief Gets the metadata of file system.
+     * @param options Optional parameters to get the metadata of this file system.
+     * @return std::map<std::string, std::string>
+     */
+    std::map<std::string, std::string> GetMetadata(
```
JinmingHu-MSFT, line 117: We don't need this API

JinmingHu-MSFT: `GetProperties` covers metadata.

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -32,54 +32,73 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+  {
+    std::string Date;
+    std::string RequestId;
+    std::string Version;
+    std::string ETag;
+    std::string LastModified;
+    DataLakeHttpHeaders HttpHeaders;
+    int64_t ContentLength = int64_t();
+    Azure::Core::Nullable<std::string> ContentRange;
+    std::string ContentMD5;
```
JinmingHu-MSFT, line 86: This is nullable in other structs. Can we make it consistent?

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -207,18 +226,71 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+     */
+    GetPathPropertiesResponse GetProperties(
+        const PathGetPropertiesOptions& options = PathGetPropertiesOptions()) const;
+
+    /**
+     * @brief Sets the metadata of a resource the path points to.
+     * @param options Optional parameters to set the metadata to the resource the path points to.
+     * @return SetMetadataResult
+     */
+    SetMetadataResult SetMetadata(
```
JinmingHu-MSFT, line 250: Sometimes this is `XxxResponse`, while sometimes this is `XxxResult`.

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -207,18 +226,71 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+        const std::map<std::string, std::string> metadata,
```
JinmingHu-MSFT, line 251: Either const ref or by value. const value doesn't make any sense

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -207,18 +226,71 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+        const SetPathMetadataOptions& options = SetPathMetadataOptions()) const;
+
+    /**
+     * @brief Get the metadata for a path.
+     * @param options Optional parameters to get the metadata from the resource the path points to.
+     * @return std::map<std::string, std::string>
+     */
+    std::map<std::string, std::string> GetMetadata(
```
JinmingHu-MSFT, line 259: We don't need this API

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -207,18 +226,71 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+
     /**
      * @brief Creates a file or directory. By default, the destination is overwritten and
      *        if the destination already exists and has a lease the lease is broken.
      * @param options Optional parameters to create the resource the path points to.
      * @return PathCreateResponse
      */
-    PathCreateResponse Create(const PathCreateOptions& options = PathCreateOptions()) const;
+    PathCreateResponse Create(
+        const PathResourceType& type,
```
JinmingHu-MSFT, line 269: is this an enum? If so, we can pass it by value.

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -207,18 +226,71 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+        const PathCreateOptions& options = PathCreateOptions()) const;
+
+    /**
+     * @brief Creates as a file. By default, the destination is overwritten and
+     *        if the destination already exists and has a lease the lease is broken.
+     * @param options Optional parameters to create the resource the path points to.
+     * @return PathCreateResponse
+     */
+    PathCreateResponse CreateAsFile(const PathCreateOptions& options = PathCreateOptions()) const
```
JinmingHu-MSFT, line 278: How do they name these two API in other language sdk?

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -639,77 +644,69 @@ namespace Azure { namespace Storage { namespace DataLake {
...
-    std::string Properties;
-    std::string LeaseDuration;
-    std::string LeaseState;
-    std::string LeaseStatus;
-    std::string XMsContentMd5;
+    Azure::Core::Nullable<std::string> Properties;
+    Azure::Core::Nullable<std::string> LeaseDuration;
+    Azure::Core::Nullable<std::string> LeaseState;
+    Azure::Core::Nullable<std::string> LeaseStatus;
+    Azure::Core::Nullable<std::string> XMsContentMd5;
```
JinmingHu-MSFT, line 661: This isn't a good name

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -639,77 +644,69 @@ namespace Azure { namespace Storage { namespace DataLake {
...
-    std::string LeaseStatus;
+    Azure::Core::Nullable<std::string> ResourceType;
+    Azure::Core::Nullable<std::string> Properties;
+    Azure::Core::Nullable<std::string> Owner;
+    Azure::Core::Nullable<std::string> Group;
+    Azure::Core::Nullable<std::string> Permissions;
+    Azure::Core::Nullable<std::string> ACL;
+    Azure::Core::Nullable<std::string> LeaseDuration;
+    Azure::Core::Nullable<std::string> LeaseState;
+    Azure::Core::Nullable<std::string> LeaseStatus;
```
JinmingHu-MSFT, line 684: Both status and state should be enum

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -8,18 +8,20 @@
 #include "http/http.hpp"
```
JinmingHu-MSFT, line 8: For the record, i didn't review this file

sdk/storage/src/datalake/file_system_client.cpp
```diff
@@ -126,21 +138,35 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+        m_dfsUri.ToString(), *m_pipeline, options.Context, protocolLayerOptions);
+    auto metadata = Details::DeserializeMetadata(result.Properties);
+    return FileSystemProperties{
         std::move(result.Date),
         std::move(result.ETag),
         std::move(result.LastModified),
         std::move(result.RequestId),
         std::move(result.Version),
-        Details::DeserializeMetadata(result.Properties),
+        metadata.HasValue() ? std::move(metadata.GetValue()) : std::map<std::string, std::string>(),
```
JinmingHu-MSFT, line 169: Why not make metadata always non-nullable?

sdk/storage/src/datalake/path_client.cpp
```diff
@@ -33,10 +33,11 @@ namespace Azure { namespace Storage { namespace DataLake {
       return std::string(begin, end);
     }
 
-    std::pair<int64_t, int64_t> GetOffsetLength(const std::string& rangeString)
+    std::pair<int64_t, Azure::Core::Nullable<int64_t>> GetOffsetLength(
```
JinmingHu-MSFT, line 36: When will this pair.second be null?

sdk/storage/src/datalake/path_client.cpp
```diff
@@ -251,37 +267,61 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+    returnVal.ETag = std::move(result.ETag);
+    returnVal.LastModified = std::move(result.LastModified);
+    returnVal.HttpHeaders = std::move(result.HttpHeaders);
+    returnVal.ContentLength = result.ContentLength;
+    returnVal.ContentRange = std::move(result.ContentRange);
+    returnVal.ContentMD5 = std::move(result.ContentMD5);
+    auto metadata = Details::DeserializeMetadata(result.Properties);
+    if (metadata.HasValue())
+    {
+      returnVal.Metadata = metadata.GetValue();
```
JinmingHu-MSFT, line 302: Why not make metadata in protocol layer non-nullable?

sdk/storage/test/datalake/file_system_client_test.hpp
```diff
@@ -0,0 +1,30 @@
...
+        bool recursive,
+        const std::string& directory = std::string());
+
+    static DataLake::DataLakeHttpHeaders GetInterestingHttpHeaders();
+
+    static std::shared_ptr<DataLake::FileSystemClient> m_fileSystemClient;
+    static std::string m_fileSystemName;
+
+    // Path related
+    static std::vector<std::string> m_PathNameSetA;
```
JinmingHu-MSFT, line 24: `m_PascalCase` or `m_camelCase` for member variables?

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+  std::string FileSystemClientTest::m_fileSystemName;
+
+  std::vector<std::string> FileSystemClientTest::m_PathNameSetA;
+  std::vector<std::string> FileSystemClientTest::m_PathNameSetB;
+  std::string FileSystemClientTest::m_DirectoryA;
+  std::string FileSystemClientTest::m_DirectoryB;
+
+  void FileSystemClientTest::SetUpTestSuite()
+  {
+    m_fileSystemName = LowercaseRandomString(10);
```
JinmingHu-MSFT, line 22: The default value is `10`, so what's the pointe of specifying it explicitly?

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+    m_fileSystemClient = std::make_shared<DataLake::FileSystemClient>(
+        DataLake::FileSystemClient::CreateFromConnectionString(
+            ADLSGen2ConnectionString(), m_fileSystemName));
+    m_fileSystemClient->Create();
+
+    m_DirectoryA = LowercaseRandomString(10);
+    m_DirectoryB = LowercaseRandomString(10);
+    for (unsigned i = 0; i < c_PATH_TEST_SIZE; ++i)
```
JinmingHu-MSFT, line 30: `size_t` or `std::size_t`? because `c_PATH_TEST_SIZE` is `size_t`

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+        options.Continuation = continuation;
+      }
+    } while (!continuation.empty());
+    return result;
+  }
+
+  DataLake::DataLakeHttpHeaders FileSystemClientTest::GetInterestingHttpHeaders()
+  {
+    static DataLake::DataLakeHttpHeaders result{
+        std::string("no-cache"),
```
JinmingHu-MSFT, line 74: aggregate initialization is forbidden.

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+        std::string("attachment"),
+        std::string("deflate"),
+        std::string("en-US"),
+        std::string("text/html; charset=UTF-8")};
```
JinmingHu-MSFT, line 78: Is is for `Content-Type`? I remember server side will return error if you upload a binary file with this content type.

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+    return result;
+  }
+
+  TEST_F(FileSystemClientTest, CreateDeleteFileSystems)
+  {
+    {
+      // Normal create/delete.
+      std::vector<DataLake::FileSystemClient> fileSystemClient;
+      for (int32_t i = 0; i < 5; ++i)
```
JinmingHu-MSFT, line 87: why not just `int`?

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+      }
+      for (const auto& client : fileSystemClient)
+      {
+        EXPECT_NO_THROW(client.Delete());
+      }
+    }
+    {
+      // Normal delete with access condition.
+      std::vector<DataLake::FileSystemClient> fileSystemClient;
+      for (int32_t i = 0; i < 5; ++i)
```
JinmingHu-MSFT, line 102: We never say `int` is forbidden everywhere. This `i` veriable is not gonna be sent out over the wire.

JinmingHu-MSFT:
https://github.com/Azure/azure-sdk-for-cpp/pull/252 has been merged, please rebease on the latest master branch, resolve conflicts.

sdk/storage/test/datalake/service_client_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+  std::string DataLakeServiceClientTest::m_fileSystemPrefixA;
+  std::string DataLakeServiceClientTest::m_fileSystemPrefixB;
+
+  void DataLakeServiceClientTest::SetUpTestSuite()
+  {
+    m_dataLakeServiceClient = std::make_shared<DataLake::ServiceClient>(
+        DataLake::ServiceClient::CreateFromConnectionString(ADLSGen2ConnectionString()));
+    m_fileSystemPrefixA = LowercaseRandomString(10);
+    m_fileSystemPrefixB = LowercaseRandomString(10);
+    for (unsigned i = 0; i < c_FILE_SYSTEM_TEST_SIZE; ++i)
```
JinmingHu-MSFT, line 24: ditto

sdk/storage/inc/datalake/path_client.hpp
katmsft: typo

sdk/storage/inc/datalake/datalake_options.hpp
katmsft: https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/update
Currently not supported.

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
katmsft: Changed to non-nullable.

sdk/storage/inc/datalake/file_system_client.hpp
katmsft: However, `GetProperties` returns more than Metadata, that customer might not need. I don't see the harm in adding this API.

sdk/storage/inc/datalake/path_client.hpp
katmsft: Changed all return value to either be a `XxxResponse` or something like `FileSystemProperties`.

sdk/storage/inc/datalake/path_client.hpp
katmsft: It is just much nicer to have a `GetMetadata` returning directly the `metadata`. Calling `GetProperties` is not intuitive enough for retrieving metadata.

sdk/storage/inc/datalake/path_client.hpp
katmsft: Nice catch.

sdk/storage/inc/datalake/path_client.hpp
katmsft: `CreateFile` or `CreateDirectory`. However other languages don't have the issue of name clashing with what is defined in `fileapi.h` :(

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
katmsft: Refined this to `ContentMd5` and the other to `TransactionalMd5`.

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
katmsft: This is another bug in current DataLake swagger that's not consistent with blob's swagger, sigh.
Modified the swagger to make this work.

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
katmsft: For the record, you need to review this file T_T.

sdk/storage/src/datalake/file_system_client.cpp
katmsft: Done.

sdk/storage/src/datalake/path_client.cpp
katmsft: When the range string is in a format of `"bytes=1024-"`.

sdk/storage/src/datalake/path_client.cpp
katmsft: Did that.

sdk/storage/test/datalake/file_system_client_test.hpp
katmsft: Typo.

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: Removed. However, the caller should not be aware of the default value. If someone changes the default value to other values, the 10 will work.
It is not a big deal in test code.

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: `size_t`, more concise.

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: Because there is no guideline asking to write `int` here.

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: True but I don't think int32_t is forbidden.

katmsft:
@JinmingHu-MSFT Had to force push for the updated master branch. Only the last commit has actual code change since your review.

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: Changed to `application/octet-stream`.

sdk/storage/test/datalake/file_system_client_test.cpp
JinmingHu-MSFT: Well, I do. I cannot figure out which variable you're trying to assign the value to.

sdk/storage/src/datalake/path_client.cpp
JinmingHu-MSFT: Can you name one API that would return this in response?

sdk/storage/inc/datalake/path_client.hpp
JinmingHu-MSFT: Does compiler complain about clashes?

Things in `fileapi.h` are C API. We're writing C++ functions, inside namespace. I think that's why namespace was invented.

sdk/storage/inc/datalake/file_system_client.hpp
JinmingHu-MSFT: No other SDK is doing so. Plus I do not think we can add API at will

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
JinmingHu-MSFT: That's 3300 lines of code change.

sdk/storage/test/datalake/file_system_client_test.cpp
JinmingHu-MSFT: whatever

sdk/storage/test/datalake/file_system_client_test.cpp
JinmingHu-MSFT: whatever

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -115,14 +126,14 @@ namespace Azure { namespace Storage { namespace DataLake {
      *        operation only if the resource has been modified since the
      (        specified date and time.
     */
-    std::string IfModifiedSince;
+    Azure::Core::Nullable<std::string> IfModifiedSince;
```
JinmingHu-MSFT, line 129: https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/storage/inc/common/access_conditions.hpp

We have `LastModifiedTimeAccessConditions` and `ETagAccessConditions`

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -286,17 +273,17 @@ namespace Azure { namespace Storage { namespace DataLake {
     /**
      * @brief The lease ID must be specified if there is an active lease.
      */
-    std::string LeaseId;
+    Azure::Core::Nullable<std::string> LeaseId;
```
JinmingHu-MSFT, line 276: https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/storage/inc/common/access_conditions.hpp

We have `LeaseAccessConditions`

sdk/storage/inc/datalake/datalake_utilities.hpp
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "common/storage_url_builder.hpp"
+#include "nullable.hpp"
```
JinmingHu-MSFT, line 7: Why do we need this header?

sdk/storage/inc/datalake/datalake_utilities.hpp
```diff
@@ -4,6 +4,7 @@
...
 
 #include <map>
 #include <string>
```
JinmingHu-MSFT, line 10: We didn't use `std::vector` in this file.

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
JinmingHu-MSFT: Chrome is even slowing down and lagging when reviewing this file on GitHub.

JinmingHu-MSFT: I skimmed it. I think it's fine.

sdk/storage/inc/datalake/file_system_client.hpp
katmsft: I can delete this if we are rejected in Jeffrey's review.

JeffreyRichter: Yes, GetProperties should be good enough

sdk/storage/inc/datalake/path_client.hpp
katmsft: Well, you are actually right. However, VS is having some issues doing the IntelliSense when I changed the name to CreateFile/CreateDirectory, tricking me into believing the name to introduce an issue. Changed in a newer iteration.

sdk/storage/src/datalake/path_client.cpp
katmsft: I think you are correct, this method is used for parsing response so we shouldn't be considering the request's format.
Changed in latest commit.

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: Then can you please provide an easy solution for this code that inline initializing the result. This way for each call a copy is made and we don't need to re-initialize the strings.

JinmingHu-MSFT: make it a member variable?

sdk/storage/inc/datalake/datalake_options.hpp
katmsft: Integrated access condition.

sdk/storage/inc/datalake/datalake_options.hpp
katmsft: Integrated access condition.

katmsft:
@JinmingHu-MSFT Reset your vote and re-request review. Thanks!

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: Not ideal since the only place to initialize it is `SetUpTestSuite()` and it is called for each individual test.

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -32,56 +32,75 @@ namespace Azure { namespace Storage { namespace DataLake {
...
     std::string RequestId;
     std::string Version;
     std::string ResourceType;
-    std::string LeaseDuration;
-    std::string LeaseState;
-    std::string LeaseStatus;
-    std::string ContentMd5;
+    Azure::Core::Nullable<std::string> LeaseDuration;
+    Azure::Core::Nullable<LeaseStateType> LeaseState;
+    Azure::Core::Nullable<LeaseStatusType> LeaseStatus;
```
JinmingHu-MSFT, line 48: Should these two be nullable?

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -18,23 +20,42 @@ namespace Azure { namespace Storage { namespace DataLake {
...
+   * @brief Specifies access conditions for a file system.
+   */
+  struct FileSystemAccessConditions : public LastModifiedTimeAccessConditions
+  {
+  };
+
+  /**
+   * @brief Specifies access conditions for a path.
+   */
+  struct PathAccessConditions : public LastModifiedTimeAccessConditions,
```
JinmingHu-MSFT, line 55: Are you sure all path operations support these 5 access conditions?

sdk/storage/inc/datalake/datalake_options.hpp
```diff
@@ -18,23 +20,42 @@ namespace Azure { namespace Storage { namespace DataLake {
...
   {
-    std::vector<std::unique_ptr<Azure::Core::Http::HttpPolicy>> policies;
+    std::vector<std::unique_ptr<Azure::Core::Http::HttpPolicy>> PerOperationPolicies;
+    std::vector<std::unique_ptr<Azure::Core::Http::HttpPolicy>> PerRetryPolicies;
+  };
+
+  /**
+   * @brief Specifies access conditions for a file system.
+   */
+  struct FileSystemAccessConditions : public LastModifiedTimeAccessConditions
```
JinmingHu-MSFT, line 48: Are you sure all filesystem operations support these 2 access conditions?

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -32,56 +32,75 @@ namespace Azure { namespace Storage { namespace DataLake {
...
-    std::string LeaseDuration;
-    std::string LeaseState;
-    std::string LeaseStatus;
+    Azure::Core::Nullable<std::string> ResourceType;
+    Azure::Core::Nullable<std::string> Owner;
+    Azure::Core::Nullable<std::string> Group;
+    Azure::Core::Nullable<std::string> Permissions;
+    Azure::Core::Nullable<std::vector<Acl>> Acls;
+    Azure::Core::Nullable<std::string> LeaseDuration;
+    Azure::Core::Nullable<LeaseStateType> LeaseState;
```
JinmingHu-MSFT, line 71: Ditto

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -207,18 +226,70 @@ namespace Azure { namespace Storage { namespace DataLake {
     /**
      * @brief Sets the properties of a resource the path points to.
      * @param options Optional parameters to set the properties to the resource the path points to.
-     * @return PathUpdateResponse
+     * @return SetPathPropertiesResponse
      */
-    PathUpdateResponse SetProperties(
+    SetPathPropertiesResponse SetProperties(
```
JinmingHu-MSFT, line 231: I remember we've changed this to `SetHttpHeaders`.

sdk/storage/inc/datalake/path_client.hpp
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/pull/257#discussion_r452608647

sdk/storage/src/datalake/path_client.cpp
```diff
@@ -47,6 +46,7 @@ namespace Azure { namespace Storage { namespace DataLake {
         {
           length = std::stoll(GetSubstringTillDelimiter('\n', subRangeString, cur)) - offset + 1;
         }
+        // else raise exception?
```
JinmingHu-MSFT, line 49: Yes, please throw an exception if parse failed

sdk/storage/test/datalake/file_system_client_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  DataLake::DataLakeHttpHeaders FileSystemClientTest::GetInterestingHttpHeaders()
+  {
+    static DataLake::DataLakeHttpHeaders result{
+        std::string("no-cache"),
+        std::string("attachment"),
+        std::string("deflate"),
+        std::string("en-US"),
+        std::string("application/octet-stream")};
```
JinmingHu-MSFT, line 78:
```suggestion
    static DataLake::DataLakeHttpHeaders result = [](){
        DataLake::DataLakeHttpHeaders ret;
        ret.Foo = "no-cache";
        ret.Bar = "attachment";
        return ret;
    }();

```

sdk/storage/inc/datalake/datalake_options.hpp
katmsft: Yes.

katmsft: Let me clarify this. All path operation that supports `PathAccessConditions` supports these 5 access conditions.

sdk/storage/inc/datalake/datalake_options.hpp
katmsft: Yes, all filesystem operation that supports access condition supports exactly those two access conditions.

sdk/storage/inc/datalake/datalake_options.hpp
JinmingHu-MSFT: okay then

sdk/storage/inc/datalake/path_client.hpp
katmsft: Yes, if there is no lease, those won't be returned.

sdk/storage/inc/datalake/path_client.hpp
katmsft: See the other comment.

sdk/storage/inc/datalake/path_client.hpp
katmsft: This API can also set metadata.

sdk/storage/inc/datalake/path_client.hpp
JinmingHu-MSFT: I doubt that. can you give me an example of HTTP response?

sdk/storage/test/datalake/file_system_client_test.cpp
katmsft: Can you elaborate on why this resolves the issue? each call to result() will still create a map.

sdk/storage/src/datalake/path_client.cpp
katmsft: I'll throw an exception on this.

sdk/storage/inc/datalake/path_client.hpp
JinmingHu-MSFT: Please keep aligned with other SDK

sdk/storage/test/datalake/file_system_client_test.cpp
JinmingHu-MSFT: `result` is a local variable, not a function. It cannot be **called**

katmsft: My bad. Did not see the last `()`.

sdk/storage/inc/datalake/path_client.hpp
katmsft: You are actually right, only the lease duration is optional, changed in next commit.

sdk/storage/inc/datalake/path_client.hpp
katmsft: Sure, changed to SetHttpHeaders and removed the functionality to set metadata.

sdk/storage/inc/datalake/path_client.hpp
```diff
@@ -206,19 +224,72 @@ namespace Azure { namespace Storage { namespace DataLake {
...
     /**
      * @brief Sets the properties of a resource the path points to.
-     * @param options Optional parameters to set the properties to the resource the path points to.
-     * @return PathUpdateResponse
+     * @param options Optional parameters to set the http headers to the resource the path points
+     * to.
+     * @return SetPathHttpHeadersResponse
+     */
+    SetPathHttpHeadersResponse SetHttpHeaders(
+        const SetPathHttpHeadersOptions& options = SetPathHttpHeadersOptions()) const;
```
JinmingHu-MSFT, line 232: This function has a mandatory option.

sdk/storage/inc/datalake/protocol/datalake_rest_client.hpp
```diff
@@ -238,14 +245,14 @@ namespace Azure { namespace Storage { namespace DataLake {
   {
     std::string Name;
     std::string LastModified;
-    std::string ETag;
+    std::string Etag;
```
JinmingHu-MSFT, line 248: Why changed?

katmsft: When parsing the response body, server end tag is `etag` not `ETag`.
Also, server is returning `lastModified` so it's camel case and making `etag` one word.

Issue #253, opened 2020-07-08
Telemetry Policy


RickWinter:
Add telemetry policy.

C++ is identified with "cpp" in telemetry.
https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

PR #264, opened 2020-07-11
Telemetry policy

antkmsft:
Closes https://github.com/Azure/azure-sdk-for-cpp/issues/253

### DIFF
--- a/sdk/core/azure-core/inc/credentials/policy/policies.hpp
+++ b/sdk/core/azure-core/inc/credentials/policy/policies.hpp
@@ -24,40 +24,33 @@ namespace Azure { namespace Core { namespace Credentials { namespace Policy {
     void operator=(BearerTokenAuthenticationPolicy const&) = delete;
 
   public:
-    BearerTokenAuthenticationPolicy(
-        std::shared_ptr<TokenCredential const> const& credential,
-        std::string const& scope)
-        : m_credential(credential)
+    explicit BearerTokenAuthenticationPolicy(
+        std::shared_ptr<TokenCredential const> credential,
+        std::string scope)
+        : m_credential(std::move(credential))
     {
-      m_scopes.push_back(scope);
+      m_scopes.emplace_back(std::move(scope));
     }
 
-    BearerTokenAuthenticationPolicy(
-        std::shared_ptr<TokenCredential const> const& credential,
-        std::vector<std::string> const& scopes)
-        : m_credential(credential), m_scopes(scopes)
-    {
-    }
-
-    BearerTokenAuthenticationPolicy(
-        std::shared_ptr<TokenCredential const> const& credential,
-        std::vector<std::string> const&& scopes)
-        : m_credential(credential), m_scopes(std::move(scopes))
+    explicit BearerTokenAuthenticationPolicy(
+        std::shared_ptr<TokenCredential const> credential,
+        std::vector<std::string> scopes)
+        : m_credential(std::move(credential)), m_scopes(std::move(scopes))
     {
     }
 
     template <typename ScopesIterator>
-    BearerTokenAuthenticationPolicy(
-        std::shared_ptr<TokenCredential const> const& credential,
+    explicit BearerTokenAuthenticationPolicy(
+        std::shared_ptr<TokenCredential const> credential,
         ScopesIterator const& scopesBegin,
         ScopesIterator const& scopesEnd)
-        : m_credential(credential), m_scopes(scopesBegin, scopesEnd)
+        : m_credential(std::move(credential)), m_scopes(scopesBegin, scopesEnd)
     {
     }
 
-    HttpPolicy* Clone() const override
+    std::unique_ptr<HttpPolicy> Clone() const override
     {
-      return new BearerTokenAuthenticationPolicy(m_credential, m_scopes);
+      return std::make_unique<BearerTokenAuthenticationPolicy>(m_credential, m_scopes);
     }
 
     std::unique_ptr<Http::RawResponse> Send(
--- a/sdk/core/azure-core/inc/http/policy.hpp
+++ b/sdk/core/azure-core/inc/http/policy.hpp
@@ -145,7 +145,7 @@ namespace Azure { namespace Core { namespace Http {
         std::string const& componentName,
         std::string const& componentVersion,
         std::string const& applicationId)
-        : m_telemetryId(std::move(BuildTelemetryId(componentName, componentVersion, applicationId)))
+        : m_telemetryId(BuildTelemetryId(componentName, componentVersion, applicationId))
     {
     }
 
--- a/sdk/core/azure-core/inc/http/policy.hpp
+++ b/sdk/core/azure-core/inc/http/policy.hpp
@@ -154,7 +154,7 @@ namespace Azure { namespace Core { namespace Http {
       return std::make_unique<TelemetryPolicy>(*this);
     }
 
-    std::unique_ptr<Response> Send(Context& ctx, Request& request, NextHttpPolicy nextHttpPolicy)
+    std::unique_ptr<RawResponse> Send(Context& ctx, Request& request, NextHttpPolicy nextHttpPolicy)
         const override;
   };
 
--- a/sdk/core/azure-core/inc/http/policy.hpp
+++ b/sdk/core/azure-core/inc/http/policy.hpp
@@ -25,7 +25,7 @@ namespace Azure { namespace Core { namespace Http {
         Request& request,
         NextHttpPolicy policy) const = 0;
     virtual ~HttpPolicy() {}
-    virtual HttpPolicy* Clone() const = 0;
+    virtual std::unique_ptr<HttpPolicy> Clone() const = 0;
 
   protected:
     HttpPolicy() = default;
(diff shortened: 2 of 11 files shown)


sdk/core/azure-core/inc/http/policy.hpp
```diff
@@ -117,4 +125,37 @@ namespace Azure { namespace Core { namespace Http {
...
+        std::string const& componentVersion,
+        std::string const& applicationId);
+
+  public:
+    explicit TelemetryPolicy(std::string const& componentName, std::string const& componentVersion)
+        : TelemetryPolicy(componentName, componentVersion, g_emptyApplicationId)
+    {
+    }
+
+    explicit TelemetryPolicy(
```
JeffreyRichter, line 144: Can you make this work more like this: https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/zc_policy_telemetry.go

Specifically, the ctor should take 1 options parameter which has 1 field which shoul dbe call application_id to match this guidelines: https://azure.github.io/azure-sdk/general_azurecore.html (the go code calls this value but the go code should change to application_id)

Then, the rest of the string should be created by the policy itself so that all callers don't have to redo the same work all the time and this also ensures that the string is correct and consistent when sent to the the service.

sdk/storage/inc/common/common_headers_request_policy.hpp
```diff
@@ -12,7 +12,10 @@ namespace Azure { namespace Storage {
     explicit CommonHeadersRequestPolicy() {}
     ~CommonHeadersRequestPolicy() override {}
 
-    HttpPolicy* Clone() const override { return new CommonHeadersRequestPolicy(*this); }
+    std::unique_ptr<HttpPolicy> Clone() const override
```
JeffreyRichter, line 15: What does this CommonHeadersRequestPolicy class even do? 
There is no comment explaining it. Why do we need this class?

JinmingHu-MSFT: It's used to add headers that apply to every HTTP request. But for the moment, I guess there's no other headers besides `x-ms-date`. So I guess we should rename it to `DatePolicy`?

sdk/core/azure-core/inc/credentials/policy/policies.hpp
```diff
@@ -24,40 +24,33 @@ namespace Azure { namespace Core { namespace Credentials { namespace Policy {
     void operator=(BearerTokenAuthenticationPolicy const&) = delete;
 
   public:
-    BearerTokenAuthenticationPolicy(
-        std::shared_ptr<TokenCredential const> const& credential,
-        std::string const& scope)
-        : m_credential(credential)
+    explicit BearerTokenAuthenticationPolicy(
```
RickWinter, line 27: Nit, This looks unrelated to the telemetry policy feature.  Easier to track history when commits are broken into unique bug/feature commits.

sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -0,0 +1,41 @@
...
+    std::string const& componentName,
+    std::string const& componentVersion,
+    std::string const& applicationId)
+{
+  // Spec: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy
+  std::ostringstream telemetryId;
+
+  if (!applicationId.empty())
+  {
+    telemetryId << applicationId.substr(0, 24) << " ";
```
RickWinter, line 22: The applicationId might have whitespace at the end.  Trim the whitespace.

antkmsft: Resolved.

sdk/core/azure-core/inc/http/policy.hpp
antkmsft: Per our standup conversation, I am not doing this. The behavior is copied from .NET SDK, and the Go factory you mentioned does create the policy for a specific module, this can be done in C++, but should not be done at this (core) level.

sdk/core/azure-core/inc/credentials/policy/policies.hpp
antkmsft: Yes, it ended up being more than I realized. In the future, I'll be splitting into smaller PRs.

sdk/storage/inc/common/common_headers_request_policy.hpp
antkmsft: Not doing anything in this PR, as there is a clarification, and it is not clear whether we will need to add new headers, in which case we may need to do the rename again, please reactivate this comment if you think we should rename it now.

sdk/core/azure-core/inc/credentials/policy/policies.hpp
antkmsft: Resolving.

sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <http/policy.hpp>
+
+#include <sstream>
```
RickWinter, line 6: This OSVersion helper should be in utils.hpp and utils.cpp within Azure::Core::Utilities namespace

sdk/storage/inc/common/shared_key_policy.hpp
```diff
@@ -17,7 +17,10 @@ namespace Azure { namespace Storage {
 
     ~SharedKeyPolicy() override {}
 
-    HttpPolicy* Clone() const override { return new SharedKeyPolicy(m_credential); }
+    std::unique_ptr<HttpPolicy> Clone() const override
```
RickWinter, line 20: Ensure the storage team reviews and signs off on this change.

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: I disagree - I would keep it here until there's another place that needs to use this function.

antkmsft: And there's currently no utils.cpp / hpp. That would need to be an internal header, because we don't want to ship it as public API. What would be the other place that uses it? I don't think that we should move this function there at this moment.

sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -0,0 +1,143 @@
...
+            {
+              osVersionInfo << (first ? "" : " ")
+                            << std::string(valueBuf, valueBuf + (valueBufSize - 1));
+              first = false;
+            }
+          }
+        }
+      }
+
+      RegCloseKey(regKey);
```
RickWinter, line 56: We only need to close the key if we successfully opened the key.

antkmsft: That's what we do - it is inside the `if(OpenKey)`

sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -0,0 +1,143 @@
...
+              "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion",
+              0,
+              KEY_READ,
+              &regKey)
+          == ERROR_SUCCESS)
+      {
+        regKeyOpened = true;
+
+        auto first = true;
+        char const* regValues[]{
```
RickWinter, line 36: The array of values we look for should be "static constexpr"

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: Oh, it wasn't! I'll fix.

sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -0,0 +1,143 @@
...
+  return osVersionInfo.str();
+}
+
+} // namespace
+
+#else
+
+#include <sys/utsname.h>
+
+namespace {
```
RickWinter, line 78: If GetOSVersion doesn't move to Utilities namespace then it needs to be in Azure::Core::Http::Details namespace

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: Done.

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: it is in unnamed namespace. Detail namespace is only for headers.

antkmsft: Because you can't (shouldn't) do unnamed namespaces in headers - they don't have the intended effect, and at the same time you need to make things available in the headers, but to your code. If the declaration is in the unnamed namespace in .cpp file, it is perfectly accessible the the code in the .cpp file, and only to that code, and the symbolas are not exported by linker (i.e. it is similar to `static` functions in C)

antkmsft:
In teams conversation @RickWinter agreed with the current changes, but only asked me to add tests. I added them. I am pinging him again to unblock the PR.

Issue #243, opened 2020-07-04
[Storage Blobs Service] ListBlobsByHierarchy API


JinmingHu-MSFT:

JinmingHu-MSFT:
done in https://github.com/Azure/azure-sdk-for-cpp/pull/265

PR #265, opened 2020-07-13
[Storage Blobs Service] ListBlobsByHierarchy

JinmingHu-MSFT:
https://github.com/Azure/azure-sdk-for-cpp/issues/243

### DIFF
--- a/sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
+++ b/sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
@@ -1,4 +1,3 @@
-
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
@@ -1178,7 +1177,7 @@ namespace Azure { namespace Storage { namespace Blobs {
           response = ListContainersSegmentFromXml(reader);
         }
         response.Version = httpResponse.GetHeaders().at("x-ms-version");
-        response.Date = httpResponse.GetHeaders().at("Date");
+        response.Date = httpResponse.GetHeaders().at("date");
         response.RequestId = httpResponse.GetHeaders().at("x-ms-request-id");
         auto response_client_request_id_iterator
             = httpResponse.GetHeaders().find("x-ms-client-request-id");
@@ -1270,7 +1269,7 @@ namespace Azure { namespace Storage { namespace Blobs {
           response = UserDelegationKeyFromXml(reader);
         }
         response.Version = httpResponse.GetHeaders().at("x-ms-version");
-        response.Date = httpResponse.GetHeaders().at("Date");
+        response.Date = httpResponse.GetHeaders().at("date");
         response.RequestId = httpResponse.GetHeaders().at("x-ms-request-id");
         auto response_client_request_id_iterator
             = httpResponse.GetHeaders().find("x-ms-client-request-id");
@@ -1780,7 +1779,7 @@ namespace Azure { namespace Storage { namespace Blobs {
           throw StorageError::CreateFromResponse(std::move(pHttpResponse));
         }
         response.Version = httpResponse.GetHeaders().at("x-ms-version");
-        response.Date = httpResponse.GetHeaders().at("Date");
+        response.Date = httpResponse.GetHeaders().at("date");
         response.RequestId = httpResponse.GetHeaders().at("x-ms-request-id");
         auto response_client_request_id_iterator
             = httpResponse.GetHeaders().find("x-ms-client-request-id");
@@ -1788,8 +1787,8 @@ namespace Azure { namespace Storage { namespace Blobs {
         {
           response.ClientRequestId = response_client_request_id_iterator->second;
         }
-        response.ETag = httpResponse.GetHeaders().at("ETag");
-        response.LastModified = httpResponse.GetHeaders().at("Last-Modified");
+        response.ETag = httpResponse.GetHeaders().at("etag");
+        response.LastModified = httpResponse.GetHeaders().at("last-modified");
         return response;
       }
 
@@ -1858,7 +1857,7 @@ namespace Azure { namespace Storage { namespace Blobs {
           throw StorageError::CreateFromResponse(std::move(pHttpResponse));
         }
         response.Version = httpResponse.GetHeaders().at("x-ms-version");
-        response.Date = httpResponse.GetHeaders().at("Date");
+        response.Date = httpResponse.GetHeaders().at("date");
         response.RequestId = httpResponse.GetHeaders().at("x-ms-request-id");
         auto response_client_request_id_iterator
             = httpResponse.GetHeaders().find("x-ms-client-request-id");
@@ -1939,7 +1938,7 @@ namespace Azure { namespace Storage { namespace Blobs {
           throw StorageError::CreateFromResponse(std::move(pHttpResponse));
         }
         response.Version = httpResponse.GetHeaders().at("x-ms-version");
-        response.Date = httpResponse.GetHeaders().at("Date");
+        response.Date = httpResponse.GetHeaders().at("date");
         response.RequestId = httpResponse.GetHeaders().at("x-ms-request-id");
         auto response_client_request_id_iterator
             = httpResponse.GetHeaders().find("x-ms-client-request-id");
@@ -1947,8 +1946,8 @@ namespace Azure { namespace Storage { namespace Blobs {
         {
           response.ClientRequestId = response_client_request_id_iterator->second;
         }
-        response.ETag = httpResponse.GetHeaders().at("ETag");
-        response.LastModified = httpResponse.GetHeaders().at("Last-Modified");
+        response.ETag = httpResponse.GetHeaders().at("etag");
+        response.LastModified = httpResponse.GetHeaders().at("last-modified");
         for (auto i = httpResponse.GetHeaders().lower_bound("x-ms-meta-");
              i != httpResponse.GetHeaders().end() && i->first.substr(0, 10) == "x-ms-meta-";
              ++i)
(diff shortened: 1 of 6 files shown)


sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
```diff
@@ -361,6 +360,11 @@ namespace Azure { namespace Storage { namespace Blobs {
     throw std::runtime_error("cannot convert " + blob_lease_status + " to BlobLeaseStatus");
   }
 
+  struct BlobPrefix
```
katmsft, line 363: Why is this a separate struct?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
```diff
@@ -1100,7 +1119,7 @@ namespace Azure { namespace Storage { namespace Blobs {
         unused(body);
         auto request = Azure::Core::Http::Request(Azure::Core::Http::HttpMethod::Get, url);
         request.AddHeader("Content-Length", "0");
```
katmsft, line 1121: If the request/response header keys are all lower case, please be consistent for setting/parsing the response.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
```diff
@@ -2059,25 +2078,124 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+      }; // struct ListBlobsFlatOptions
+
+      static Azure::Core::Http::Request ListBlobsFlatConstructRequest(
+          const std::string& url,
+          BodyStreamPointer& body,
+          const ListBlobsFlatOptions& options)
+      {
+        unused(body);
+        auto request = Azure::Core::Http::Request(Azure::Core::Http::HttpMethod::Get, url);
+        request.AddHeader("Content-Length", "0");
```
katmsft, line 2097: Do we have an issue tracking this?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
```diff
@@ -2059,25 +2078,124 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+          if (bodyStream->Length() == -1)
+          {
+            bodyContent = Azure::Core::Http::BodyStream::ReadToEnd(context, *bodyStream);
+          }
+          else
+          {
+            bodyContent.resize(static_cast<std::size_t>(bodyStream->Length()));
+            Azure::Core::Http::BodyStream::ReadToCount(
+                context, *bodyStream, &bodyContent[0], bodyStream->Length());
+          }
```
katmsft, line 2151: Ideally with `if (bodyStream->Length() < 0)`.
Why not just `ReadToEnd`?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: aligned with other sdk

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: what's wrong with this line of code?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: In addition, in the [RFC](https://tools.ietf.org/html/rfc2616) it's defined `Content-Length` rather than `content-length`. So what's wrong with writing it correctly?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: If we know the body size beforehand, we can allocate the buffer in one shot.

JinmingHu-MSFT: We don't need to expand the buffer again and again

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
katmsft: Then why we are parsing as `content-length`? I don't think we are writing the correct code when we are sending or parsing with different header keys.
In an ideal world, `"Content-Length"` will be defined as a header key string constant, and it is intuitive to set/get using this constant. It is difficult for anyone who maintains this library to remember that set/get needs to use different header keys for the same headers for requests and responses.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
katmsft: Why do we need to explicitly add `"Content-Length: 0"` to each request? And even for `GET` request?
I know we are working around a bug, but I was asking if we have an issue tracking this bug.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: These are two different things. Let's focus on the first one. Do you think it's reasonable at all to tell an HTTP client that we want to set header `Content-Length`?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
katmsft: Sure, but I don't know if these extra lines of code are worthwhile due to we don't have the perf counter yet.
I'm not saying we shouldn't do it, I'm saying that we should have a same standard on optimization in code level by the look of it, not the actual number.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
katmsft: I don't think it makes sense, I don't want to always assume that other SDKs are carefully thought about.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/307

JinmingHu-MSFT: So why don't you clarify what this issue is at the first comment?

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
katmsft: > 4.2 Message Headers
> 
>    HTTP header fields, which include general-header (section 4.5),
>    request-header (section 5.3), response-header (section 6.2), and
>    entity-header (section 7.1) fields, follow the same generic format as
>    that given in Section 3.1 of RFC 822 [9]. Each header field consists
>    of a name followed by a colon (":") and the field value. **Field names
>    are case-insensitive.** The field value MAY be preceded by any amount
>    of LWS, though a single SP is preferred. Header fields can be
>    extended over multiple lines by preceding each extra line with at
>    least one SP or HT. Applications ought to follow "common form", where
>    one is known or indicated, when generating HTTP constructs, since
>    there might exist some implementations that fail to accept anything
RFC also says that `Field names
are case-insensitive.` So `Content-Length` and `content-length` even `CONTent-LenGtH` should work at all time.
I think what I want to have is case-insensitive compares when getting the response header value, then there won't be any problems anymore.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: Yes. In `ListBlobs` API, the result can be up to several megabytes. So I think it's necessary to optimize the code to avoid copying several mb of data.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: Since it's case-insensitive, I can set header by `Content-Length` and get header by `content-length`.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT:
```
    <BlobPrefix>  
      <Name>blob-prefix</Name>  
    </BlobPrefix>  
```
here's a snippet of the HTTP response. `Name` is just a property of `BlobPrefix`, although currently the only one. We cannot rule out the possibility of more properties being added in the future.

So I think it makes sense to define a struct that maps to this xml node.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
JinmingHu-MSFT: To some extent, i agree with you on the map with case-insensitive comparator idea. Anyway, it's not priority.

sdk/storage/inc/blobs/internal/protocol/blob_rest_client.hpp
katmsft: > I don't think it's necessary to benchmark before making every optimization. If we can foresee that the overhead is not trivial, we can directly optimize it, right?

And I think it's always a good habit to write optimized code.

Anyway, then I think it is only the issue with the if statement. I don't like `if (bodyStream->Length() == -1)` because it has implications on the implementation outside our control. I like my proposal better. Not critical so I'll sign-off the PR first.

JinmingHu-MSFT: > And I think it's always a good habit to write optimized code.

No, I cannot agree on this. It's all about tradeoff between optimization and readability/maintainability. Writing assembly is the most optimized but we hardly ever do that because it's not readable.

JinmingHu-MSFT: Your proposal is `if (bodyStream->Length() < 0)`. What's the difference here? Returning `-1` on unknown stream length is defined behavior. It's not going to return other value for now.

Let's say in the future it'll return other negative values, in which way, one is going to be better than the other?

PR #312, opened 2020-07-15
suggested changes in API review

JinmingHu-MSFT:


sdk/storage/src/blobs/blob_client.cpp
```diff
@@ -48,7 +48,8 @@ namespace Azure { namespace Storage { namespace Blobs {
     {
       policies.emplace_back(std::unique_ptr<Azure::Core::Http::HttpPolicy>(p->Clone()));
     }
-    // TODO: Retry policy goes here
+    policies.emplace_back(
```
katmsft, line 51: Need to add a test case to validate that the retry policy works after when we have a mocking mechanism. Create an issue or add TODO, please.

JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/318

PR #319, opened 2020-07-16
add perf benchmark code

JinmingHu-MSFT:


sdk/storage/test/blobs/performance_benchmark.cpp
```diff
@@ -36,4 +43,69 @@ namespace Azure { namespace Storage { namespace Test {
     }
   }
 
+  TEST_F(BlobContainerClientTest, DISABLED_MultiThreadPerf)
+  {
+    constexpr int concurrency = 64;
```
katmsft, line 48: This number should be tuned in the future. 64 is fine for this release.

sdk/storage/test/blobs/performance_benchmark.cpp
```diff
@@ -36,4 +43,69 @@ namespace Azure { namespace Storage { namespace Test {
...
+              "MultiThreadPerf" + RandomString()));
+    }
+
+    constexpr std::size_t bufferSize = static_cast<std::size_t>(1_GB);
+    std::vector<uint8_t> buffer = RandomBuffer(bufferSize);
+
+    {
+      std::vector<std::future<void>> futures;
+      Blobs::UploadBlobOptions options;
+      options.ChunkSize = 8_MB;
```
katmsft, line 65: This number should be tuned in the future. 8MB is fine for this release.

sdk/storage/test/blobs/performance_benchmark.cpp
JinmingHu-MSFT: We can never find a number suitable for all kinds of hardware configuration. So you tune it for a specific machine, it still won't work for other machines.

sdk/storage/test/blobs/performance_benchmark.cpp
JinmingHu-MSFT: We can never find a number suitable for all kinds of hardware configuration. So you tune it for a specific machine, it still won't work for other machines.

sdk/storage/test/blobs/performance_benchmark.cpp
katmsft: We can also specify the recommended machines here. It is difficult for maintainers to know how to run the perf test perfectly, but at least this would help achieve that goal.

PR #328, opened 2020-07-21
integrate TelemetryPolicy into Blob Service

JinmingHu-MSFT:


katmsft:
Since we are releasing different libraries, should we follow .NET's folder structure in general?
Ref: https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/storage
Each package has it's own `inc`, `src`, `sample` etc.

JinmingHu-MSFT:
> Since we are releasing different libraries, should we follow .NET's folder structure in general?
> Ref: https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/storage
> Each package has it's own `inc`, `src`, `sample` etc.

I don't know. I don't feel it's necessary though.

katmsft:
> > Since we are releasing different libraries, should we follow .NET's folder structure in general?
> > Ref: https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/storage
> > Each package has it's own `inc`, `src`, `sample` etc.
> 
> I don't know. I don't feel it's necessary though.

I think we should separate the folders for future Vcpkg releases. Can you please create an item to track as this is not going to fit into our private preview release?

katmsft:
[Feature Request] Reconstruct the folder hierachy for Azure Storage. #339

PR #336, opened 2020-07-22
Refined DataLake sample

katmsft:


sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+
+  // Creates a file under the directory.
+  auto fileClient = fileSystemClient.GetFileClient(directoryName + "/" + fileName);
+  fileClient.Create();
+
+  // Append/flush/read data from the client.
+  // Append data
+  size_t bufferSize = 64;
+  std::vector<uint8_t> buffer(bufferSize);
+  for (size_t i = 0; i < bufferSize; ++i)
```
JinmingHu-MSFT, line 60: it would be better if you append text, and twice. Such like `Hello` and `World`, and then download it back and print it.

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+  {
+    fileSystemClient.Create();
+  }
+
+  // Create a directory.
+  auto directoryClient = fileSystemClient.GetDirectoryClient(directoryName);
+  directoryClient.Create();
+
+  // Creates a file under the directory.
+  auto fileClient = fileSystemClient.GetFileClient(directoryName + "/" + fileName);
```
JinmingHu-MSFT, line 53:
```suggestion
  auto fileClient = directoryClient.GetFileClient(directoryName + "/" + fileName);
```

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+  fileClient.AppendData(bufferStream.get(), 0);
+
+  // Flush
+  fileClient.FlushData(bufferSize);
+
+  // Read
+  auto result = fileClient.Read();
+  Azure::Core::Context context;
+  auto downloaded = Azure::Core::Http::BodyStream::ReadToEnd(context, *(result->Body));
+  // downloaded contains your downloaded data.
```
JinmingHu-MSFT, line 76: How to read the downloaded content

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+
+  // Append/flush/read data from the client.
+  // Append data
+  size_t bufferSize = 64;
+  std::vector<uint8_t> buffer(bufferSize);
+  for (size_t i = 0; i < bufferSize; ++i)
+  {
+    buffer[i] = static_cast<uint8_t>(i >> 3);
+  }
+  auto bufferStream = std::make_unique<Azure::Core::Http::MemoryBodyStream>(
```
JinmingHu-MSFT, line 64: Here you're trying to give users an impression that bodyStream must be allocated on heap, or managed by smart pointer, which is not the case. We've decided that input bodyStream takes a raw pointer, so that our users can have more flexible memory management methods, which means they can choose to allocate it on stack or on heap, to manage it via raw pointer or leave it to smart pointer.

Sample file is usually the first place our new users are going to look at. So it's very important for us to convey accurate information. It's not like unittest code, which is fine as long as it works.

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+  // Create file systems.
+  auto fileSystemClient = serviceClient.GetFileSystemClient(fileSystemName);
+  auto iter = std::find_if(
+      fileSystems.begin(), fileSystems.end(), [&fileSystemName](const FileSystem& fileSystem) {
+        return fileSystem.Name == fileSystemName;
+      });
+  if (iter == fileSystems.end())
+  {
+    fileSystemClient.Create();
+  }
```
JinmingHu-MSFT, line 46: I feel this is too complicated as a sample. Sample is supposed to be easy and succinct, so that users can grasp the main points in a short time.

Creating a new filesystem and finding duplicate, which is the one we want to tell a beginner first?

JinmingHu-MSFT, line 46: I feel this is too complicated as a sample. Sample is supposed to be easy and succinct, so that users can grasp the main points in a short time.

Creating a new filesystem and finding duplicate, which is the one we want to tell a beginner first?

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+  do
+  {
+    auto response = serviceClient.ListFileSystems();
+    if (response->Continuation.HasValue())
+    {
+      continuation = response->Continuation.GetValue();
+    }
+    fileSystems.insert(
+        fileSystems.end(), response->Filesystems.begin(), response->Filesystems.end());
+  } while (!continuation.empty());
```
JinmingHu-MSFT, line 35: This should go after create filesystem

JinmingHu-MSFT, line 35: This should go after create filesystem

JinmingHu-MSFT, line 35: This should go after create filesystem

JinmingHu-MSFT, line 35: This should go after create filesystem

JinmingHu-MSFT, line 35: This should go after create filesystem

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+  size_t bufferSize = 64;
+  std::vector<uint8_t> buffer(bufferSize);
+  for (size_t i = 0; i < bufferSize; ++i)
+  {
+    buffer[i] = static_cast<uint8_t>(i >> 3);
+  }
+  auto bufferStream = std::make_unique<Azure::Core::Http::MemoryBodyStream>(
+      Azure::Core::Http::MemoryBodyStream(buffer));
+
+  fileClient.AppendData(bufferStream.get(), 0);
```
JinmingHu-MSFT, line 67: I may want to know what `0` is for at the first glance.

JinmingHu-MSFT: It will help if you add a piece of comment here.

It may also help if you append two chunks of data, with different offsets, and then flush them. That better demonstrates how `Append` and `Flush` is supposed to be used.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: I don't think it's necessary, but I don't have strong objection as well, so modified the sample per your request.

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,80 @@
...
+  fileClient.FlushData(bufferSize);
+
+  // Read
+  auto result = fileClient.Read();
+  Azure::Core::Context context;
+  auto downloaded = Azure::Core::Http::BodyStream::ReadToEnd(context, *(result->Body));
+  // downloaded contains your downloaded data.
+
+  // Delete file system.
+  fileSystemClient.Delete();
```
JinmingHu-MSFT, line 79: It throws an exception when user tries to run samples twice. `Create` would fail because the filesystem is being deleted. Then the user would get confused and try again. Then it succeeds because the filesystem was already deleted after a little time. Then the user gets more confused.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Changed to `auto fileClient = directoryClient.GetFileClient(fileName)`

sdk/storage/sample/datalake_getting_started.cpp
katmsft: By `Azure::Core::Http::BodyStream::ReadToEnd` or ? I thought it was already clear enough. I'll make the return value to be <uint8_t> to be more specific.

JinmingHu-MSFT: it may help a little bit if you write the return type out instead of using `auto`.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: I don't think it should be covered by Storage's sample, if we need to cover every bit of usage of Azure Core, the effort will be huge for us.
Anyway, I think it is not a bad idea to provide more information in the sample, so I changed how `MemoryBodyStream` can be initialized and showed two approaches.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: dito.

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: I didn't mean we should cover everything of storage, not at all. We can only convey very limited information from sample, so we should make it concise, accurate and less misunderstanding.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Already changed to this.

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: Sorry I cannot agree.

What's the first thing we want our users to learn about DataLake SDK? I'd say to create a filesystem, and after that maybe create a directory or file.

So you used 9 lines of code here, only 1 line is useful for our users to know how to create a filesystem.

Finding a specific string in a vector of strings, why would somebody want to know how to do that?

sdk/storage/inc/datalake/directory_client.hpp
```diff
@@ -65,6 +65,13 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
         const std::string& directoryUri,
         const DirectoryClientOptions& options = DirectoryClientOptions());
 
+    /**
+     * @brief Create a FileClient from current DirectoryClient
+     * @param path Path of the file under the directory.
+     * @return FileClient
+     */
+    FileClient GetFileClient(const std::string& path) const;
```
JinmingHu-MSFT, line 73: Is there something like `GetSubdirectory` in other SDK? If we have function to get a file under a directory, maybe it's also reasonable to get a subdirectory.

JinmingHu-MSFT: If we do have this function in other SDK, I think you dont have to do it now, creating an issue to track it is also fine.

katmsft: Yes, but Jeffrey did not bring this up during the API review so I don't think it's currently necessary to add. 
Created https://github.com/Azure/azure-sdk-for-cpp/issues/338 to track.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Added exception parsing logic.

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,117 @@
...
+  // downloaded contains your downloaded data.
+  std::cout << "Downloaded data was:\n" + std::string(downloaded.begin(), downloaded.end())
+            << std::endl;
+
+  // Delete file system.
+  try
+  {
+    fileSystemClient.Delete();
+  }
+  catch (const Azure::Storage::StorageError& e)
```
JinmingHu-MSFT, line 108: Usually the exception won't be thrown here, because we're using key authentication and it's not common if a key has permission to create a filesystem while doesn't have permission to delete a filesystem.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: I don't think so, the sequence is sample for ServiceClient->FileSystemClient->Path/Directory/FileClient. I think it is better.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Because we don't have `CreateIfNotExist` or anywhere to tell our customers that how you can find anything in a listed result.
Again, I think with the newly added comment make the code very discoverable.

JinmingHu-MSFT: Still cannot agree, maybe we should escalate to @vinjiang

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: CreateFilesystem is very basic, ListFilesystem is more advanced. Plus the list result is very likely to be empty.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: > so we should make it concise, accurate and less misleading.

Well, I think it is at least equally important for our customers to know the buffer can be passed in by either a raw pointer or reference, so I made the sample to cover both. I still don't think the previous implementation is misleading, as we can always use a dereference operator or a reference operator to transform one into another.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Sure, I don't have problem with that.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Yes, but among all the other things in this sample, this is the one that gets asked most in Emails/Github issues from my feeling.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Actually it is quite possible if the customer is trying with SAS and limited access granted. I recalled that track 1 SDK has all the calls caught in a try/catch block so I changed to this as well.

JinmingHu-MSFT: fine

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: I cannot remember, can you share some of those Emails/GitHub issues?

Even if it was true, it still couldn't justify that ListFileSystem is a more important API than CreateFileSystem and it should come first.

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,100 @@
...
+      fileSystems.insert(
+          fileSystems.end(), response->Filesystems.begin(), response->Filesystems.end());
+    } while (!continuation.empty());
+
+    // Find if the file systems already exist.
+    auto fileSystemClient = serviceClient.GetFileSystemClient(fileSystemName);
+    auto iter = std::find_if(
+        fileSystems.begin(), fileSystems.end(), [&fileSystemName](const FileSystem& fileSystem) {
+          return fileSystem.Name == fileSystemName;
+        });
```
JinmingHu-MSFT, line 44: new line here

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,100 @@
...
+
+    // Creates a file under the directory.
+    auto fileClient = directoryClient.GetFileClient(fileName);
+    fileClient.Create();
+
+    // Append/flush/read data from the client.
+    // Append data
+    std::string str1 = "Hello ";
+    std::string str2 = "World!";
+    std::vector<uint8_t> buffer(str1.begin(), str1.end());
```
JinmingHu-MSFT, line 63: You're making a copy of all the data to be uploaded, this can be misleading. Because several of our Track1 customers ever had this concern of too many copies.

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: i'm sure we're not talking about the same thing.

JinmingHu-MSFT: I'm concerned about `MemoryBodyStream` being managed by `std::unique_ptr`, not which constructor of `MemoryBodyStream` you're calling.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: This is just to show that our customers can also use the buffer if they want to. We shouldn't care how the buffer ifsconstructed and customers shouldn't reference this sample code to see how the raw data, which is customer input, is constructed.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Removed.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: I believe you just handled a case from Nara V and we have databox reporting similar issues of listing returning empty value.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: I see, then I actually agree with you. Changed to stack-allocated.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: After discussion, I think it's one vote against two to stay as is, so I changed the sequence.

sdk/storage/sample/datalake_getting_started.cpp
```diff
@@ -3,11 +3,110 @@
...
+    // Create file systems and ignore the already exist error.
+    try
+    {
+      fileSystemClient.Create();
+    }
+    catch (Azure::Storage::StorageError& e)
+    {
+      if (e.ErrorCode != "ContainerAlreadyExists")
+      {
+        throw std::move(e);
```
JinmingHu-MSFT, line 38: What is this?

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: Then how did you tell our customers that they shouldn't reference this sample?

katmsft: Well, it's to me a common sense not to reference how *input* parameters are initialized for any samples. If we should also be careful about that, then should we also comment on `GetConnectionString()` not to be the recommended way of setting/getting connection strings?

sdk/storage/sample/datalake_getting_started.cpp
katmsft: Cannot `throw e` as it contains a `unique_ptr` disabling the copy constructor.

JinmingHu-MSFT: rethrow

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: So some users have some questions about a specific API, so what? We should put every API that some customer ever asked before at the very beginning of this sample?

sdk/storage/sample/datalake_getting_started.cpp
JinmingHu-MSFT: Sorry, but this doesn't look like a common sense to me.

If you have any good suggestions about how to set connection string in sample, we can discuss about it.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: First, it is already not at the beginning of the latest commit, since I was out-voted for what is the most important in the sample. Second, I want to avoid answering the questions we keep getting asked for track 1 SDK by putting it into the most obvious places in the sample, I don't feel wrong for it.
Also, I never encountered even one occasion where a customer complains about the sample containing too much information, it's usually the other way around where customers complain about the sample not containing enough information.

sdk/storage/sample/datalake_getting_started.cpp
katmsft: I'm not saying current `GetConnectionString()`, what track 1 does (hard-coded), or what we are doing here (getting from system env) are both fine to me as it is an input parameter. I don't think we should go an extra mile to regulate what customer tries to do with input parameter which is transparent to storage SDK as long as the type is right and data is valid. 
> Because several of our Track1 customers ever had this concern of too many copies.

Track 1 SDK customer complains about the copies within the SDK, I don't think it's rational to complain about copies in customer code, which is managed by themselves.
>Sorry, but this doesn't look like a common sense to me.

Well, if you don't think it's common sense, I can somehow understand why I should comment that this is not what we suggest our customer to do. Added the comment for clarity.

JinmingHu-MSFT:
approved to unblock private preview release

PR #337, opened 2020-07-22
Rename XxxRespone->XxxInfo

JinmingHu-MSFT:


sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -192,7 +181,7 @@ namespace Azure { namespace Storage { namespace Blobs {
      * parameters to execute this function.
      * @return A SetAccessTierResponse on successfully setting the tier.
```
katmsft, line 182: Not consistent with the actual return type.

sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -220,7 +209,7 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @param options Optional parameters to execute this function.
      * @return A AbortCopyBlobResponse on successfully aborting.
      */
-    Azure::Core::Response<AbortCopyBlobResponse> AbortCopyFromUri(
+    Azure::Core::Response<AbortCopyBlobInfo> AbortCopyFromUri(
```
katmsft, line 212: What about `AbortCopy` API? Are they having the same name?

sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -283,7 +272,7 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @param options Optional parameters to execute this function.
      * @return A DeleteBlobResponse on successfully deleting.
```
katmsft, line 273: ditto.

katmsft: And others.

sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -192,7 +181,7 @@ namespace Azure { namespace Storage { namespace Blobs {
      * parameters to execute this function.
      * @return A SetAccessTierResponse on successfully setting the tier.
      */
-    Azure::Core::Response<SetAccessTierResponse> SetAccessTier(
+    Azure::Core::Response<SetAccessTierInfo> SetAccessTier(
```
katmsft, line 184: Can we be absolutely sure that there won't be any container access tier in the future? If not, then there would be breaking change or `SetAccessTierInfo` and `SetContainerAccessTierInfo`.

sdk/storage/inc/blobs/blob_responses.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "protocol/blob_rest_client.hpp"
+
+#include <limits>
+#include <string>
+#include <utility>
+
+namespace Azure { namespace Storage { namespace Blobs {
+
+  struct BlobDownloadInfo
```
katmsft, line 14: `DeleteContainerInfo`
`DeleteBlobInfo`
`UndeleteBlobInfo`
Those are all `<Verb><Object>Info`
But `BlobDownloadInfo` is `<Object><Verb>Info`.

sdk/storage/inc/blobs/blob_responses.hpp
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "protocol/blob_rest_client.hpp"
+
+#include <limits>
```
katmsft, line 8: What's it for?

sdk/storage/inc/blobs/blob_responses.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <string>
+#include <utility>
```
katmsft, line 10: What's it for?

sdk/storage/inc/blobs/blob_responses.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+  struct BlobDownloadInfo
+  {
+    std::string ETag;
+    std::string LastModified;
+    int64_t ContentLength = 0;
+    BlobHttpHeaders HttpHeaders;
+    std::map<std::string, std::string> Metadata;
+    Blobs::BlobType BlobType = Blobs::BlobType::Unknown;
+    Azure::Core::Nullable<bool> ServerEncrypted;
```
katmsft, line 22: Why is it `Nullable`?

sdk/storage/inc/blobs/blob_responses.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+  struct PageRange
+  {
+    int64_t Offset;
+    int64_t Length;
+  };
+
+  struct PageRangesInfo
+  {
+    std::string RequestId;
```
katmsft, line 34: Do we still need to have the fabulous 4 here?

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: Why would a container have access tier? It doesn't make any sense.

sdk/storage/inc/blobs/blob_responses.hpp
JinmingHu-MSFT: align with .Net SDK

sdk/storage/inc/blobs/blob_responses.hpp
JinmingHu-MSFT: I cannot remember. But I changed it from non-nullable to nullable. There must be an API that doesn't return this value in response headers.

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: There isn't such an API

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_responses.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_responses.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_client.hpp
katmsft: There might be the case that customer wants to manage the access tier at the container level. I'm not sure if that's going to be added in the future.
What I want to point out is that `SetAccessTierInfo` does not limit the scope to be `Blob`, then any addition to the scope under the `Azure::Storage::Blobs` namespace would be breaking change in the future. If this is considered not possible in your mind, I'm fine with the current naming.

sdk/storage/inc/blobs/blob_responses.hpp
katmsft: Then I don't think .NET SDK is doing the correct thing here. @JeffreyRichter what do you think? Is it going to be challenged by the review board in the future?

sdk/storage/inc/blobs/blob_responses.hpp
katmsft: fixed.

sdk/storage/inc/blobs/blob_responses.hpp
katmsft: Maybe `GetBlob` REST with content range does not return this. Please give it a try when you have time. Not priority.

sdk/storage/inc/blobs/blob_client.hpp
katmsft: https://docs.microsoft.com/en-us/rest/api/storageservices/abort-copy-blob

JinmingHu-MSFT: There's only one Copy and one Abort Copy API in client API. This aligns with other sdk

sdk/storage/inc/blobs/blob_responses.hpp
JinmingHu-MSFT: I think it's fine.

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: fixed

PR #340, opened 2020-07-23
Added Readme/Changelog and 3rd party notice.

katmsft:


sdk/storage/CHANGELOG.md
```diff
@@ -0,0 +1,71 @@
+# Release History
+
+## 1.0.0-preview.1 (Unreleased)
+
+* Added support for Blob features:
```
JinmingHu-MSFT, line 5: It's not common to use past tense.

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Need Help?
+
+Be sure to check out the [Azure Storage Forum](https://social.msdn.microsoft.com/Forums/azure/en-US/home?forum=windowsazuredata) on MSDN if you need help, or use [StackOverflow](http://stackoverflow.com/questions/tagged/azure).
+
+## Collaborate & Contribute
+
+We gladly accept community contributions.
+
+- **Issues:** Report bugs on the [Issues page](https://github.com/Azure/azure-sdk-for-cpp/issues) in GitHub. Ideally, please add an "[Azure-Storage]" prefix to the title for easier categorizing.
```
JinmingHu-MSFT, line 31: `[Storage]` prefix? Because everything in this repo is about Azure.

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+#### Unix Platforms
+
+You can use the package manager on different Unix platforms to install the dependencies. The dependencies to installs are:
+
+CMake 3.15.0 and greater.
+Libxml2.
+OpenSSL.
+Libcurl.
```
JinmingHu-MSFT, line 56: check the preview

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+On Windows, dependencies are managed by [Vcpkg](https://github.com/microsoft/vcpkg). You can reference the [Quick Start](https://github.com/microsoft/vcpkg#quick-start-windows) to quickly set yourself up.
+After Vcpkg is initialized and bootstrapped, you can install the dependencies:
+```BatchFile
+vcpkg.exe install libxml2:x64-windows curl:x64-windows
+```
+
+#### Unix Platforms
+
+You can use the package manager on different Unix platforms to install the dependencies. The dependencies to installs are:
```
JinmingHu-MSFT, line 51: > The dependencies to installs are

maybe typo or something?

sdk/storage/README.md
JinmingHu-MSFT, line 56:
```suggestion
libcurl.
```

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+First, download the repository to your local folder:
+```BatchFile
+git clone https://github.com/Azure/azure-sdk-for-cpp.git
+```
+
+#### Windows
+
+##### Use CMake to generate the solution file
+```BatchFile
+cmake . -A x64 -DCMAKE_TOOLCHAIN_FILE=<YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake -DBUILD_CURL_TRANSPORT=ON -DBUILD_TESTING=ON -DBUILD_STORAGE_SAMPLES=ON -DCMAKE_CXX_FLAGS="/MP /EHsc"
```
JinmingHu-MSFT, line 69:
```suggestion
cmake . -A x64 -DCMAKE_TOOLCHAIN_FILE=<YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake
```

JinmingHu-MSFT: don't make things too complicated for our users

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+Then you can open the generated `az.sln` to open the solution and build the library.
+The built library will be in `<ProjectDir>\<Setup>\` respectively for Azure Core and Azure Storage. e.g. `azure_core.lib` will be in `sdk\core\azure-core\Debug` for debug setup.
```
JinmingHu-MSFT, line 73: Is it called configuration or setup?

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+If Vcpkg is not globally integrated, then you need to open CMakeSettings.json and change the `Make toolchain file to be <YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake` and save.
+Then you can build Azure Storage libraries by selecting the target in Visual Studio, or simply build all.
+The libraries will be in `<ProjectRoot>\out\build\<Setup>\sdk\<LibraryName>` respectively.
+
+#### Unix Platforms
+
+You can run the following command in the downloaded code's root folder to build the code.
+
+```BatchFile
+cmake . -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON -DBUILD_STORAGE_SAMPLES=ON
```
JinmingHu-MSFT, line 87:
```suggestion
cmake . -DCMAKE_BUILD_TYPE=Debug
```

JinmingHu-MSFT: it's a very very bad habit to run cmake command it root directory.

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+make
```
JinmingHu-MSFT, line 88:
```suggestion
cmake --build .
```

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+#### Windows
+
+##### Use CMake to generate the solution file
+```BatchFile
+cmake . -A x64 -DCMAKE_TOOLCHAIN_FILE=<YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake -DBUILD_CURL_TRANSPORT=ON -DBUILD_TESTING=ON -DBUILD_STORAGE_SAMPLES=ON -DCMAKE_CXX_FLAGS="/MP /EHsc"
+```
+
+Then you can open the generated `az.sln` to open the solution and build the library.
```
JinmingHu-MSFT, line 72: It's kind of unfriendly for our customers to use command line first and then switch to GUI. Why not stick to command line all the time in this section and use GUI all the time in the next section?

sdk/storage/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Dependencies
+
+### Azure Core SDK
+[Reference documentation.](https://github.com/Azure/azure-sdk-for-cpp/blob/master/README.md)
+
+### nlohmann/json
+[Official Web Page](https://github.com/nlohmann/json)
+
+### Libxml2
+[Official Web Page](http://xmlsoft.org/)
```
JinmingHu-MSFT, line 111: [Azure Core SDK](https://github.com/Azure/azure-sdk-for-cpp/blob/master/README.md)

[nlohmann/json](https://github.com/nlohmann/json)

[Libxml2](http://xmlsoft.org/)

sdk/storage/CHANGELOG.md
katmsft: https://github.com/Azure/azure-storage-ruby/blob/master/ChangeLog.md
https://github.com/Azure/azure-storage-net/blob/master/Blob/Changelog.txt
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/CHANGELOG.md
I think it is pretty common for Azure Storage SDK to use past tense. Can you elaborate why do you think it is not common to use past tense?

sdk/storage/README.md
katmsft: Sure, why not :).

sdk/storage/README.md
katmsft: Nice catch.

sdk/storage/README.md
katmsft: Interesting, the official name is indeed libcurl with lowercase `l`. Good catch!

sdk/storage/README.md
katmsft: Default for `BUILD_CURL_TRANSPORT` is OFF so if we want it to work by default, we need to leave it here at least. However, I aggree that we can remove the other ones.

sdk/storage/README.md
katmsft: It's configuration, good catch, I'll change them.

sdk/storage/README.md
katmsft: I never knew this, can you elaborate?

sdk/storage/README.md
katmsft: Well, I didn't think about this. I think you are probably right, ignore the GUI build and added `cmake --build .` in this guide as well.

sdk/storage/CHANGELOG.md
JinmingHu-MSFT: well i have counter example
https://github.com/microsoft/vcpkg/blob/master/CHANGELOG.md

sdk/storage/README.md
JinmingHu-MSFT: the generated files would be mixed up with source files and other files that already exist in the repo.

sdk/storage/README.md
```diff
@@ -0,0 +1,116 @@
...
+After Vcpkg is initialized and bootstrapped, you can install the dependencies:
+```BatchFile
+vcpkg.exe install libxml2:x64-windows curl:x64-windows
+```
+
+#### Unix Platforms
+
+You can use the package manager on different Unix platforms to install the dependencies. The dependencies to be installed are:
+
+  - CMake 3.15.0 and greater.
```
JinmingHu-MSFT, line 53: higher?

sdk/storage/README.md
```diff
@@ -0,0 +1,116 @@
...
+git clone https://github.com/Azure/azure-sdk-for-cpp.git
+```
+
+#### Windows
+
+##### Use CMake to generate the solution file
+
+In a new folder you created under the root directory:
+```BatchFile
+cmake .. -A x64 -DCMAKE_TOOLCHAIN_FILE=<YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake -DBUILD_CURL_TRANSPORT=ON
```
JinmingHu-MSFT, line 71: You've tried that if we don't add `-DBUILD_CURL_TRANSPORT=ON` it won't work, right?

sdk/storage/README.md
```diff
@@ -0,0 +1,116 @@
...
+cmake --build .
+```
```
JinmingHu-MSFT, line 73: Not now, but maybe in the future, you may want to add more options. Reference to https://github.com/Azure/azure-storage-cpplite#other-build-options

sdk/storage/README.md
```diff
@@ -0,0 +1,116 @@
...
+WIP
+TODO when ready.
+
+## Dependencies
+
+### [Azure Core SDK](https://github.com/Azure/azure-sdk-for-cpp/blob/master/README.md)
+
+### [nlohmann/json](https://github.com/nlohmann/json)
+
+### [Libxml2](http://xmlsoft.org/)
```
JinmingHu-MSFT, line 110: I meant it seems not necessary to make these 3 headings. You may want to try and compare, you can pick whichever you think it's prettier.

sdk/storage/README.md
katmsft: It works now but it shouldn't. The fact that this works without `DBUILD_CURL_TRANSPORT=ON` should be a bug. If that is fixed in the future, the README would then be invalid.

sdk/storage/README.md
katmsft: Yeah, created a refinement issue for it. https://github.com/Azure/azure-sdk-for-cpp/issues/349

sdk/storage/README.md
JinmingHu-MSFT: I don't think it's a bug

JinmingHu-MSFT: There's gonna be one default transport, either libcurl or winHTTP or something else, right?

sdk/storage/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Via Vcpkg
+WIP
+TODO when ready.
+
+## Dependencies
+
+  - [Azure Core SDK](https://github.com/Azure/azure-sdk-for-cpp/blob/master/README.md)
+  - [nlohmann/json](https://github.com/nlohmann/json)
+  - [Libxml2](http://xmlsoft.org/)
```
JinmingHu-MSFT, line 108: Be consistent, `Libxml2` or `libxml2`

sdk/storage/README.md
```diff
@@ -0,0 +1,114 @@
...
+If Vcpkg is not globally integrated, then you need to open CMakeSettings.json and change the `Make toolchain file to be <YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake` and save.
+Then you can build Azure Storage libraries by selecting the target in Visual Studio, or simply build all.
+The libraries will be in `<ProjectRoot>\out\build\<Configuration>\sdk\<LibraryName>` respectively.
+
+#### Unix Platforms
+
+You can run the following command in a new folder created under the downloaded code's root folder to build the code.
+
+```BatchFile
+cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON -DBUILD_STORAGE_SAMPLES=ON
```
JinmingHu-MSFT, line 89: we don't want our customers to build test by default.

JinmingHu-MSFT: if you have BUILD_STORAGE_SAMPLES turned on, also do the same for Windows build commands.

sdk/storage/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+If Vcpkg is not globally integrated, then you need to open CMakeSettings.json and change the `Make toolchain file to be <YOUR_VCPKG_INSTALL_DIR>/scripts/buildsystems/vcpkg.cmake` and save.
+Then you can build Azure Storage libraries by selecting the target in Visual Studio, or simply build all.
+The libraries will be in `<ProjectRoot>\out\build\<Configuration>\sdk\<LibraryName>` respectively.
+
+#### Unix Platforms
+
+You can run the following command in a new folder created under the downloaded code's root folder to build the code.
+
+```BatchFile
```
JinmingHu-MSFT, line 88: This is not BatchFile

sdk/storage/README.md
katmsft: Removed and created https://github.com/Azure/azure-sdk-for-cpp/issues/350 for refinement from Core side.

sdk/storage/CHANGELOG.md
katmsft: I will leave it as is and can you propose this in next discussion meeting?

PR #353, opened 2020-07-27
Add Account SAS and Blob SAS

JinmingHu-MSFT:


sdk/storage/inc/blobs/blob_sas_builder.hpp
```diff
@@ -0,0 +1,109 @@
...
+  {
+    Container,
+    Blob,
+    BlobSnapshot,
+    BlobVersion,
+  };
+
+  enum class BlobContainerSasPermissions
+  {
+    None = 0,
```
katmsft, line 23: Why is it necessary to have `None` here?

JinmingHu-MSFT: So that we can write code like this https://github.com/Azure/azure-sdk-for-cpp/blob/daa6854717c7322edfcee7d905945d0d2f2316c8/sdk/storage/src/blobs/blob_sas_builder.cpp#L14

JinmingHu-MSFT: I can remove these `None`s if you have better way to do that.

sdk/storage/inc/blobs/blob_sas_builder.hpp
```diff
@@ -0,0 +1,109 @@
...
+    std::string StartsOn;
+    std::string ExpiresOn;
+    std::string IPRange;
+    std::string Identifier;
+    std::string ContainerName;
+    std::string BlobName;
+    std::string Snapshot;
+    BlobSasResource Resource;
+
+    std::string CacheControl;
```
katmsft, line 91: Why not `BlobHttpHeaders`?

sdk/storage/inc/blobs/blob_sas_builder.hpp
```diff
@@ -0,0 +1,109 @@
...
+  {
+    using type = std::underlying_type_t<BlobSasPermissions>;
+    return static_cast<BlobSasPermissions>(static_cast<type>(lhs) & static_cast<type>(rhs));
+  }
+
+  struct BlobSasBuilder
+  {
+    std::string Version = c_APIVersion;
+    SasProtocol Protocol;
+    std::string StartsOn;
```
katmsft, line 82: How would a customer use this? Is there a helper function?

sdk/storage/inc/blobs/blob_sas_builder.hpp
JinmingHu-MSFT: I don't know, which one do you prefer?

sdk/storage/inc/blobs/blob_sas_builder.hpp
JinmingHu-MSFT: Nope, Azure Core team is supposed to provide a `DataTime` class.

sdk/storage/inc/blobs/blob_sas_builder.hpp
```diff
@@ -0,0 +1,109 @@
...
+    std::string CacheControl;
+    std::string ContentDisposition;
+    std::string ContentEncoding;
+    std::string ContentLanguage;
+    std::string ContentType;
+
+    void SetPermissions(BlobContainerSasPermissions permissions);
+    void SetPermissions(BlobSasPermissions permissions);
+
+    std::string ToSasQueryParameters(const SharedKeyCredential& credential);
```
katmsft, line 100: `SharedKeyCredential`'s constructor takes `std::string accountName, std::string accountKey`. Is it better to make the parameters be `const std::string& sharedKey,
        const std::string& accountName` to be more clear of what is needed and more consistent with the function below?

sdk/storage/inc/common/account_sas_builder.hpp
```diff
@@ -0,0 +1,108 @@
...
+  enum class AccountSasResource
+  {
+    None = 0,
+    Service = 1,
+    Container = 2,
+    Object = 4,
+    All = ~0,
+  };
+
+  inline AccountSasResource operator|(AccountSasResource lhs, AccountSasResource rhs)
```
katmsft, line 28: I personally don't like operator overloading outside of the class. But it's minor.

sdk/storage/inc/blobs/blob_sas_builder.hpp
JinmingHu-MSFT: ah, I remember, we cannot override `Content-MD5`

sdk/storage/inc/blobs/blob_sas_builder.hpp
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/242

sdk/storage/inc/blobs/blob_sas_builder.hpp
```diff
@@ -0,0 +1,109 @@
...
+    std::string Snapshot;
+    BlobSasResource Resource;
+
+    std::string CacheControl;
+    std::string ContentDisposition;
+    std::string ContentEncoding;
+    std::string ContentLanguage;
+    std::string ContentType;
+
+    void SetPermissions(BlobContainerSasPermissions permissions);
```
katmsft, line 97: Hopefully, your document will mark that calling different `SetPermissions` on the same `BlobSasBuilder` is undefined behavior.

sdk/storage/inc/common/storage_credential.hpp
```diff
@@ -24,17 +30,19 @@ namespace Azure { namespace Storage {
       m_accountKey = std::move(accountKey);
     }
 
-    std::string AccountName;
+    const std::string AccountName;
 
   private:
     friend class SharedKeyPolicy;
-    std::string GetAccountKey()
+    friend struct Blobs::BlobSasBuilder;
```
katmsft, line 37: If Sas builders are decoupled from `SharedKeyCredential`, we don't need this anymore. Please consider it.

sdk/storage/test/test_base.hpp
```diff
@@ -59,4 +61,7 @@ namespace Azure { namespace Storage { namespace Test {
 
   void DeleteFile(const std::string& filename);
 
+  std::string ToISO8601(const std::chrono::system_clock::time_point& time_point);
```
katmsft, line 64: nit: `ToIso8601` and `ToRfc1123`?

sdk/storage/inc/common/account_sas_builder.hpp
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  inline AccountSasPermissions operator&(AccountSasPermissions lhs, AccountSasPermissions rhs)
+  {
+    using type = std::underlying_type_t<AccountSasPermissions>;
+    return static_cast<AccountSasPermissions>(static_cast<type>(lhs) & static_cast<type>(rhs));
+  }
+
+  struct AccountSasBuilder
+  {
```
katmsft, line 91: I feel somethings here should be `Nullable` right? e.g. `IpRange`.

sdk/storage/inc/common/account_sas_builder.hpp
JinmingHu-MSFT: enum class is not class, they cannot have member functions. Where else do you want to put them?

sdk/storage/inc/blobs/blob_sas_builder.hpp
JinmingHu-MSFT: We already have an abstraction of account name and shared key, is there any reason you don't want to use it? Plus key is supposed to be a secret thing. I don't think it's good practice to pass this secret around everywhere.

JinmingHu-MSFT: So I got the same question for you. In [filesystem client](https://github.com/Azure/azure-sdk-for-cpp/blob/573ffdf1f73b4a03861dcf1c3fd0a761cce588d0/sdk/storage/inc/datalake/file_system_client.hpp#L48), why don't you have the constructor accept account name and shared key, but use `SharedKeyCredential` instead?

sdk/storage/inc/common/storage_credential.hpp
JinmingHu-MSFT: i'd rather not do that. see comments https://github.com/Azure/azure-sdk-for-cpp/pull/353#discussion_r460931362.

sdk/storage/test/test_base.hpp
JinmingHu-MSFT: We don't have guideline for that yet. Why not bring this up in the next weekly meeting?

sdk/storage/inc/common/account_sas_builder.hpp
JinmingHu-MSFT: I don't think it's necessary. What are the benefits? Plus in .Net SDK, they're ordinary string.

sdk/storage/inc/blobs/blob_sas_builder.hpp
JinmingHu-MSFT: it's defined behavior, the last one shall prevail.

sdk/storage/inc/common/account_sas_builder.hpp
JinmingHu-MSFT: Changed IPRange to nullable

sdk/storage/inc/blobs/blob_sas_builder.hpp
katmsft: I usually use `if ((permissions | BlobContainerSasPermissions::Read) == permissions)`.

sdk/storage/inc/blobs/blob_sas_builder.hpp
katmsft: The only concern is the overhead introduced by `SharedKeyCredential` because it has mutex in it. It has no additional usage of the class itself besides the fact that it holds the account name and key.

sdk/storage/inc/common/account_sas_builder.hpp
katmsft: Look forward to the Core's extensible enum, if that solves this problem.

sdk/storage/inc/blobs/blob_sas_builder.hpp
katmsft: As we talked offline, filesystem client holds `SharedKeyCredential` and uses other functionalities provided by the class, including thread safety insurance.

JinmingHu-MSFT: i'm not gonna take that as a convincing argument unless you have evidence shows that the performance is impacted.

PR #366, opened 2020-07-29
GetBlobServiceProperties and SetBlobServiceProperties

JinmingHu-MSFT:


sdk/storage/inc/blobs/protocol/blob_rest_client.hpp
```diff
@@ -354,6 +363,12 @@ namespace Azure { namespace Storage { namespace Blobs {
     std::string Name;
   }; // struct BlobPrefix
 
+  struct BlobRetentionPolicy
+  {
+    bool Enabled = false;
+    Azure::Core::Nullable<int32_t> Days;
```
katmsft, line 369: Retention policy itself can be nullable but not `Days` right?

katmsft: The swagger is having different behavior with the doc, with swagger stating the parameter being required but doc does not:

Required only if a retention policy is enabled. Indicates the number of days that metrics or logging data should be retained. All data older than this value will be deleted. The minimum value you can specify is 1; the largest value is 365 (one year).
--

katmsft: I think we should follow the doc, not swagger.

sdk/storage/inc/blobs/protocol/blob_rest_client.hpp
```diff
@@ -1098,7 +1154,222 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+          const std::string& url,
+          const GetPropertiesOptions& options)
+      {
+        auto request = Azure::Core::Http::Request(Azure::Core::Http::HttpMethod::Get, url);
+        request.AddQueryParameter("restype", "service");
+        request.AddQueryParameter("comp", "properties");
+        request.AddHeader("x-ms-version", c_APIVersion);
+        if (options.Timeout.HasValue())
+        {
+          request.AddQueryParameter("timeout", std::to_string(options.Timeout.GetValue()));
```
katmsft, line 1174: I thought we aggreed that timeout should be added by pipeline.

JinmingHu-MSFT: I'll create an issue to track that

JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/370

sdk/storage/inc/blobs/protocol/blob_rest_client.hpp
JinmingHu-MSFT: oh great, another testimony that swagger is not perfect. i'm going to write it down.

sdk/storage/inc/blobs/protocol/blob_rest_client.hpp
```diff
@@ -1473,6 +1830,302 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+            else if (path.size() == 1 && path[0] == XmlTagName::k_Days)
+            {
+              ret.Days = std::stoi(node.Value);
+            }
+          }
+        }
+        return ret;
+      }
+
+      static BlobStaticWebsite BlobStaticWebsiteFromXml(XmlReader& reader)
```
katmsft, line 2053: I remember we had a conversation regarding if the name should contain `Blob` in it. `Azure::Storage::Blobs::BlobSomething` just feels redundant. I think we want to bring this up to Jeffrey to avoid getting challenged during official API review.

JinmingHu-MSFT: Sure, we can certainly discuss about this with Jeffrey. But currently it aligns with .Net, so I guess we're unlikely to be challenged.

PR #371, opened 2020-07-30
[Storage Blob Service] API GetAccountInfo

JinmingHu-MSFT:


sdk/storage/src/blobs/blob_service_client.cpp
```diff
@@ -159,4 +159,20 @@ namespace Azure { namespace Storage { namespace Blobs {
...
 
+  Azure::Core::Response<AccountInfo> BlobServiceClient::GetAccountInfo(
+      const GetAccountInfoOptions& options) const
+  {
+    BlobRestClient::Service::GetAccountInfoOptions protocolLayerOptions;
+    return BlobRestClient::Service::GetAccountInfo(
+        options.Context, *m_pipeline, m_serviceUrl.ToString(), protocolLayerOptions);
+  }
+
+  Azure::Core::Response<BlobServiceStatistics> BlobServiceClient::GetStatistics(
```
katmsft, line 170: Please don't provide API that does not work right now.

JinmingHu-MSFT: removed

PR #381, opened 2020-08-03
[Storage Blob Service] Get/SetBlobContainerAccessPolicy

JinmingHu-MSFT:


sdk/storage/inc/blobs/blob_sas_builder.hpp
```diff
@@ -281,6 +286,12 @@ namespace Azure { namespace Storage { namespace Blobs {
      */
     void SetPermissions(BlobSasPermissions permissions);
 
+    /**
+     * @brief Clears the permissions for the blob SAS.
+     * 
+     */
+    void ResetPermisions() { Permissions.clear(); }
```
katmsft, line 293: Why do we have this here?

JinmingHu-MSFT: mainly for test purpose

katmsft: It's a struct, just clear in test code.

JinmingHu-MSFT: removed

PR #384, opened 2020-08-03
Added support for get/set file service properties.

katmsft:


JinmingHu-MSFT:
I don't think it should be called `service_client.hpp`. We would have multiple `service_client.hpp`. How about `datalake_service_client.hpp` and `share_servic_client.hpp`? .Net SDK does so

katmsft:
> I don't think it should be called `service_client.hpp`. We would have multiple `service_client.hpp`. How about `datalake_service_client.hpp` and `share_servic_client.hpp`? .Net SDK does so

Sure, I'll change them in a new PR to align with Blob service and .NET SDK.

PR #397, opened 2020-08-05
Updates to Context/ContextSharedState

CaseyCarter:
Per discussion on 2020-07-28: Casey to `const`-qualify non-mutating member functions of `Context` and provide mutual exclusion for accesses to `Context::ContextSharedState::CancelAt`. 

There are also some drive-by changes. Two micro-optimizations:

  1. `ContextSharedState` accepts a `shared_ptr` argument by `const&` and makes a copy, increasing its reference count. This is suboptimal when the argument is an rvalue, which we could move into `ContextSharedState::parent` without touching reference counts. If the constructor is changed to instead take the `shared_ptr` by value and move it into `parent` then the caller will make a copy when the argument is an lvalue and no one makes a copy when the argument is an rvalue - optimally handling both cases.

  2. Several functions (`Context::operator[]`, `Context::HasKey`, and `Context::CancelWhen`) iterate over the chain of `parent` pointers rooted in the `Context`'s associated `ContextSharedState` member `m_contextSharedState`. They do so by using a `shared_ptr` iteration variable into which successive `shared_ptr` values from the chain of `parent`s are copied. Each of these `shared_ptr` copy assignments decreases the reference count of the old value and increases the reference count of the new value. However, the chain of parent pointers cannot be modified while we iterate over it so the reference count manipulation is unnecessary. If we instead use a raw pointer for the iteration variable and assign it successive values of `parent->get()` we can iterate the same sequence of parent objects more cheaply.

and one minor correctness fix:

  1. `storage_uri_builder.cpp` calls `std::back_inserter` without including the associated header `<iterator>`. This may compile - a standard library header may transitively include other headers - but is subject to breaking with different Standard Libraries or future versions. Including `<iterator>` is the obvious fix.
  
I've added no test coverage since these are modifications to existing behavior and not additions of behavior.

### DIFF
--- a/sdk/core/azure-core/inc/context.hpp
+++ b/sdk/core/azure-core/inc/context.hpp
@@ -173,13 +173,13 @@ namespace Azure { namespace Core {
 
     Context& operator=(const Context&) = default;
 
-    Context WithDeadline(time_point cancelWhen)
+    Context WithDeadline(time_point cancelWhen) const
     {
       return Context{std::make_shared<ContextSharedState>(
           m_contextSharedState, cancelWhen, std::string(), ContextValue{})};
     }
 
-    Context WithValue(const std::string& key, ContextValue&& value)
+    Context WithValue(const std::string& key, ContextValue&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
           m_contextSharedState, time_point::max(), key, std::move(value))};
@@ -187,7 +187,7 @@ namespace Azure { namespace Core {
 
     time_point CancelWhen() const;
 
-    const ContextValue& operator[](const std::string& key)
+    const ContextValue& operator[](const std::string& key) const
     {
       if (!key.empty())
       {
@@ -204,7 +204,7 @@ namespace Azure { namespace Core {
       return empty;
     }
 
-    bool HasKey(const std::string& key)
+    bool HasKey(const std::string& key) const
     {
       if (!key.empty())
       {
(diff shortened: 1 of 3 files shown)


sdk/core/azure-core/inc/context.hpp
```diff
@@ -145,20 +146,23 @@ namespace Azure { namespace Core {
     struct ContextSharedState
     {
       std::shared_ptr<ContextSharedState> Parent;
-      time_point CancelAt;
+      time_point CancelAt; // access guarded by Mtx
       std::string Key;
       ContextValue Value;
+      mutable std::mutex Mtx;
```
CaseyCarter, line 152: This abbreviated name is inconsistent with the naming guidance in the Guidelines; `Mutex` would be better,

RickWinter: Name it Mutex, 

The mutex should be private to the SharedState object.

CaseyCarter: It doesn't make sense for a mutex that guards accesses to a data member to have differing accessibility than that data member. Everyone who accesses `CancelAt` must first lock `Mutex`, so it doesn't make sense to allow access to `CancelAt` but not `Mutex`. I'll make them both private.

sdk/core/azure-core/inc/context.hpp
```diff
@@ -219,7 +232,10 @@ namespace Azure { namespace Core {
       return false;
     }
 
-    void Cancel() { m_contextSharedState->CancelAt = time_point::min(); }
+    void Cancel()
+    {
+      m_contextSharedState->CancelNow();
+    }
```
CaseyCarter, line 238: Style: this should revert to the original one-liner style.

CaseyCarter:
From the merge conflicts, this was largely superseded by #378; I'll drop this PR.

PR #402, opened 2020-08-05
[Storage Blob Service] Container Lease and Blob Lease

JinmingHu-MSFT:


sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -284,6 +284,76 @@ namespace Azure { namespace Storage { namespace Blobs {
     Azure::Core::Response<UndeleteBlobInfo> Undelete(
         const UndeleteBlobOptions& options = UndeleteBlobOptions()) const;
 
+    /**
+     * @brief Acquires a lease on the blob.
+     *
+     * @param proposedLeaseId
+     * Proposed lease ID, in a GUID string format.
```
katmsft, line 291: The format of the comment here is really weird.

JinmingHu-MSFT: It is weird

katmsft: Can you try to fix it?

JinmingHu-MSFT: not a priority

PR #416, opened 2020-08-07
[Storage Blobs Service] Blob Tags

Jinming-Hu:


sdk/storage/test/blobs/append_blob_client_test.cpp
```diff
@@ -88,4 +88,64 @@ namespace Azure { namespace Storage { namespace Test {
...
+    tags[c3] = v3;
+
+    auto downloadedTags = *appendBlobClient.GetTags();
+    EXPECT_TRUE(downloadedTags.empty());
+    appendBlobClient.SetTags(tags);
+    downloadedTags = *appendBlobClient.GetTags();
+    EXPECT_EQ(downloadedTags, tags);
+
+    /*
+    FilterBlobs doens't work for now because HTTP Client provided by Core SDK doesn't encode query
```
JinmingHu-MSFT, line 116: https://github.com/Azure/azure-sdk-for-cpp/issues/417

sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -357,13 +357,36 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+        const SetBlobTagsOptions& options = SetBlobTagsOptions()) const;
+
+    /**
+     * @brief Gets the tags associated with the underlying blob.
+     * 
+     * @param
+     * options Optional parameters to execute this function.
+     * @return Tags on successfully
+     * getting tags.
+     */
```
katmsft, line 386: Please fix the format.
To be more specific, any comment should not be split into multiple lines before the collum limit is hit. Also, each new line should be approximately aligned in all multiple-line comments.

sdk/storage/inc/blobs/blob_options.hpp
```diff
@@ -20,12 +20,24 @@ namespace Azure { namespace Storage { namespace Blobs {
...
 
+  /**
+   * @brief Specifies HTTP options for conditioanal requests based on tags.
+   */
+  struct TagAccessConditions
+  {
+    /**
+     * @brief Optional SQL statement to apply to the tags of the Blob.
+     */
+    std::string TagConditions;
```
katmsft, line 31: Leave a sample or a reference link here please.

JinmingHu-MSFT:
blocked by https://github.com/Azure/azure-sdk-for-cpp/issues/407

sdk/storage/inc/blobs/blob_options.hpp
Jinming-Hu: fixed

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -20,12 +20,26 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+   * @brief Specifies HTTP options for conditioanal requests based on tags.
+   */
+  struct TagAccessConditions
+  {
+    /**
+     * @brief Optional SQL statement to apply to the tags of the Blob. Refer to
+     * https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations#tags-predicate-syntax
+     * for the format of SQL statements.
+     */
+    std::string TagConditions;
```
katmsft, line 33: Please create a work item to do some convenience enhancement to provide functionality for customers to construct the tag condition query string.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -993,7 +1058,9 @@ namespace Azure { namespace Storage { namespace Blobs {
     /**
      * @brief Optional conditions that the source must meet to perform this operation.
      */
-    BlobAccessConditions SourceConditions;
+    struct : public LastModifiedTimeAccessConditions, public ETagAccessConditions
```
katmsft, line 1061: I don't know if the unnamed struct pattern is a good idea for something that might be reused. Can we bring this to review before we decide that we can write code in this style?

katmsft (approved):
Approve but there are some pending discussion/items. Also, I think it is important to have a sample for tag's usage as it is not obvious, but it can be added later.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
Jinming-Hu: language alignment is one of our most sought in Track2. Is there any source that suggests they're going to do so in any other language?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
Jinming-Hu: We defined separate structs for those that might be reused for example `BlobAccessConditions `. Access conditions for this API is very special. None of any other API has the same set of fields as it. So in this case I think it's fine to use unnamed struct. If in the future, we had a new API that share the same fields, we could define a new struct, it's not even a breaking change.

PR #425, opened 2020-08-10
[Storage Blobs Service] Add test cases for LastModifiedTime and ETag and lease id

JinmingHu-MSFT:


sdk/storage/src/blobs/blob_container_client.cpp
```diff
@@ -174,6 +174,11 @@ namespace Azure { namespace Storage { namespace Blobs {
     protocolLayerOptions.Metadata = metadata;
     protocolLayerOptions.LeaseId = options.AccessConditions.LeaseId;
     protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
+    if (options.AccessConditions.IfUnmodifiedSince.HasValue())
+    {
```
katmsft, line 178: I thought we don't want to throw exceptions on pre-condition check?
https://azure.github.io/azure-sdk/cpp_design.html#pre-conditions

JinmingHu-MSFT: I don't think this counts as precondition

JinmingHu-MSFT: This is actually server side loss-of-function. All other operations either support both `If-Modified-Since` and `If-Unmodified-Since` or neither. But this `SetMetadata` is very strange that it only supports one of them. So it's perfectly reasonable that some time in the future, this API will support both.

This is very different from the cases in the definition of precondition

> for example by passing an out-of-range value or a null pointer

katmsft: It is also violating the rule of 'Do not check server behavior'. If we don't add this check here and sends if-unmodified-since, what does server return? If the server returns a clear indication of the problem, we can remove the check here.

JinmingHu-MSFT: Server silently ignores it, which is not good.

JinmingHu-MSFT: If the server rejected it and returned 4xx, I wouldn't throw this exception here.

katmsft: > Server silently ignores it, which is not good.

Please leave a comment clarifying the server behavior and the fact this should be removed when the service finally decided to align its behavior or return 4XX.

JinmingHu-MSFT: done

PR #427, opened 2020-08-10
[Storage Blobs Service] Customer-Provided Key (CPK-V) and Encryption Scope (CPK-N)

JinmingHu-MSFT:


JinmingHu-MSFT:
Hi  @JeffreyRichter , these two features involve some major API changes. I'd like to have your approval before merging this PR into master branch.

API Changes are:
1. In both `BlobClient` and `BlobContainerClient`, two members variables are added. see [here](https://github.com/Azure/azure-sdk-for-cpp/blob/b14ac6fa53e8c561063b354b89e3a476a1ca35bc/sdk/storage/inc/blobs/blob_client.hpp#L370). This aligns with .Net SDK.

2. In both `BlobClientOptions` and `BlobContainerClinetOpionts`, which are used to initialize a `BlobClient` and `BlobContainerClient` respectively, two member variables are added. see [here](https://github.com/Azure/azure-sdk-for-cpp/blob/b14ac6fa53e8c561063b354b89e3a476a1ca35bc/sdk/storage/inc/blobs/blob_options.hpp#L226). This also aligns with .Net SDK.

sdk/storage/src/blobs/append_blob_client.cpp
```diff
@@ -83,6 +83,13 @@ namespace Azure { namespace Storage { namespace Blobs {
     protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
     protocolLayerOptions.IfMatch = options.AccessConditions.IfMatch;
     protocolLayerOptions.IfNoneMatch = options.AccessConditions.IfNoneMatch;
+    if (m_customerProvidedKey.HasValue())
+    {
+      protocolLayerOptions.EncryptionKey = m_customerProvidedKey.GetValue().Key;
+      protocolLayerOptions.EncryptionKeySha256 = m_customerProvidedKey.GetValue().KeyHash;
```
katmsft, line 89: Was the key hash created with hashing `EncryptionKey`? If so, I suppose we should not ask customer to input both right?

sdk/storage/src/common/crypt.cpp
```diff
@@ -150,6 +214,16 @@ namespace Azure { namespace Storage {
 
 #else
 
+  std::string Sha256(const std::string& text)
```
katmsft, line 217: I strongly recommend self-written hashing/encryption algorithms, although leveraging other dependencies, to implement some predefined input/output UT for validation purposes. Then if a maintainer changes any code that may have an impact on the output, it will raise more alarms.

sdk/storage/src/blobs/append_blob_client.cpp
JinmingHu-MSFT: I had the same thought as you. But current impl aligns with other SDK.

sdk/storage/src/common/crypt.cpp
JinmingHu-MSFT: what do you mean `self-written hashing/encryption algorithms`? can you clarify?

sdk/storage/inc/blobs/blob_client.hpp
```diff
@@ -367,12 +367,18 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    Azure::Core::Nullable<EncryptionKey> m_customerProvidedKey;
+    Azure::Core::Nullable<std::string> m_encryptionScope;
 
   private:
     explicit BlobClient(
         UriBuilder blobUri,
-        std::shared_ptr<Azure::Core::Http::HttpPipeline> pipeline)
-        : m_blobUrl(std::move(blobUri)), m_pipeline(std::move(pipeline))
+        std::shared_ptr<Azure::Core::Http::HttpPipeline> pipeline,
+        Azure::Core::Nullable<EncryptionKey> customerProvidedKey,
```
JeffreyRichter, line 377: Should these just be in the options struct and not another ctor overload?

sdk/storage/inc/blobs/blob_options.hpp
```diff
@@ -197,6 +219,16 @@ namespace Azure { namespace Storage { namespace Blobs {
      * are applied to every retrial.
      */
     std::vector<std::unique_ptr<Azure::Core::Http::HttpPolicy>> PerRetryPolicies;
+
+    /**
+     * @brief Sets the customer provided key used when making requests.
```
JeffreyRichter, line 224: Sets or "holds"? This is a field so it doesn't set anything; it holds what should be used, right?

sdk/storage/inc/blobs/blob_options.hpp
```diff
@@ -197,6 +219,16 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+     */
+    Azure::Core::Nullable<EncryptionKey> CustomerProvidedKey;
+
+    /**
+     * @brief Sets the encryption scope used when making requests.
```
JeffreyRichter, line 229: Same comment

sdk/storage/inc/blobs/blob_options.hpp
```diff
@@ -219,6 +251,17 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @brief Name-value pairs to associate with the container as metadata.
      */
     std::map<std::string, std::string> Metadata;
+
+    /**
+     * @brief The encryption scope to set as the default on the container.
```
JeffreyRichter, line 256: set --> use?

sdk/storage/inc/blobs/blob_options.hpp
```diff
@@ -430,6 +473,16 @@ namespace Azure { namespace Storage { namespace Blobs {
      * are applied to every retrial.
      */
     std::vector<std::unique_ptr<Azure::Core::Http::HttpPolicy>> PerRetryPolicies;
+
+    /**
+     * @brief Sets the customer provided key used when making requests.
```
JeffreyRichter, line 478: Sets?

sdk/storage/inc/blobs/blob_options.hpp
```diff
@@ -430,6 +473,16 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+     */
+    Azure::Core::Nullable<EncryptionKey> CustomerProvidedKey;
+
+    /**
+     * @brief Sets the encryption scope used when making requests.
```
JeffreyRichter, line 483: Sets?

sdk/storage/src/common/crypt.cpp
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/430

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: This is a private constructor only for `BlobContainerClient::GetBlobClient()`, for internal use only. Customers cannot directly use it.

We cannot have an options struct as a parameter for this constructor because in `BlobContainerClient`, we have `std::shared_ptr<Azure::Core::Http::HttpPipeline>` which contains all pipeline policies. We cannot separate them into per-operation policies and per-retry policies, thus cannot construct a `BlobClientOptions` struct.

sdk/storage/inc/blobs/blob_options.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_options.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_options.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_options.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_options.hpp
JinmingHu-MSFT: fixed

sdk/storage/inc/blobs/blob_client.hpp
JinmingHu-MSFT: I'm going to proceed merging this PR, but we can still keep discussing about this.

PR #436, opened 2020-08-11
Rename UUID -> Uuid

RickWinter:
Update header references


sdk/core/azure-core/inc/uuid.hpp
```diff
@@ -3,34 +3,35 @@
 
 #pragma once
 
+#include <memory>
```
antkmsft, line 6: For std::memcpy, it is surprisingly `<cstring>` that you need to include: http://www.cplusplus.com/reference/cstring/memcpy/

PR #440, opened 2020-08-12
[Storage Common] Replace old RFC1123 impl which cannot be built until GCC 5

JinmingHu-MSFT:
we need to support GCC 4.8 https://azure.github.io/azure-sdk/cpp_implementation.html#linux

### DIFF
--- a/sdk/storage/src/common/common_headers_request_policy.cpp
+++ b/sdk/storage/src/common/common_headers_request_policy.cpp
@@ -4,8 +4,6 @@
 #include "common/common_headers_request_policy.hpp"
 
 #include <ctime>
-#include <iomanip>
-#include <sstream>
 
 namespace Azure { namespace Storage {
 
@@ -29,10 +27,16 @@ namespace Azure { namespace Storage {
 #else
       gmtime_r(&t, &ct);
 #endif
-      std::stringstream dateString;
-      dateString.imbue(std::locale("C"));
-      dateString << std::put_time(&ct, "%a, %d %b %Y %H:%M:%S GMT");
-      request.AddHeader(c_HttpHeaderXMsDate, dateString.str());
+      static const char* weekdays[] = {"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
+      static const char* months[]
+          = {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
+      std::string rfc1123Format = "%a, %d %b %Y %H:%M:%S GMT";
+      rfc1123Format.replace(rfc1123Format.find("%a"), 2, weekdays[ct.tm_wday]);
+      rfc1123Format.replace(rfc1123Format.find("%b"), 2, months[ct.tm_mon]);
+      char datetimeStr[32];
+      std::strftime(datetimeStr, sizeof(datetimeStr), rfc1123Format.data(), &ct);
+
+      request.AddHeader(c_HttpHeaderXMsDate, datetimeStr);
     }
 
     return nextHttpPolicy.Send(ctx, request);


JinmingHu-MSFT:
Hi @JeffreyRichter about this [CommonHeadersRequestPolicy](https://github.com/Azure/azure-sdk-for-cpp/blob/471970f8b6dd120d40f91ab97e98715622c51eb1/sdk/storage/src/common/common_headers_request_policy.cpp#L10), originally @katmsft and I thought we might have some headers that apply to all requests of all services. But now I find that we don't have any other such headers except for `x-ms-date`. So I want to rename it to something like `DatePolicy` or `DateHeaderPolicy`. What do you think?

antkmsft:
We are going to need this too at some point, see https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/src/http/retry_policy.cpp#L45

JinmingHu-MSFT:
> We are going to need this too at some point, see https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/src/http/retry_policy.cpp#L45

So why not Core team write a `DateTime` class that provides conversion from RFC-1123 and ISO8601. Both Core and storage team can benefit from it.

JeffreyRichter:
> Hi @JeffreyRichter about this [CommonHeadersRequestPolicy](https://github.com/Azure/azure-sdk-for-cpp/blob/471970f8b6dd120d40f91ab97e98715622c51eb1/sdk/storage/src/common/common_headers_request_policy.cpp#L10), originally @katmsft and I thought we might have some headers that apply to all requests of all services. But now I find that we don't have any other such headers except for `x-ms-date`. So I want to rename it to something like `DatePolicy` or `DateHeaderPolicy` or `DataTimePolicy`. What do you think?

First, I don't think you need to support x-ms-date at all; instead just use the standard HTTP date header. And the policy should set this and so it doesn't have to be exposed to customers at all.

And the Core team will create a DateTime class (copied from CPP REST with some renamed functions) which can be used.

RickWinter:
@JeffreyRichter Does storage respect the http Date header?  From my past experience with the API there were certain APIs which had different behaviors based on if you did(n't) send the x-ms-date header.

JeffreyRichter:
I seem to remember someone (maybe it was you) saying this to me before. Is it that x-ms-date supports higher resolution or some thing? Whichever header we decide to use, it should not be exposed to a customer.

JinmingHu-MSFT:
> I seem to remember someone (maybe it was you) saying this to me before. Is it that x-ms-date supports higher resolution or some thing? Whichever header we decide to use, it should not be exposed to a customer.

This policy that sets `x-ms-date` header isn't exposed to customers now, it's for internal use only. And i'm looking for suggestions about the name, I think `CommonHeadersRequestPolicy` is not descriptive.

JeffreyRichter:
I guess I don't understand why there is a name at all. This looks fine to me:  request.AddHeader(c_HttpHeaderXMsDate, datetimeStr);

JinmingHu-MSFT:
> I guess I don't understand why there is a name at all. This looks fine to me: request.AddHeader(c_HttpHeaderXMsDate, datetimeStr);

We have to update the date on every retry, so it has to be a per-retry policy

JeffreyRichter:
Does the policy only update the x-ms-date header or does it do more than this?

JeffreyRichter:
Since this policy is Storage specific and might do all kinds of stuff over time, I suggest you just call it the StoragePerTryPolicy. Then, over time you can shove more stuff (headers or query parameters) into to without adding additional policies to the pipeline.

JinmingHu-MSFT:
> Since this policy is Storage specific and might do all kinds of stuff over time, I suggest you just call it the StoragePerTryPolicy. Then, over time you can shove more stuff (headers or query parameters) into to without adding additional policies to the pipeline.

Sure.

PR #442, opened 2020-08-12
EnvironmentCredential with ClientSecretCredential support

### LINKED ISSUE #79: Implement Environment Credential
https://github.com/Azure/azure-sdk-for-cpp/issues/5

antkmsft:
Closes #79

### DIFF
--- a/sdk/core/azure-core/inc/credentials/credentials.hpp
+++ b/sdk/core/azure-core/inc/credentials/credentials.hpp
@@ -3,8 +3,10 @@
 
 #pragma once
 
-#include <chrono>
 #include <context.hpp>
+
+#include <chrono>
+#include <memory>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -36,10 +38,10 @@ namespace Azure { namespace Core { namespace Credentials {
   private:
     static std::string const g_aadGlobalAuthority;
 
-    std::string const m_tenantId;
-    std::string const m_clientId;
-    std::string const m_clientSecret;
-    std::string const m_authority;
+    std::string m_tenantId;
+    std::string m_clientId;
+    std::string m_clientSecret;
+    std::string m_authority;
 
   public:
     explicit ClientSecretCredential(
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/credentials/credentials.cpp
```diff
@@ -196,3 +198,70 @@ AccessToken Azure::Core::Credentials::ClientSecretCredential::GetToken(
     throw AuthenticationException("unknown error");
   }
 }
+
+Azure::Core::Credentials::EnvironmentCredential::EnvironmentCredential()
```
RickWinter, line 202: Should these be in Azure::Core or Azure::Identity?

antkmsft: Parts of this *will* become Identity - https://github.com/Azure/azure-sdk-for-cpp/issues/428 I want to do this in a separate PR.

antkmsft: Not sure about this specific exception going to Identity or remaining in Core - I'll need to think about it (the very basics DO live in Core in other SDKs too; For example, abstract class TokenCredential lives in Core, while ClientSecretCredential lives in Identity).

sdk/core/azure-core/inc/credentials/credentials.hpp
```diff
@@ -36,10 +39,10 @@ namespace Azure { namespace Core { namespace Credentials {
   private:
     static std::string const g_aadGlobalAuthority;
 
-    std::string const m_tenantId;
-    std::string const m_clientId;
-    std::string const m_clientSecret;
-    std::string const m_authority;
+    std::string m_tenantId;
```
RickWinter, line 42: Why do we want to modify these?  I'd expect you initialize in the constructor and never change them after.  If you wanted a credential with a different value then you make a new credential

antkmsft: https://azure.github.io/azure-sdk/cpp_design.html#const-and-reference-members + Billy doesn't recommend these due to unexpected UB described in CWG1116 and CWG1776.

RickWinter: I've added an agenda item to Tuesday's meeting for const on non-union members.  We can add this back after that discussion.

antkmsft:
@RickWinter Thank you for the approval. As I said, if we come to another conclusion regarding const, I will create another PR to update this place and some others.

PR #449, opened 2020-08-13
Rename return types and option types

JinmingHu-MSFT:
disabled datalake build


sdk/storage/inc/blobs/block_blob_client.hpp
```diff
@@ -124,7 +124,7 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @param options Optional parameters to execute this function.
      * @return A BlobContentInfo describing the state of the updated block blob.
```
katmsft, line 125: BlobContentInfo changed to UploadBlockBlobResult

katmsft: Ditto for other comments.

JinmingHu-MSFT: fixed

PR #450, opened 2020-08-13
Added support for share service

katmsft:


sdk/storage/inc/shares/protocol/share_rest_client.hpp
```diff
@@ -2557,18 +2558,41 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
         {
           // Success
           ShareGetPropertiesResponse result;
-          result.Metadata = response.GetHeaders().at(Details::c_HeaderMetadata);
+
+          for (auto i = response.GetHeaders().lower_bound(Details::c_HeaderMetadata);
```
JinmingHu-MSFT, line 2562: `c_HeaderMetadata` should be defined as `x-ms-meta-`

sdk/storage/test/shares/share_client_test.cpp
```diff
@@ -54,4 +66,120 @@ namespace Azure { namespace Storage { namespace Test {
...
+      EXPECT_EQ(metadata2, result);
+    }
+  }
+
+  TEST_F(FileShareClientTest, ShareQuota)
+  {
+    const int32_t quota32GB = 32;
+    const int32_t quota64GB = 64;
+    const int32_t quota5120GB = 5120;
+    const int32_t quotaInvalid = 5121;
```
JinmingHu-MSFT, line 108: if max quota is changed in the future, you'll have to fix this test case.

JinmingHu-MSFT: and you're testing server behavior

sdk/storage/inc/shares/protocol/share_rest_client.hpp
katmsft: Swagger uses `"x-ms-meta"` as the name of metadata. I think it's better to align.

sdk/storage/test/shares/share_client_test.cpp
katmsft: Sure, I'll remove this.

PR #456, opened 2020-08-14
[Storage Blobs Service] Transactional MD5 and CRC64

### LINKED ISSUE #295: [Storage Blob Service] Transactional Checksum Support in storage SDK
MD5 and CRC64

JinmingHu-MSFT:
closes #295


JinmingHu-MSFT:
closes #295

sdk/storage/inc/common/crypt.hpp
```diff
@@ -3,13 +3,65 @@
...
+  class Md5 {
+  public:
+    Md5();
+    ~Md5();
+
+    void Update(const uint8_t* data, std::size_t length);
+
+    std::string Digest() const;
+
+    static std::string Hash(const uint8_t* data, std::size_t length)
```
katmsft, line 23: A one-liner saying that this API returns Base64 encoded string would be great.

sdk/storage/src/common/crypt.cpp
```diff
@@ -82,96 +97,164 @@ namespace Azure { namespace Storage {
...
   {
+    std::string buffer;
+    BCRYPT_HASH_HANDLE hashHandle = nullptr;
+    std::size_t hashLength = 0;
+  };
 
-    static AlgorithmProviderInstance AlgorithmProvider(AlgorithmType::Hmac_Sha256);
+  Md5::Md5()
+  {
+    static Details::AlgorithmProviderInstance AlgorithmProvider(Details::AlgorithmType::Md5);
```
katmsft, line 201: Why pascal case?

katmsft: Ditto for other local variables. I see snake case/pascal case/camel cases in different areas.

sdk/storage/inc/common/crypt.hpp
JinmingHu-MSFT: It returns binary

sdk/storage/src/common/crypt.cpp
```diff
@@ -275,4 +389,823 @@ namespace Azure { namespace Storage {
 
 #endif
 
+  static constexpr uint64_t poly = 0x9A6C9329AC4BC9B5ULL;
```
katmsft, line 392: Good thing is that track 1 proved the below code's integrity. Otherwise, I don't know how to maintain it.

JinmingHu-MSFT: abstruse mathematical principles behind it. Don't try to comprehend it. just use it.

sdk/storage/src/common/crypt.cpp
JinmingHu-MSFT: fixed

PR #462, opened 2020-08-17
[Storage Blobs Service] Undelete Container

JinmingHu-MSFT:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/298

### DIFF
--- a/sdk/storage/inc/blobs/blob_container_client.hpp
+++ b/sdk/storage/inc/blobs/blob_container_client.hpp
@@ -149,6 +149,19 @@ namespace Azure { namespace Storage { namespace Blobs {
         const DeleteContainerOptions& options = DeleteContainerOptions()) const;
 
     /**
+     * @brief Restores a previously deleted container.
+     *
+     * @param deletedContainerName The name of the previously deleted container.
+     * @param deletedContainerVersion The version of the previously deleted container.
+     * @param options Optional parameters to execute this function.
+     * @return An UndeleteContainerResult if successful.
+     */
+    Azure::Core::Response<UndeleteContainerResult> UndeleteContainer(
+        const std::string& deletedContainerName,
+        const std::string& deletedContainerVersion,
+        const UndeleteContainerOptions& options = UndeleteContainerOptions()) const;
+
+    /**
      * @brief Returns all user-defined metadata and system properties for the specified
      * container. The data returned does not include the container's list of blobs.
      *
--- a/sdk/storage/inc/blobs/blob_options.hpp
+++ b/sdk/storage/inc/blobs/blob_options.hpp
@@ -123,7 +123,7 @@ namespace Azure { namespace Storage { namespace Blobs {
     /**
      * @brief Specifies that the container's metadata be returned.
      */
-    ListBlobContainersIncludeOption Include = ListBlobContainersIncludeOption::None;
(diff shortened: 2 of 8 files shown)


JinmingHu-MSFT:
https://github.com/Azure/azure-sdk-for-cpp/issues/463

JinmingHu-MSFT:
@katmsft Maybe I should move this function to service client. .Net SDK puts it in service client.

katmsft:
> @katmsft Maybe I should move this function to service client. .Net SDK puts it in service client.
> 
> But .Net sdk also has `CreateContainer` and `DeleteContainer` in service client, neither of which we have in our service client.

I think it's a convenient feature, but not essential. If we have this in service client, it's difficult to argue that we should not have create page/append blob in container client or delete blob in container client.

JinmingHu-MSFT:
> > @katmsft Maybe I should move this function to service client. .Net SDK puts it in service client.
> > But .Net sdk also has `CreateContainer` and `DeleteContainer` in service client, neither of which we have in our service client.
> 
> I think it's a convenient feature, but not essential. If we have this in service client, it's difficult to argue that we should not have create page/append blob in container client or delete blob in container client.

The problem is, they don't have `Undelete` in container client.

katmsft:
> > > @katmsft Maybe I should move this function to service client. .Net SDK puts it in service client.
> > > But .Net sdk also has `CreateContainer` and `DeleteContainer` in service client, neither of which we have in our service client.
> > 
> > 
> > I think it's a convenient feature, but not essential. If we have this in service client, it's difficult to argue that we should not have create page/append blob in container client or delete blob in container client.
> 
> The problem is, they don't have `Undelete` in container client.

Wow, then it is a bigger design decision than I thought, maybe we should bring this up.

JeffreyRichter:
> Hi @JeffreyRichter , can you share you opinion about where this `UndeleteContainer` should be, container client or service client? Thanks.

I don't see an REST API for undelete container; only undelete blob (https://docs.microsoft.com/en-us/rest/api/storageservices/undelete-blob). The methods should be on the client with the matching URL. Since undelete blob requires a URL to a blob, the method should be on the BlobClient. 

The reason why .NET has ALSO put methods on the parent Client is because of the difference in pattern between REST and OOP. In a REST world, you get the URL first and then you can create the resource at that URL via an HTTP PUT. For OOP people, this doesn't feel right. For OOP, when you construct something (like a BlobClient), it feels like the blob (not just the URL) should also be created at that time. So, .NET repeated some methods on the parent Client to be more OOP.

While C++ is an OOP language, I think you can forgo the extra methods (we could always add them later if customers really want them). We have to have the methods on the Client with the corresponding URL. So, I'd put undelete blob on the BlobClient and NOT on the ContainerClient.

JinmingHu-MSFT:
> I don't see an REST API for undelete container; only undelete blob (https://docs.microsoft.com/en-us/rest/api/storageservices/undelete-blob). The methods should be on the client with the matching URL. Since undelete blob requires a URL to a blob, the method should be on the BlobClient.

UndeleteContainer is a new feature. REST API hasn't been updated yet.

JinmingHu-MSFT:
> The methods should be on the client with the matching URL.

Then I'll keep this API in container client.

PR #464, opened 2020-08-17
Rename return types for Datalake and file services.

katmsft:


katmsft:
Also fixes https://github.com/Azure/azure-sdk-for-cpp/issues/461.

sdk/storage/inc/datalake/datalake_responses.hpp
```diff
@@ -10,26 +10,27 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
...
   };
 
-  using FileSystemInfo = FileSystemCreateResponse;
+  using CreateFileSystemResult = FileSystemCreateResult;
+  using SetFileSystemMetadataResult = FileSystemCreateResult;
 
   // PathClient models:
 
-  using PathDeleteInfo = PathDeleteResponse;
+  using DeletePathResult = PathDeleteResult;
```
JinmingHu-MSFT, line 33: what? cannot you just name in `DeletePathResult` in protocol layer?

sdk/storage/inc/shares/share_responses.hpp
```diff
@@ -9,19 +9,21 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+  using DeleteShareResult = ShareDeleteResult;
+  using CreateShareSnapshotResult = ShareCreateSnapshotResult;
+  using GetSharePropertiesResult = ShareGetPropertiesResult;
+  using SetShareQuotaResult = ShareSetQuotaResult;
+  using SetShareMetadataResult = ShareSetMetadataResult;
+  using SetShareAccessPolicyResult = ShareSetAccessPolicyResult;
+  using GetShareStatisticsResult = ShareGetStatisticsResult;
+  using CreateSharePermissionResult = ShareCreatePermissionResult;
+  using GetShareAccessPolicyResult = ShareGetAccessPolicyResult;
+  using GetSharePermissionResult = ShareGetPermissionResult;
```
JinmingHu-MSFT, line 27: I feel awful about these.

sdk/storage/inc/datalake/datalake_responses.hpp
katmsft: I cannot.
E.g. for `ShareCreateSnapshotResult`, it comes from 3 parts: `Share` is the service name, `CreateSnapshot` is what swagger defines as the method's name and `Result` is appended by the generator.
If I changed the protocol layer to `DeletePathResult`, then `ShareCreateSnapshotResult` will be `CreateSnapshotShareResult`, which is way worse than the current name.

sdk/storage/inc/shares/share_responses.hpp
katmsft: See the other comments.

sdk/storage/inc/datalake/datalake_responses.hpp
JinmingHu-MSFT: @JeffreyRichter said this naming rule only applies to convenience layer. So I guess we can name protocol layer variables whatever we like.

PR #486, opened 2020-08-19
[Storage Blob Service] Concurrent upload for Page Blob

JinmingHu-MSFT:


sdk/storage/src/blobs/page_blob_client.cpp
```diff
@@ -175,9 +177,97 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    {
+      createOptions.EncryptionKey = m_customerProvidedKey.GetValue().Key;
+      createOptions.EncryptionKeySha256 = m_customerProvidedKey.GetValue().KeyHash;
+      createOptions.EncryptionAlgorithm = m_customerProvidedKey.GetValue().Algorithm;
+    }
+    createOptions.EncryptionScope = m_encryptionScope;
+    auto createResult = BlobRestClient::PageBlob::Create(
+        options.Context, *m_pipeline, m_blobUrl.ToString(), createOptions);
+
+    constexpr int64_t c_defaultChunkSize = 8 * 1024 * 1024;
```
katmsft, line 200: I think the default page size is 4MB in track 1 SDK. Can we make sure 8MB is the best value?

katmsft (approved):

sdk/storage/src/blobs/page_blob_client.cpp
JinmingHu-MSFT: i don't know, I didn't say 8MB is the best. But we have to have a default value anyway.

JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/497

PR #510, opened 2020-08-23
Added support for directory/file client for file service.

katmsft:


sdk/storage/inc/shares/share_file_attribute.hpp
```diff
@@ -0,0 +1,211 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
JinmingHu-MSFT, line 1: Why not define these in protocol layer?

sdk/storage/inc/common/constants.hpp
```diff
@@ -20,4 +20,7 @@ namespace Azure { namespace Storage { namespace Details {
   constexpr static const char* c_defaultSasVersion = "2019-12-12";
 
   constexpr int c_reliableStreamRetryCount = 3;
+  constexpr int64_t c_FileUploadDefaultChunkSize = 4 * 1024 * 1024;
+  constexpr int64_t c_DownloadDefaultChunkSize = 4 * 1024 * 1024;
```
JinmingHu-MSFT, line 24: why do we define these variables here? Are they going to be used at more than once place?

sdk/storage/inc/shares/protocol/share_rest_client.hpp
```diff
@@ -832,7 +833,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
   {
     std::string ETag;
     std::string LastModified;
-    bool IsServerEncrypted = bool();
+    bool RequestIsServerEncrypted = bool();
```
JinmingHu-MSFT, line 836: This variable name is kind of weird. .Net it's named `IsServerEncrypted`. In js, it's named `isServerEncrypted`. In Python, it's named `server_encrypted`.

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include <string>
+
+namespace Azure { namespace Storage { namespace Files { namespace Shares {
+
+  class FileClient;
+
+  class DirectoryClient {
+  public:
+    /**
+     * @brief Create from connection string
```
JinmingHu-MSFT, line 26: the doc is way too brief. It doesn't provide any useful information.

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+     * @param shareName The name of a file share.
+     * @param directoryPath The path of a directory.
+     * @param options Optional parameters used to initialize the client.
+     * @return ShareClient The client that can be used to manage a share resource.
+     */
+    static DirectoryClient CreateFromConnectionString(
+        const std::string& connectionString,
+        const std::string& shareName,
+        const std::string& directoryPath,
+        const ShareClientOptions& options = ShareClientOptions());
```
JinmingHu-MSFT, line 37:
```suggestion
        const DirectoryClientOptions& options = ShareClientOptions());
```

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * @brief Shared key authentication client.
+     * @param shareDirectoryUri The URI of the directory this client's request targets.
+     * @param credential The shared key credential used to initialize the client.
+     * @param options Optional parameters used to initialize the client.
+     */
+    explicit DirectoryClient(
+        const std::string& shareDirectoryUri,
+        std::shared_ptr<SharedKeyCredential> credential,
+        const ShareClientOptions& options = ShareClientOptions());
```
JinmingHu-MSFT, line 48: ditto

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+     * @return ShareClient The client that can be used to manage a share resource.
+     */
+    static DirectoryClient CreateFromConnectionString(
+        const std::string& connectionString,
+        const std::string& shareName,
+        const std::string& directoryPath,
+        const ShareClientOptions& options = ShareClientOptions());
+
+    /**
+     * @brief Shared key authentication client.
```
JinmingHu-MSFT, line 40: too brief

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * @brief Bearer token authentication client.
+     * @param shareDirectoryUri The URI of the directory this client's request targets.
+     * @param credential The token credential used to initialize the client.
+     * @param options Optional parameters used to initialize the client.
+     */
+    explicit DirectoryClient(
+        const std::string& shareDirectoryUri,
+        std::shared_ptr<Core::Credentials::TokenCredential> credential,
+        const ShareClientOptions& options = ShareClientOptions());
```
JinmingHu-MSFT, line 59: ditto

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+        const GetDirectoryPropertiesOptions& options = GetDirectoryPropertiesOptions()) const;
+
+    /**
+     * @brief Sets the properties of the directory.
+     * @param smbProperties The SMB properties to be set to the directory.
+     * @param options Optional parameters to set this directory's properties.
+     * @return Azure::Core::Response<SetDirectoryPropertiesResult> containing the properties of the
+     * directory returned from the server.
+     */
+    Azure::Core::Response<SetDirectoryPropertiesResult> SetProperties(
```
JinmingHu-MSFT, line 127: not align with other sdk

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+    Azure::Core::Response<DeleteDirectoryResult> Delete(
+        const DeleteDirectoryOptions& options = DeleteDirectoryOptions()) const;
+
+    /**
+     * @brief Gets the properties of the directory.
+     * @param options Optional parameters to get this directory's properties.
+     * @return Azure::Core::Response<GetDirectoryPropertiesResult> containing the properties of the
+     * directory returned from the server.
+     */
+    Azure::Core::Response<GetDirectoryPropertiesResult> GetProperties(
```
JinmingHu-MSFT, line 117: not align with other sdk

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+        const std::map<std::string, std::string>& metadata,
+        const SetDirectoryMetadataOptions& options = SetDirectoryMetadataOptions()) const;
+
+    /**
+     * @brief List files and directories under the directory.
+     * @param options Optional parameters to list the files and directories under this directory.
+     * @return Azure::Core::Response<ListFilesAndDirectoriesSegmentedResult> containing the
+     * information of the operation, directory, share and the listed result.
+     */
+    Azure::Core::Response<ListFilesAndDirectoriesSegmentedResult> ListFilesAndDirectoriesSegmented(
```
JinmingHu-MSFT, line 149: not align with other sdk

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * @brief Closes a handle or handles opened on a directory at the service.
+     * @param handleId The ID of the handle to be closed.
+     * @param options Optional parameters to close one of or all this directory's open handles.
+     * @return Azure::Core::Response<ForceCloseDirectoryHandlesResult> containing the information
+     * of the closed handles
+     * @remark This operation may return a marker showing that the operation can be continued.
+     */
+    Azure::Core::Response<ForceCloseDirectoryHandlesResult> ForceCloseHandles(
```
JinmingHu-MSFT, line 171: close Handle or Handle**s**?

sdk/storage/inc/shares/share_directory_client.hpp
```diff
@@ -0,0 +1,189 @@
...
+        const std::string& handleId,
+        const ForceCloseDirectoryHandlesOptions& options
+        = ForceCloseDirectoryHandlesOptions()) const;
+
+  private:
+    UriBuilder m_shareDirectoryUri;
+    std::shared_ptr<Azure::Core::Http::HttpPipeline> m_pipeline;
```
JinmingHu-MSFT, line 178: these two should be protected?

sdk/storage/inc/shares/share_file_client.hpp
```diff
@@ -0,0 +1,357 @@
...
+     * @param shareName The name of a file share.
+     * @param filePath The path of a file.
+     * @param options Optional parameters used to initialize the client.
+     * @return ShareClient The client that can be used to manage a share resource.
+     */
+    static FileClient CreateFromConnectionString(
+        const std::string& connectionString,
+        const std::string& shareName,
+        const std::string& filePath,
+        const ShareClientOptions& options = ShareClientOptions());
```
JinmingHu-MSFT, line 36:
```suggestion
        const FileClientOptions& options = ShareClientOptions());
```

JinmingHu-MSFT: ditto for others

sdk/storage/inc/shares/share_file_client.hpp
```diff
@@ -0,0 +1,357 @@
...
+    Azure::Core::Response<AbortCopyFileResult> AbortCopy(
+        std::string copyId,
+        const AbortCopyFileOptions& options = AbortCopyFileOptions()) const;
+
+    /**
+     * @brief Gets the properties of a file.
+     * @param options Optional parameters to get the properties of this file.
+     * @return Azure::Core::Response<GetFilePropertiesResult> containing the file properties.
+     */
+    Azure::Core::Response<GetFilePropertiesResult> GetProperties(
```
JinmingHu-MSFT, line 193: not align with other sdk

JinmingHu-MSFT: ditto for other APIs

JinmingHu-MSFT: you'd better have a good reason when you decide to do something different from other Track2 sdk

sdk/storage/inc/shares/share_responses.hpp
```diff
@@ -26,4 +27,93 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+     * @brief If specified, the provided file attributes shall be set. Default value:
+     * FileAttribute::Archive for file and FileAttribute::Directory for directory.
+     * FileAttribute::None can also be specified as default.
+     */
+    FileAttributes Attributes = static_cast<FileAttributes>(0);
+
+    /**
+     * @brief Creation time for the file/directory. Default value: Now.
+     */
+    std::string FileCreationTime = std::string("now");
```
JinmingHu-MSFT, line 75: now this should be defined as constant somewhere

sdk/storage/inc/shares/share_responses.hpp
```diff
@@ -26,4 +27,93 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+  using UploadFileRangeResult = FileUploadRangeResult;
+  using ClearFileRangeResult = FileUploadRangeResult;
+  using UploadFileRangeFromUrlResult = FileUploadRangeFromURLResult;
+  using GetFileRangeListResult = FileGetRangeListResult;
+  using ListFileHandlesSegmentedResult = ListDirectoryHandlesSegmentedResult;
+  using ForceCloseFileHandlesResult = FileForceCloseHandlesResult;
+  using AcquireFileLeaseResult = FileAcquireLeaseResult;
+  using ReleaseFileLeaseResult = FileReleaseLeaseResult;
+  using BreakFileLeaseResult = FileBreakLeaseResult;
+  using ChangeFileLeaseResult = FileChangeLeaseResult;
```
JinmingHu-MSFT, line 102: I feel terrible about these meaningless typedefs.

sdk/storage/inc/shares/share_responses.hpp
```diff
@@ -26,4 +27,93 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+
+  struct DownloadFileToResult
+  {
+    std::string ETag;
+    std::string LastModified;
+    int64_t ContentLength = 0;
+    FileShareHttpHeaders HttpHeaders;
+    std::map<std::string, std::string> Metadata;
+    Azure::Core::Nullable<bool> IsServerEncrypted;
```
JinmingHu-MSFT, line 111: So this variable has different names in different result structs.

sdk/storage/src/shares/share_client.cpp
```diff
@@ -217,4 +233,43 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
         m_shareUri.ToString(), *m_pipeline, options.Context, protocolLayerOptions);
   }
 
+  Azure::Core::Response<ListFilesAndDirectoriesSegmentedResult>
+  ShareClient::ListFilesAndDirectoriesSegmented(
```
JinmingHu-MSFT, line 237: not align with other sdk

sdk/storage/src/shares/share_directory_client.cpp
```diff
@@ -0,0 +1,270 @@
...
+    }
+  }
+
+  DirectoryClient::DirectoryClient(
+      const std::string& shareDirectoryUri,
+      std::shared_ptr<SharedKeyCredential> credential,
+      const ShareClientOptions& options)
+      : m_shareDirectoryUri(shareDirectoryUri)
+  {
+
```
JinmingHu-MSFT, line 46: extra new line

sdk/storage/src/shares/share_directory_client.cpp
```diff
@@ -0,0 +1,270 @@
...
+    std::vector<std::unique_ptr<Azure::Core::Http::HttpPolicy>> policies;
+    policies.emplace_back(std::make_unique<Azure::Core::Http::TelemetryPolicy>(
+        Azure::Storage::Details::c_FileServicePackageName, FileServiceVersion));
```
JinmingHu-MSFT, line 49: client request id policy

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,877 @@
...
+    {
+      protocolLayerOptions.FilePermission = options.FilePermission.GetValue();
+    }
+    else if (!options.SmbProperties.FilePermissionKey.empty())
+    {
+      protocolLayerOptions.FilePermissionKey = options.SmbProperties.FilePermissionKey;
+    }
+    else
+    {
+      protocolLayerOptions.FilePermission = "inherit";
```
JinmingHu-MSFT, line 140: this should be defined as constant

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,877 @@
...
+      if (!options.HttpHeaders.GetValue().CacheControl.empty())
+      {
+        protocolLayerOptions.FileCacheControl = options.HttpHeaders.GetValue().CacheControl;
+      }
+      if (!options.HttpHeaders.GetValue().ContentDisposition.empty())
+      {
+        protocolLayerOptions.FileContentDisposition
+            = options.HttpHeaders.GetValue().ContentDisposition;
+      }
+    }
```
JinmingHu-MSFT, line 774: why do you have to check `HasValue()` and `.empty()` here?

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,877 @@
...
+    protocolLayerOptions.FileAttributes = FileAttributesToString(options.SmbProperties.Attributes);
+    if (protocolLayerOptions.FileAttributes.empty())
+    {
+      protocolLayerOptions.FileAttributes = FileAttributesToString(FileAttributes::None);
+    }
+    protocolLayerOptions.FileCreationTime = options.SmbProperties.FileCreationTime;
+    protocolLayerOptions.FileLastWriteTime = options.SmbProperties.FileLastWriteTime;
+    if (options.FilePermission.HasValue())
+    {
+      protocolLayerOptions.FilePermission = options.FilePermission.GetValue();
```
JinmingHu-MSFT, line 741: So this `FilePermission` is nullable in convenience layer while non-nullable in protocol layer?

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,877 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
JinmingHu-MSFT, line 1: some functions in this file are so long

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,877 @@
...
+    }
+    else if (!options.SmbProperties.FilePermissionKey.empty())
+    {
+      protocolLayerOptions.FilePermissionKey = options.SmbProperties.FilePermissionKey;
+    }
+    else
+    {
+      protocolLayerOptions.FilePermission = "inherit";
+    }
+    if (options.HttpHeaders.HasValue())
```
JinmingHu-MSFT, line 751: in the first place. why is `Httpheaders` nullable?

sdk/storage/test/shares/share_file_client_test.hpp
```diff
@@ -0,0 +1,22 @@
...
+  class FileShareFileClientTest : public FileShareDirectoryClientTest {
+  protected:
+    static void SetUpTestSuite();
+    static void TearDownTestSuite();
+
+    static std::shared_ptr<Files::Shares::FileClient> m_fileClient;
+    static std::string m_fileName;
+    static std::vector<uint8_t> m_fileContent;
+
+    static void RandomizeContent();
```
JinmingHu-MSFT, line 19: how is this different from https://github.com/Azure/azure-sdk-for-cpp/blob/70eeec598496de78ba5f6a5ccd7bdb4b4724fd87/sdk/storage/test/test_base.hpp#L57

sdk/storage/test/shares/share_file_client_test.cpp
```diff
@@ -0,0 +1,473 @@
...
+        futures.emplace_back(std::async(std::launch::async, testUploadFromFile, c, l));
+      }
+    }
+    for (auto& f : futures)
+    {
+      f.get();
+    }
+  }
+
+  TEST_F(FileShareFileClientTest, ConcurrentDownload)
```
JinmingHu-MSFT, line 329: i know you copied this from blob test case. But that one is so slow that i was gonna rewrite it.

sdk/storage/test/shares/share_file_client_test.cpp
```diff
@@ -0,0 +1,473 @@
...
+    aLease = *m_fileClient->AcquireLease(CreateUniqueLeaseId());
+    brokenLease = *m_fileClient->BreakLease();
+    EXPECT_FALSE(brokenLease.ETag.empty());
+    EXPECT_FALSE(brokenLease.LastModified.empty());
+    m_fileClient->BreakLease();
+  }
+
+  TEST_F(FileShareFileClientTest, ConcurrentUpload)
+  {
+    std::vector<uint8_t> fileContent = RandomBuffer(static_cast<std::size_t>(8_MB));
```
JinmingHu-MSFT, line 258: i know you copied this from blob test case. But that one is so slow that i was gonna rewrite it.

JinmingHu-MSFT (changes requested):
see in comments

sdk/storage/inc/common/constants.hpp
katmsft: `c_FileUploadDefaultChunkSize` is used in two functions and `c_DownloadDefaultChunkSize` is used in four functions. I think it should be defined in a place for an easier change in the future.

JinmingHu-MSFT: which four functions?

sdk/storage/inc/shares/protocol/share_rest_client.hpp
katmsft: There are two kinds of `IsServerEncrypted`, one with header key `x-ms-server-encrypted` and `x-ms-request-server-encrypted`, they mean different things.

sdk/storage/inc/shares/share_directory_client.hpp
katmsft: It can close multiple handles.

sdk/storage/inc/shares/protocol/share_rest_client.hpp
JinmingHu-MSFT: and what's the difference?

sdk/storage/inc/shares/share_directory_client.hpp
katmsft: Why? I don't think there will should child class for `DirectoryClient`.

sdk/storage/inc/shares/share_directory_client.hpp
JinmingHu-MSFT: my mistake

sdk/storage/inc/shares/share_responses.hpp
katmsft: To align with the naming convention.

sdk/storage/inc/shares/share_responses.hpp
katmsft: Yes, it is by design.

JinmingHu-MSFT: which design?

JinmingHu-MSFT: do we have an agreement or conclusion in some meeting?

sdk/storage/src/shares/share_client.cpp
katmsft: Aligns with JS: https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/storage/storage-file-share/src/ShareDirectoryClient.ts#L1179.

I don't think .NET is using a good name. I would imagine `Get` something returns the content, not the catalog.

JinmingHu-MSFT: list is fine. I didn't mean the function name.

JinmingHu-MSFT: the client

sdk/storage/src/shares/share_file_client.cpp
katmsft: Even if `HttpHeaders` is not null, some of the headers can still be empty, this check avoids sending them as empty headers.

sdk/storage/src/shares/share_file_client.cpp
katmsft: Yes, `FilePermission` and `FilePermissionKey` cannot both be null, but either must be null if the other one is not.

sdk/storage/test/shares/share_file_client_test.hpp
katmsft: The memory is reused.

sdk/storage/test/shares/share_file_client_test.cpp
katmsft: Then I look forward to your rewrites. :)

sdk/storage/inc/shares/share_file_attribute.hpp
katmsft: Swagger does not have this defined.

sdk/storage/inc/common/constants.hpp
katmsft: `c_DownloadDefaultChunkSize ` should be used in Blob's concurrent download and File's concurrent download, unless there is a good reason for them to have different chunk size.

sdk/storage/inc/shares/protocol/share_rest_client.hpp
katmsft: They are genuinely same at the moment, but see this:

>The Create File, Create Directory, Put Range, Set Directory Metadata, Set File Metadata, and Set File Properties operations now return the x-ms-request-server-encrypted response header. This header is set to true if the contents of the request have been successfully encrypted.

> The Get File, Get File Properties, and Get Directory Properties operations now return the x-ms-server-encrypted response header. This header is set to true if the file data and application metadata are completely encrypted. If the file is not encrypted, or if only parts of the file/application metadata are encrypted, this header is set to false.

 https://docs.microsoft.com/en-us/rest/api/storageservices/version-2017-04-17

Service deliberately made this difference for some reason.

sdk/storage/src/shares/share_file_client.cpp
JinmingHu-MSFT: 1. The `HttpHeaders` is non-null. This is a conclusion from the private review meeting with Jeffrey. If you are not okay with this, bring it up and we can revisit this conclusion. I don't think silently ignore a conclusion is a good idea.

2. I don't think these headers should be nullable in protocol layer. But since we don't have any regulation about protocol layer, it's totally up to you how to define them. My suggestion is to make these non-nullable and add this HTTP header if it's not empty. In this way, your convenience layer code will be much simpler. But it's still totally up to you whether or not take this suggestion.

sdk/storage/src/shares/share_file_client.cpp
JinmingHu-MSFT: I think if something is nullable in convenience layer, it should also be nullable in protocol layer.

sdk/storage/test/shares/share_file_client_test.hpp
JinmingHu-MSFT: then how about https://github.com/Azure/azure-sdk-for-cpp/blob/70eeec598496de78ba5f6a5ccd7bdb4b4724fd87/sdk/storage/test/test_base.hpp#L56?

with just one reinterpret_cast

sdk/storage/inc/shares/share_file_attribute.hpp
JinmingHu-MSFT: another defect of swagger.

sdk/storage/inc/shares/protocol/share_rest_client.hpp
JinmingHu-MSFT: I don't think `RequestIsServerEncrypted ` is a good name. so either keep align with other SDK or escalate.

sdk/storage/inc/common/constants.hpp
JinmingHu-MSFT: `c_FileUploadDefaultChunkSize ` is only used by File Service, why is it also in this file?

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,877 @@
...
+  {
+    auto protocolLayerOptions = ShareRestClient::File::DeleteOptions();
+    protocolLayerOptions.LeaseIdOptional = options.AccessConditions.LeaseId;
+    return ShareRestClient::File::Delete(
+        m_shareFileUri.ToString(), *m_pipeline, options.Context, protocolLayerOptions);
+  }
+
+  Azure::Core::Response<DownloadFileResult> FileClient::Download(
+      const DownloadFileOptions& options) const
+  {
```
JinmingHu-MSFT, line 183: reliable stream

JinmingHu-MSFT: so either do it in this PR or create an issue to track, in case we'd forget

sdk/storage/inc/shares/share_responses.hpp
katmsft: By service's design. I can raise it to Jeffrey if you think it's necessary. We can continue the discussion in the other comment raising the same issue.

sdk/storage/inc/common/constants.hpp
katmsft: I will remove it and move it to file-only constants.

sdk/storage/inc/shares/protocol/share_rest_client.hpp
katmsft: How about `IsRequestServerEncrypted`?

sdk/storage/test/shares/share_file_client_test.hpp
katmsft: I'm OK with it.

sdk/storage/inc/shares/share_directory_client.hpp
JinmingHu-MSFT: SetHttpHeaders

sdk/storage/inc/shares/protocol/share_rest_client.hpp
JinmingHu-MSFT: No, I don't see  any semantic difference between them

JinmingHu-MSFT:
there's no `preserve` definition in swagger. This is a defect.

sdk/storage/inc/shares/share_client.hpp
```diff
@@ -179,6 +199,19 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
         const std::string& permissionKey,
         const GetSharePermissionOptions& options = GetSharePermissionOptions()) const;
 
+    /**
```
JinmingHu-MSFT, line 202: This may be going to be removed

sdk/storage/inc/shares/share_file_client.hpp
```diff
@@ -0,0 +1,359 @@
...
+     * @brief Sets the properties of the file.
+     * @param newSizeInByte Byte value of the size the file to resize to. If the specified byte
+     * value is less than the current size of the file, then all ranges above the specified byte
+     * value are cleared.
+     * @param options Optional parameters to resize the file.
+     * @return Azure::Core::Response<ResizeFileResult> containing the properties of the
+     * file returned from the server.
+     */
+    Azure::Core::Response<ResizeFileResult> Resize(
+        int64_t newSizeInByte,
```
JinmingHu-MSFT, line 219: Do we need `InByte` here? Other length variables dont end up with `InBytes`

sdk/storage/inc/shares/share_file_client.hpp
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * @brief Sets the properties of the file.
+     * @param smbProperties The SMB properties to be set to the file.
+     * @param options Optional parameters to set this file's properties.
+     * @return Azure::Core::Response<SetFilePropertiesResult> containing the properties of the
+     * file returned from the server.
+     */
+    Azure::Core::Response<SetFilePropertiesResult> SetProperties(
+        FileShareSmbProperties smbProperties,
```
JinmingHu-MSFT, line 206: feel kind of weird

sdk/storage/inc/shares/share_options.hpp
```diff
@@ -202,4 +221,675 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+    /**
+     * @brief Context for cancelling long running operations.
+     */
+    Azure::Core::Context Context;
+
+    /**
+     * @brief The snapshot parameter is an opaque DateTime value that, when present, specifies the
+     * share snapshot to query.
+     */
+    Azure::Core::Nullable<std::string> ShareSnapshot;
```
JinmingHu-MSFT, line 267: Why is it here? Shouldn't we have `WithSnapshot` for share client, directory client and file client?

JinmingHu-MSFT: ditto for others

sdk/storage/inc/shares/share_options.hpp
```diff
@@ -202,4 +221,675 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+     */
+    Azure::Core::Context Context;
+
+    /**
+     * @brief An MD5 hash of the content. This hash is used to verify the integrity of the data
+     * during transport. When the ContentMD5 parameter is specified, the File service compares the
+     * hash of the content that has arrived with the header value that was sent. If the two hashes
+     * do not match, the operation will fail with error code 400 (Bad Request).
+     */
+    Azure::Core::Nullable<std::string> ContentMD5;
```
JinmingHu-MSFT, line 622:
```suggestion
    Azure::Core::Nullable<std::string> ContentMd5;
```

sdk/storage/src/shares/share_directory_client.cpp
```diff
@@ -0,0 +1,272 @@
...
+    protocolLayerOptions.FileAttributes = FileAttributesToString(options.SmbProperties.Attributes);
+    if (protocolLayerOptions.FileAttributes.empty())
+    {
+      protocolLayerOptions.FileAttributes = FileAttributesToString(FileAttributes::Directory);
+    }
+    protocolLayerOptions.FileCreationTime = options.SmbProperties.FileCreationTime;
+    protocolLayerOptions.FileLastWriteTime = options.SmbProperties.FileLastWriteTime;
+    if (options.DirectoryPermission.HasValue())
+    {
+      protocolLayerOptions.FilePermission = options.DirectoryPermission.GetValue();
```
JinmingHu-MSFT, line 149: still, if in convenience layer one variable is nullable, why isn't it nullable in protocol layer?

sdk/storage/src/shares/share_file_client.cpp
```diff
@@ -0,0 +1,870 @@
...
+    return ShareRestClient::File::StartCopy(
+        m_shareFileUri.ToString(), *m_pipeline, options.Context, protocolLayerOptions);
+  }
+
+  Azure::Core::Response<AbortCopyFileResult> FileClient::AbortCopy(
+      std::string copyId,
+      const AbortCopyFileOptions& options) const
+  {
+    auto protocolLayerOptions = ShareRestClient::File::AbortCopyOptions();
+    protocolLayerOptions.CopyId = copyId;
```
JinmingHu-MSFT, line 241: you pass `copyId` by value, but you didn't `std::move` from it.

JinmingHu-MSFT: ditto for `copySource`

sdk/storage/src/shares/share_file_client.cpp
JinmingHu-MSFT: reference the constant variable

JinmingHu-MSFT (changes requested):
CI failed

sdk/storage/inc/shares/protocol/share_rest_client.hpp
katmsft: I'll ask for Jeffrey's idea.

sdk/storage/inc/shares/share_file_client.hpp
katmsft: Removed.

sdk/storage/test/shares/share_file_client_test.cpp
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/pull/513

sdk/storage/inc/shares/share_options.hpp
katmsft: Align with .NET.

sdk/storage/src/shares/share_file_client.cpp
katmsft: Implemented.

sdk/storage/inc/shares/share_constants.hpp
```diff
@@ -0,0 +1,19 @@
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+namespace Azure { namespace Storage { namespace Files { namespace Shares {
+
+  constexpr static const char* c_FileDefaultTimeValue = "now";
+  constexpr static const char* c_FileDefaultPermission = "inherit";
```
JinmingHu-MSFT, line 10: I don't think `default` is a good name. For copy operation, preserve is default, while for create operation, inherit is default.

sdk/storage/src/shares/share_file_client.cpp
katmsft: I think I'm going to make HttpHeaders non-null, but keep protocol layer headers nullable for those headers that can be empty.

sdk/storage/src/shares/share_file_client.cpp
katmsft: It is not necessary that the convenience layer aligns with the protocol layer, like the FileCreationTime we discussed.

sdk/storage/src/shares/share_directory_client.cpp
katmsft: Not necessary, but this one should be changed.

sdk/storage/inc/shares/share_constants.hpp
katmsft: Changed to `c_FileInheritPermission`.

katmsft:
Need to force push to rebase to master.

JinmingHu-MSFT (approved):

sdk/storage/inc/shares/share_client.hpp
katmsft: Reverted this change.

sdk/storage/inc/shares/protocol/share_rest_client.hpp
katmsft: This change is reverted to align with other SDK, however still pending for Jeffrey's input.

PR #511, opened 2020-08-24
Remove context from StorageError::CreateFromResponse

JinmingHu-MSFT:
because that function doesn't involve any I/O operation.


JinmingHu-MSFT:
@katmsft please fix the build errors in share_rest_client

katmsft (approved):

PR #514, opened 2020-08-24
Restructure the azure inc folder

RickWinter:
Move the inc folder to be under inc/azure/core.  This allows us to have a structured include where azure-core includes are accessed by "azure/core/..../<....header.hpp....>

ex:
Previously
`#include "context.hpp"`
now
`#include "azure/core/context.hpp"`


ahsonkhan:
> sdk/core/azure-core/inc/azure.hpp → sdk/core/azure-core/inc/azure/core/azure.hpp

I lost count of the number of "cores" 😁

Seems excessive.

Should it be a lot flatter, like this instead?
`sdk/core/inc/...`?

Why do we have `/azure/core/` under `/inc/` and why do we have an `/azure-core/` subdirectory?

JinmingHu-MSFT (approved):

JinmingHu-MSFT:
> Why do we have `/azure/core/` under `/inc/` and also have an `/azure-core/` subdirectory?

yeah, i also have the same question. `sdk/core/inc/azure/core/foo/bar` seems good enough.

RickWinter:
> > sdk/core/azure-core/inc/azure.hpp → sdk/core/azure-core/inc/azure/core/azure.hpp
> 
> I lost count of the number of "cores" 😁
> 
> Seems excessive.
> 
> Should it be a lot flatter, like this instead?
> `sdk/core/inc/...`?
> 
> Why do we have `/azure/core/` under `/inc/` and also have an `/azure-core/` subdirectory?

This layout follows our track2 design guidelines. 
https://azure.github.io/azure-sdk/policies_repostructure.html#sdk-directory-layout

Core does end up with core repeated a few times.  This is unfortunate but it does keep us consistent with all other languages.  I don't see good enough reasons to break from the design guidelines.
ex:
`https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/core/azure-core/src/main/java/com/azure/core`
`https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/core/azure-core/azure/core`

As to each core's purpose:
We are planning to ship the SDK as packages.  The first "core" is the service.  The second "core" is the package.  The third "core" is for layout of headers in the inc folder.  The structure provides good completion experience when typing out includes.  Makes it very easy to know you are including a storage header `azure/storage/blobs` vs a core header `azure/core/`.

`sdk/core/azure-core/inc/azure/core/azure.hpp`
`sdk/<service>/<package>/inc/...`
is 
`sdk` where sources reside in track2
`core` the service team, in this case the AzureSDK team
`azure-core` the package name
The inc folder then creates a structure that will lay out the include files into system inc directory accordingly.  Everything is under azure, hence the azure under the inc folder.
ex:
`inc/azure/core` the include folder for core "services"
`inc/azure/storage/blobs/` the include folder for "storage blobs"

ahsonkhan:
Thanks @RickWinter, that's starting to make more sense now.

Unlike Python, Java, (and now C++), I envy that .NET get's away with not having the extra `/azure/core/`
`https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/core/Azure.Core`

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
```diff
@@ -3,7 +3,7 @@
 
 #pragma once
 
-#include <context.hpp>
+#include "azure/core/context.hpp"
```
ahsonkhan, line 6: When do we use `<>` vs `""`?

Should this be:
```suggestion
#include <azure/core/context.hpp>
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -3,10 +3,10 @@
 
 #pragma once
 
-#include "body_stream.hpp"
+#include "azure/core/http/body_stream.hpp"
+#include "azure/core/internal/contract.hpp"
```
ahsonkhan, line 7: Same here/elsehere. Why did we change from `<internal/contract.hpp>` to `"..."`?

sdk/core/azure-core/inc/azure/core/logging/logging.hpp
```diff
@@ -3,7 +3,7 @@
 
 #pragma once
 
-#include "azure.hpp"
+#include "azure/core/azure.hpp"
```
ahsonkhan, line 6: Unrelated to this PR, why do we have a header file named `azure.hpp`. We should rename it.

sdk/core/azure-core/test/ut/file_upload.cpp
```diff
@@ -11,9 +11,10 @@
 #include <Windows.h>
 #endif // Windows
 
+#include "azure/core/http/http.hpp"
+
 #include "transport_adapter.hpp"
-//#include <iostream>
-#include <response.hpp>
```
ahsonkhan, line 16: Was this include unnecessary before?

sdk/core/azure-core/inc/azure/core/logging/logging.hpp
RickWinter: At one point there was a desire to have a top level header.  The desire was to have one header to include all services present, i.e. `#include "azure/azure.hpp"` and get everything azure.  Similarly with `#include "storage/blobs.hpp"` to get storage blobs.

sdk/core/azure-core/test/ut/file_upload.cpp
RickWinter: Good catch, I cleaned out a commented include and added back the wrong header.

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
RickWinter: Within the project we use "".  External includes we use <>.  I happened to fix a couple that were wrong and then stopped b/c I didn't want the PR to have that many changes.  There is an issue to cleanup the includes which I'll address in a follow up PR

sdk/core/azure-core/inc/azure/core/http/http.hpp
RickWinter: See above response

Issue #339, opened 2020-07-23
[Feature Request] Reconstruct the folder hierachy for Azure Storage.


katmsft:
Needs to be azure_storage_blob, azure_storage_common etc.

katmsft:
Add options to build blob/datalake/file service separately.

PR #536, opened 2020-08-26
Restructure storage directory

JinmingHu-MSFT:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/339


JinmingHu-MSFT:
@danieljurek @Azure/azure-sdk-eng Hi, can you take a look at the eng/sys related failure? We restructured the directory hierarchy and this error comes up.

JinmingHu-MSFT:
@danieljurek @Azure/azure-sdk-eng Hi, I made some changes to eng/sys code, can you take time to review it? Thanks

sdk/storage/CMakeLists.txt
```diff
@@ -8,174 +8,31 @@ set(CMAKE_CXX_STANDARD 14)
...
-
-add_library(azure-storage-file-share ${AZURE_STORAGE_SHARES_HEADER} ${AZURE_STORAGE_SHARES_SOURCE})
-target_link_libraries(azure-storage-file-share azure-storage-common)
-
-add_library(azure::storage::common ALIAS azure-storage-common)
-add_library(azure::storage::blob ALIAS azure-storage-blob)
-add_library(azure::storage::file::datalake ALIAS azure-storage-file-datalake)
-add_library(azure::storage::file::share ALIAS azure-storage-file-share)
 
+option(BUILD_STORAGE_SAMPLES "Build storage sample codes" ON)
```
RickWinter, line 26: Nit, place the options at the top of the CMakeList so they are easy to find

sdk/storage/azure-storage-blobs/version.txt
```diff
@@ -0,0 +1 @@
+1.0.0-beta.1
```
RickWinter, line 1: Get rid of this file and use a version.h file in the `azure/storage/blobs/version.h`.  See azure-core folder for the contents of that file.

sdk/storage/azure-storage-common/version.txt
```diff
@@ -0,0 +1 @@
+1.0.0-beta.1
```
RickWinter, line 1: See comment about using a version.h file above.  There needs to be a version.h file in the `inc/azure/<service>/` folder for each service

sdk/storage/azure-storage-files-datalake/version.txt
```diff
@@ -0,0 +1 @@
+1.0.0-beta.1
```
RickWinter, line 1: See comment on version.h files

sdk/storage/azure-storage-blobs/version.txt
JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/version.txt

JinmingHu-MSFT: azure core folder also has this file.

JinmingHu-MSFT: I think we should save the version information only at one place, either in version.txt or in version.hpp. I'm fine either way.

JinmingHu-MSFT: https://github.com/Azure/azure-sdk-for-cpp/issues/543 I created an issue to track that. This is not a big deal. I think we can figure it out in the future.

sdk/storage/CMakeLists.txt
JinmingHu-MSFT: fixed

PR #540, opened 2020-08-26
Reference the specific header

RickWinter:
Fix the reference to use response.hpp rather than all of http.hpp
Correct the CMake which was inappropriately including both Curl and WinHttp


sdk/core/azure-core/CMakeLists.txt
```diff
@@ -29,11 +27,25 @@ add_library (
...
+  )
+endif()
+if(BUILD_WINHTTP_TRANSPORT)
+  set(
+    AZURE_CORE_WINHTTP_TRANSPORT_SOURCE
+    src/http/winhttp/win_http_transport.cpp
+  )
+endif()
+
+add_library (${TARGET_NAME} ${AZURE_CORE_SOURCE} ${AZURE_CORE_CURL_TRANSPORT_SOURCE} ${AZURE_CORE_WINHTTP_TRANSPORT_SOURCE})
```
ahsonkhan, line 47: nit/questions: What is the string value of `${AZURE_CORE_WINHTTP_TRANSPORT_SOURCE}` when `BUILD_WINHTTP_TRANSPORT` is false?

 I am guessing it's empty. Is having an empty `''` parameter in `add_library` OK to do?

Maybe there is a way to chain the calls to `add_library` and only call `add_library ${AZURE_CORE_WINHTTP_TRANSPORT_SOURCE}` inside the above if-block.

sdk/core/azure-core/test/ut/file_upload.cpp
```diff
@@ -11,7 +11,7 @@
 #include <Windows.h>
 #endif // Windows
 
-#include "azure/core/http/http.hpp"
+#include "azure/core/response.hpp"
```
ahsonkhan, line 14: How come this didn't cause CI to fail earlier?

RickWinter: This was an old file with an old include.  Not all of response is used and response includes http.hpp.

PR #654, opened 2020-09-14
Response<T> stores Nullable<T>

CaseyCarter:
Per design discussion with Jeffrey last week. It's convenient to the Poller design to allow a `Response<T>` to sometimes contain no `T`. Since we envision this feature being useful to Client code as well, we landed on having `Response<T>` always store a `Nullable<T>` rather than using `Response<Nullable<T>>` only for our specific use case.

Drive-by: Replace `Nullable(const T&)` with `Nullable(T)` to avoid extra copy when initialized with an rvalue.


RickWinter (approved):

sdk/core/azure-core/inc/azure/core/response.hpp
```diff
@@ -9,6 +9,9 @@
 #pragma once
 
 #include <azure/core/http/http.hpp>
+#include <azure/core/http/nullable.hpp>
```
RickWinter, line 12:
```suggestion
#include <azure/core/nullable.hpp>
```

CaseyCarter: Ouch - just a wee bit of copy-pasta here.

vhvb1989 (approved):

ahsonkhan (approved):

Jinming-Hu (approved):

PR #655, opened 2020-09-15
[Storage Blob Service] Fix bug: encode blob name

Jinming-Hu:


katmsft (approved):

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -27,8 +27,8 @@ namespace Azure { namespace Storage { namespace Blobs {
   {
     auto parsedConnectionString = Details::ParseConnectionString(connectionString);
     auto blobUri = std::move(parsedConnectionString.BlobServiceUri);
-    blobUri.AppendPath(containerName, true);
-    blobUri.AppendPath(blobName, true);
+    blobUri.AppendPath(containerName);
```
ahsonkhan, line 30: Do we have tests for all these encoding fixes?

Jinming-Hu: i'm going to add tests tomorrow

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/664

PR #656, opened 2020-09-15
Resolved an issue where path is not encoded.

katmsft:


Jinming-Hu (approved):

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_file_client.hpp
```diff
@@ -233,9 +233,10 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     explicit FileClient(
```
Jinming-Hu, line 233: Just for the record, `FileClient` contains a `BlockBlobClient` and a `BlobClient`, I don't think this is right. But I don't care as long as it works.

PR #667, opened 2020-09-17
Update CODEOWNERS

JeffreyRichter:


RickWinter (approved):

joshfree (approved):
in a subsequent PR, we should update the catch all line and add some other folks there, like Ahson

PR #673, opened 2020-09-19
Mention VCPKG_DEFAULT_TRIPLET in Build instruction

ku-sourav:
Cmake fails to find packages(curl, etc) if VCPKG_DEFAULT_TRIPLET is not set to x64-windows-static. If not set, cmake sets VCPKG_TARGET_TRIPLET to x64-windows instead of x64-windows-static. Since curl was installed using x64-windows-static in vcpkg, cmake fails to find the package. Adding set VCPKG_DEFAULT_TRIPLET to x64-windows-static explicitly in build instructions so that it is helpful to avoid cmake error(could not find CURL)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ can install them with Homebrew.
 ### Building and Testing
 
 #### Building the project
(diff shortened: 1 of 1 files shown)


vhvb1989 (approved):
Thank you!

RickWinter (approved):

PR #697, opened 2020-09-30
Integrate storage retry policy.

katmsft:


sdk/storage/azure-storage-files-datalake/src/datalake_path_client.cpp
```diff
@@ -114,12 +118,15 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     {
       policies.emplace_back(p->Clone());
     }
-    policies.emplace_back(
-        std::make_unique<Azure::Core::Http::RetryPolicy>(Azure::Core::Http::RetryOptions()));
+    StorageRetryWithSecondaryOptions dfsRetryOptions = options.RetryOptions;
+    dfsRetryOptions.SecondaryHostForRetryReads
+        = Details::GetDfsUriFromUri(options.RetryOptions.SecondaryHostForRetryReads);
```
Jinming-Hu, line 123: `GetDfsUriFromUri` can return an empty string. Will everything still be able to work if so?

Jinming-Hu (approved):

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_utilities.hpp
```diff
@@ -8,7 +8,7 @@
 
 namespace Azure { namespace Storage { namespace Files { namespace DataLake { namespace Details {
   std::string GetBlobUriFromUri(const std::string& uri);
-  std::string GetDfsUriFromUri(const std::string& uri);
+  std::string GetDfsUriFromUri(const std::string& uri, bool returnEmptyOnNotFound = true);
```
Jinming-Hu, line 11: This is really ugly.

katmsft: There is no better way to do so.

Jinming-Hu: Let me ask you one question. When is it useful to return an empty uri if the original uri doesn't contain `.dfs.`?

In [this](https://github.com/Azure/azure-sdk-for-cpp/blob/a898fd819aa0fa987358f751f246f4d8a8b85157/sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp#L62) use case, `m_dfsUri` is empty if the original uri is a customized domain, how is an empty uri better than the original one?

katmsft: You are actually right. I'll remove this.

katmsft:
Re-submitted https://github.com/Azure/azure-sdk-for-cpp/pull/722 to work-around the check enforcer hang.

PR #721, opened 2020-10-09
Avoid test case throw meaningless exception when sys env is not set.

katmsft:


ahsonkhan (approved):

katmsft:
Pushed to a wrong remote branch, will resubmit this PR.

PR #722, opened 2020-10-09
Integrate storage retry policy.

katmsft:
Code is same as https://github.com/Azure/azure-sdk-for-cpp/pull/697, resubmit due to a blocked check-enforcer.


vinjiang (approved):
PR for kicking off the enforcer again.

PR #728, opened 2020-10-09
Update code owners to better manage default notifications.

ahsonkhan:


ahsonkhan:
@weshaggard, @mitchdenny - the pipelines are not getting triggered here (tried re-opening the PR).

weshaggard:
@ahsonkhan I think for this PR there is no pipelines that trigger for codeowner file changes. I think we can just override check-enforcer in this case and merge.

weshaggard (approved):

ahsonkhan:
>I think for this PR there is no pipelines that trigger for codeowner file changes.

Nice. So, do we have some filters in place that make sure the CI only runs for code changes? Do markdown changes trigger the CI legs (such as link verification/etc.)? Which file modifications are ignored?

weshaggard:
See https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/storage/ci.yml#L10 as an example the pipelines are triggering based on paths.

>Do markdown changes trigger the CI legs (such as link verification/etc.)

We don't currently but that is something in the works.

PR #745, opened 2020-10-12
Upload/Create is non-overwriting by default

Jinming-Hu:


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -1144,6 +1149,11 @@ namespace Azure { namespace Storage { namespace Blobs {
      */
     Azure::Core::Context Context;
 
+    /**
+     * @brief Whether the existing blob should be deleted and recreated.
```
katmsft, line 1153: Why block blob's comment is different with page blob/append blob's? I think block blob is also deleted and recreated if overwritten.

Jinming-Hu: align with .Net

katmsft (approved):

PR #746, opened 2020-10-12
Live test framework

katmsft:


weshaggard:
Closing this PR and going back to https://github.com/Azure/azure-sdk-for-cpp/pull/608 so we don't lose the discussion there.

PR #747, opened 2020-10-12
Removed Nullable return value that never returns.

katmsft:


Jinming-Hu (approved):

Jinming-Hu:
> /azp run cpp - storage

No tests were actually run.

katmsft:
> > /azp run cpp - storage
> 
> No tests were actually run.

I know, it is the current build only CI.

PR #748, opened 2020-10-12
Extract part of credentials to Identity module

antkmsft:
Closes https://github.com/Azure/azure-sdk-for-cpp/issues/428.

This mimics other SDKs. It will allow us to have flexibility to bring an auth library (MSAL) at any point in the future, should there become an opportunity, without the dependencies requiring a client change; while core will remain lightweight without the dependency on MSAL.


sdk/identity/azure-identity/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+add_library(azure-identity ${AZURE_IDENTITY_HEADER} ${AZURE_IDENTITY_SOURCE})
+target_include_directories(azure-identity PUBLIC inc)
+target_link_libraries(azure-identity azure-core)
+
+# TODO: Do not read from version.txt, use version.hpp and replace this file
+# command with:
+# get_az_version("${CMAKE_CURRENT_SOURCE_DIR}/inc/azure/identity/<package-name>/version.hpp")
+# Access the variable from this function using ${AZ_LIBRARY_VERSION}
+# Look at the CMakeLists.txt for template package for example usage
+file(READ version.txt AZURE_IDENTITY_PACKAGE_VERSION)
```
RickWinter, line 25: Fix before commiting this change.  There is code I wrote in C which you can use to grab the version in CMake.

`
if(EXISTS ${AZ_SDK_VERSION_H_FILE})
    file (STRINGS ${AZ_SDK_VERSION_H_FILE} VERSION_H_CONTENT)
    message(STATUS "Read version from ${AZ_SDK_VERSION_H_FILE}")

    #Find and read the AZ_SDK_VERSION_STRING define
    string(REGEX MATCHALL "#define AZ_SDK_VERSION_STRING \"([0-9]+)\.([0-9]+)\.([0-9]+)(\-[^\"\-]+)?\"" VERSION_PARTS ${VERSION_H_CONTENT})

    #Ensure we matched as expected.
    #   MAJOR.AZ_VERSION_MINOR.PATCH are required.
    #   PRERELEASE is optional.
    if(NOT CMAKE_MATCH_1 AND NOT CMAKE_MATCH_2 AND NOT CMAKE_MATCH_3)
        message(FATAL_ERROR "Unexpected version format in ${AZ_SDK_VERSION_H_FILE}")
    endif()
    set(AZ_SDK_VERSION_MAJOR ${CMAKE_MATCH_1} )
    set(AZ_SDK_VERSION_MINOR ${CMAKE_MATCH_2} )
    set(AZ_SDK_VERSION_PATCH ${CMAKE_MATCH_3} )
    if(CMAKE_MATCH_4)
        #Trim off the "-" from the start of pre-release portion
        # https://cmake.org/cmake/help/v3.10/command/string.html?highlight=substring#substring
        string(SUBSTRING ${CMAKE_MATCH_4} 1 -1 PRERELEASE_SUBSTRING )
        set(AZ_SDK_VERSION_PRERELEASE ${PRERELEASE_SUBSTRING})
    endif()
    message(STATUS "AZ_SDK_VERSION_MAJOR " ${AZ_SDK_VERSION_MAJOR})
    message(STATUS "AZ_SDK_VERSION_MAJOR " ${AZ_SDK_VERSION_MINOR})
    message(STATUS "AZ_SDK_VERSION_PATCH " ${AZ_SDK_VERSION_PATCH})
    message(STATUS "AZ_SDK_VERSION_PRERELEASE " ${AZ_SDK_VERSION_PRERELEASE})
else()
  message(FATAL_ERROR "Missing Version file ${AZ_SDK_VERSION_H_FILE}")
endif()
`

sdk/storage/ci.yml
```diff
@@ -11,6 +11,7 @@ trigger:
       - eng/
       - CMakeLists.txt
       - sdk/core
+      - sdk/identity
```
RickWinter, line 14: Does Storage need the identity package?

RickWinter (changes requested):

sdk/identity/azure-identity/CMakeLists.txt
antkmsft: This is what every other C++ SDK lib currently does. We need to do it, but it needs to be a separate PR.
I just opened an issue to track this - #750.

sdk/storage/ci.yml
antkmsft: Yes - they use ClientSecretCredential, that sits in Identity.
In .NET SDK, it sits in Identity, too.
And that is fine, when you think that Storage depends even on HTTP adapters, Identity dependence is much more "native", natural, simple and lightweight dependency.

sdk/identity/azure-identity/CMakeLists.txt
antkmsft: Done for the identity module/this PR.

antkmsft: I updated the issue I opened with the list of four Storage libraries still using the version.txt method:
* azure-storage-blobs
* azure-storage-common
* azure-storage-files-datalake
* azure-storage-files-shares

sdk/storage/ci.yml
antkmsft: I think my answer resolves the conversation.

sdk/template/azure-template/inc/azure/template/version.hpp
```diff
@@ -18,12 +18,12 @@ namespace Azure { namespace Template {
...
     static constexpr int Patch = AZURE_TEMPLATE_VERSION_PATCH;
     static std::string const PreRelease;
-    static std::string const VersionString();
+    static std::string VersionString();
 
   private:
-    //To avoid leaking out the #define values we smuggle out the value
+    // To avoid leaking out the #define values we smuggle out the value
     // which will later be used to initialize the PreRelease std::string
-    static constexpr const char* secret = AZURE_TEMPLATE_VERSION_PRERELEASE;
```
RickWinter, line 26: Keep this as secret b/c the value is special and only exists to allow us to smuggle the data out

RickWinter (approved):

sdk/template/azure-template/inc/azure/template/version.hpp
antkmsft: I don't understand what we're achieving by naming it "cryptically", not following the naming guidelines, since it is already private, but I don't have that strong of an opinion as you have, and I don't think it is that important - it's just a name, and everything is pretty much is clear as a day, no matter how you name it.

sdk/core/azure-core/inc/azure/core/credentials.hpp
```diff
@@ -3,19 +3,77 @@
 
 /**
  * @file
- * @brief Authentication policies.
+ * @brief Credentials used for authentication with many (not all) Azure SDK client libraries.
```
ahsonkhan, line 6:
```suggestion
 * @brief Credentials used by Azure SDK client libraries for authentication.
```

sdk/core/azure-core/inc/azure/core/version.hpp
```diff
@@ -38,7 +38,7 @@ namespace Azure { namespace Core {
      * @brief The version in string format used for telemetry following the `semver.org` standard
      * (https://semver.org).
      */
-    static std::string const VersionString();
```
ahsonkhan, line 41: Why remove `const`? Is it because it is already being returned by value?

sdk/identity/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+cmake_minimum_required (VERSION 3.12)
```
ahsonkhan, line 4: Be consistent with the rest
```suggestion
cmake_minimum_required (VERSION 3.13)
```

sdk/identity/azure-identity/README.md
```diff
@@ -0,0 +1,73 @@
+# Azure SDK Identity Library for C++
+
+Azure::Identity (`azure-identity`) provides shared primitives, abstractions, and helpers for modern Azure SDK client libraries written in the C++. These libraries follow the [Azure SDK Design Guidelines for C++][azure_sdk_cpp_development_guidelines].
```
ahsonkhan, line 3:
```suggestion
Azure::Identity (`azure-identity`) provides shared primitives, abstractions, and helpers for modern Azure SDK client libraries written in C++. These libraries follow the [Azure SDK Design Guidelines for C++][azure_sdk_cpp_development_guidelines].
```

sdk/identity/azure-identity/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The library contains commonly (but not universally) used credential types.
+
+## Getting started
+
+Typically, you will not need to download `azure-identity`; it will be downloaded for you as a dependency of the client libraries.  In case you want to download it explicitly (to implement your own client library, for example), you can find the source
+in here.
```
ahsonkhan, line 10:
```suggestion
Typically, you will not need to download `azure-identity`; it will be downloaded for you as a dependency of the client libraries. In case you want to download it explicitly (to implement your own client library, for example), you can find the source in here.
```

PR #755, opened 2020-10-13
Make sure CI pipelines legs get triggered when cmake modules are modified.

ahsonkhan:
Certain PRs, like this - https://github.com/Azure/azure-sdk-for-cpp/pull/753 weren't triggering any CI pipelines to run, when they should.

Following suggestion from Wes: https://github.com/Azure/azure-sdk-for-cpp/pull/728#issuecomment-707251860

@Jinming-Hu, @katmsft - I think we should make a similar change to the storage ci.yml file.

### DIFF
--- a/sdk/core/ci.yml
+++ b/sdk/core/ci.yml
@@ -8,6 +8,7 @@ trigger:
       - hotfix/*
   paths:
     include:
+      - cmake-modules/
       - eng/
       - CMakeLists.txt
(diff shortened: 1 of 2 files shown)


Jinming-Hu:
https://github.com/Azure/azure-sdk-for-cpp/pull/756

Jinming-Hu (approved):

PR #773, opened 2020-10-14
[Storage] revert non-overwriting changes

Jinming-Hu:


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -930,11 +930,6 @@ namespace Azure { namespace Storage { namespace Blobs {
      */
     Azure::Core::Context Context;
 
-    /**
```
ahsonkhan, line 933: Is it feasible to use `git revert` in this case to do a clean revert of https://github.com/Azure/azure-sdk-for-cpp/pull/745 directly rather than a custom revert (this commit: 5d1a8f4cbc2be18e06bba47048a6c49b09110b22)? Or does it contain changes we want to keep?

Jinming-Hu: i did something else in that PR so I cannot revert the whole commit

katmsft: This is better to be avoided in the future.

katmsft (approved):

PR #788, opened 2020-10-15
DateTime: Allow dates from 1601 to 1900

### LINKED ISSUE #769: Datetime needs to support Win32 epoch.
Current Datetime implementation supports date time later than 1900/01/01-00:00:00, however, storage service supports Win32 epoch, which is 1601/01/01-00:00:00.

antkmsft:
Closes #769.


antkmsft:
CI/CD reports green, but is not actually running unit tests. I need to do more fixes, and I'll reopen this PR.

PR #793, opened 2020-10-16
replace std::tolower with Azure::Core::ToLower

Jinming-Hu:


vhvb1989 (approved):
Great, tks

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp
```diff
@@ -3473,11 +3474,7 @@ namespace Azure { namespace Storage { namespace Blobs {
         std::set<std::string> metadataKeys;
         for (const auto& pair : options.Metadata)
         {
-          std::string key = pair.first;
-          std::transform(key.begin(), key.end(), key.begin(), [](unsigned char c) {
-            return static_cast<char>(std::tolower(c));
-          });
-          if (metadataKeys.insert(key).second == false)
+          if (metadataKeys.insert(Azure::Core::Details::ToLower(pair.first)).second == false)
```
RickWinter, line 3477: Should not be using Details namespace outside of the assembly.  We need to move our ToLower function if you are going to take a dependency on it.

RickWinter (changes requested):

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp
Jinming-Hu: `ToLower` is a function provided in Core sdk. So this is a change core team should do, not storage team.

Jinming-Hu: Or are you suggesting that we should just keep things as it is. Should I just close this PR?

ahsonkhan: FYI, @JeffreyRichter

PR #796, opened 2020-10-16
Resolved the issue in datalake/file where file/path/directory/share/f…

katmsft:
…ilesystem name are not encoded.


sdk/storage/azure-storage-files-shares/test/share_client_test.cpp
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 #include "share_client_test.hpp"
+#include "azure/storage/common/crypt.hpp"
```
Jinming-Hu, line 5: reformat

Jinming-Hu (approved):

sdk/storage/azure-storage-files-shares/test/share_client_test.cpp
katmsft: Don't know why my Clang formatter does not catch this one.

PR #799, opened 2020-10-16
DateTime: Allow dates from 1601 to 1900

### LINKED ISSUE #769: Datetime needs to support Win32 epoch.

antkmsft:
Closes #769.

### DIFF
--- a/sdk/core/azure-core/src/datetime.cpp
+++ b/sdk/core/azure-core/src/datetime.cpp
@@ -56,10 +56,8 @@ constexpr int SecondsInDay = 24 * SecondsInHour;
 
 constexpr int DaysInYear = 365;
 
-constexpr ComputeYearResult ComputeYear(int64_t secondsSince1900)
+constexpr ComputeYearResult ComputeYear(int64_t secondsSince1601)
 {
-  constexpr int64_t SecondsFrom1601To1900 = 9435484800LL;
-
   constexpr int DaysIn4Years = DaysInYear * 4 + 1;
 
   constexpr int DaysIn100Years = DaysIn4Years * 25 - 1;
@@ -71,8 +69,7 @@ constexpr ComputeYearResult ComputeYear(int64_t secondsSince1900)
   constexpr int64_t SecondsIn100Years = static_cast<int64_t>(SecondsInDay) * DaysIn100Years;
   constexpr int64_t SecondsIn400Years = static_cast<int64_t>(SecondsInDay) * DaysIn400Years;
 
-  int64_t secondsLeft
-      = secondsSince1900 + SecondsFrom1601To1900; // shift to start of this 400 year cycle
+  int64_t secondsLeft = secondsSince1601;
 
   int year400 = static_cast<int>(secondsLeft / SecondsIn400Years);
   secondsLeft -= year400 * SecondsIn400Years;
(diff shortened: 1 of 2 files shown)


Jinming-Hu (approved):

RickWinter:
cc: @BillyONeal adding you as you mentioned using 1900 vs 1601 was a design decision in the original impl

sdk/core/azure-core/test/ut/datetime.cpp
```diff
@@ -457,3 +455,17 @@ TEST(DateTime, ParseTimeInvalid2)
...
 }
+
+TEST(DateTime, ParseDatesBefore1900)
+{
+  TestDateTimeRoundtrip("1899-01-01T00:00:00Z");
+  auto dt1 = DateTime::FromString("1899-01-01T00:00:00Z", DateTime::DateFormat::Iso8601);
+  auto dt2 = DateTime::FromString("Sun, 1 Jan 1899 00:00:00 GMT", DateTime::DateFormat::Rfc1123);
+  EXPECT_EQ(dt1.ToInterval(), dt2.ToInterval());
+
+  TestDateTimeRoundtrip("1601-01-01T00:00:00Z");
```
ahsonkhan, line 466: Please also add a test for a date prior to 1600 to capture edge case behavior.

sdk/core/azure-core/test/ut/datetime.cpp
```diff
@@ -97,7 +97,7 @@ TEST(DateTime, ParseTimeRoundrip6)
   TestDateTimeRoundtrip("2013-11-19T14:30:59.5Z");
 }
 
-TEST(DateTime, ParseTimeRoundripYear1900) { TestDateTimeRoundtrip("1900-01-01T00:00:00Z"); }
```
ahsonkhan, line 100: Please keep this test case, in addition to the new 1601 one.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -320,9 +315,6 @@ constexpr int CountLeapYears(int yearsSince1900)
 
   result += tmpYears / 4;
 
-  // subtract off leap years from 1601
-  result -= 72;
```
ahsonkhan, line 324: Neat

ahsonkhan (approved):

sdk/core/azure-core/test/ut/datetime.cpp
antkmsft: "time zone underflow" and "year too small" above are updated to test that.

sdk/core/azure-core/test/ut/datetime.cpp
```diff
@@ -440,8 +439,7 @@ TEST(DateTime, ParseTimeInvalid2)
       "1971-01-01T00:60:00Z", // minute too big
       "1971-01-01T00:00:70Z", // second too big
       "1971-01-01T00:00:61Z",
-      "1899-01-01T00:00:00Z", // underflow
```
ahsonkhan, line 443: Keep this
```suggestion
      "1599-01-01T00:00:00Z", // underflow
```

antkmsft:
> cc: @BillyONeal adding you as you mentioned using 1900 vs 1601 was a design decision due to leap years in the original impl. https://en.wikipedia.org/wiki/Leap_year

 I can say about this, that in this implementation ticks do match Windows ticks between 1601 and 1900 and seconds match .NET seconds you get for the same period, so two implementations agree on the number of seconds passed between 1601 and 1900 using the same rules as after 1900.

sdk/core/azure-core/test/ut/datetime.cpp
antkmsft: See 1600-01-01 above.
There was duplication in the original test. so this instance is removed completely.

Original implementation was allowing starting 1900, this one starting 1601, so the corresponding underflow date to test would be 1600, not 1599.

antkmsft:
For the future: if we ever allow year 0001 (ISO and RFC specify it as the earliest), I'd say, make the internal representation (`m_interval`) go negative for these, so that 0 is the default and a start of the NT epoch, which effectively appears in Azure services, and is an earliest possible date in Storage.

PR #831, opened 2020-10-21
Add prepare pipelines.

mitchdenny:
This PR adds the prepare pipelines YAML used for streamlining pipeline creation.


eng/pipelines/prepare-pipelines.yml
```diff
@@ -0,0 +1,10 @@
+trigger: none
+
+extends:
+  template: /eng/common/pipelines/templates/steps/prepare-pipelines.yml
+  parameters:
+    Repository: $(Build.Repository.Name)
+    Prefix: cpp
+    CIConventionOptions: ''
+    UPConventionOptions: '--variablegroup 58 --variablegroup 76 --variablegroup 56'
+    TestsConventionOptions: ''
```
ahsonkhan, line 10: Add a new line character to avoid `missing eof` warning.

eng/pipelines/prepare-pipelines.yml
```diff
@@ -0,0 +1,10 @@
+trigger: none
+
+extends:
+  template: /eng/common/pipelines/templates/steps/prepare-pipelines.yml
+  parameters:
+    Repository: $(Build.Repository.Name)
+    Prefix: cpp
+    CIConventionOptions: ''
+    UPConventionOptions: '--variablegroup 58 --variablegroup 76 --variablegroup 56'
```
ahsonkhan, line 9: What do these values mean? Also, what is UP?

eng/pipelines/prepare-pipelines.yml
```diff
@@ -0,0 +1,10 @@
+trigger: none
```
ahsonkhan, line 1: With no trigger, what triggers it? Manual?

eng/pipelines/prepare-pipelines.yml
mitchdenny: ```up``` means _Unified Pipeline_. It ties back to a convention supported in Pipeline Generator (the tool we use to auto generate your pipelines from ```ci.yml``` files.

mitchdenny: The ```--variablegroup``` arguments relate the IDs of variable groups in Azure Pipelines that we map to the pipeline. For example the release pipelines need access to certain secrets which are exposed in variable groups.

eng/pipelines/prepare-pipelines.yml
mitchdenny: The pipeline is triggered by a service team which is onboarding a new service for the first time. The way this will work is that Check Enforcer will detect that a PR hasn't run any pipelines. It will then add a comment to the PR with some helpful tips, one of those tips is to "prepare pipelines" for the new service, which the customer would do by issuing this command:

```/azp run prepare-pipelines```

This file is the definition that sits behind this.

eng/pipelines/prepare-pipelines.yml
mitchdenny: Done.

ahsonkhan:
Interesting, how did we merge this without an explicit approval? I thought this GitHub repo required at least one approval before the button was green.

Like:
image

PR #837, opened 2020-10-22
move storage headers

Jinming-Hu:


vhvb1989 (approved):

sdk/storage/azure-storage-files-datalake/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 cmake_minimum_required (VERSION 3.13)
 
 set (AZURE_STORAGE_DATALAKE_HEADER
-    inc/azure/storage/files/datalake/datalake.hpp
+    inc/azure/storage/datalake.hpp
```
Jinming-Hu, line 7: do you think this should be `file_datalake.hpp` or just `datalake.hpp`?
also `file_shares.hpp` or `shares.hpp`?

sdk/storage/azure-storage-files-datalake/inc/azure/storage/datalake.hpp
```diff
@@ -7,4 +7,5 @@
 #include "azure/storage/files/datalake/datalake_file_client.hpp"
```
ahsonkhan, line 7: So, if you expect there might be naming collisions with datalake or shares that are not related to files, within storage (now or in the future), it would be best to keep the `/files/` directory.

Similarly, is knowing this is related to files, as a domain-specific concept, import for the user?

```diff
-sdk/storage/azure-storage-files-datalake/inc/azure/storage/datalake.hpp
+sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake.hpp
```

sdk/storage/azure-storage-files-shares/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set (AZURE_STORAGE_SHARES_HEADER
     inc/azure/storage/files/shares/share_responses.hpp
     inc/azure/storage/files/shares/share_sas_builder.hpp
     inc/azure/storage/files/shares/share_service_client.hpp
-    inc/azure/storage/files/shares/shares.hpp
+    inc/azure/storage/shares.hpp
```
ahsonkhan, line 17: Same here, keep the `/files/` subdirectory:
```suggestion
    inc/azure/storage/files/shares.hpp
```

ahsonkhan (changes requested):
Move the headers up only **one** directory level.

sdk/storage/azure-storage-files-datalake/CMakeLists.txt
ahsonkhan: Keep the `/files/` subdirectory.
```suggestion
    inc/azure/storage/files/datalake.hpp
```

ahsonkhan:
Please add a changelog entry for this change.

sdk/storage/azure-storage-files-datalake/CMakeLists.txt
Jinming-Hu: While, I watched the replay again, this is not we agreed on in the meeting this week.

Jinming-Hu: what we've agreed on is `inc/azure/storage/datalake.hpp`

ahsonkhan (approved):

sdk/storage/azure-storage-files-datalake/CMakeLists.txt
ahsonkhan: I see what you are saying. Given the "package name" already contains `azure-storage-files-datalake`, having it show up again in the include-path _seems_ redundant.

However, moving it all the way to the root, like you observed, introduces the question of whether we need to keep `files` in the header name or not.

I don't feel too strongly about this, but my mental model when bringing this up was to limit larger directory structure changes and primarily focus on simplifying the include header uses case from the user.

`#include <azure/storage/files_datalake.hpp>` vs `#include <azure/storage/files/datalake.hpp>` isn't materially different to warrant moving datalake.hpp up anymore than just one level, and there is value in keeping the `/files/` structure clear from an organizational perspective. There's another small benefit by doing this, and that's avoiding underscore in the simplified header.

I think where we have landed in this PR now is great :)

katmsft (approved):

PR #841, opened 2020-10-22
Add a test that forces core.hpp to be included

RickWinter:
Adding a test that includes core.hpp for building.  It is not intended to be an exhaustive check of all the types in core.hpp.


ahsonkhan:
Is the goal of this PR different from https://github.com/Azure/azure-sdk-for-cpp/pull/824?

PR #892, opened 2020-11-03
Centralize the docindex scripts and regenerate Github IO landing page.

sima-zhu:


eng/pipelines/doc-index.yml
```diff
@@ -8,25 +8,23 @@ jobs:
...
         workingDirectory: $(Build.SourcesDirectory)
         displayName: Download and Extract DocFX
-
-      - pwsh: >-
-          $(Build.SourcesDirectory)/eng/docs/index/Generate-DocIndex.ps1
-          -RepoRoot $(Build.SourcesDirectory)
-          -DocGenDir "$(Build.SourcesDirectory)/eng/docs/index"
-          -verbose
-        displayName: 'Generate Doc Index'
+        name: setupDocfxTool
```
weshaggard, line 17: What is the name for here?

weshaggard (approved):

eng/pipelines/doc-index.yml
sima-zhu: I have tested over the variable $(docfxToolPath), the only way to make it work is to use the step name to call it.

`$(setupDocfxTool.docfxPath)`

sima-zhu: The variable shared across steps.

weshaggard: Oh... that is because you set the "isOutput=true" for this use case that isn't needed but also don't really hurt anything.

PR #894, opened 2020-11-03
Added publish to GH pages steps.

sima-zhu:


eng/pipelines/doc-index.yml
```diff
@@ -35,16 +34,23 @@ jobs:
...
 
-      - task: CopyFiles@2
-        displayName: Copy HTML to Artifacts Directory
-        inputs:
-          sourceFolder: $(Build.SourcesDirectory)/docfx_project/
-          content: '**\*'
-          targetFolder: $(Build.ArtifactStagingDirectory)/docfx_project
-          overWrite: true
+      - pwsh: |
+          git clone --depth 1 --branch gh-pages --single-branch https://github.com/Azure/azure-sdk-for-cpp.git $(GHPagesDirectory)
```
weshaggard, line 38: As we talked lets try to avoid a duplicated clone.

eng/pipelines/doc-index.yml
```diff
@@ -35,16 +34,23 @@ jobs:
...
+        workingDirectory: $(Build.SourcesDirectory)
+        displayName: Git pull GH pages branch
 
-      - task: PublishPipelineArtifact@0
```
weshaggard, line 46: We should keep the publishing of the artifacts to help with debugging.

eng/pipelines/doc-index.yml
```diff
@@ -25,7 +25,6 @@ jobs:
             -RepoRoot $(Build.SourcesDirectory)
             -DocGenDir "$(Build.SourcesDirectory)/eng/docs/index"
             -verbose
-
       - task: UsePythonVersion@0
```
weshaggard, line 28: Lets see about changing the output directory to somewhere outside of the SourcesDirectory, something like the artifact directory.

weshaggard: We might also need to change the output path to the docfx tool as well so it isn't in the source tree.

sima-zhu: What about we put it `$(Build.BinariesDirectory)`

eng/pipelines/doc-index.yml
```diff
@@ -11,40 +11,63 @@ jobs:
...
       - task: PublishPipelineArtifact@0
         condition: succeeded()
         inputs:
           artifactName: "Doc.Index"
-          targetPath: $(Build.ArtifactStagingDirectory)/docfx_project/_site
\ No newline at end of file
+          targetPath: $(Build.ArtifactStagingDirectory)/docfx_project/_site
+
+      - pwsh: |
+          git checkout -b gh-pages-local --track origin/gh-pages-root
```
weshaggard, line 50: This can be just git checkout gh-pages-root.

eng/pipelines/doc-index.yml
```diff
@@ -11,40 +11,63 @@ jobs:
...
+
+      - task: PowerShell@2
+        displayName: Push the Docs to GH-Pages
+        condition: succeeded()
+        inputs:
+          pwsh: true
+          workingDirectory: $(Build.SourcesDirectory)
+          filePath: $(gitPushScriptPath)
+          arguments: >
+            -PRBranchName "gh-pages1"
```
weshaggard, line 69: This is fine testing but you will want this to be gh-pages once testing is done.

eng/pipelines/doc-index.yml
```diff
@@ -11,40 +11,63 @@ jobs:
...
+          targetPath: $(Build.ArtifactStagingDirectory)/docfx_project/_site
+
+      - pwsh: |
+          git checkout -b gh-pages-local --track origin/gh-pages-root
+        workingDirectory: $(Build.SourcesDirectory)
+        displayName: Git pull GH pages branch
+
+      - pwsh: |
+          Copy-Item -Path $(Build.ArtifactStagingDirectory)/docfx_project/_site/* -Destination ./ -Recurse -Force
+          git add -A
```
weshaggard, line 56: You don't need to add here as that will happen in the git-branch-push script.

eng/pipelines/doc-index.yml
sima-zhu: We have to track the remote gh-pages-root, otherwise it will inherit the commits from master. We have checkout step which pulls master

eng/pipelines/doc-index.yml
sima-zhu: All files copied to the gh-pages-local branch is untracked from git. git-branch-push will create a new branch which inherit gh-pages-local which did not include the new changes.

sima-zhu: I tested over. If not add the line, the branch did not get update to the right changes.

weshaggard: OK if this is needed for new untracked files then I'm good with having it here.

eng/pipelines/doc-index.yml
```diff
@@ -11,40 +11,62 @@ jobs:
...
 
-      - task: CopyFiles@2
-        displayName: Copy HTML to Artifacts Directory
-        inputs:
-          sourceFolder: $(Build.SourcesDirectory)/docfx_project/
-          content: '**\*'
-          targetFolder: $(Build.ArtifactStagingDirectory)/docfx_project
-          overWrite: true
+      - pwsh: |
+          Copy-Item -Path $(Build.SourcesDirectory)/eng/common/scripts/git-branch-push.ps1 -Destination ./ -Recurse -Force
```
weshaggard, line 38: Lets be sure to copy the entire "eng" directory here as well.

sima-zhu: Sure. If so, we will have the eng/common folder show up in gh-pages.

sima-zhu: Or we put the entire folder under `$(Build.BinariesDirectory)`

weshaggard (approved):

PR #920, opened 2020-11-09
Removed IPv6 not compatible SAS test cases.

katmsft:


vhvb1989 (approved):

Jinming-Hu (approved):

sdk/storage/azure-storage-blobs/test/blob_sas_test.cpp
```diff
@@ -237,9 +237,10 @@ namespace Azure { namespace Storage { namespace Test {
       auto sasToken = builder2.ToSasQueryParameters(*keyCredential);
       EXPECT_THROW(verify_blob_create(sasToken), StorageError);
 
-      builder2.IPRange = "0.0.0.0-255.255.255.255";
-      sasToken = builder2.ToSasQueryParameters(*keyCredential);
-      EXPECT_NO_THROW(verify_blob_create(sasToken));
+      // TODO: Add this test case back with support to contain IPv6 ranges when service is ready.
```
ahsonkhan, line 240: There should be issues filed with any such TODOs in source to track them along with a comment linking to the issue. Please avoid TODOs like these in source going forward.

Jinming-Hu: I don't think so. There should be some TODOs for server side, but not client side. Theoretically, once server sides supports IPV6, client side will support it without making any changes.

ahsonkhan: My feedback is a general purpose one. We shouldn't have commented out code in source/test and TODO comments that don't have a tracking issue, especially as we get closer to GA and the quality bar rises. If there is some test that needs to be enabled/added, or a TODO scenario that needs to be implemented, please file an issue for it and remove the comments. If you want to keep a TODO comment, add the link to the issue to add back the test, feature, etc.

I have no opinions on the details of this particular scenario around IPV6. The specific details you mentioned here belong on that issue :)

PR #938, opened 2020-11-10
DateTime: API review feedback

antkmsft:


sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+   * @brief An exception that gets thrown when @DateTime error occurs.
+   */
+  class DateTimeException : public std::runtime_error {
+  public:
+    /**
+     * @brief Construct with message string.
+     *
+     * @param msg Message string.
+     */
+    explicit DateTimeException(std::string const& msg) : std::runtime_error(msg) {}
```
ahsonkhan, line 26:
```suggestion
     * @param message Message string.
     */
    explicit DateTimeException(std::string const& message) : std::runtime_error(message) {}
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
-    typedef uint64_t IntervalType;
-
-  private:
-    //  Number of seconds between 01-01-1970 and 01-01-1601.
-    static constexpr IntervalType WindowsToPosixOffsetSeconds = 11644473600LL;
+    /**
+     * @brief Represents time duration.
+     */
+    class Duration {
+      friend class DateTime;
```
ahsonkhan, line 38: Is this needed? And if so, why? Can we avoid this tight coupling?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+
+      int64_t m_100nsIntervals;
+      constexpr explicit Duration(int64_t intervalsOf100ns) : m_100nsIntervals(intervalsOf100ns) {}
+
+      static constexpr int64_t IntervalsOf100nsPerMicrosecond = 10;
+      static constexpr int64_t IntervalsOf100nsPerMillisecond
+          = 1000 * IntervalsOf100nsPerMicrosecond;
+      static constexpr int64_t IntervalsOf100nsPerSecond = 1000 * IntervalsOf100nsPerMillisecond;
+      static constexpr int64_t IntervalsOf100nsPerMinute = 60 * IntervalsOf100nsPerSecond;
+      static constexpr int64_t IntervalsOf100nsPerHour = 60 * IntervalsOf100nsPerMinute;
```
ahsonkhan, line 48: Just to verify, these are private, correct? Do we have no distinction between private and public constexpr in their names?

cc @JeffreyRichter

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
-    //  Number of seconds between 01-01-1970 and 01-01-1601.
-    static constexpr IntervalType WindowsToPosixOffsetSeconds = 11644473600LL;
+    /**
+     * @brief Represents time duration.
+     */
+    class Duration {
+      friend class DateTime;
+
+      int64_t m_100nsIntervals;
+      constexpr explicit Duration(int64_t intervalsOf100ns) : m_100nsIntervals(intervalsOf100ns) {}
```
ahsonkhan, line 41: Do we need to expose the resolution in the argument name? What if we ant to increase that in the future?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+      constexpr Duration() : m_100nsIntervals(0) {}
+
+      static constexpr Duration FromHours(int hours)
+      {
+        return Duration(hours * IntervalsOf100nsPerHour);
+      }
+
+      void constexpr AddHours(int hours) { m_100nsIntervals += (hours * IntervalsOf100nsPerHour); }
+
+      static constexpr Duration FromMinutes(long long minutes)
```
ahsonkhan, line 60: Why `long long` and not int, or just `int64_t`?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+      }
+
+      static constexpr Duration FromNanoseconds(long long nanoseconds)
+      {
+        if (nanoseconds % 100 != 0)
+        {
+          throw DateTimeException("Can't convert from nanoseconds");
+        }
+
+        return Duration(nanoseconds / 100);
```
ahsonkhan, line 107: Use named constant for this rather than "magic number".

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+      {
+        if (nanoseconds % 100 != 0)
+        {
+          throw DateTimeException("Can't convert from nanoseconds");
+        }
+
+        m_100nsIntervals += (nanoseconds / 100);
+      }
+
+      long long constexpr GetNanoseconds() const { return m_100nsIntervals * 100; }
```
ahsonkhan, line 120: Same here/elsewhere.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+      }
+
+      constexpr bool operator<(Duration const& other) const
+      {
+        return (m_100nsIntervals < other.m_100nsIntervals);
+      }
+
+      constexpr bool operator<=(Duration const& other) const
+      {
+        return (*this < other) || (*this == other);
```
ahsonkhan, line 146: +1 for using existing operators to implement new ones :)

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,157 @@
...
+      }
+
+      constexpr bool operator!=(Duration const& other) const { return !(*this == other); }
```
ahsonkhan, line 149: Same here. 👍

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,60 +188,52 @@ namespace Azure { namespace Core {
...
+     * @brief Construct an instance of @DateTime.
+     *
+     * @param year Year.
+     * @param month Month.
+     * @param day Day.
+     * @param hour Hour.
+     * @param minute Minute.
+     * @param seconds Seconds.
+     *
+     * @throw DateTimeException If any paramter is invalid.
```
ahsonkhan, line 210: Document the default date if I call: `DateTime(0)`

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,60 +188,52 @@ namespace Azure { namespace Core {
...
     /**
      * @brief Create @DateTime from a string representing time in UTC in the specified format.
      *
-     * @param timeString A string with the date and time.
+     * @param dateTime A string with the date and time.
      * @param format A format to which /p timeString adheres to.
      *
-     * @return @DateTime that was constructed from the \p timeString; Uninitialized
-     * (!@IsInitialized()) @DateTime if parsing \p timeString was not successful.
+     * @return @DateTime that was constructed from the \p timeString.
```
ahsonkhan, line 220:
```suggestion
     * @return @DateTime that was constructed from the \p dateTime string.
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -113,107 +243,77 @@ namespace Azure { namespace Core {
...
-    IntervalType ToInterval() const { return m_interval; }
+    DateTime& operator+=(Duration const& value)
+    {
+      m_since1601 += value;
+      return *this;
+    }
+
+    DateTime& operator-=(Duration const& value)
+    {
+      m_since1601 += value;
```
ahsonkhan, line 271: Add? This should be subtract. Hopefully there are test that catch this.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -113,107 +243,77 @@ namespace Azure { namespace Core {
...
+    {
+      return this->m_since1601 >= dt.m_since1601;
+    }
 
     /// Compare the chronological order of two @DateTime instances.
-    bool operator<=(const DateTime& dt) const { return this->m_interval <= dt.m_interval; }
-
-    /// Create an interval from milliseconds.
-    static IntervalType FromMilliseconds(unsigned int milliseconds)
+    constexpr bool operator<=(const DateTime& dt) const
```
ahsonkhan, line 309: This should be implemented as `! > `. Same elsewhere that is possible.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -610,13 +614,13 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
     // parse month
     if (!IsDigit<1>(str[0]) || !IsDigit(str[1]))
     {
-      return result;
+      throw DateTimeException("Error parsing DateTime.");
```
ahsonkhan, line 617: Consider adding contextual detail to all these error messages since we know where the parsing error occurred.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -777,6 +782,75 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  // from this constructor
+  if (year > 9999)
+  {
+    throw DateTimeException("The requested year exceeds the year 9999.");
+  }
+
+  if (year < 1601)
+  {
+    throw DateTimeException("The requested year is less than the year 1601.");
+  }
```
ahsonkhan, line 802:
```suggestion
  if (year > 9999 || year < 1601)
  {
    string message = year > 9999 ? "The requested year exceeds the year 9999." : "The requested year is less than the year 1601.";
    throw DateTimeException(message);
  }
```

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -777,6 +782,75 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  result.m_since1601.m_100nsIntervals
+      = static_cast<int64_t>(secondsSince1601 * Duration::IntervalsOf100nsPerSecond + fracSec);
   return result;
 }
+
+DateTime::DateTime(int year, int month, int day, int hour, int minute, int seconds)
+{
+  // We should combine creation/validation logic, so it is reusable from both parsing functions and
+  // from this constructor
+  if (year > 9999)
```
ahsonkhan, line 794: Use named constants.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -777,6 +782,75 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  }
+
+  if (year < 1601)
+  {
+    throw DateTimeException("The requested year is less than the year 1601.");
+  }
+
+  if (month <= 0 || month > 12)
+  {
+    throw DateTimeException("Invalid month number.");
```
ahsonkhan, line 806: FWIW, .NET uses these exception messages:
`Year, Month, and Day parameters describe an un-representable DateTime.`
`Hour, Minute, and Second parameters describe an un-representable DateTime.`

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -777,6 +782,75 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  {
+    throw DateTimeException("The requested year exceeds the year 9999.");
+  }
+
+  if (year < 1601)
+  {
+    throw DateTimeException("The requested year is less than the year 1601.");
+  }
+
+  if (month <= 0 || month > 12)
```
ahsonkhan, line 804: Also check `day >= 1`

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -777,6 +782,75 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  {
+    throw DateTimeException("Invalid seconds value.");
+  }
+
+  if (!ValidateDay(day, month - 1, year))
+  {
+    throw DateTimeException("Invalid day value.");
+  }
+
+  char outBuffer[38]{}; // Thu, 01 Jan 1970 00:00:00 GMT\0
```
ahsonkhan, line 829: Where did 38 come from? Those strings in the comment look like ~30 characters.

sdk/core/azure-core/test/ut/datetime.cpp
```diff
@@ -9,49 +9,54 @@
...
+} // namespace
+
 TEST(DateTime, ParseDateAndTimeBasic)
 {
-  auto dt1 = DateTime::FromString("20130517T00:00:00Z", DateTime::DateFormat::Iso8601);
-  EXPECT_NE(0u, dt1.ToInterval());
-
-  auto dt2 = DateTime::FromString("Fri, 17 May 2013 00:00:00 GMT", DateTime::DateFormat::Rfc1123);
-  EXPECT_NE(0u, dt2.ToInterval());
+  auto dt1 = DateTime::Parse("20130517T00:00:00Z", DateTime::DateFormat::Rfc3339);
```
ahsonkhan, line 21: If I pass in Rfc1123 here, will it throw? And same below? Worth adding that as a test (mix-match format and input string).

sdk/core/azure-core/test/ut/simplified_header.cpp
```diff
@@ -16,7 +16,7 @@
 TEST(Logging, simplifiedHeader)
 {
   EXPECT_NO_THROW(Azure::Core::Context c);
-  EXPECT_NO_THROW(Azure::Core::DateTime::DateTime::FromSeconds(10));
```
ahsonkhan, line 19: What required changing this?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -113,107 +243,77 @@ namespace Azure { namespace Core {
...
     {
-      return milliseconds * TicksPerMillisecond;
+      return this->m_since1601 <= dt.m_since1601;
     }
 
-    /// Create an interval from seconds.
-    static IntervalType FromSeconds(unsigned int seconds) { return seconds * TicksPerSecond; }
```
ahsonkhan, line 167: What happened to the `Year`/`Month`/`Hour`/etc. constants, so that the user can do `Datetime + 24 * Hour`?

ahsonkhan (changes requested):

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan:
```suggestion
      m_since1601 -= value;
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,60 +188,52 @@ namespace Azure { namespace Core {
       /// RFC 1123.
       Rfc1123,
 
-      /// ISO 8601.
-      Iso8601,
+      /// RFC 3339.
+      Rfc3339,
```
ahsonkhan, line 192: Unlike ISO, RFC 3339 doesn't limit the number of fractional second digits. This will need some implementation change (can be done in a separate PR).

I recommend supporting up to 16 fractional digits (ignoring digit 8-16), throwing for anything more than that, but only using the first 7 as meaningful digits.

This RFC also supports ' ', 'T', 'Z', 't', 'z' for separators. What is our behavior for those? It is OK to only accept 'T' and ' ', if you like, and reject the lower case versions.

Look at this for more info on the subtle differences between ISO 8601 and RFC 3339: https://docs.microsoft.com/en-us/dotnet/standard/datetime/system-text-json-support#support-for-parsing

antkmsft: Yes, that will go as a separate PR, now we want clean API interface.
I also looked at .NET implementation, it does rounding on 8th digit, >=5 is always up. Years 0001 - 9999 are allowed.

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: That's fine. What's the default date if I only provide the year? For example `DateTime(2020)` - This creates 01-01-2020 00:00:00.00... correct? That might be helpful to document.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -430,12 +437,12 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
     else
     {
-      return result;
+      throw DateTimeException("Error parsing DateTime: Day of the month.");
     }
 
     if (monthDay == 0)
     {
-      return result;
+      throw DateTimeException("Error parsing DateTime: Invalid month day number (0).");
```
ahsonkhan, line 445: Missing argument for `{0}`?

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -449,33 +456,33 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
-      return result;
+      throw DateTimeException("Error parsing DateTime: Year.");
     }
 
     int year = (str[0] - '0') * 1000 + (str[1] - '0') * 100 + (str[2] - '0') * 10 + (str[3] - '0');
-    if (year < 1601)
+    if (year < MinYear)
     {
-      return result;
+      throw DateTimeException("Error parsing DateTime: Year is less than 1601.");
```
ahsonkhan, line 479: Use constants like `MinYear` within the exception message rather than hard cording it. Same for 9999.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -777,6 +785,77 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  if (year > MaxYear || year < MinYear)
+  {
+    throw DateTimeException(
+        year > MaxYear ? "The requested year exceeds the year 9999."
+                       : "The requested year is less than the year 1601.");
+  }
+
+  if (month <= 0 || month > 12)
+  {
+    throw DateTimeException("Invalid month number.");
```
ahsonkhan, line 806: nit: consistency with other messages below
```suggestion
    throw DateTimeException("Invalid month value.");
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,158 @@
...
+
+    public:
+      constexpr Duration() : m_100nsIntervals(0) {}
+
+      static constexpr Duration FromHours(int hours)
+      {
+        return Duration(hours * IntervalsOf100nsPerHour);
+      }
+
+      void constexpr AddHours(int hours) { m_100nsIntervals += (hours * IntervalsOf100nsPerHour); }
```
ahsonkhan, line 59: When do single line methods get collapsed into one line vs have three separate lines? Is it because of line length?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,158 @@
...
+      {
+        if (nanoseconds % NanosecondResolution != 0)
+        {
+          throw DateTimeException("Can't convert from nanoseconds");
+        }
+
+        m_100nsIntervals += (nanoseconds / NanosecondResolution);
+      }
+
+      long long constexpr GetNanoseconds() const { return m_100nsIntervals * 100; }
```
ahsonkhan, line 121:
```suggestion
      long long constexpr GetNanoseconds() const { return m_100nsIntervals * NanosecondResolution; }
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,60 +189,52 @@ namespace Azure { namespace Core {
...
+     * @param month Month.
+     * @param day Day.
+     * @param hour Hour.
+     * @param minute Minute.
+     * @param seconds Seconds.
+     *
+     * @throw DateTimeException If any paramter is invalid.
      */
-    DateTime() : m_interval(0) {}
+    DateTime(int year, int month = 1, int day = 1, int hour = 0, int minute = 0, int seconds = 0);
```
ahsonkhan, line 213: Why are all parameters singular except for seconds?

```suggestion
    DateTime(int year, int month = 1, int day = 1, int hour = 0, int minute = 0, int second = 0);
```

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -127,19 +132,20 @@ constexpr char const monthNames[] = "Jan\0Feb\0Mar\0Apr\0May\0Jun\0Jul\0Aug\0Sep
 
 } // namespace
 
-std::string DateTime::ToString(DateFormat format, TimeFractionFormat fractionFormat) const
+std::string DateTime::GetString(DateFormat format, TimeFractionFormat fractionFormat) const
 {
-  if (m_interval > 2650467743999999999LL)
+  if (m_since1601.m_100nsIntervals > 2650467743999999999LL)
```
ahsonkhan, line 137: magic number - turn to named constant

vhvb1989 (approved):

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Doxygen should be populating defaults. Too bad I can't check at the moment, as the build pipeline is broken.

sdk/core/azure-core/src/datetime.cpp
antkmsft: to do that, we ill need to bring ostringstream to format messages, I don't think it is worth the cost. It is ok to use constants so that you can distinguish between to different identical values, but it does not mean that you can change 1601 to 0001, or switch the resolution, and expect it to work. You'll need to review all the occurrences, just as we did when we rebesed from year 1900 to 1601. I think this implementation should get rewritten at some point, and there we should think of the most effective ways to make it extensible.

sdk/core/azure-core/src/datetime.cpp
antkmsft: No, it literally says - "the number is invalid (ant that number is zero) :)"

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,60 +189,52 @@ namespace Azure { namespace Core {
       /// RFC 1123.
       Rfc1123,
 
-      /// ISO 8601.
-      Iso8601,
+      /// RFC 3339.
+      Rfc3339,
```
ahsonkhan, line 193: Based on this (https://github.com/Azure/azure-sdk-for-cpp/issues/776#issuecomment-724582360), it looks like we need to support the extended ISO 8601-1:2019 profile with all levels of granularity, and not just the RFC 3339 subset.

That is what the current implementation is doing anyway as shown by this test:
https://github.com/Azure/azure-sdk-for-cpp/blob/a2ddbc8cc913ff123c186bb32ab8b72f8d6265f0/sdk/core/azure-core/test/ut/datetime.cpp#L47-L53

For example, according to RFC 3339, this is invalid: `"yyyy'-'MM'-'dd"`

Therefore, we should revert this back to ISO.

```suggestion
      /// ISO 8601.
      Iso8601,
```

For additional context:
https://docs.microsoft.com/en-us/dotnet/standard/datetime/system-text-json-support#support-for-parsing

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -575,30 +583,31 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
       }
 
       secondsSince1601
           = AdjustTimezone(secondsSince1601, static_cast<unsigned char>(tzCh), tzHours, tzMinutes);
 
       if (secondsSince1601 < 0)
       {
-        return result;
+        throw DateTimeException(
+            "Error parsing DateTime: year is < 1601 ather time zone adjustments.");
```
ahsonkhan, line 595:
```suggestion
            "Error parsing DateTime: year is < 1601 after time zone adjustments.");
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: Should we use explicit inttypes (http://www.cplusplus.com/reference/cstdint/) like int32_t and int64_t in public surface area instead of int/long long? We use the inttypes in C as well.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -113,107 +244,68 @@ namespace Azure { namespace Core {
...
-    std::string ToString(DateFormat format = DateFormat::Rfc1123) const
+    std::string GetString(DateFormat format) const
     {
-      return ToString(format, TimeFractionFormat::DropTrailingZeros);
+      return GetString(format, TimeFractionFormat::DropTrailingZeros);
     };
 
     /**
-     * @brief Get a string representation of the @DateTime formated with ISO8601.
+     * @brief Get a string representation of the @DateTime formated with ISO 8601.
```
ahsonkhan, line 253:
```suggestion
     * @brief Get a string representation of the @DateTime formatted with ISO 8601.
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -57,53 +196,45 @@ namespace Azure { namespace Core {
...
+     * @brief Construct an instance of @DateTime.
+     *
+     * @param year Year.
+     * @param month Month.
+     * @param day Day.
+     * @param hour Hour.
+     * @param minute Minute.
+     * @param second Seconds.
+     *
+     * @throw DateTimeException If any paramter is invalid.
```
ahsonkhan, line 211:
```suggestion
     * @throw DateTimeException If any parameter is invalid.
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: See https://azure.github.io/azure-sdk/cpp_design.html#integer-sizes

antkmsft: As I said, for the data that does not gets sent over the wire, I don't think we have to. But to move on for now, I updated the types.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,161 @@
...
+      constexpr Duration() : m_100nsIntervals(0) {}
+
+      static constexpr Duration FromHours(int32_t hours)
+      {
+        return Duration(hours * IntervalsOf100nsPerHour);
+      }
+
+      void constexpr AddHours(int32_t hours)
+      {
+        m_100nsIntervals += (hours * IntervalsOf100nsPerHour);
```
ahsonkhan, line 61: We probably need integer overflow checks here/elsewhere. Otherwise, a large enough value could "subtract" hours instead.

Verify the behavior for all such public APIs in this PR which are used for multiplication/addition.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,161 @@
...
+      }
+
+      static constexpr Duration FromMinutes(int64_t minutes)
```
ahsonkhan, line 64: Add a comment why these smaller time values are int64_t while `FromHours` and larger time values are int32_t.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,161 @@
...
-    /// A type that represents tick spans.
-    typedef uint64_t IntervalType;
-
-  private:
-    //  Number of seconds between 01-01-1970 and 01-01-1601.
-    static constexpr IntervalType WindowsToPosixOffsetSeconds = 11644473600LL;
+    /**
+     * @brief Represents time duration.
+     */
+    class Duration {
```
ahsonkhan, line 37: Why is this class nested within DateTime rather than it being standalone?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,161 @@
...
+      {
+        if (nanoseconds % NanosecondResolution != 0)
+        {
+          throw DateTimeException("Can't convert from nanoseconds");
+        }
+
+        return Duration(nanoseconds / NanosecondResolution);
+      }
+
+      void constexpr AddNanoseconds(int64_t nanoseconds)
```
ahsonkhan, line 114: Add input validation for negative values?

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: It is an implementation detail, I do not want to make this a part of the API surface by documenting it.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: It is known behavior for integers to behave this way. And this is not a regression from the previous implementation.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: It does not have much use outside DateTime, they are coupled, and it is the same scope as prefious typedef uint64_t duration.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: No, since we only have Add() method, the way you subtract is by providing negative value to add.

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: `FromMinutes` is taking an int64_t but `FromHours` is taking int32_t. Those are public API differences already, no?

All that said, I am suggesting a `//` comment explaining the reason for this implementation detail, not doxygen doc comments.

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: I see. Gotcha. Do we have public API documentation explaining this?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,161 @@
...
+      static constexpr int64_t IntervalsOf100nsPerMillisecond
+          = 1000 * IntervalsOf100nsPerMicrosecond;
+      static constexpr int64_t IntervalsOf100nsPerSecond = 1000 * IntervalsOf100nsPerMillisecond;
+      static constexpr int64_t IntervalsOf100nsPerMinute = 60 * IntervalsOf100nsPerSecond;
+      static constexpr int64_t IntervalsOf100nsPerHour = 60 * IntervalsOf100nsPerMinute;
+
+    public:
+      constexpr Duration() : m_100nsIntervals(0) {}
+
+      static constexpr Duration FromHours(int32_t hours)
```
ahsonkhan, line 54: Add doxygen docs to all public surface area.

ahsonkhan (approved):
Other than adding docs to public APIs, looks good.

There are a couple of other issues around the implementation detail, but they are out of the scope of this PR and need more review.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,19 +12,161 @@
 #include <string>
 
 namespace Azure { namespace Core {
+
+  /**
+   * @brief An exception that gets thrown when @DateTime error occurs.
+   */
+  class DateTimeException : public std::runtime_error {
```
JeffreyRichter, line 19: For recoverable errors, we should be inheriting from std::exception.
For non-recoverable (precondition-like) errors, we should be inheriting from std::runtime_error.
We should double check this for all exception types we define.

sdk/core/azure-core/inc/azure/core/datetime.hpp
JeffreyRichter: I still question the nested-ness. I get that these frequently used together but this isn't mandatory. For example, we could have a function that accepts a timeout as a Duration (no DateTime is used).

sdk/core/azure-core/inc/azure/core/datetime.hpp
JeffreyRichter: I hope these are private but I find it hard to believe that we use all of these. They don't seem necessary to me.

sdk/core/azure-core/inc/azure/core/datetime.hpp
JeffreyRichter: I didn't think we were going to have functions like this at all.
I thought we could just set a Duration variable like this: Duration d = 2 * Duration::Hour

And then, instead of "AddHours" function just do:  d += 3 * Duration:Hour

So, I just would have expected a really simple Duration type that wraps a signed int64, overloads a few operators (+, -, ==, !=, <, >, <=, >=) and defines some constants (Second, Minutes, Hour) for whatever units the  int64 is in. Nothing else; no methods.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -50,6 +52,9 @@ struct ComputeYearResult
   int SecondsLeftThisYear;
 };
 
+constexpr auto MinYear = 1601;
+constexpr auto MaxYear = 9999;
+
 constexpr int SecondsInMinute = 60;
 constexpr int SecondsInHour = SecondsInMinute * 60;
 constexpr int SecondsInDay = 24 * SecondsInHour;
```
JeffreyRichter, line 60: Do not have symbols for days; not all days have 24 hours in them.
Do we need these symbols at all?

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -50,6 +52,9 @@ struct ComputeYearResult
   int SecondsLeftThisYear;
 };
 
+constexpr auto MinYear = 1601;
+constexpr auto MaxYear = 9999;
```
JeffreyRichter, line 56: I don't think these need to be public. we can just document the range.

ahsonkhan: I agree. However, they aren't public. Just named constants as part of implementation detail.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -113,107 +247,68 @@ namespace Azure { namespace Core {
...
-     * @brief Get a string representation of the @DateTime formated with ISO8601.
+     * @brief Get a string representation of the @DateTime formatted with ISO 8601.
      *
-     * @param fractionFormat The format that is applied to the fraction part from the ISO8601 date.
+     * @param fractionFormat The format that is applied to the fraction part from the ISO 8601 date.
      *
      * @throw DateTimeException If year exceeds 9999, or if \p fractionFormat is not recognized.
      */
-    std::string ToIso8601String(TimeFractionFormat fractionFormat) const
+    std::string GetIso8601String(TimeFractionFormat fractionFormat) const
```
RickWinter, line 262: This should be an overload of GetString with a DateFormat

ahsonkhan: I provided similar feedback here: https://github.com/Azure/azure-sdk-for-cpp/pull/936#discussion_r520269392

Which is why this seems out of the scope of the PR.

antkmsft: RFC 1123 does not have fractional seconds / TimeFractionFormat has no effect with it, so there are two methods: `GetString(DateTimeFormat)`, and `GetIso8601String(TimeFractionFormat)`.

sdk/core/azure-core/src/datetime.cpp
antkmsft: These are private, not visible to the outside. They are used by the cpprest's implementation which we derived ours from. There will likely be a time when we rewrite it, but I would not touch them now, cpprest works as we need it to. Good thing we have tests, so we later can make these modifications. In cpprest's logic, days are 24 hours, and it works correctly.

I think it is okay to assume that days have 24 hours in limited contexts. Days don't have 24 hours in DST-affected time, but we store time as UTC. Same goes for leap seconds - this class can parse them, but does not store them. In a context of this class being able to represent calendar dates for sending them over the wire, it works. This class operates on calendar dates, durations are not meant to be precisely accurate in the terms of physical time durations. Nevertheless, it works accurately for the time zones/DST, and leap years. It does not work timewise accurately for leap seconds, but calendar-wise, it is accurate. It is also probably not perfectly accurate from historical point either, but  for representing the datetimes in ISO/RFC formats, for information exchange, it is good implementation. For instance, in ISO format, you can create a date that is prior to the date when Gregorian calendar was even created, not to mention rules for leap days.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -13,7 +13,7 @@
 - Added `GetPort()` to `Url`.
 - Added `TransportPolicyOptions`.
 - Added `TelemetryPolicyOptions`.
-- Added `DateFormat::ToIso8601String(TimeFractionFormat)`.
+- Added `DateFormat::ToRfc3339String(TimeFractionFormat)`.
```
RickWinter, line 16: I would expect we have ToString with an overload that takes a format.  Is there a reason to have a specific method named for this?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,170 +46,195 @@ namespace Azure { namespace Core {
...
      *
-     * @throw DateTimeException If year exceeds 9999, or if \p format is not recognized.
+     * @throw std::length_error If year exceeds 9999.
+     * @throw std::invalid_argument If \p format is not recognized.
      */
-    std::string ToString(DateFormat format = DateFormat::Rfc1123) const
+    std::string GetString(DateFormat format) const
     {
-      return ToString(format, TimeFractionFormat::DropTrailingZeros);
+      return GetString(format, TimeFractionFormat::DropTrailingZeros);
```
RickWinter, line 114: Why GetString over ToString?

sdk/core/azure-core/CHANGELOG.md
antkmsft: It's because we only support this for RFC 3339. Other RFC doesn't even have fractional seconds.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: This is what we decided during the API review. There was no clear winner, but we wanted consistency. Uuid had GetString(), so this one has the same name.

antkmsft: We followed up in the chat, we will discuss wit the team whether all GetString()s should be ToString()s.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -127,19 +133,20 @@ constexpr char const monthNames[] = "Jan\0Feb\0Mar\0Apr\0May\0Jun\0Jul\0Aug\0Sep
...
 
-std::string DateTime::ToString(DateFormat format, TimeFractionFormat fractionFormat) const
+std::string DateTime::GetString(DateFormat format, TimeFractionFormat fractionFormat) const
 {
-  if (m_interval > 2650467743999999999LL)
+  static constexpr auto const EndOfYear9999 = 2650467743999999999LL;
+  if (m_since1601.count() > EndOfYear9999)
   {
-    throw DateTimeException("The requested year exceeds the year 9999.");
+    throw std::length_error("The requested year exceeds the year 9999.");
```
ahsonkhan, line 141: format error instead?

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -774,9 +779,84 @@ DateTime DateTime::FromString(std::string const& dateString, DateFormat format)
...
+  {
+    throw std::invalid_argument("Invalid hour value.");
+  }
+
+  if (minute < 0 || minute > 60)
+  {
+    throw std::invalid_argument("Invalid minute value.");
+  }
+
+  if (second < 0 || second > 60)
```
ahsonkhan, line 825: Is 60 valid?

sdk/core/azure-core/CHANGELOG.md
ahsonkhan: Why don't we have just a single method, and throw for format enum values that don't mix-match well? Having different named methods like this doesn't scale if and when we add more format support and options for formatting.

```C#
// If the user provides incompatible arguments - validate them and throw:
GetString(DateFormat, TimeFractionFormat);
```

Ideally, we would just have one options paramater:
```C#
GetString(DateFormatOptions = Some Default Value);
```

ahsonkhan (approved):
Some small nits. However, we should be keeping just one `GetString()` method and not have other named ones.

Barring that issue, looks good.

sdk/core/azure-core/src/datetime.cpp
BillyONeal: +1 -- length_error is "you tried to form a container whose length is unrepresentable" e.g. `std::string("hello").append(SIZE_MAX, 'a')`

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,170 +46,195 @@ namespace Azure { namespace Core {
...
     /**
      * @brief Get a string representation of the @DateTime.
      *
      * @param format The representation format to use.
      * @param fractionFormat The format for the fraction part of the Datetime. Only supported by
-     * ISO8601
+     * RFC 3339.
      *
-     * @throw DateTimeException If year exceeds 9999, or if \p format is not recognized.
+     * @throw std::length_error If year exceeds 9999.
```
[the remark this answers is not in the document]
BillyONeal: It looks like it was documented because it was explicitly thrown

sdk/core/azure-core/src/datetime.cpp
antkmsft: format_error is C++20.

sdk/core/azure-core/CHANGELOG.md
antkmsft: Why throw, if the exception can be prevented in compile-time?
But let's discuss the formatting overall, later.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -17,28 +17,24 @@ namespace Azure { namespace Core {
...
-    /// A type that represents tick spans.
-    typedef uint64_t IntervalType;
-
-  private:
-    //  Number of seconds between 01-01-1970 and 01-01-1601.
-    static constexpr IntervalType WindowsToPosixOffsetSeconds = 11644473600LL;
+    /**
+     * @brief Units of measurement the difference between instances of @DateTime.
+     */
+    typedef std::chrono::duration<int64_t, std::ratio<1, 10000000>> Duration;
```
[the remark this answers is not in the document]
antkmsft: 1 / 10,000,000 of a second. 1 equals to 100ns.

sdk/core/azure-core/src/datetime.cpp
ahsonkhan: Missed that. `runtime_error` then? The main point is, pick a reasonable exception type here, that fits, other than `length_error`.

BillyONeal: I'd probably make your own type for this (IIRC at one time there was a plan for a shared az_core exception hierarchy? I don't think this integer overflow case matches any of the existing ones well)

sdk/core/azure-core/inc/azure/core/datetime.hpp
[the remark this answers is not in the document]
antkmsft: format_error is C++20. we have 3 candidates - length_error, out_of_range, and invalid_argument. I did not want to throw invalid_argument when it is not based on the method input parameter (assuming `this` is not one of them).
I see if you don't like `length_error`, you probably won't like `out_of_range` either, so let me to throw only `invalid_argument`.

sdk/core/azure-core/src/datetime.cpp
antkmsft: format_error is C++20. we have 3 candidates - length_error, out_of_range, and invalid_argument. I did not want to throw invalid_argument when it is not based on the method input parameter (assuming `this` is not one of them).
I see if you don't like `length_error`, you probably won't like `out_of_range` either, so let me to throw only `invalid_argument`.

sdk/core/azure-core/src/datetime.cpp
antkmsft: 60 is allowed, it is leap seconds, parser would accept it. But not for minutes, good catch.

sdk/core/azure-core/src/datetime.cpp
antkmsft: @BillyONeal, yesterday we haven't decided finally, but leaning heavily towards not having the AzureSDK exception hierarchy, and even reusing std::exception where suitable. It would've been great if you participated in that discussion. But so far, yesterday we decided on throwing plain std:: exceptions in this case.

sdk/core/azure-core/CHANGELOG.md
vhvb1989: >  this doesn't scale if and when we add more format support

What if adding another date format now also support/requires other specific TimeFractionFormat or other type of options?
Will you add one extra parameter to `GetString` and now validate the combination of 3 parameters?
The way that escalates is nasty.

Agree with Anton, if we can prevent things that shouldn't be called from compilation time, let's do it like that. Otherwise, what's the benefit giving valid enum input that would do nothing but fail?

antkmsft:
> Some small nits. However, we should be keeping just one `GetString()` method and not have other named ones.
Thank you for the review! I think we should discuss GetRfc3339String() it and not fix it as part of this PR.
Also, Rick thinks we should name `GetString()` as `ToString()` (for other classes as well), and I agree with him.

sdk/core/azure-core/inc/azure/core/datetime.hpp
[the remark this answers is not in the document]
antkmsft: I'll put a non-doxygen comment.

sdk/core/azure-core/CHANGELOG.md
ahsonkhan: > Why throw, if the exception can be prevented in compile-time?

Trade-off to avoid API explosion.

> What if adding another date format now also support/requires other specific TimeFractionFormat or other type of options?

This is why the single struct argument that captures all future customizations is most extensible. We have this pattern elsewhere in the SDK, so customers should be familiar with it.

What would you do in that case? Add more differently named methods taking various other enums and options? So, the DateTime will have a growing list of GetXStringY(some parameter) APIs?

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Sorry I didn't leave a comment, but resolved. Yes - any line expansion is/was due to autoformatting.

sdk/core/azure-core/CHANGELOG.md
antkmsft: I'm not opposed to having full formatString support; But I realize it is likely never going to be used.

antkmsft: Jeffrey opposed to having default value, at least in this specific case.

I don't think we are going to reach an agreement on this today, more discussions are needed. I unresolved all the comments that might provide food for discussions / info for the future - on this topics and others.

I am going to complete this PR. After all, GetRfc3339String() is not a breaking change, should we want to make `GetString(DateFormat, Fractionaloptions)` public in the future. Maybe it won't look as elegant, but technically we even still can be making breaking changes in this stage.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -20,6 +20,7 @@ namespace Azure { namespace Core {
     /**
      * @brief Units of measurement the difference between instances of @DateTime.
      */
+    // 1 == 100 ns (1 / 10,000,000 of a second, 7 fractional digits).
```
ahsonkhan, line 23: Good comment!

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -95,8 +96,7 @@ namespace Azure { namespace Core {
      * @param fractionFormat The format for the fraction part of the Datetime. Only supported by
      * RFC 3339.
      *
-     * @throw std::length_error If year exceeds 9999.
-     * @throw std::invalid_argument If \p format is not recognized.
+     * @throw std::length_error If year exceeds 9999, or if \p format is not recognized.
```
ahsonkhan, line 99: wrong exception type :(

should be `invalid_argument`

antkmsft: Oops, the comment is wrong.

antkmsft: https://github.com/Azure/azure-sdk-for-cpp/pull/950

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Doxygen does provide values for defaults: 
image

image

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -50,170 +47,193 @@ namespace Azure { namespace Core {
...
+    }
 
-    /// Checks whether this instance of @DateTime is initialized.
-    bool IsInitialized() const { return m_interval != 0; }
+    /**
+     * @brief Get this @DateTime representation as a @Duration from the start of the
+     * implementation-defined epoch.
+     * @return @Duration since the start of the implementation-defined epoch.
+     */
+    constexpr explicit operator Duration() const { return m_since1601; }
```
JeffreyRichter, line 232: Do we need this? Why would someone want to all this?

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: BTW, our implementation would allow it, see line 47 in the unit tests:
```
    auto dt = DateTime::Parse("2013-05-17", DateTime::DateFormat::Rfc3339);
    EXPECT_NE(0, static_cast<DateTime::Duration>(dt).count());
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: I think you're right, we can remove it. It was a map of the cpprest's Interval() method. It is used in tests, but I don't see customer using it. For the tests, we still can get it by doing `auto duration(dateTime - DateTime(1601))`.

antkmsft: https://github.com/Azure/azure-sdk-for-cpp/pull/967

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: > What makes you think that 'yyyy'-'MM'-'dd' is invalid for RFC 3339?

Because of what the ABNF says and things like:
https://validator.w3.org/feed/docs/error/InvalidRFC3339Date.html
> Note that RFC 3339 is more strict than the W3C Date and Time Format. In particular, all portions of the date and time must be present, with the exception of the fractional portions of the second, which may be omitted.

https://stackoverflow.com/a/522281/12509023
> Most notably RFC 3339 specifies a complete representation of date and time (only fractional seconds are optional).

But maybe that's semantics, and maybe our parser can be more lenient and allow it. That would be OK.

Its possible the strict adherence to what RFC 3339 is what a customer expects, as a feature, of our DateTime, so either we reject 2020-11-12, or allow those sub-forms, and document it explicitly. In that case, it seems strange to use that name for the enum value. The Cpp Rest SDK called it ISO 8601 and supported that flexibility in parsing. Why is that?

JeffreyRichter: OK, well the Azure HTTP guidelines explicitly state to use RFC 3339 and so I think we can expect the time portion and fail if the string is not truly RFC 3339.

antkmsft: I don't see any value on failing when just "yyyy-MM-dd" is provided, and I also don't see anything in the links above that could mean that yyyy-MM-dd is incorrect.

ahsonkhan: > I also don't see anything in the links above that could mean that yyyy-MM-dd is incorrect.

Hmm... 

@antkmsft - I quoted the relevant snippets in my post above:

>  In particular, **all portions** of the date and time **must be present**

PR #946, opened 2020-11-11
Resolve review comments from Core team.

katmsft:


sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
```diff
@@ -342,7 +342,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
   {
     std::string LastModified;
     std::string Etag;
-    int32_t Quota = int32_t();
+    int64_t Quota = int64_t();
```
Jinming-Hu, line 345: this is in GB, right? I think int32 is enough

Jinming-Hu, line 345: Why you changed the other to `QuotaInGiB` while keep this unchanged?

Jinming-Hu (approved):
some minor comments

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
katmsft: This is generated by swagger and swagger is calling it Quota. It is not visible to customers as well.

katmsft: But the service does not doc that it is an int32 value. Maybe this SDK is still used 100 years later and Quota exceeds what int32 supports.

PR #965, opened 2020-11-12
Moved model types in file share service into Models namespace.

katmsft:


sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_file_attribute.hpp
```diff
@@ -5,207 +5,213 @@
...
-    lhs = lhs & rhs;
-    return lhs;
-  }
-
-  inline FileAttributes FileAttributesFromString(const std::string& fileAttributesString)
-  {
-    FileAttributes result = static_cast<FileAttributes>(0);
-
-    if (fileAttributesString == "ReadOnly")
+  namespace Models {
```
Jinming-Hu, line 8: Are you sure you've re-formatted with clang-format? It seems to me that this `namespace Models {` line should come in the same line as other namespaces

Jinming-Hu (approved):

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_file_attribute.hpp
katmsft: It shouldn't be, there are `Models` and `Details` namespace within this file under namespace `Azure::Storage::Files::Shares`, that will new up a line for `Models` or `Details` namespace.

Issue #742, opened 2020-10-12
latest master branch build failure


Jinming-Hu:
```
[ 15%] Built target azure-core
[ 25%] Built target azure-storage-common
[ 30%] Built target azure-storage-files-shares
[ 38%] Built target azure-storage-blobs
[ 46%] Built target azure-storage-files-datalake
[ 50%] Built target azure-storage-sample
[ 52%] Built target gtest
[ 54%] Built target gmock
[ 56%] Built target gtest_main
[ 76%] Built target azure-storage-test
[ 86%] Built target azure-core-test
[ 88%] Built target azure_hang
[ 89%] Linking CXX executable azure_core_storage_issue_248
/usr/bin/ld: ../../libazure-core.a(curl.cpp.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [sdk/core/azure-core/test/e2e/CMakeFiles/azure_core_storage_issue_248.dir/build.make:105: sdk/core/azure-core/test/e2e/azure_core_storage_issue_248] Error 1
make[1]: *** [CMakeFiles/Makefile2:597: sdk/core/azure-core/test/e2e/CMakeFiles/azure_core_storage_issue_248.dir/all] Error 2
make: *** [Makefile:160: all] Error 2
```

RickWinter:
We are unable to reproduce this failure.  If you have a repro, please provide details on the OS, targeted arch, and any build flags/options you are passing.

Jinming-Hu:
```
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON
make -j
```

Jinming-Hu:
Ubuntu 20.04 LTS
gcc version 9.3.0
cmake version 3.18.0

Jinming-Hu:
I can also reproduce it on CentOS

Jinming-Hu:
seems it's caused by not linking to pthread.

can core sdk try to link to pthread like this

```
set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
set(THREADS_PREFER_PTHREAD_FLAG TRUE)
find_package(Threads REQUIRED)
target_link_libraries(YOUR_TARGET Threads::Threads)
```

[...]

PR #980, opened 2020-11-13
Fix linker error of missing pthread on Linux

lordgamez:
Fixes error of issue #742

### DIFF
--- a/sdk/core/azure-core/CMakeLists.txt
+++ b/sdk/core/azure-core/CMakeLists.txt
@@ -8,9 +8,13 @@ project(${TARGET_NAME} LANGUAGES CXX)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
+set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads REQUIRED)
+
 # min version for `CURLSSLOPT_NO_REVOKE`
 # https://curl.haxx.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
-set(CURL_MIN_REQUIRED_VERSION 7.44) 
+set(CURL_MIN_REQUIRED_VERSION 7.44)
 find_package(CURL ${CURL_MIN_REQUIRED_VERSION} CONFIG)
 if(NOT CURL_FOUND)
   find_package(CURL ${CURL_MIN_REQUIRED_VERSION} REQUIRED)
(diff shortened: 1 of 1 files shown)


Jinming-Hu (approved):

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -8,9 +8,13 @@ project(${TARGET_NAME} LANGUAGES CXX)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
+set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads REQUIRED)
```
ahsonkhan, line 13: Is this a linux only package? If so, should this be under a non-windows ifdef?

Jinming-Hu: https://cmake.org/cmake/help/latest/module/FindThreads.html

It's **NOT** a linux only package. So it shouldn't under a non-windows ifdef.

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -58,7 +62,7 @@ add_library (Azure::Core ALIAS ${TARGET_NAME})
 create_code_coverage(core ${TARGET_NAME} "${TARGET_NAME}-test")
 
 target_include_directories(${TARGET_NAME} PUBLIC ${CURL_INCLUDE_DIRS})
-target_link_libraries(${TARGET_NAME} PRIVATE CURL::libcurl)
+target_link_libraries(${TARGET_NAME} PRIVATE CURL::libcurl Threads::Threads)
```
ahsonkhan, line 65: `<thread>` is used in storage retry policy as well. Should we link `::Threads` there too, or is it sufficient to rely on azure-core linking it (which storage takes a dependency on)?

sdk/core/azure-core/CMakeLists.txt
ahsonkhan: I understand the `Threads` package isn't windows only. But what about the `pthread` related CMAKE flags. Those are posix only, right? What do `CMAKE_THREAD_PREFER_PTHREAD` and `THREADS_PREFER_PTHREAD_FLAG` being true mean on Windows? Are they ignored or will they look for pthread specifically and fail? Presumably on windows, the call to `find_package` will set `CMAKE_USE_WIN32_THREADS_INIT`.

sdk/core/azure-core/CMakeLists.txt
ahsonkhan: > Should we link `::Threads` there too, or is it sufficient to rely on azure-core linking it (which storage takes a dependency on)?

I just realized, storage common already links the `::Threads` library.

https://github.com/Azure/azure-sdk-for-cpp/blob/a716e094fe4c17f872181cc755e13843f53ead8d/sdk/storage/azure-storage-common/CMakeLists.txt#L44-L50

ahsonkhan (approved):

ahsonkhan:
Thanks for the fix and contribution @lordgamez, and thank you @Jinming-Hu for the proposed solution.

sdk/core/azure-core/CMakeLists.txt
Jinming-Hu: whatever. It's a code change to core sdk. You guys decide what do with it.

Issue #964, opened 2020-11-12
Default EndpointSuffix used in storage connection string causes failure


lordgamez:
**Describe the bug**
Instantiating a BlobContainerClient (or any storage client) using a storage connection string with missing EndpointSuffix creates a faulty client with invalid endpoint. Changing the default EndpointSuffix value at sdk/storage/azure-storage-common/src/storage_credential.cpp:54 from ".core.windows.net" to "core.windows.net" fixes the issue.

***Exception or Stack Trace***
The problem is due to an additional '.' character in the default EndpointSuffix:
```
AzureBlobStorageTests.cpp:26: FAILED:
due to unexpected exception with message:
  Fail to get a new connection for: ggyimesiminifi.blob..core.windows.net.
  Couldn't resolve host name
```

**To Reproduce**
1. Create connection string without EndpointSuffix
2. Initialize BlobContainerClient
3. Try to upload blob

***Code Snippet***
```
std::string connection_string = "DefaultEndpointsProtocol=https;AccountName=<account>;AccountKey=<key>";  // No EndpointSuffix
std::string container_name = "test-container";
std::string data = "data";
std::vector<uint8_t> buffer(data.begin(), data.end());
auto container_client = Azure::Storage::Blobs::BlobContainerClient::CreateFromConnectionString(connection_string, container_name)
auto blob_client = container_client.GetBlockBlobClient("blob.txt");
Azure::Core::Http::MemoryBodyStream data(buffer);
blob_client.Upload(&data);
```

**Expected behavior**
Upload should succeed.

**Screenshots**
-

**Setup (please complete the following information):**
 - OS: Ubuntu 20.04
 - IDE : VSCode
 - Version of the Library used: azure-storage-blobs_1.0.0-beta.4

**Additional context**
-

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [X] Bug Description Added
- [X] Repro Steps Added
- [X] Setup information Added

Jinming-Hu:
fixed by https://github.com/Azure/azure-sdk-for-cpp/pull/981

PR #981, opened 2020-11-13
Fix default EndpointSuffix when parsing connection string

lordgamez:
Creating a storage client with a connection string where no EndpointSuffix is defined results in faulty service URI containing two '.' characters between the uri and the suffix. This goal of this change is to correct this issue. Addresses issue #964

### DIFF
--- a/sdk/storage/azure-storage-common/CMakeLists.txt
+++ b/sdk/storage/azure-storage-common/CMakeLists.txt
@@ -69,6 +69,7 @@ target_sources(
     PRIVATE
     test/bearer_token_test.cpp
     test/crypt_functions_test.cpp
+    test/storage_credential_test.cpp
     test/test_base.cpp
     test/test_base.hpp
 )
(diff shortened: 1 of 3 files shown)


ahsonkhan:
@Jinming-Hu  and @katmsft please take a look.

I am wondering, what happens if the `EndpointSuffix` is provided? Would we thn miss the `.` being removed in this change?

Jinming-Hu:
> @Jinming-Hu and @katmsft please take a look.
> 
> I am wondering, what happens if the `EndpointSuffix` is provided? Would we thn miss the `.` being removed in this change?

i'm going to take a look next Monday

Jinming-Hu (approved):

Jinming-Hu:
> I am wondering, what happens if the `EndpointSuffix` is provided? Would we thn miss the `.` being removed in this change?

```
DefaultEndpointsProtocol=https;AccountName=targettest;AccountKey=***;EndpointSuffix=core.windows.net
```

This is what connection string looks like on Azure Portal. It comes without that `.` before `core.windows.net`

PR #989, opened 2020-11-16
unify blob client options

Jinming-Hu:


Jinming-Hu:
@katmsft Please fix build errors in datalake service

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_container_client.hpp
```diff
@@ -313,8 +313,12 @@ namespace Azure { namespace Storage { namespace Blobs {
   private:
     explicit BlobContainerClient(
         Azure::Core::Http::Url containerUri,
-        std::shared_ptr<Azure::Core::Http::HttpPipeline> pipeline)
-        : m_containerUrl(std::move(containerUri)), m_pipeline(std::move(pipeline))
+        std::shared_ptr<Azure::Core::Http::HttpPipeline> pipeline,
+        Azure::Core::Nullable<EncryptionKey> customerProvidedKey,
```
katmsft, line 317: `Nullable` parameters cannot be ignored when being called. Consider `Azure::Core::Nullable<EncryptionKey> customerProvidedKey = Azure::Core::Nullable<EncryptionKey>()`.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_container_client.hpp
```diff
@@ -313,8 +313,12 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+        Azure::Core::Nullable<std::string> encryptionScope)
```
katmsft, line 318: Ditto.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_container_client.hpp
Jinming-Hu: This is a **private** constructor for internal use only.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_container_client.hpp
Jinming-Hu: ditto

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_container_client.hpp
katmsft: Should every caller that calls this function pass in those values? If not, I'm OK with the change.

Jinming-Hu: > Should every caller that calls this function pass in those values?

Yes, that's why it is in parameter list without any default value.

katmsft: I don't think the statement is valid. DataLake service does not support this feature. Checked .NET/JS. The parameters are meaningless at least for DataLake service.

Jinming-Hu: > I don't think the statement is valid.

which statement specifically?

> DataLake service does not support this feature

which feature?

katmsft: > > Should every caller that calls this function pass in those values?
> 
> Yes, that's why it is in parameter list without any default value.

This statement.

DataLake does not support CPK, so when calling this API, DataLake should not be aware of what it is.

Jinming-Hu: Fine, but I don't think adding default value for those two parameters actually solves this problem.

I do think the caller should be aware of every parameter of the functions it calls.

katmsft: > I do think the caller should be aware of every parameter of the functions it calls.
I think we can agree to disagree. :)

Jinming-Hu: > > I do think the caller should be aware of every parameter of the functions it calls.
> 
> I think we can agree to disagree. :)

If a blob container has enabled encryption-scope, then it's not accessible anymore to DataLake service. This doesn't make any sense. So it's foreseeable that DataLake service is also going to support encryption scope to improve interoperaibility. Based on this premise, my proposal would be DataLake calls this constructor with `Azure::Core::Nullable<EncryptionKey>()`, which will reduce our workload once these encryption stuff is supported in DataLake.

katmsft: > which will reduce our workload once these encryption stuff is supported in DataLake.

If we make them optional with default value at the moment, once datalake support it default value should be removed, then compile error would tell us that DataLake should also be changed to put them in from client options. I think this approach would be better tbh.

Anyway, feel free to leave it as is, not a major concern.

Jinming-Hu: I don't see why there will be compiler error

katmsft (approved):

PR #1034, opened 2020-11-30
block id list now is a vector of strings

Jinming-Hu:


RickWinter (commented):
Add an entry into the Changelog

vhvb1989 (approved):
Add changelog if this is a breaking change

Issue #1026, opened 2020-11-25
rename variables for access condition


Jinming-Hu:
```C++
Nullable<int64_t> MaxSize;
```
#
In .NET we call this `IfMaxSizeLessThanOrEqual`.

```C++
Nullable<int64_t> AppendPosition;
```
#
In .NET we call this `IfAppendPositionEqual`.

originally posted by @tg-msft

Jinming-Hu:
I think `IfMaxSizeLessThanOrEqualTo` might be better than `IfMaxSizeLessThanOrEqual`. What do you guys think?

PR #1035, opened 2020-11-30
rename variables for append blob access condition

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/1026


vhvb1989 (approved):
changelog for this breaking change before merging please

Issue #1031, opened 2020-11-27
CreateIfNotExists, DeleteIfExists


Jinming-Hu:
```c  
Azure::Storage::Blobs::BlobContainerClient.Create:const_Azure::Storage::Blobs::CreateContainerOptions_&::Azure::Core::Response<CreateContainerResult>
```
#
No `CreateIfNotExists`?

PR #1059, opened 2020-12-02
CreateIfNotExists and DeleteIfExists

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/1031

### DIFF
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
@@ -118,6 +118,17 @@ namespace Azure { namespace Storage { namespace Blobs {
         const CreateAppendBlobOptions& options = CreateAppendBlobOptions()) const;
 
     /**
+     * @brief Creates a new 0-length append blob. The content keeps unchanged if the blob already
+     * exists.
+     *
+     * @param options Optional parameters to execute this function.
+     * @return A CreateAppendBlobResult describing the newly created append blob. Null if the blob
+     * already exists.
+     */
+    Azure::Core::Response<Models::CreateAppendBlobResult> CreateIfNotExists(
+        const CreateAppendBlobOptions& options = CreateAppendBlobOptions()) const;
+
+    /**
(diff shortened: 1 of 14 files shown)


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
```diff
@@ -117,6 +117,17 @@ namespace Azure { namespace Storage { namespace Blobs {
     Azure::Core::Response<Models::CreateAppendBlobResult> Create(
         const CreateAppendBlobOptions& options = CreateAppendBlobOptions()) const;
 
+    /**
+     * @brief Creates a new 0-length append blob. The content keeps unchanged if the blob already
+     * exists.
+     *
+     * @param options Optional parameters to execute this function.
+     * @return A CreateAppendBlobResult describing the newly created append blob. Null if the blob
```
katmsft, line 125: I think Null is a bit misleading to our customers. How about `Raw response if the blob already exists.`

Also, we agreed to remark that not checking if the response has value before accessing it may cause process to crash. Please also document that.

katmsft (changes requested):

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
Jinming-Hu: it always returns the raw response

Jinming-Hu: plus it aligns with .Net https://github.com/Azure/azure-sdk-for-net/blob/dfbc7aa9ee5f053d2d84f85d4d142152886f53ed/sdk/storage/Azure.Storage.Blobs/src/BlobContainerClient.cs#L822

katmsft (approved):
Approve with comment.

Jinming-Hu:
> Approve with comment.

Apparently we cannot come up with a good doc description that satisfies both of us. In this case, I'd suggest we move forward merging this PR, and come back to discuss again once anybody has some better idea.

PR #1072, opened 2020-12-03
Fix MSVC warning D9025 ("overriding '/W3' with '/W4'")

antkmsft:
See https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin

--

### DIFF
--- a/cmake-modules/global_compile_options.cmake
+++ b/cmake-modules/global_compile_options.cmake
@@ -8,6 +8,9 @@ if(MSVC)
     set(WARNINGS_AS_ERRORS_FLAG "/WX")
   endif()
 
+  # https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin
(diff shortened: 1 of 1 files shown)


RickWinter (approved):

cmake-modules/global_compile_options.cmake
```diff
@@ -8,6 +8,9 @@ if(MSVC)
     set(WARNINGS_AS_ERRORS_FLAG "/WX")
   endif()
 
+  # https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin
+  string(REGEX REPLACE "/W3" "/W4" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
ahsonkhan, line 12: So, if I understand this correctly, we are looking at instances of `/W3` in `CMAKE_CXX_FLAGS`, and replacing them with `/W4`.

**Questions:**
1) Why do we need to do this when below on line 15 we are adding the `/W4` flag anyway?
2) What happens if no instances of `/W3` are found. It isn't clear from the docs: https://cmake.org/cmake/help/latest/command/string.html#regex-replace

PR #1078, opened 2020-12-04
Make SDK buildable in VS 2017

antkmsft:

### DIFF
--- a/sdk/core/azure-core/src/http/body_stream.cpp
+++ b/sdk/core/azure-core/src/http/body_stream.cpp
@@ -23,6 +23,7 @@
 #include <stdexcept>
 #include <vector>
 
+using Azure::Core::Context;
 using namespace Azure::Core::Http;
 
 // Keep reading until buffer is all fill out of the end of stream content is reached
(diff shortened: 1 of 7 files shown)


vhvb1989 (approved):

sdk/core/azure-core/src/http/body_stream.cpp
```diff
@@ -53,7 +54,7 @@ std::vector<uint8_t> BodyStream::ReadToEnd(Context const& context, BodyStream& b
 
   for (auto chunkNumber = 0;; chunkNumber++)
   {
-    buffer.resize((chunkNumber + 1) * chunkSize);
+    buffer.resize((static_cast<decltype(buffer)::size_type>(chunkNumber) + 1) * chunkSize);
```
ahsonkhan, line 57: Why is this necessary and what is this cast doing exactly?

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -7,7 +7,7 @@
 #include <chrono>
 #include <sstream>
 
-using namespace Azure::Core;
+using Azure::Core::Context;
```
ahsonkhan, line 10: @CaseyCarter - the guidelines don't mention anything about this, so I am wondering, is it fine to use using statement declarations like this?

ahsonkhan (commented):
Looks good.

sdk/core/azure-core/src/http/body_stream.cpp
antkmsft: We remove a warning, stating our intentions more clearly. `resize()` takes `vector<uint8_t>::size_type`, which is defined by implementation, but realistically is `uint64`. `chunkNumber` is `int`.

PR #1086, opened 2020-12-07
Try to remove the dependency on a curl header in client secret credential implementation

ahsonkhan:
Ideally, there should be no direct dependency on a specific transport adapter, even in the implementation.

### DIFF
--- a/sdk/identity/azure-identity/src/client_secret_credential.cpp
+++ b/sdk/identity/azure-identity/src/client_secret_credential.cpp
@@ -3,7 +3,6 @@
 
 #include "azure/identity/client_secret_credential.hpp"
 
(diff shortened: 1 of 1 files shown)


antkmsft (approved):

antkmsft (commented):
Have you reviewed other files for not having the same unnecessary include?

ahsonkhan:
> Have you reviewed other files for not having the same unnecessary include?

Nope, will do that separately.

vhvb1989 (approved):

ahsonkhan:
From a quick glance, didn't find any other frivolous includes specific to curl.hpp:
https://github.com/Azure/azure-sdk-for-cpp/search?l=C%2B%2B&p=2&q=curl

PR #1092, opened 2020-12-08
Use CONFIG QUIET when looking for packages

antkmsft:

### DIFF
--- a/sdk/core/azure-core/CMakeLists.txt
+++ b/sdk/core/azure-core/CMakeLists.txt
@@ -15,7 +15,7 @@ find_package(Threads REQUIRED)
 # min version for `CURLSSLOPT_NO_REVOKE`
 # https://curl.haxx.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
 set(CURL_MIN_REQUIRED_VERSION 7.44)
-find_package(CURL ${CURL_MIN_REQUIRED_VERSION} CONFIG)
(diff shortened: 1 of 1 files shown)


Jinming-Hu (approved):

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ find_package(Threads REQUIRED)
 # min version for `CURLSSLOPT_NO_REVOKE`
 # https://curl.haxx.se/libcurl/c/CURLOPT_SSL_OPTIONS.html
 set(CURL_MIN_REQUIRED_VERSION 7.44)
-find_package(CURL ${CURL_MIN_REQUIRED_VERSION} CONFIG)
+find_package(CURL ${CURL_MIN_REQUIRED_VERSION} CONFIG QUIET)
```
ahsonkhan, line 18: Why are we doing this? What's the harm in keeping the warning visible?

ahsonkhan (approved):
Other than understanding the motivation, the change itself looks good.

sdk/core/azure-core/CMakeLists.txt
antkmsft: Victor recommended it, so yes we remove the warning, but the very next line ("REQUIRED") fails, no need to have an additional warning then.

PR #1106, opened 2020-12-09
Add context.IsCanceled to provide a non-throwing alternative to ThrowIfCanceled.

ahsonkhan:


sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -391,12 +391,17 @@ namespace Azure { namespace Core {
           = ContextSharedState::ToMsecSinceEpoch(time_point::min());
     }
 
+    /**
+     * @brief Checks if the context was canceled.
```
antkmsft, line 395: I think doxygen comments for C++ should describe verb/action instead of a function, so I would put it as "Check if the context is canceled" - because it is what you see when you invoke intellisense. You can see examples throughout this file and other places, although I don't think there is formal agreement.

I would fill out the section completely and not leave that to future doc reviews, by adding "`@return "`true` if this context is canceled, `false` otherwise.`"

antkmsft (approved):

vhvb1989 (approved):
What about moving this API to `Details::`
I am thinking if there's a benefit for end-users to have this method public.
As an end-user, I just want to either call Cancel() or set a timeout right?  so, why do I need to have something like `isCanceled()` which seems like an implementation details.

Anyway, just a though about it. It seems right for the SDK implementation to be able to check to decide what we want to throw, but it is kind of weird for public API maybe.

Please wait for @RickWinter approval before merging as he's the author.

JeffreyRichter (approved):

JeffreyRichter:
It's OK for IsCancelled to be public; customers might want to call it.

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -391,12 +391,17 @@ namespace Azure { namespace Core {
...
+     */
+    bool IsCanceled() const { return CancelWhen() < std::chrono::system_clock::now(); }
```
RickWinter, line 397: Tests need to be added for this new method.

RickWinter (changes requested):
Add tests for the new method

sdk/core/azure-core/inc/azure/core/context.hpp
ahsonkhan: I expected existing unit tests for `ThrowIfCanceled` to also test this method indirectly, but looks like we don't have tests for that method either. :)

ahsonkhan: Oh... no tests or even any callers for `WithDeadline()` :(

ahsonkhan: https://github.com/Azure/azure-sdk-for-cpp/issues/1146

RickWinter (approved):

ahsonkhan:
> I am thinking if there's a benefit for end-users to have this method public.

@vhvb1989 since we have a public throwing version of this API `ThrowIfCanceled`, it is useful to add a non-throwing one as an alternative. This enables the customer to write code that follows the "tester-doer" pattern:
https://www.oreilly.com/library/view/api-in-c/9781800207813/video3_8.html

PR #1108, opened 2020-12-09
Make changes towards making libraries individually buildable

antkmsft:


sdk/storage/azure-storage-blobs/CMakeLists.txt
```diff
@@ -2,24 +2,22 @@
...
-    inc/azure/storage/blobs/blob_container_client.hpp
-    inc/azure/storage/blobs/blob_options.hpp
-    inc/azure/storage/blobs/blob_responses.hpp
-    inc/azure/storage/blobs/blob_sas_builder.hpp
-    inc/azure/storage/blobs/blob_service_client.hpp
-    inc/azure/storage/blobs/block_blob_client.hpp
-    inc/azure/storage/blobs/page_blob_client.hpp
-    inc/azure/storage/blobs/protocol/blob_rest_client.hpp
-    inc/azure/storage/blobs/version.hpp
-)
```
Jinming-Hu, line 20: I don't understand, why don't you want to include header files?

Jinming-Hu: Is there any documents saying that it's a good practice?

sdk/storage/azure-storage-common/CMakeLists.txt
```diff
@@ -2,26 +2,27 @@
...
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../../cmake-modules")
+
+include(az_version)
+include(CodeCoverage)
+include(DefineTransportAdapter)
+include(doxygen_common)
+include(global_compile_options)
+
+set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
```
Jinming-Hu, line 20: Does `azure-core`  depend on `Threads` already? If so, we don't need to repeat it here.

sdk/storage/azure-storage-blobs/CMakeLists.txt
antkmsft: They don't need to be included, because source files include them. You don't build headers, you build sources.

Jinming-Hu: Is there any drawbacks to include them?

antkmsft: No, other than that you probably need to update the list each time a file is added (but if you forget, nothing will fail).
Do you want to keep them? I can revert this specific change.

sdk/storage/azure-storage-common/CMakeLists.txt
antkmsft: I think if I remove it from here, it would apply when building Core, but not when building this library. Let me to move it to global_compile_options.

Jinming-Hu: but `azure-storage-common` depends on `azure-core`, it will get these compile options from there.

antkmsft: If you import via vcpkg, and not via building by loading the entire Azure SDK project and building it, it is possible. Plus, unless you use `PARENT_SCOPE` when setting cmake variables, they won't be seen by siblings or predecessors.

sdk/storage/azure-storage-blobs/CMakeLists.txt
Jinming-Hu: The benefit of doing so is that we can see these header files in the project, so it's easier for us to navigate and edit. image

Jinming-Hu: I cannot see header files for `azure-core`.
image

antkmsft: Thank you, I will revert, and add headers for core.

Jinming-Hu: Well, you don't have to do the same for core, as long as core developers are fine with current state. I don't want to act like meddling...

sdk/storage/azure-storage-common/CMakeLists.txt
Jinming-Hu: OK, i'm not familiar with how vcpkg works.

sdk/storage/azure-storage-blobs/CMakeLists.txt
vhvb1989: @Jinming-Hu, you can change the project explorer view to see the entire source instead.
It's a personal choice at the end I think, I don't really like the cmake project view from VS. It goes away upon making changes to the CMakeLists files which is a shame hhahaha.

sdk/identity/azure-identity/CMakeLists.txt
```diff
@@ -1,33 +1,39 @@
...
-add_library(azure-identity ${AZURE_IDENTITY_HEADER} ${AZURE_IDENTITY_SOURCE})
-target_include_directories(azure-identity PUBLIC inc)
-target_link_libraries(azure-identity azure-core)
+# make sure that users can consume the project as a library.
+add_library(azure::identity ALIAS azure-identity)
 
 # Uncomment once identity have tests
 # coverage. Has no effect if BUILD_CODE_COVERAGE is OFF
 # create_code_coverage(identity azure-identity "azure-identity-test")
+#create_code_coverage(core azure-identity "azure-identity-test")
```
vhvb1989, line 32: this line is already there. and the package (first arg is identity)

vhvb1989 (approved):
Just beautiful!

Really like how the project structure looks here! Thank you very much!

vhvb1989:
Take a look to Storage CI results. The Coverage report is gone:
https://github.com/Azure/azure-sdk-for-cpp/pull/1108/checks?check_run_id=1521668150

sdk/storage/azure-storage-blobs/CMakeLists.txt
antkmsft: Very good feedback, thank you @Jinming-Hu! Including all files when generating project file is a valuable scenario, I updated all the libraries to use this approach. I also checked for it to include all the files, and I ordered the list in the order you see them in IDE, so that it is easier to keep in sync/updated. Would you ok the changes now?

Jinming-Hu (approved):

sdk/storage/azure-storage-blobs/CMakeLists.txt
Jinming-Hu: yeah, I think this is perfect

PR #1150, opened 2020-12-11
Following the guidelines, use quotes for files within the same project, and angle brackets for external dependencies.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1148


vhvb1989 (approved):
Thank you.
I think the most important is to make sure we follow the guidelines from header files. But it is OK to also do it from source.

PR #1167, opened 2020-12-14
Fix build error and small improvements

Jinming-Hu:
error message:

```
[ 29%] Built target azure-core
[ 36%] Built target azure-identity
[ 37%] Building CXX object sdk/storage/azure-storage-common/CMakeFiles/azure-storage-common.dir/src/xml_wrapper.cpp.o
/home/jamis/azure-sdk-for-cpp/sdk/storage/azure-storage-common/src/xml_wrapper.cpp:6:10: fatal error: libxml/xmlreader.h: No such file or directory
    6 | #include <libxml/xmlreader.h>
      |          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [sdk/storage/azure-storage-common/CMakeFiles/azure-storage-common.dir/build.make:212: sdk/storage/azure-storage-common/CMakeFiles/azure-storage-common.dir/src/xml_wrapper.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:311: sdk/storage/azure-storage-common/CMakeFiles/azure-storage-common.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
```

### DIFF
--- a/sdk/storage/azure-storage-blobs/CMakeLists.txt
+++ b/sdk/storage/azure-storage-blobs/CMakeLists.txt
@@ -65,7 +65,6 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-storage-common-cpp_INCLUDE_DIRS}
 )
 
 target_link_libraries(azure-storage-blobs PUBLIC Azure::azure-storage-common)
--- a/sdk/storage/azure-storage-common/CMakeLists.txt
+++ b/sdk/storage/azure-storage-common/CMakeLists.txt
@@ -73,10 +73,9 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-core-cpp_INCLUDE_DIRS}
 )
 
-target_include_directories(azure-storage-common INTERFACE ${LIBXML2_INCLUDE_DIRS})
+target_include_directories(azure-storage-common PUBLIC ${LIBXML2_INCLUDE_DIRS})
 target_link_libraries(azure-storage-common PUBLIC Azure::azure-core)
 target_link_libraries(azure-storage-common INTERFACE Threads::Threads ${LIBXML2_LIBRARIES})
 
--- a/sdk/storage/azure-storage-files-datalake/CMakeLists.txt
+++ b/sdk/storage/azure-storage-files-datalake/CMakeLists.txt
@@ -64,7 +64,6 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-storage-blobs-cpp_INCLUDE_DIRS}
 )
 
 target_link_libraries(azure-storage-files-datalake PUBLIC Azure::azure-storage-blobs)
--- a/sdk/storage/azure-storage-files-shares/CMakeLists.txt
+++ b/sdk/storage/azure-storage-files-shares/CMakeLists.txt
@@ -61,7 +61,6 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-storage-common-cpp_INCLUDE_DIRS}
 )
 
 target_link_libraries(azure-storage-files-shares PUBLIC Azure::azure-storage-common)


sdk/storage/azure-storage-blobs/CMakeLists.txt
```diff
@@ -65,7 +65,6 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-storage-common-cpp_INCLUDE_DIRS}
```
Jinming-Hu, line 68: Hi @antkmsft , storage team would like to understand why this include directory is necessary.

sdk/storage/azure-storage-common/CMakeLists.txt
```diff
@@ -73,10 +73,9 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-core-cpp_INCLUDE_DIRS}
 )
 
-target_include_directories(azure-storage-common INTERFACE ${LIBXML2_INCLUDE_DIRS})
+target_include_directories(azure-storage-common PUBLIC ${LIBXML2_INCLUDE_DIRS})
```
Jinming-Hu, line 78: Hi @antkmsft , this is a changed made in PR https://github.com/Azure/azure-sdk-for-cpp/pull/1013. Can you help us understand why `INTERFACE` is perferred over `PUBLIC`?

We found this caused a build error of `libxml2` not being found. It can build if dependencies are installed via `vcpkg` just because `libxml2` headers happen to be in the same directory as other dependencies.

ahsonkhan:
> Fix build error and small improvements

@Jinming-Hu, can you please describe the build error you are seeing that this PR is fixing either in the PR description/title or as a comment?

sdk/storage/azure-storage-blobs/CMakeLists.txt
antkmsft: It plays a role when you only build blobs library (load `sdk/storage/azure-storage-blobs/CMakeLists.txt` and NOT the CmakeLists in the root folder), and get storage-common as find_package(), i.e. when you build a vcpackage. When you don't build a single library, but the whole SDK, `AZ_ALL_LIBRARIES` is defined, so we don't try to find_package(azure-storage-common-cpp), and build as usual, plus then `azure-storage-common-cpp_INCLUDE_DIRS` is not defined, having no effect.

antkmsft: Do not remove this line, it will break vcpkg build. It is an empty string when not defined, should be harmless for you either way.

sdk/storage/azure-storage-common/CMakeLists.txt
antkmsft: It is the other way around - when something is PUBLIC, it carries the dependencies with it. From what I can see, switching to `PUBLIC` from `INTERFACE` is ok from the perspective of building vcpkg. You don't want to do that if you don't use libxml2 in your public headers (#1009), but then it should be `PRIVATE`.
Right now, it is not this library that fails compilation, it is probably the application that consumes the SDK fails to build, in which case, if the library is consumed as `INTERFACE`, it should be doing find_package() in it's own cmakefile. 

In short, it is fine to switch it from INTERFACE to PUBLIC, but if #1009 gets fixed, it should become PRIVATE.

sdk/storage/azure-storage-files-datalake/CMakeLists.txt
```diff
@@ -64,7 +64,6 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-storage-blobs-cpp_INCLUDE_DIRS}
```
antkmsft, line 67: Don't remove.

sdk/storage/azure-storage-files-shares/CMakeLists.txt
```diff
@@ -61,7 +61,6 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-storage-common-cpp_INCLUDE_DIRS}
```
antkmsft, line 64: Don't remove. Does it break any builds?

antkmsft (changes requested):
INTERFACE => PUBLIC is fine. Don't remove `${..._INCLUDE_DIRS}`

sdk/storage/azure-storage-common/CMakeLists.txt
```diff
@@ -73,10 +73,9 @@ target_include_directories(
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
       $<INSTALL_INTERFACE:include>
-      ${azure-core-cpp_INCLUDE_DIRS}
```
antkmsft, line 76: Don't remove.

sdk/storage/azure-storage-common/CMakeLists.txt
Jinming-Hu: I guess we have different understanding for `INTERFACE` and `PUBLIC`.

Be it as it may, it doesn't compile if it was `INTERFACE`. I also agree after #1009 is fixed, this should be changed to `PRIVATE`.

Jinming-Hu: > Right now, it is not this library that fails compilation, it is probably the application that consumes the SDK fails to build

I added the error message, it failed when building `azure-storage-common`.

antkmsft: I am surprised, because CI builds successfully, and it builds successfully in the isolation as well (the screenshots in this PR: https://github.com/Azure/azure-sdk-for-cpp/pull/1013)

sdk/storage/azure-storage-blobs/CMakeLists.txt
Jinming-Hu: Shouldn't the include dirs of `azure-storage-common` be set already via `target_link_libraries(azure-storage-blobs PUBLIC Azure::azure-storage-common)`?

sdk/storage/azure-storage-common/CMakeLists.txt
Jinming-Hu: as I already said, CI succeeded because we only test the scenario of installing dependencies via vcpkg. In this case, libxml headers happen to be in the same directory as other dependencies.

sdk/storage/azure-storage-blobs/CMakeLists.txt
antkmsft: That only adds link instructions, but does not configure include directory. It works when building the entire SDK, because in that case all the include directories are accessible. After all, why then do we have to do target_include_directories(${LIBXML2_INCLUDE_DIRS})? When building as vcpkg, azure-storage-common to azure-storage-blobs is no different than libxml2 or any other 3rd party lib.

sdk/storage/azure-storage-common/CMakeLists.txt
antkmsft: I am fine if you switch INERFACE to PUBLIC. I don't think you should be doing that, but it won't break vcpkg build, the rest is up to you. The ideal is PRIVATE, after all. My only objection is the removal of `target_include_dirs(${azure-*_INCLUDE_DIRS})`.

Jinming-Hu: I'll resolve this conversation.

sdk/storage/azure-storage-blobs/CMakeLists.txt
Jinming-Hu: > That only adds link instructions, but does not configure include directory.

I don't think so. It says in [this page](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements) `The usage requirements of a target can transitively propagate to dependents.`

[This section](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#build-specification-and-usage-requirements) and [this section](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-properties) have a better definition for `usage requirements`.

sdk/storage/azure-storage-common/CMakeLists.txt
Jinming-Hu: duplicate, discuss in another conversation

sdk/storage/azure-storage-files-shares/CMakeLists.txt
Jinming-Hu: duplicate, discuss in another conversation

sdk/storage/azure-storage-files-datalake/CMakeLists.txt
Jinming-Hu: duplicate, discuss in another conversation

katmsft (approved):

sdk/storage/azure-storage-blobs/CMakeLists.txt
Jinming-Hu: @antkmsft ping...

This PR is blocked here. Can you take a look? Thanks!

antkmsft (approved):
Your PR does build as vcpkg. Approving.

Jinming-Hu:
> Your PR does build as vcpkg. Approving.

Thank you!

Issue #531, opened 2020-08-26
Poller support in Azure::Core


RickWinter:
Add a generic solution for polling

Jinming-Hu:
Do we have a ETA for this item? Can we expect it before GA?

PR #1186, opened 2020-12-15
Add Operation<T>

RickWinter:
Add support for Operation<T>


sdk/core/azure-core/inc/azure/core/long_running_operation_state.hpp
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+namespace Azure { namespace Core {
+
+  class LongRunningOperationState {
+    std::string m_value;
+
+  public:
+    // Provide `explicit` conversion from string or types convertible to string:
+    explicit LongRunningOperationState(const std::string& value) : m_value(value) {}
```
vhvb1989, line 24: I have asked @antkmsft  what's the difference between 
`explicit LongRunningOperationState(const std::string& value) : m_value(value)`
vs
`explicit LongRunningOperationState(std::string value) : m_value(std::move(value))`

For what I understand, the second way provides a better and clearer intention about the method will make a copy of the string that we are passing.
While the first one, seems like the method is only reading the value (and we make a copy of it behind the scenes)

So, the preferred way should be the second approach

sdk/core/azure-core/src/long_running_operation_state.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include "azure/core/long_running_operation_state.hpp"
+
+namespace Azure { namespace Core {
+
+  const LongRunningOperationState LongRunningOperationState::NotStarted("NOT_STARTED");
+  const LongRunningOperationState LongRunningOperationState::InProgress{"IN_PROGRESS"};
+  const LongRunningOperationState LongRunningOperationState::SuccessfullyCompleted{"SUCCESSFULLY_COMPLETED"};
+  const LongRunningOperationState LongRunningOperationState::Failed{"FAILED"};
+  const LongRunningOperationState LongRunningOperationState::UserCancelled{"USER_CANCELLED"};
```
vhvb1989, line 12: change curly bracket to parenthesis?

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -37,8 +37,9 @@ add_executable (
     logging.cpp
     main.cpp
     nullable.cpp
+    operation.cpp
     pipeline.cpp
-    policy.cpp
+    policy.cpp
```
vhvb1989, line 42: Fix this empty lines at the end

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+  };
+
+  class StringClient {
+  public:
+    StringOperation StartStringUpdate()
+    {
+      //Make initial String call
+      StringOperation operation = StringOperation(this);
+      return operation;
```
vhvb1989, line 70:
```suggestion
      return StringOperation(this);
```

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,74 @@
...
+  class StringClient;
+
+  class StringOperation : Operation<std::string> {
+
+  private:
+    StringClient* m_client;
+    bool m_operationComplete = false;
+    std::string m_operationToken;
+
+    int count = 0;
```
vhvb1989, line 23:
```suggestion
    int m_count = 0;
```

vhvb1989 (commented):
Looks good.

We could consider creating a simple service somewhere to tests this with network

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <azure/core/operation_status.hpp>
+#include <azure/core/response.hpp>
+
+using namespace Azure::Core;
+
+namespace Azure { namespace Core { namespace Test {
+
+  class StringClient;
+
+  class StringOperation : public Operation<std::string> {
```
CaseyCarter, line 18: Missing `#include <string>`.

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+#pragma once
+
+#include <azure/core/context.hpp>
+#include <azure/core/nullable.hpp>
+#include <azure/core/operation.hpp>
+#include <azure/core/operation_status.hpp>
+#include <azure/core/response.hpp>
+
+using namespace Azure::Core;
```
CaseyCarter, line 12: Normally I would yell about a using-directive in a header, but since this is test code it's not a HUGE issue. I will point out that this particular using-directive is unnecessary: All the code in this header is defined in a subnamespace of `Azure::Core`, so unqualified name lookup will find things in `Azure::Core` anyway.

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+  class StringClient;
+
+  class StringOperation : public Operation<std::string> {
+
+  private:
+    StringClient* m_client;
+    std::string m_operationToken;
+    std::string m_value;
+
+  private:
```
CaseyCarter, line 25: This `private` is redundant with line 20:
```suggestion
```

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+    int count = 0;
+
+  public:
+    StringOperation(StringClient* client){ m_client = client; }
```
CaseyCarter, line 29:
```suggestion
    explicit StringOperation(StringClient* client) : m_client{client} {}
```

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+
+    std::string GetResumeToken() { return m_operationToken; }
```
CaseyCarter, line 32:
```suggestion
    std::string GetResumeToken() const { return m_operationToken; }
```

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    std::unique_ptr<Http::RawResponse> Poll(Context& context = Context())
```
CaseyCarter, line 34: Ditto "`Context& context = Context()` is no bueno". Also, it's good practice to annotate these virtual functions with `override` to indicate that they are overriding some virtual signature from a base class: it tells readers there's action-at-a-distance going on, and tells compilers to error if we failed to properly override the base's signature.
```suggestion
    std::unique_ptr<Http::RawResponse> Poll(Context& context) override
```

CaseyCarter, line 34: Missing `#include <memory>`.

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+        m_status = OperationStatus::Succeeded;
+      }
+      // Tests fake out the network and might throw
+      //  context.ThrowIfCanceled();
+
+      return std::make_unique<Http::RawResponse>(
+          (uint16_t)1, (uint16_t)0, Http::HttpStatusCode(200), "OK");
+    }
+
+    std::string Value()
```
CaseyCarter, line 49:
```suggestion
    std::string Value() override
```

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+      std::unique_ptr<Http::RawResponse> response;
+      while (!Done())
+      {
+        // Sleep for the period
+        // std::this_thread::sleep(period); // Update to use retry header if it exists
+
+        response = Poll(context);
+      }
+
+      m_value = "OperationCompleted";
```
CaseyCarter, line 68: Shouldn't this happen in `Poll` instead?

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,83 @@
...
+      //  context.ThrowIfCanceled();
+
+      return std::make_unique<Http::RawResponse>(
+          (uint16_t)1, (uint16_t)0, Http::HttpStatusCode(200), "OK");
+    }
+
+    std::string Value()
+    {
+      if (m_status != OperationStatus::Succeeded)
+        throw std::runtime_error("InvalidOperation");
```
CaseyCarter, line 52: Don't the guidelines call for fully-bracing substatements?

CaseyCarter (changes requested):

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,41 @@
...
+    bool HasValue()
+    {
+      return (m_status == OperationStatus::Succeeded);
+    }
+
+    virtual T Value() = 0;
+
+    virtual std::string GetResumeToken() = 0;
+
+    virtual std::unique_ptr<Http::RawResponse> Poll(Context& context = Context()) = 0;
```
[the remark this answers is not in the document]
RickWinter: I initially was going to use what you described.  Then decided I could be crafty with optional args.  I wasn't aware that wasn't possible with lvalue refs.  Too bad the compiler doesn't warn me about it :).

sdk/core/azure-core/test/ut/operation_test.hpp
RickWinter: An old habit of writing code where visibility was specified for members and functions with each being treated separately.

sdk/core/azure-core/test/ut/operation_test.hpp
RickWinter: It normally would so I'll update the test case to be a good citizen

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
CaseyCarter: > Too bad the compiler doesn't warn me about it

There's some weirdness here since it's a class template and the function is pure virtual. This:
```c++
struct Context {};

struct S {
  virtual void meow(Context& = Context()) = 0;
};
```
gets a diagnostic (https://godbolt.org/z/fb6vff) although MSVC only complains with `-permissive-` (the "try to be more standard-conforming" flag) since this is notoriously an area where legacy MSVC was broken. (Apparently someone tried to portray this as a feature instead of a bug in ancient times earning it the nickname "Evil Extension".)

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#pragma once
+
+#include "azure/core/context.hpp"
+#include "azure/core/operation_status.hpp"
+#include "azure/core/response.hpp"
+
+#include <chrono>
+#include <string>
+#include <memory>
```
CaseyCarter, line 12:
```suggestion
#include <memory>
#include <string>
```

sdk/core/azure-core/test/ut/operation_test.hpp
RickWinter: It might, but I couldn't find it.  I fixed it as I know having the { } isn't a violation

sdk/core/azure-core/inc/azure/core/long_running_operation_state.hpp
ahsonkhan: Keep in mind, these are public APIs so we should try to be idiomatic and avoid unnecessary string copies.

sdk/core/azure-core/inc/azure/core/operation.hpp
ahsonkhan: Or whatever order Ctrl+K Ctrl+D in VS formats this as :)

sdk/core/azure-core/test/ut/operation_test.hpp
ahsonkhan: How do you catch such things in code review @CaseyCarter ?

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,41 @@
...
+          m_status == OperationStatus::Cancelled ||
+          m_status == OperationStatus::Failed);
+    }
+
+    bool HasValue()
+    {
+      return (m_status == OperationStatus::Succeeded);
+    }
+
+    virtual T Value() = 0;
```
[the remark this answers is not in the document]
ahsonkhan: Removing `const` on a virtual method in the future isn't considered breaking, correct?

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: This is OK to do if the parameter was marked as const, correct, if we want an optional argument with a default value?

We do this quite often for our `Options` types. For example:
https://github.com/Azure/azure-sdk-for-cpp/blob/ed1908efa149e2dd8b567c4a36d4dcc56b754769/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp#L37-L41

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
CaseyCarter: Removing `const` would most definitely be breaking since it's part of the function signature; "old" implementations of `T Value() const;` in derived classes would not override the "new" `virtual T Value();` in the base class. (I should have made this more clear in my initial comment.)

sdk/core/azure-core/test/ut/operation_test.hpp
CaseyCarter: 1. Memorize which Standard Library header declares each entity. (This is really a product of decades of experience moreso than rote memorization, at which I'm not great.)
2. Cross-check each library name I run across against the list of headers (which is ideally short enough to memorize) at the beginning of the file.

Ironically I get a little better at (1) and a little worse at (2) every year.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
CaseyCarter: > This is OK to do if the parameter was marked as const, correct, if we want an optional argument with a default value?

Yes, a `const` lvalue reference can bind to an rvalue. I've clarified this in the original comment.

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+      return (
+          m_status == OperationStatus::Succeeded || m_status == OperationStatus::Cancelled
+          || m_status == OperationStatus::Failed);
+    }
+
+    bool HasValue() const { return (m_status == OperationStatus::Succeeded); }
+
+    virtual T Value() const = 0;
+
+    virtual std::string GetResumeToken() const = 0;
```
ahsonkhan, line 40: Is this the equivalent of `Id` in .NET?
https://github.com/Azure/azure-sdk-for-net/blob/13aff100da135d11cd593965e3397b899ac5807d/sdk/core/Azure.Core/src/OperationOfT.cs#L25

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+    virtual Response<T> PollUntilDoneInternal(std::chrono::milliseconds period, Context& context)
+        = 0;
+
+  protected:
+    OperationStatus m_status = OperationStatus::NotStarted;
+
+  public:
+    OperationStatus Status() const { return m_status; }
+
+    bool Done() const
```
ahsonkhan, line 29:
```suggestion
    bool IsDone() const
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    // Provide a "Get" accessor
+    const std::string& Get() const noexcept { return m_value; }
+
+    // Provide your example values as static const members
+    static const OperationStatus NotStarted;
+    static const OperationStatus Running;
+    static const OperationStatus Succeeded;
+    static const OperationStatus Cancelled;
+    static const OperationStatus Failed;
```
ahsonkhan, line 43: Doxygen docs needed for public APIs.

sdk/core/azure-core/src/operation_status.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include "azure/core/operation_status.hpp"
+
+namespace Azure { namespace Core {
+
+  const OperationStatus OperationStatus::NotStarted("NotStarted");
+  const OperationStatus OperationStatus::Running{"Running"};
+  const OperationStatus OperationStatus::Succeeded{"Succeeded"};
+  const OperationStatus OperationStatus::Failed{"Failed"};
+  const OperationStatus OperationStatus::Cancelled{"Cancelled"};
```
ahsonkhan, line 12: Be consistent when using `()` vs `{}`

Also, out of curiosity, why are the enum values in the cpp file and not in the header? I see the guidelines tell us to do it this way, but want to understand why:
https://azure.github.io/azure-sdk/cpp_design.html#enumerations
@CaseyCarter?

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -42,8 +42,9 @@ add_executable (
...
```
ahsonkhan, line 47: nit: revert extra spacing added here.

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
+// ************ CUSTOMER CODE **********************
```
ahsonkhan, line 1: Needs license header at the start of file.

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "gtest/gtest.h"
```
ahsonkhan, line 3: Angle brackets
```suggestion
#include <gtest/gtest.h>
```

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
...
+  StringClient client;
+  auto operation = client.StartStringUpdate();
+
+  EXPECT_FALSE(operation.Done());
+  EXPECT_FALSE(operation.HasValue());
+
+  while(!operation.Done())
+  {
+    EXPECT_FALSE(operation.HasValue());
+    auto response = operation.Poll();
```
ahsonkhan, line 26: What does calling `operation.Value()`after this line do? Add an assert for what we expect there.

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  EXPECT_TRUE(operation.Done());
+  EXPECT_TRUE(operation.HasValue());
+
+  auto result = operation.Value();
+  EXPECT_TRUE(result == "StringOperation-Completed");
+}
+
+TEST(Operation, PollUntilDone)
```
ahsonkhan, line 36: Need `OperationStatus` tests, including equality comparison.

ahsonkhan: Also need `GetResumeToken` tests.

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
...
+  EXPECT_FALSE(operation.HasValue());
+  
+  auto start = std::chrono::high_resolution_clock::now();
+  auto response = operation.PollUntilDone(500ms);
+  auto end = std::chrono::high_resolution_clock::now();
+  std::chrono::duration<double, std::milli> elapsed = end - start;
+  //StringOperation test code is implemented to poll 2 times
+  EXPECT_TRUE(elapsed >= 1s);
+
+  EXPECT_TRUE(operation.Done());
```
ahsonkhan, line 51: We should also test that `Done` returns true for other operation statuses like canceled due to a context timeout.

sdk/core/azure-core/test/ut/operation_test.hpp
```diff
@@ -0,0 +1,87 @@
...
+      while (!Done())
+      {
+        // Sleep for the period
+        // Actual clients should respect the retry after header if present
+        std::this_thread::sleep_for(period);
+
+        response = Poll(context);
+      }
+
+      return Response<std::string>(m_value, std::move(response));
```
ahsonkhan, line 58: Should we reset `m_count` to 0 at this point, so that a second call to `PollInternal` still waits for a second?

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
...
+  EXPECT_TRUE(operation.HasValue());
```
ahsonkhan, line 52: What happens if we call `PollUntilDone` again after operation is done? I would imagine that should return almost right away (or one first poll) and keep the state unchanged. Consider adding that test.

ahsonkhan (changes requested):
Add doxygen docs and maybe add more test cases, especially around the edges.

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    // Provide a "Get" accessor
+    const std::string& Get() const noexcept { return m_value; }
+
+    // Provide your example values as static const members
+    const static OperationStatus NotStarted;
+    const static OperationStatus Running;
+    const static OperationStatus Succeeded;
+    const static OperationStatus Cancelled;
+    const static OperationStatus Failed;
```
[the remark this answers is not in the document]
ahsonkhan: If that is more idiomatic, we need to update the guidelines because they are currently misleading:
https://azure.github.io/azure-sdk/cpp_design.html#enumerations

We also use `const static` here, which we may want to fix (@vhvb1989) :
https://github.com/Azure/azure-sdk-for-cpp/blob/ed1908efa149e2dd8b567c4a36d4dcc56b754769/sdk/core/azure-core/src/http/url.cpp#L79

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <chrono>
+#include <string>
+#include <memory>
+
+namespace Azure { namespace Core {
+
+  //@param <T> The type of the final result of the long-running operation.
+  template <class T> class Operation {
+  private:
+    virtual std::unique_ptr<Http::RawResponse> PollInternal(Context& context) = 0;
```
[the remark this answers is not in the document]
RickWinter: They are private methods in the class.  There is no accessibility to them so naming isn't dictated other than Pascal

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+    virtual Response<T> PollUntilDoneInternal(std::chrono::milliseconds period, Context& context)
```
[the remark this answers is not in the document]
RickWinter: That is the dotnet guidelines not ours.

That said the testcase displays how it would be used and eventually the storage team will add usage patterns for each Operation

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: Then it is certainly a great call out. We don't want to too eagerly mark something as const if we expect it might need to be changed in the future.

What about removing const for regular non-virtual public methods? Is that fine to do?

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: Why are they pure virtual then if they aren't mean to be accessible outside the class?

http://www.gotw.ca/publications/mill18.htm
> Guideline #2: Prefer to make virtual functions private.
> This lets the derived classes override the function to customize the behavior as needed, without further exposing the virtual functions directly by making them callable by derived classes (as would be possible if the functions were just protected). The point is that virtual functions exist to allow customization; unless they also need to be invoked directly from within derived classes' code, there's no need to ever make them anything but private. 

From https://stackoverflow.com/a/2170696/12509023

Also from https://stackoverflow.com/a/3978552/12509023
> The confusion is common enough, that C++ FAQ advocated against using private virtuals, for a long time, because confusion seemed to be a bad thing.

Though I can't find mention of that in the provided link. @CaseyCarter - thoughts?

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: > That is the dotnet guidelines not ours.

I think the API design principle holds though :)

> That said the testcase displays how it would be used and eventually the storage team will add usage patterns for each Operation

Sounds good.

sdk/core/azure-core/test/ut/operation.cpp
RickWinter: Its an implementation detail for the operation.  Most Operations will throw, but we don't enforce that.

sdk/core/azure-core/test/ut/operation_test.hpp
RickWinter: No, this is a test not an actual operation.  That behavior, like m_count is an implementation detail

sdk/core/azure-core/test/ut/operation.cpp
RickWinter: Its an implementation detail for the operation

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    bool HasValue() const { return (m_status == OperationStatus::Succeeded); }
+
+    virtual T Value() const = 0;
+
+    virtual std::string GetResumeToken() const = 0;
+
+    std::unique_ptr<Http::RawResponse> Poll()
+    {
+      Context context;
```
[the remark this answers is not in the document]
RickWinter: Its not used by the test b/c we don't use a pipeline.  Jeffrey wants the pipelines to do the Cancel checks so that makes context uninteresting from the point of view of testing operation.

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  //@param <T> The type of the final result of the long-running operation.
+  template <class T> class Operation {
+  private:
+    virtual std::unique_ptr<Http::RawResponse> PollInternal(Context& context) = 0;
+    virtual Response<T> PollUntilDoneInternal(std::chrono::milliseconds period, Context& context)
+        = 0;
+
+  protected:
+    OperationStatus m_status = OperationStatus::NotStarted;
```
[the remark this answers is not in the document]
RickWinter: The derived types are the ones that process the response.  Its the responsibility of the derived type to set if the Operation failed or succeeded.

sdk/core/azure-core/test/ut/operation.cpp
RickWinter: The equality comparison is implicitly tested inside of Done().
m_status == OperationStatus::Succeeded

I'll add some specific testing,

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
JeffreyRichter: private & virtual doesn't make sense to me. Pure makes less sense as a derived class must implement it but it can't because it's private.
Maybe I'm missing something specific to C++ here, but C# wouldn't even compile this.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
JeffreyRichter: Rick & I reversed the params to make context be optional but that may no longer apply now if an overload that doesn't require context was added.

sdk/core/azure-core/inc/azure/core/operation.hpp
JeffreyRichter: I'm OK with IsDone.

sdk/core/azure-core/inc/azure/core/operation.hpp
JeffreyRichter: Yes. It's OK if we want to call it ID but token is probably a better name.

JeffreyRichter: Here's why this function is required: A machine initiates the LRO and then can get a token (or ID), pass the token(Id) to another machine. The other machine can use its XxxClient and call an overload of the StartXxx method that accepts the token/id. This overload also returns an Operation<T> and now the other machine can poll until done.

sdk/core/azure-core/test/ut/operation.cpp
JeffreyRichter: Which should probably mandate a specific exception be thrown and put it in the guidelines so that customers know exactly how to recover from this kind of failure. Although, calling Value when HasValue is false is a contract/precondition failure. In other words, a programmer bug, that should not have code written to recover from it. So, perhaps a C++ standard library runtime_error-derived type should be thrown here.

sdk/core/azure-core/test/ut/operation.cpp
JeffreyRichter: It should return immediately.
If we leave too much to be an "implementation detail" then we will get different implementations and this would be good for customers. I think we should err on the side of being prescriptive here with behavior and put it in our guidelines.

Also, the more code we put in the base class and the less code client library devs have to write in the derived classes helps ensure this while making it easier for client lib devs to implement thier derived class.

JeffreyRichter (approved):

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
RickWinter: Operation<T> is a template.  The derived type must provide an implementation for these two functions.  They are private b/c we do not want them directly called.  Instead it is the overloads on Operation<T> which will call the derived types methods.

sdk/core/azure-core/inc/azure/core/operation.hpp
RickWinter: There is no consistency across languages on the Operation type.  Every language did their own thing.  IsDone is close to what what JS used.  I'm ok with that.  I'll make a change

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,56 @@
...
+  {
+    EXPECT_FALSE(operation.HasValue());
+    auto response = operation.Poll();
+  }
+
+  EXPECT_TRUE(operation.Done());
+  EXPECT_TRUE(operation.HasValue());
+
+  auto result = operation.Value();
+  EXPECT_TRUE(result == "StringOperation-Completed");
```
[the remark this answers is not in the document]
RickWinter: It doesn't work for all types

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: I am a bit confused by this. Should the context value be set to something though? Maybe root context with infinite timeout?

My question about not passing an uninitialized local in the source code isn't really about the testing.

sdk/core/azure-core/test/ut/operation.cpp
ahsonkhan: > then we will get different implementations and this would be good for customers. 

I am assuming, @JeffreyRichter, that you meant: this would **not** be good for customers.

sdk/core/azure-core/test/ut/operation.cpp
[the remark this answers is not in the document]
ahsonkhan: Applies broadly to all the new tests added in this PR, not just that one instance.

sdk/core/azure-core/test/ut/operation_status.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  status = OperationStatus::Succeeded;
+  EXPECT_TRUE(status == OperationStatus::Succeeded);
+  EXPECT_TRUE(status.Get() == "Succeeded");
+  EXPECT_FALSE(status != OperationStatus::Succeeded);
+}
+
+TEST(OperationStatus, Custom)
+{
+  OperationStatus status2("CustomValue");
```
ahsonkhan, line 40: nit: why start with 2 instead of 1?
```suggestion
  OperationStatus status1("CustomValue");
```

sdk/core/azure-core/test/ut/operation_status.cpp
```diff
@@ -0,0 +1,56 @@
...
+  EXPECT_TRUE(status2.Get() == "CustomValue");
+  EXPECT_TRUE(status2 != OperationStatus::NotStarted);
+
+  OperationStatus status3 = OperationStatus("CustomValue");
+  EXPECT_TRUE(status3.Get() == "CustomValue");
+  EXPECT_TRUE(status3 != OperationStatus::NotStarted);
```
ahsonkhan, line 46: Two different `OperationStatus` instances with the same string value are considered equal, correct? Add a test assert:

```suggestion
  EXPECT_TRUE(status3 != OperationStatus::NotStarted);
  EXPECT_EQ(status3, status2);
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Added a WinHTTP-based `HttpTransport` called `WinHttpTransport` and use that as the default `TransportPolicyOptions.Transport` on Windows when sending and receiving requests and responses over the wire.
 - Added `Range` type to `Azure::Core::Http` namespace. 
+- Added support for long-running operations with `Operation<T>`.
```
ahsonkhan, line 9: nit: fully qualify the type
```suggestion
- Added support for long-running operations with `Azure::Core::Operation<T>`.
```

ahsonkhan (changes requested):
Blocking until we have doxygen docs, unless you plan to add those later.

sdk/core/azure-core/test/ut/operation.cpp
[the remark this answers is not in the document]
ahsonkhan: > It doesn't work for all types

Yep, I understand. For things like std::strings though, or anything that has a custom defined `==` it should do the right thing.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: OK, so @JeffreyRichter are you suggesting we have `Context` as the first parameter in `PollUntilDoneInternal` or leave it as is?

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: > The derived type must provide an implementation for these two functions. They are private b/c we do not want them directly called.

Right, gotcha. So, since they are not _actually_ private (i.e. we can't rename or break them), the original questions come back:
- Is this a common pattern to have private virtual methods suffixed with `Internal`? Should we name it differently or are we establishing a pattern here which we may want to add to the guidelines?
- Would we want such APIs to show up in APIView?

Also, I understand we want it to be virtual and private. Does it have to be **pure** virtual though?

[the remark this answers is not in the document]
vhvb1989: > The derived type must provide an implementation

Then we need it to be pure virtual right?  ` = 0`  otherwise overriding it is optional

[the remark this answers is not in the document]
vhvb1989: It should be protected, not private

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
vhvb1989: context is mandatory as first parameter and for any I/O API.
That's how we have built all other APIs so far. Let's not break this now :)

sdk/core/azure-core/inc/azure/core/operation.hpp
ahsonkhan: For context, my suggestion was less about consistency with other languages and more about what I would consider a naming convention for bool returning "test" methods. I, as the caller, am asking the API a question, and it returns a yes/no bool answer, so the name shouldn't just be a verb or noun but rather a query on the verb, like `IsFinished, or `HasValue`, etc.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
vhvb1989: yeah, this doesn't make sense here.
The Poll API should take the context as a parameter. It is an I/O operation.

sdk/core/azure-core/test/ut/operation.cpp
ahsonkhan: > I can add a check for how the test case is implemented. But note that is all that is being verified is how the test was written,

Right, it isn't all that useful testing the test implementation.

> it does not enforce a contract on the Operation

This is what I meant, that we should establish some contract (maybe through documentation) so the implementer of `Operation<T>` knows how to provide an implementation that adheres to the contract so others can call it and get consistent behavior. And then add that test :)

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    std::unique_ptr<Http::RawResponse> Poll(Context& context) { return PollInternal(context); }
+
+    Response<T> PollUntilDone(std::chrono::milliseconds period)
+    {
+      Context context;
+      return PollUntilDoneInternal(period, context);
+    }
+
+    Response<T> PollUntilDone(std::chrono::milliseconds period, Context& context)
```
vhvb1989, line 56: context is always the fist one

sdk/core/azure-core/src/operation_status.cpp
vhvb1989: @ahsonkhan only the definition is in the header.  The initialization is done on source files.
It's similar to function signatures.

sdk/core/azure-core/test/ut/CMakeLists.txt
vhvb1989: LoL

image

...

image

sdk/core/azure-core/inc/azure/core/operation.hpp
ahsonkhan: >  A machine initiates the LRO and then can get a token (or ID), pass the token(Id) to another machine. The other machine can use its XxxClient and call an overload of the StartXxx method that accepts the token/id. This overload also returns an Operation and now the other machine can poll until done.

Can we add some unit test that verifies our `Operation<T>` behaves this way and that the `GetResumeToken` returns something meaningful, or are we planning to leave all that to the implementer of the `Operation` and the client?

Should the ID/token always be unique each time it is called (like a Guid), or be constant/stable for a specific instance of an operation?

sdk/core/azure-core/test/ut/operation.cpp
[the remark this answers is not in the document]
vhvb1989: > It doesn't work for all types

When it comes to have something for all types, we should use `EXPECT_PRED2`, `EXPECT_PRED3` ..  etc
where the number at the end is the number of parameters we want to use for the predicate, the we need to pass a predicate and then the arguments

```cpp
EXPECT_PRED2([](TypeA x, TypeB y){
   return x.a == y.b;
}, type_a_var, type_b_var);
```
This is my preferred method when it comes to custom checks where EXPECT_EQ is not enoug

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    virtual std::string GetResumeToken() const = 0;
+
+    std::unique_ptr<Http::RawResponse> Poll()
+    {
+      Context context;
+      return PollInternal(context);
+    }
+
+    std::unique_ptr<Http::RawResponse> Poll(Context& context) { return PollInternal(context); }
```
ahsonkhan, line 48: I don't understand why we are accepting context if we are not going to do anything with it. We should consider checking or `context.IsCanceled()` and setting the operationstatus to cancelled before calling `PollInternal` and not call it for such cases.

**Is the point that context is just a pass through and it's the implementer's responsibility to do with it what they wish?** Is there any reason why they wouldn't honor `context.IsCanceled()`?

I don't know what we'd do for returning the raw response type in that case, but something like:
```C+
    std::unique_ptr<Http::RawResponse> Poll(Context& context)
    {
      if (conext.IsCanceled())
      {
        m_status = OperationStatus::Cancelled;
        return ???; // or do we still end u p calling the user function? both are odd
      }
      return PollInternal(context);
    }
```
Same approach for `PollUntilDone`.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: Making it private restricts even derived types from calling it, which has some value. They can override the behavior, but when and how it is called is controlled by the base type. That can be a useful mechanism to scope and isolate what the derived types can do.

The problem I see is, currently we aren't doing anything significant to control that behavior in some way (like checking context/etc.) and end up just calling their overridden method directly without any prologue/epilogue ceremony at the call site.

sdk/core/azure-core/src/operation_status.cpp
ahsonkhan: I understand we do that generally, but sometimes, especially for functions with small implementations, we end up writing it in the header file directly. We are even doing that in this PR, for example:
https://github.com/Azure/azure-sdk-for-cpp/blob/3fea31fee2d549b7070732a06d23b40c7aa52206/sdk/core/azure-core/inc/azure/core/operation.hpp#L29-L34
https://github.com/Azure/azure-sdk-for-cpp/blob/3fea31fee2d549b7070732a06d23b40c7aa52206/sdk/core/azure-core/inc/azure/core/operation_status.hpp#L25-L33

So, why not do the same for enums. It looks like the guidelines specifically went out of the way to call out and suggest providing the initialization of these enum values in the cpp file. Why?

When should we leave the implementation in the header file versus move to the source?

sdk/core/azure-core/test/ut/operation_status.cpp
RickWinter: No they won't be equal and hence the check.  status3 is a custom value and hence it should not be equal to some other status value.

sdk/core/azure-core/inc/azure/core/operation.hpp
RickWinter: This is a Mocked up Operation<T> for test purposes.  In real implementations the PollInternal method would trigger the pipeline via the client, i.e. pipeline->Send(context, ...._) method.  The pipeline is the place we check the context and the pipeline throws if the context is cancelled.

What I've written for the test is to eliminate the client and pipeline from my test implementation of Operation.  They are extra dependencies that the test doesn't need to validate an Operation builds, runs, functions as expected

sdk/core/azure-core/inc/azure/core/operation.hpp
RickWinter: That won't always be the case (per Jeffrey).  Though now that I've eliminate the default parameter for context, there is no reason not to have it be the first

sdk/core/azure-core/test/ut/operation_status.cpp
RickWinter: Cause I had a 1 initially that didn't make sense as a test case and updating the numbering was extra work that didn't provide any value :)

sdk/core/azure-core/inc/azure/core/operation.hpp
JeffreyRichter: The Poll function makes an I/O request to the service. ALL service I/O requests take a context and honor it already. So, the derive type that implements Poll will take the Context and pass it to the pipeline which will honor it; there is no need for anything in Operation to check the Context.

In short, the context should be checked immediately before an I/O operation and we have 2 kinds of I/O operations. Just before making an HTTP request and we're going to have the Transport policy check the context for this. The other is when reading from a BodyStream and we're going to have BodyStream's Read method do this check.

We should not be checking context cancellation anywhere else in the SDK. 
Someday, we might have 1 exception to this: we might check Cancellation in a long-running compute-bound loop that does no I/O operations. Today, we don't have any of these I don't think we ever will.

sdk/core/azure-core/inc/azure/core/operation.hpp
JeffreyRichter: I think Context should be first now that we have the overload.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
vhvb1989: > Making it private restricts even derived types from calling it

That's partially true.  The derived class can only call its own implementation.

> but when and how it is called is controlled by the base type.

No, the derived class can call its own implementation whenever it wants. You can have this:
```cpp
class Base {
private:
  virtual int Foo() = 0;
};

class Vh : public Base {
private:
  int Foo() override { return 1; }

public:
  // As derived class, I can call my implementation whenever I want
  int DoSomething() { return Foo(); }
};
```

The only thing different from making the pure virtual method `protected` and not `private` is to block derived classes from calling the virtual base function directly. But this would make no sense at all!
Like
```cpp
class Base {
protected:
  virtual int Foo() = 0;
};

class Vh : public Base {
private:
  int Foo() override { return 1; }

public:
  // There would have no sense to call the base to redirect to my own implementation
  int DoSomething() { return Base::Foo(); }
};
```

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,61 @@
...
+    virtual Response<T> PollUntilDoneInternal(std::chrono::milliseconds period, Context& context)
+        = 0;
+
+  protected:
+    OperationStatus m_status = OperationStatus::NotStarted;
+
+  public:
+    OperationStatus Status() const { return m_status; }
+
+    bool IsDone() const
```
vhvb1989, line 29: doxygen for all public code please

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
RickWinter: Customers call Poll() when they are doing the polling.  We do not want to force a customer to pass a context.  Its an additional concept and the customer may not care about it.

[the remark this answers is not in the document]
vhvb1989: > We do not want to force a customer to pass a context

We are doing that already. All Service APIs takes options which contains the context.
Let's follow the same pattern and add `PollOptions` as an optional parameter for calling `Poll()`.
We would have the context inside the PollOptions and only the cancellation-interested end-users could use this options structure.
It also offers the bag of options for any future setting we might need for polling(). 

@ahsonkhan the context is not `Uninitialized`, it is initialized based on its default constructor:
`Context() : m_contextSharedState(std::make_shared<ContextSharedState>()) {}`

We can always create and use a context like this. It basically becomes a root node. 
However, it is usually simpler to use the `Azure::Core::GetApplicationContext()` instead of creating a new context

sdk/core/azure-core/inc/azure/core/operation.hpp
RickWinter: refresh, they've been added :)

sdk/core/azure-core/inc/azure/core/operation.hpp
RickWinter: Updated

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
```
vhvb1989, line 3: mising file doxygen like

```
/**
 * @file
 * @brief bla bla bla.
 */
```
like all the other headers

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
```
vhvb1989, line 3: missing file docs

RickWinter: Good catch, its duplicated info for the one type in the file.  I forgot to add it but will.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
ahsonkhan: > In this case then, we should have an overload that takes context & period and another overload that just takes period.

When creating function overloads, this ordering style is not ideal. You generally want overloads to grow in the number of parameters, leaving the order of overlapping arguments, consistent.

Do we need an overload where context is optional? Why not just have one that takes both, or is it crucial that we not force customers to pass a context for cancellation?
```C++
virtual Response<T> PollUntilDoneInternal(Context& context, std::chrono::milliseconds period)
```

sdk/core/azure-core/inc/azure/core/operation.hpp
ahsonkhan: Makes sense. Should multiple calls to `GetResumeToken()` always return the exact same string?

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,46 @@
...
+    bool operator!=(const OperationStatus& other) const noexcept
+    {
+      return !(*this == other);
+    }
+
+    // Provide a "Get" accessor
+    const std::string& Get() const noexcept { return m_value; }
+
+    // Provide your example values as static const members
+    static const OperationStatus NotStarted;
```
vhvb1989, line 39: constexpr?

sdk/core/azure-core/test/ut/operation_status.cpp
ahsonkhan: Hmm... status2 and status3 had the exact same `"CustomValue"` string though.

sdk/core/azure-core/inc/azure/core/operation.hpp
ahsonkhan: > The pipeline is the place we check the context and the pipeline throws if the context is cancelled.

> The Poll function makes an I/O request to the service. ALL service I/O requests take a context and honor it already. So, the derive type that implements Poll will take the Context and pass it to the pipeline which will honor it; there is no need for anything in Operation to check the Context.

This is a really important design detail for folks like me to understand the relationship between context and operation. Thank you!

I am definitely interested in seeing how the concrete derived types end up looking like in the client library and how  this all flows through the SDK.

sdk/core/azure-core/src/operation_status.cpp
vhvb1989: > When should we leave the implementation in the header file versus move to the source?

For functions:  Only when the implementation is kind of short (a few lines) and does not require/use external libraries which will require including an external header (like using json, or something).
It is usually for set/get method implementation where there's just a return or something simple.

For variables (like this case): Only when the variable is a `constexpr`, which means we can assign a constant value duting the definition. This is true for non Heap-allocated stuff (int, enums, char*, etc). But when it is a reference to a Class (like this example), we can't create a constant object of the class statically.  So we just define the symbol and assign the value in source

sdk/core/azure-core/inc/azure/core/operation.hpp
JeffreyRichter: I can imagine an implementation that does always return teh same string but, in reality, they all should.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
JeffreyRichter: I agree that ordering is not ideal; that is why Rick & I swapped the order but now that conflicts with our "Context is always 1 param" guideline.

We could ship with just 1 overload and add the other later based on customer feedback. It's just this puts Context in their face which we usually do not do for the mainstream APIs.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/531

sdk/core/azure-core/test/ut/CMakeLists.txt
RickWinter: They keep coming back even though they aren't showing up in my VS changes,.  I believe they are gone now.

sdk/core/azure-core/CHANGELOG.md
antkmsft: I don't think we should do it, it is less readable. Unless we move things from namespace to namespace.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
antkmsft: Private virtual is a legit C++ concept called NVI (non-virtual interface), Herb Sutter is a huge fan.

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
RickWinter: I switched to use the GetApplication(), it makes the code easier to read.  Thanks for the suggestion @vhvb1989 .

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,125 @@
...
+    bool IsDone() const noexcept
+    {
+      return (
+          m_status == OperationStatus::Succeeded || m_status == OperationStatus::Cancelled
+          || m_status == OperationStatus::Failed);
+    }
+
+    /**
+     * @brief Returns true if the long-running operation completed successfully and has produced a
+     * final result.  The final result is accessible from Value().
```
ahsonkhan, line 71: consistency nit:
```suggestion
     * @brief Returns true if the long-running operation completed successfully and has produced a
     * final result. The final result is accessible from Value().
```

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,125 @@
...
+     * @brief Calls the server to get updated status of the long-running operation.
+     *
+     * @param context #Context allows the user to cancel the long-running operation.
+     *
+     * @return An HTTP #RawResponse returned from the service.
+     */
+    std::unique_ptr<Http::RawResponse> Poll(Context& context) { return PollInternal(context); }
+
+    /**
+     * @brief Periodically calls the server till the long-running operation completes;
```
ahsonkhan, line 99:
```suggestion
     * @brief Periodically calls the server till the long-running operation completes.
```

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * @param period Time in milliseconds to wait between polls
```
ahsonkhan, line 101:
```suggestion
     * @param period Time in milliseconds to wait between polls.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ * @brief Valid states for long-running Operations.  Services can extend upon the default set of
```
ahsonkhan, line 6:
```suggestion
 * @brief Valid states for long-running Operations. Services can extend upon the default set of
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * @brief Long-running operation states.
+   */
+  class OperationStatus {
+    std::string m_value;
+
+  public:
+    /**
+     * @brief Construct a #OperationStatus with \p value.
```
ahsonkhan, line 27:
```suggestion
     * @brief Construct an #OperationStatus with \p value.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @param value A non-absent value to initialize with.
```
ahsonkhan, line 29: What do you mean by `non-absent`?

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     */
+    explicit OperationStatus(const std::string& value) : m_value(value) {}
+    /**
+     * @brief Construct a #OperationStatus with \p value.
```
ahsonkhan, line 33:
```suggestion
     * @brief Construct an #OperationStatus with \p value.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @param value A non-absent value to initialize with.
+     */
+    explicit OperationStatus(std::string&& value) : m_value(std::move(value)) {}
+    /**
+     * @brief Construct a #OperationStatus with \p value.
```
ahsonkhan, line 39:
```suggestion
     * @brief Construct an #OperationStatus with \p value.
```

antkmsft (approved):

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @param value A non-absent value to initialize with.
+     */
+    explicit OperationStatus(const char* value) : m_value(value) {}
+
+    /**
+     * @brief Compare two #OperationStatus objects
```
ahsonkhan, line 46:
```suggestion
     * @brief Compare two #OperationStatus objects for equality.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @param other A #OperationStatus to compare with.
```
ahsonkhan, line 48:
```suggestion
     * @param other An #OperationStatus to compare with.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     */
+    bool operator==(const OperationStatus& other) const noexcept
+    {
+      return Strings::LocaleInvariantCaseInsensitiveEqual(m_value, other.m_value);
+    }
+
+    /**
+     * @brief Compare two #OperationStatus objects
+     *
+     * @param other A #OperationStatus to compare with.
```
ahsonkhan, line 60:
```suggestion
     * @param other An #OperationStatus to compare with.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @return `true` if the states have the same string representation. `false` otherwise.
+     */
+    bool operator==(const OperationStatus& other) const noexcept
+    {
+      return Strings::LocaleInvariantCaseInsensitiveEqual(m_value, other.m_value);
+    }
+
+    /**
+     * @brief Compare two #OperationStatus objects
```
ahsonkhan, line 58:
```suggestion
     * @brief Compare two #OperationStatus objects.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @param other A #OperationStatus to compare with.
+     *
+     * @return `false` if the states have the same string representation. `true` otherwise.
+     */
+    bool operator!=(const OperationStatus& other) const noexcept { return !(*this == other); }
+
+    /**
+     * @brief The std::string representation of the value
```
ahsonkhan, line 67:
```suggestion
     * @brief The std::string representation of the operation status.
```

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -0,0 +1,78 @@
...
+     */
+    const std::string& Get() const noexcept { return m_value; }
+
+    static const OperationStatus NotStarted;
+    static const OperationStatus Running;
+    static const OperationStatus Succeeded;
+    static const OperationStatus Cancelled;
+    static const OperationStatus Failed;
```
ahsonkhan, line 75: Doxygen docs missing for these public static consts.

sdk/core/azure-core/test/ut/operation.cpp
```diff
@@ -0,0 +1,90 @@
...
+    EXPECT_FALSE(operation.HasValue());
+    EXPECT_THROW(operation.Value(), std::runtime_error);
+    auto response = operation.Poll();
+  }
+
+  EXPECT_TRUE(operation.IsDone());
+  EXPECT_TRUE(operation.HasValue());
+
+  auto result = operation.Value();
+  EXPECT_TRUE(result == "StringOperation-Completed");
```
ahsonkhan, line 37:
```suggestion
  EXPECT_EQ(result, "StringOperation-Completed");
```

sdk/core/azure-core/test/ut/operation_status.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+TEST(OperationStatus, Custom)
+{
+  OperationStatus status1("CustomValue");
+  EXPECT_EQ(status1.Get(), "CustomValue");
+  EXPECT_NE(status1, OperationStatus::NotStarted);
+
+  OperationStatus status2 = OperationStatus("CustomValue");
+  EXPECT_EQ(status2.Get(), "CustomValue");
+  EXPECT_NE(status2, OperationStatus::NotStarted);
```
ahsonkhan, line 41:
```suggestion
  EXPECT_NE(status2, OperationStatus::NotStarted);
  EXPECT_EQ(status1, status2);
```

ahsonkhan (approved):
Some docs suggestions like a -> an, or add a period.
Looks good otherwise.

sdk/core/azure-core/test/ut/CMakeLists.txt
ahsonkhan: It's still there :(

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -42,8 +42,10 @@ add_executable (
     logging.cpp
     main.cpp
     nullable.cpp
+    operation.cpp
+    operation_status.cpp
     pipeline.cpp
-    policy.cpp
+    policy.cpp
```
ahsonkhan, line 48: The extra spacing issue still exists:
image

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -0,0 +1,125 @@
...
+  protected:
+    OperationStatus m_status = OperationStatus::NotStarted;
+
+  public:
+    /**
+     * @brief Final reuslt of the long-running operation.
+     *
+     * @return Response<T> the final result of the long-running operation.
+     */
+    virtual T Value() const = 0;
```
ahsonkhan, line 42: Can we consider something like this so that all implementations of `Value` in the derived types don't have to worry about providing error handling, and we can have a consistent exception type/message?

```C++
    private virtual T ValueImplementation() const = 0;
    T Value() const
    {
      if (!HasValue())
      {
        throw some exception
      }
      return ValueImplementation();
    }
```

sdk/core/azure-core/inc/azure/core/operation.hpp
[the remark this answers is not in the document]
JeffreyRichter: Thanks Anton. I found this: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Non-Virtual_Interface
So now I see that this is legit C++ but I still don't fully understand. Why not just make the methods protected instead of private?

sdk/core/azure-core/inc/azure/core/operation.hpp
vhvb1989: Think we can also do this for Nullable<>::GetValue()

ahsonkhan: This fits well as a precondition check so it can be turned off in release once the calling code is correctly written to avoid the unnecessary cost of extra checks

PR #1210, opened 2020-12-18
Fix doxygen comments for Operation<T>

RickWinter:

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/operation_status.hpp
+++ b/sdk/core/azure-core/inc/azure/core/operation_status.hpp
@@ -3,7 +3,7 @@
 
 /**
  * @file
- * @brief Valid states for long-running Operations.  Services can extend upon the default set of
+ * @brief Valid states for long-running Operations. Services can extend upon the default set of
  * values.
  */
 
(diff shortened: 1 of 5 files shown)


sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -24,28 +24,28 @@ namespace Azure { namespace Core {
...
      *
      * @param value A non-absent value to initialize with.
      */
     explicit OperationStatus(const std::string& value) : m_value(value) {}
     /**
-     * @brief Construct a #OperationStatus with \p value.
+     * @brief Construct an #OperationStatus with \p value.
      *
-     * @param value A non-absent value to initialize with.
+     * @param value The value to initialize with.
```
ahsonkhan, line 35: Make a similar fix for the two other ctors above:
https://github.com/Azure/azure-sdk-for-cpp/blob/cb32cbeeb942d3dd3f1d3490febf68c76206ff0c/sdk/core/azure-core/inc/azure/core/operation_status.hpp#L29-L41

sdk/core/azure-core/inc/azure/core/operation_status.hpp
```diff
@@ -24,28 +24,28 @@ namespace Azure { namespace Core {
 
   public:
     /**
-     * @brief Construct a #OperationStatus with \p value.
+     * @brief Construct an #OperationStatus with \p value.
      *
      * @param value A non-absent value to initialize with.
```
ahsonkhan, line 29: Apply the same fix to other uses of `non-absent`.
```suggestion
     * @param value The value to initialize with.
```

sdk/core/azure-core/test/ut/operation_status.cpp
```diff
@@ -48,4 +48,9 @@ TEST(OperationStatus, Custom)
   OperationStatus status4 = OperationStatus(std::string("CustomValue"));
   EXPECT_EQ(status4.Get(), "CustomValue");
   EXPECT_NE(status4, OperationStatus::NotStarted);
+
+  EXPECT_EQ(status1, status2);
+  EXPECT_EQ(status2, status3);
+  EXPECT_EQ(status3, status4);
+
```
ahsonkhan, line 55: Nice :)

ahsonkhan (approved):
Other than the `non-absent` fix applied everywhere, looks good.

sdk/core/azure-core/inc/azure/core/operation_status.hpp
RickWinter: good catch, fixed

ahsonkhan (approved):

ahsonkhan:
This is a follow up to https://github.com/Azure/azure-sdk-for-cpp/pull/1186

Issue #1087, opened 2020-12-07
Make the transport adapter tests `GetSuffix` retrieve the custom string from the info or have GetTransportOptions return the name as well.


ahsonkhan:
Look here for an example how the suffix string is returned as a tuple with the unit test parameter:
https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#value-parameterized-tests

Here's the relevant code: https://github.com/Azure/azure-sdk-for-cpp/blob/4cc1754b633341e3515edcc18bad87ed5c9188bc/sdk/core/azure-core/test/ut/transport_adapter_curl.cpp#L18-L37

PR #1217, opened 2020-12-18
Merge param-tests for the transport adapter

vhvb1989:
fixes: #1216 
fixes: #1087


sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <azure/core/context.hpp>
+#include <azure/core/http/policy.hpp>
+#include <azure/core/response.hpp>
+
+#include <iostream>
+#include <string>
+#include <thread>
+
+#include <http/curl/curl_connection_pool_private.hpp>
+#include <http/curl/curl_connection_private.hpp>
```
RickWinter, line 14:
```suggestion
#include <azure/core/http/curl/curl_connection_pool_private.hpp>
#include <azure/core/http/curl/curl_connection_private.hpp>
```

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,56 @@
...
+      return info.param.Prefix;
+    }
+  } // namespace
+
+  /*********************** Transporter Adapter Tests ******************************/
+  INSTANTIATE_TEST_SUITE_P(
+      TransportAdapterCurlImpl,
+      TransportAdapter,
+      testing::Values(
+#if defined(AZ_PLATFORM_WINDOWS)
```
RickWinter, line 46: Do you want to support custom transports in these adapter tests?

RickWinter (approved):

sdk/core/azure-core/test/ut/transport_adapter_curl.cpp
```diff
@@ -15,27 +17,6 @@ using testing::ValuesIn;
...
-    static Azure::Core::Http::TransportPolicyOptions GetTransportOptions()
-    {
-      Azure::Core::Http::TransportPolicyOptions options;
-      options.Transport = std::make_shared<Azure::Core::Http::CurlTransport>();
-      return options;
-    }
-
-    // When adding more than one parameter, this function should return a unique string.
-    // But since we are only using one parameter (the libcurl transport adapter) this is fine.
-    static std::string GetSuffix(const testing::TestParamInfo<TransportAdapter::ParamType>& info)
```
ahsonkhan, line 29: Also fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1087, correct?

sdk/core/azure-core/test/ut/transport_adapter_base.hpp
```diff
@@ -19,8 +19,20 @@
 
 namespace Azure { namespace Core { namespace Test {
 
-  class TransportAdapter
-      : public testing::TestWithParam<Azure::Core::Http::TransportPolicyOptions> {
+  struct TransportAdaptersTestParameter
+  {
+    std::string Prefix;
```
ahsonkhan, line 24: What's the benefit of prefix instead of suffix. It's fine, but I want to understand your motivation for the change.

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <thread>
+
+#include <http/curl/curl_connection_pool_private.hpp>
+#include <http/curl/curl_connection_private.hpp>
+
+using testing::ValuesIn;
+
+namespace Azure { namespace Core { namespace Test {
+
+  /**********************   Define the parameters for the base test and a suffix  ***************/
```
ahsonkhan, line 20: I am confused, is it prefix now or still suffix? We still have the `GetSuffix` method.
```suggestion
  /**********************   Define the parameters for the base test and a prefix  ***************/
```

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,56 @@
...
+      testing::Values(
+#if defined(AZ_PLATFORM_WINDOWS)
+          GetTransportOptions(
+              "winTransportAdapter",
+              std::make_shared<Azure::Core::Http::WinHttpTransport>()),
+#endif
+          GetTransportOptions(
+              "curlTransportAdapter",
+              std::make_shared<Azure::Core::Http::CurlTransport>())),
+      GetSuffix);
```
ahsonkhan, line 54: This should be in an `#else`.

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,56 @@
...
+  } // namespace
+
+  /*********************** Transporter Adapter Tests ******************************/
+  INSTANTIATE_TEST_SUITE_P(
+      TransportAdapterCurlImpl,
+      TransportAdapter,
+      testing::Values(
+#if defined(AZ_PLATFORM_WINDOWS)
+          GetTransportOptions(
+              "winTransportAdapter",
```
ahsonkhan, line 48:
```suggestion
              "winHttpTransportAdapter",
```

ahsonkhan (changes requested):
suffix or prefix?
Also, surround the curl one in an #else/#endif.

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
vhvb1989: No. These are private headers. The root folder is in `sdk/core/azure-core/src`.
I will add a comment about it

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
vhvb1989: I could, but it would take some time to make sure it works. I would need to implement a custom transport adapter and try to run the tests with it.
It would be easier to add some docs about how to test a custom adapter by adding some lines in here.

sdk/core/azure-core/test/ut/transport_adapter_curl.cpp
vhvb1989: correct 👍

sdk/core/azure-core/test/ut/transport_adapter_base.hpp
vhvb1989: It should be `suffix`. I have updated it.
This is the string that gtest adds to the end of the test name and that makes each test different from the same test but with a different parameter.

This is how it looks:
image

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
vhvb1989: suffix!  I have fixed it

sdk/core/azure-core/test/ut/transport_adapter_base.hpp
ahsonkhan: Does gtest allow making it a prefix for grouping reasons?

sdk/core/azure-core/test/ut/transport_adapter_curl.cpp
```diff
@@ -3,39 +3,23 @@
...
 #include <azure/core/response.hpp>
+
 #include <iostream>
 #include <string>
 #include <thread>
 
+// The next includes are from Azure Core private headers.
+// That's why the path starts from `sdk/core/azure-core/src/`
+// They are included to test the connection pool from the curl transport adapter implementation.
 #include <http/curl/curl_connection_pool_private.hpp>
```
ahsonkhan, line 16: So, should these have `src` in the path? The comment says the path starts from `sdk/core/azure-core/src/`. How?
```C++
#include <src/http/curl/curl_connection_pool_private.hpp>
```

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -25,10 +25,6 @@ if(BUILD_TRANSPORT_CURL)
   SET(CURL_TRANSPORT_TESTS transport_adapter_curl.cpp)
```
ahsonkhan, line 25: Do the same here?

sdk/core/azure-core/test/ut/transport_adapter_curl.cpp
vhvb1989: No.
The trick is done in the CMake test project where we include that directory as part of the INCLUDE path:
```cmake
# Adding private headers from CORE to the tests so we can test the private APIs with no relative paths include.
target_include_directories (azure-core-test PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../src>)
```

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -64,4 +60,6 @@ target_link_libraries(azure-core-test PRIVATE azure-core gtest gmock)
 # gtest_discover_tests will scan the test from azure-core-test and call add_test
 # for each test to ctest. This enables `ctest -r` to run specific tests directly.
 gtest_discover_tests(azure-core-test
-     TEST_PREFIX azure-core.)
+     TEST_PREFIX azure-core.
+     NO_PRETTY_TYPES
+     NO_PRETTY_VALUES)
```
ahsonkhan, line 65: Let me try out these settings locally in a bit to see how they behave, before merging.

ahsonkhan (approved):
Looks good, but please wait a couple of hours before merging to give me a chance to try the newly added gtest settings.

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+      // Can't use empty spaces or underscores (_) as per google test documentation
+      // https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#specifying-names-for-value-parameterized-test-parameters
+      return info.param.Suffix;
+    }
+  } // namespace
+
+  /*********************** Transporter Adapter Tests ******************************/
+  /*
+   * MSVC does not support adding the pre-processor `#if` condition indise the MACRO parameters,
```
ahsonkhan, line 39:
```suggestion
   * MSVC does not support adding the pre-processor `#if` condition inside the MACRO parameters,
```

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,73 @@
...
+  /* WinHttp + LibCurl */
+  INSTANTIATE_TEST_SUITE_P(
+      Test,
+      TransportAdapter,
+      testing::Values(
+          GetTransportOptions("winHttp", std::make_shared<Azure::Core::Http::WinHttpTransport>()),
+          GetTransportOptions("libCurl", std::make_shared<Azure::Core::Http::CurlTransport>())),
+      GetSuffix);
+
+#elif defined(BUILD_TRANSPORT_WINHTTP_ADAPTER) && !defined(BUILD_CURL_HTTP_TRANSPORT_ADAPTER)
```
ahsonkhan, line 52: I believe this can be simplified given the first `#if condition to just:
```suggestion
#elif defined(BUILD_TRANSPORT_WINHTTP_ADAPTER)
```

sdk/core/azure-core/test/ut/transport_adapter_implementation.cpp
```diff
@@ -0,0 +1,73 @@
...
+  /* WinHttp */
+  INSTANTIATE_TEST_SUITE_P(
+      Test,
+      TransportAdapter,
+      testing::Values(
+          GetTransportOptions("winHttp", std::make_shared<Azure::Core::Http::WinHttpTransport>())),
+      GetSuffix);
+
+#elif !defined(BUILD_TRANSPORT_WINHTTP_ADAPTER) && defined(BUILD_CURL_HTTP_TRANSPORT_ADAPTER)
```
ahsonkhan, line 61: Same here:
```suggestion
#elif defined(BUILD_CURL_HTTP_TRANSPORT_ADAPTER)
```

vhvb1989:
I love the output!

image

PR #1220, opened 2020-12-18
DateTime: add default constructor

### LINKED ISSUE #1175: DateTime: Add default constructor
Default constructor creates a date of 01-01-0001 (see #1174)

antkmsft:
Closes #1175


RickWinter (changes requested):
Need test cases for this new constructor

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -227,5 +227,11 @@ namespace Azure { namespace Core {
     // Private constructor. Use static methods to create an instance.
     explicit DateTime(Duration const& since1601) : m_since1601(since1601) {}
     Duration m_since1601;
+
+  public:
+    /**
+     * @brief Construct an instance of @DateTime.
+     */
+    DateTime() : m_since1601(0) {}
```
ahsonkhan, line 235: Didn't we settle on `01-01-0001`, from the linked issue.

ahsonkhan (changes requested):
Value of default ctor should be 01-01-0001

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: It is a separate issue planned for after the new year.

antkmsft: Even if we initialize it now with `-504911232000000000LL`, which means it is `0001-01-01`, it will be a value that you can't get a string of, can't construct via ints constructor, or can't parse from a string. Full support is needed.

ahsonkhan (approved):

RickWinter (approved):

PR #1226, opened 2020-12-18
Doclinks

RickWinter:
Fix up the link to the guidelines.


ahsonkhan:
This is no longer necessary. It was already fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/1225

PR #1239, opened 2020-12-21
Throw exception if curl_easy_init returns Null

ku-sourav:
If curl_easy_init() returns Null, no other functions can be used. So throw an exception if curl_easy_init fails.
https://curl.se/libcurl/c/curl_easy_init.html

### DIFF
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -1094,6 +1094,12 @@ std::unique_ptr<CurlNetworkConnection> CurlConnectionPool::GetCurlConnection(
   // Creating a new connection is thread safe. No need to lock mutex here.
   // No available connection for the pool for the required host. Create one
   CURL* newHandle = curl_easy_init();
+  if (!newHandle)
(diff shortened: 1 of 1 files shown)


ahsonkhan:
@ku-sourav thanks for contributing and making that fix! I am curious, how did you run across this issue and discover `curl_easy_init` was returning `NULL`? Is it possible for us to add a deterministic unit test for such a scenario where we force `curl_easy_init` to fail?

ahsonkhan (approved):

ku-sourav:
AFAIK, curl_easy_init fails when system memory is exhausted. If you look at its source, it internally calls [curl_open ](https://github.com/curl/curl/blob/98b6aa14d98a7098e328fba2745e2b666f073afc/lib/url.c#L631)which then does calloc and if it fails, curl_easy_init will return null. Another scenario is if [curl_global_init](https://github.com/curl/curl/blob/ac9a7245fc503a8d46871ab2bf1a7321392b9cda/lib/easy.c#L283)(when called from curl_easy_init) fails for any reason, null will be returned.

PR #1246, opened 2020-12-25
Integrate Core::DateTime into storage

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_service_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_service_client.hpp
@@ -104,8 +104,10 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @brief Retrieves a key that can be used to delegate Active Directory authorization to
      * shared access signatures.
      *
-     * @param startsOn Start time for the key's validity. The time should be specified in UTC.
-     * @param expiresOn Expiration of the key's validity. The time should be specified in UTC.
+     * @param startsOn Start time for the key's validity. The time should be specified in UTC, and
+     * will be truncated to second.
+     * @param expiresOn Expiration of the key's validity. The time should be specified in UTC, and
+     * will be truncated to second.
      * @param options Optional parameters to execute this function.
      * @return A deserialized GetUserDelegationKeyResult instance.
      */
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_service_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_service_client.hpp
@@ -104,16 +104,14 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @brief Retrieves a key that can be used to delegate Active Directory authorization to
      * shared access signatures.
      *
-     * @param startsOn Start time for the key's validity, in ISO date format. The time should be
-     * specified in UTC.
-     * @param expiresOn Expiration of the key's validity, in ISO date format. The time should be
-     * specified in UTC.
+     * @param startsOn Start time for the key's validity. The time should be specified in UTC.
+     * @param expiresOn Expiration of the key's validity. The time should be specified in UTC.
      * @param options Optional parameters to execute this function.
      * @return A deserialized GetUserDelegationKeyResult instance.
      */
     Azure::Core::Response<Models::GetUserDelegationKeyResult> GetUserDelegationKey(
-        const std::string& startsOn,
-        const std::string& expiresOn,
+        const Azure::Core::DateTime& startsOn,
+        const Azure::Core::DateTime& expiresOn,
         const GetUserDelegationKeyOptions& options = GetUserDelegationKeyOptions()) const;
 
     /**
--- a/sdk/storage/azure-storage-common/CHANGELOG.md
+++ b/sdk/storage/azure-storage-common/CHANGELOG.md
@@ -2,12 +2,18 @@
 
 ## 12.0.0-beta.6 (Unreleased)
 
+### New Features
+
+- Add new type `ContentHash`.
+
+### Breaking Changes
+
 - Rename `SharedKeyCredential` to `StorageSharedKeyCredential`.
 - Rename `StorageSharedKeyCredential::UpdateAccountKey` to `StorageSharedKeyCredential::Update`.
 - Move `StorageRetryPolicy`, `StoragePerRetryPolicy` and `SharedKeyPolicy` to `Details` namespace.
 - Remove `StorageRetryOptions`, use `Azure::Core::Http::RetryOptions` instead.
 - Move Account SAS into `Azure::Storage::Sas` namespace.
-- Add new type `ContentHash`.
+- All date time related strings are now changed to `Azure::Core::DateTime` type.
 
 ## 12.0.0-beta.5 (2020-11-13)
 
--- a/sdk/storage/azure-storage-common/test/test_base.cpp
+++ b/sdk/storage/azure-storage-common/test/test_base.cpp
@@ -235,66 +235,6 @@ namespace Azure { namespace Storage { namespace Test {
     return result;
   }
 
-  std::string ToIso8601(
-      const std::chrono::system_clock::time_point& timePoint,
-      int numDecimalDigits)
-  {
-    std::time_t epoch_seconds = std::chrono::system_clock::to_time_t(timePoint);
-    struct tm ct;
-#if defined(AZ_PLATFORM_WINDOWS)
-    gmtime_s(&ct, &epoch_seconds);
-#elif defined(AZ_PLATFORM_POSIX)
-    gmtime_r(&epoch_seconds, &ct);
-#endif
-    std::string time_str;
-    time_str.resize(64);
-    std::strftime(&time_str[0], time_str.length(), "%Y-%m-%dT%H:%M:%S", &ct);
-    time_str = time_str.data();
-    if (numDecimalDigits != 0)
-    {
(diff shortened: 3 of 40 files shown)


katmsft:
There are many changes due to the merged protocol layer generator. I cannot avoid that to make the PR cleaner, sorry.

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
```diff
@@ -248,6 +248,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
       const ListPathsOptions& options) const
   {
     Details::DataLakeRestClient::FileSystem::ListPathsOptions protocolLayerOptions;
+    protocolLayerOptions.Resource = Models::FileSystemResourceType::Filesystem;
```
Jinming-Hu, line 251: Why don't we need to set this before?

Why not `FileSystem`?

sdk/storage/azure-storage-files-datalake/test/datalake_file_client_test.cpp
```diff
@@ -452,8 +452,9 @@ namespace Azure { namespace Storage { namespace Test {
       options.Metadata = RandomMetadata();
       auto res
           = fileClient.UploadFrom(fileContent.data(), static_cast<std::size_t>(fileSize), options);
+      auto lastModified = fileClient.GetProperties()->LastModified;
       EXPECT_FALSE(res->ETag.empty());
-      EXPECT_FALSE(res->LastModified.empty());
+      EXPECT_EQ(res->LastModified, lastModified);
```
Jinming-Hu, line 457: How did you test `lastModifed` is correctly parsed?

sdk/storage/azure-storage-files-datalake/test/datalake_sas_test.cpp
```diff
@@ -60,8 +60,8 @@ namespace Azure { namespace Storage { namespace Test {
         std::make_shared<Azure::Identity::ClientSecretCredential>(
             AadTenantId(), AadClientId(), AadClientSecret()));
     auto userDelegationKey = *serviceClient1.GetUserDelegationKey(
-        ToIso8601(std::chrono::system_clock::now() - std::chrono::minutes(5)),
-        ToIso8601(std::chrono::system_clock::now() + std::chrono::minutes(60)));
+        Core::DateTime::Now() - std::chrono::minutes(5),
+        Core::DateTime::Now() + std::chrono::minutes(60));
```
Jinming-Hu, line 64: sasStartsOn, sasExpiresOn);

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
katmsft: It is due to the enum not being parsed as an optional value, due to it being an enum with one possible value. It was parsed as a constant.
However, to be future compatible, I don't think what used to be done is correct here, so I'm making it an enum.
The enum's string value is `filesystem`, I don't think it makes sense to call it `FileSystem`

sdk/storage/azure-storage-files-datalake/test/datalake_file_client_test.cpp
katmsft: That's actually a good question. Any proposals?

Jinming-Hu: search `IsValidDate`

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
Jinming-Hu: > The enum's string value is filesystem, I don't think it makes sense to call it FileSystem

why not?

katmsft: As talked offline, this is due to weird service behavior of this query parameter being called:
```
    "FileSystemResource": {
	  "name": "resource",
	  "in": "query",
	  "description": "The value must be \"filesystem\" for all filesystem operations.",
	  "required": true,
	  "type": "string",
	  "enum": [
		"filesystem"
	  ],
	  "x-ms-enum": {
		"name": "FileSystemResourceType",
		"modelAsString": false
	  }
	},
```
While for other query parameter:
```
          {
            "name": "action",
            "in": "query",
            "description": "Optional. If the value is \"getStatus\" only the system defined properties for the path are returned. If the value is \"getAccessControl\" the access control list is returned in the response headers (Hierarchical Namespace must be enabled for the account), otherwise the properties are returned.",
            "required": false,
            "type": "string",
            "enum": [
              "getAccessControl",
              "getStatus"
            ],
            "x-ms-enum": {
              "name": "PathGetPropertiesAction",
              "modelAsString": false
            }
          },
```
Service itself is introducing this inconsistency, there is little I can do in client.

katmsft (approved):

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_service_client.hpp
```diff
@@ -104,16 +104,16 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @brief Retrieves a key that can be used to delegate Active Directory authorization to
      * shared access signatures.
      *
-     * @param startsOn Start time for the key's validity, in ISO date format. The time should be
-     * specified in UTC.
-     * @param expiresOn Expiration of the key's validity, in ISO date format. The time should be
-     * specified in UTC.
+     * @param startsOn Start time for the key's validity. The time should be specified in UTC, and
+     * will be truncated to second.
```
ahsonkhan, line 108: What motivated this doc change from ISO to UTC and what does it mean that it will be "truncated to second"?

sdk/storage/azure-storage-common/CHANGELOG.md
```diff
@@ -2,12 +2,18 @@
 
 ## 12.0.0-beta.6 (Unreleased)
 
+### New Features
+
+- Add new type `ContentHash`.
```
ahsonkhan, line 7: Fully qualify the type when it is added, include the namespace.

sdk/storage/azure-storage-common/test/test_base.cpp
```diff
@@ -306,4 +246,12 @@ namespace Azure { namespace Storage { namespace Test {
     return secondaryUri.GetAbsoluteUrl();
   }
 
+  bool IsValidTime(const Azure::Core::DateTime& datetime)
+  {
+    // We assume datetime within a week is valid.
```
ahsonkhan, line 251: nit: Why? If we really want to assume that, the method should be named to be self-documenting
`IsValidTimeWithinAWeek`

Otherwise, if and when a test fails in the future, and the date is a valid time (well-formed), it will be difficult to debug why by looking at the callstacks or logs, that the reason was it wasn't within one week from now.

sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
```diff
@@ -52,6 +52,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
       ret.ContentEncoding = std::move(headers.ContentEncoding);
       ret.ContentLanguage = std::move(headers.ContentLanguage);
       ret.ContentType = std::move(headers.ContentType);
+      ret.ContentHash = std::move(headers.ContentHash);
```
ahsonkhan, line 55: Was this a bug since content hash was missing? Is there a unit test for this. Please avoid such changes in a large PR like this as it hides semantic differences that need to be reviewed carefully and tested.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_service_client.hpp
Jinming-Hu: It's not changed. ISO and UTC are different kind of things.

In the past, we use `std::string` to represent the datetime, so the customer had to know the format (ISO or RFC).

Jinming-Hu: > what does it mean that it will be "truncated to second"?

fractional part will be truncated, like `1:23:45.6789000` will be truncated to `1:23:45`.

sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
Jinming-Hu: @katmsft Can you answer this question?

sdk/storage/azure-storage-common/test/test_base.cpp
Jinming-Hu: Nah, I don't think so. The purpose of this function is just to check if a datetime in API response is valid or not, and there really isn't a strict criteria of what should be considered valid. We just assume that none of the test suites will being running for more than a week (which is a safe assumption), so the timestamps for the creation date, last modified date and last accessed date etc won't be older than a week.

If (although it's unlikely to happen) the assumption above is no longer valid in the future, we can totally change the criteria, like extending from a week to a month. Or if we want to adopt a stricter criteria, we can shorten to a day. All we need to do is to change the internal implementation of `IsValidTime`. The callers of this function should not make any changes, because all they want to do is still to check if the timestamp is valid or not.

PR #1247, opened 2020-12-29
Integrate `Http::Range` into blob service

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -160,17 +160,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       const DownloadBlobOptions& options) const
   {
     Details::BlobRestClient::Blob::DownloadBlobOptions protocolLayerOptions;
-    if (options.Range.HasValue() && options.Range.GetValue().Length.HasValue())
-    {
-      protocolLayerOptions.Range = std::make_pair(
-          options.Range.GetValue().Offset, options.Range.GetValue().Offset + options.Range.GetValue().Length.GetValue() - 1);
-    }
-    else if (options.Range.HasValue())
-    {
-      protocolLayerOptions.Range = std::make_pair(
-          options.Range.GetValue().Offset,
-          std::numeric_limits<std::remove_reference_t<decltype(options.Range.GetValue().Offset)>>::max());
-    }
+    protocolLayerOptions.Range = options.Range;
     protocolLayerOptions.LeaseId = options.AccessConditions.LeaseId;
     protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
     protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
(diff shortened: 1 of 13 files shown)


sdk/storage/azure-storage-blobs/test/block_blob_client_test.cpp
```diff
@@ -358,8 +358,12 @@ namespace Azure { namespace Storage { namespace Test {
       downloadBuffer.resize(static_cast<std::size_t>(downloadSize), '\x00');
       Blobs::DownloadBlobToOptions options;
       options.Concurrency = concurrency;
-      options.Offset = offset;
-      options.Length = length;
+      if (offset.HasValue() || length.HasValue())
+      {
+        options.Range = Core::Http::Range();
+        options.Range.GetValue().Offset = offset.GetValue();
```
katmsft, line 364: nit: `GetValue()` + setting the value is not intuitive. It makes the code a bit more difficult to understand. Also, `(offset.HasValue() || length.HasValue())` does not gurantee that `offset.GetValue()` won't crash.
How about `options.Range->Offset = offset;`?

katmsft (approved):

sdk/storage/azure-storage-blobs/test/block_blob_client_test.cpp
Jinming-Hu: > GetValue() + setting the value is not intuitive. It makes the code a bit more difficult to understand.

There's no way to get around.

> (offset.HasValue() || length.HasValue()) does not gurantee that offset.GetValue() won't crash.

When `length` has value, `offset` always has value. This is just test code, we don't have to be that strict.

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -198,12 +188,12 @@ namespace Azure { namespace Storage { namespace Blobs {
...
-            = (options.Offset.HasValue() ? options.Offset.GetValue() : 0) + retryInfo.Offset;
-        newOptions.Length = options.Length;
-        if (newOptions.Length.HasValue())
+        newOptions.Range = Core::Http::Range();
+        newOptions.Range.GetValue().Offset
+            = (options.Range.HasValue() ? options.Range.GetValue().Offset : 0) + retryInfo.Offset;
+        if (options.Range.HasValue() && options.Range.GetValue().Length.HasValue())
         {
-          newOptions.Length = options.Length.GetValue() - retryInfo.Offset;
+          newOptions.Range.GetValue().Length = options.Range.GetValue().Length.GetValue() - retryInfo.Offset;
```
ahsonkhan, line 196: note to self: see if there is a better way to write this rather than repeated checks to `.Range.HasValue()`.

Do we need an `Http::Range()` ctor that takes offset and length to make creating such range creation easier? 
Could a pattern like `options.Range = {1_MB, 2_MB};` be applied here?

Jinming-Hu: > Do we need an Http::Range() ctor that takes offset and length to make creating such range creation easier?

I'm not opposed to adding such constructor(s), although I don't think it's necessary. But what I want to say is, the `Length` of an `Http::Range` can be nullable. If you want to add a constructor that instantiates a range with a specific length, you might also want to add a constructor that instantiates a range without length.

Jinming-Hu: > Could a pattern like options.Range = {1_MB, 2_MB}; be applied here?

This is [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization), which is not allowed. We talked about this in one of the guideline meetings.

Issue #1202, opened 2020-12-17
Make `IsServerEncrypted` to non-nullable to be aligned with .Net


Jinming-Hu:

Jinming-Hu:
Blob service done in https://github.com/Azure/azure-sdk-for-cpp/pull/1238

PR #1281, opened 2021-01-07
Make IsServerEncrypted non-nullable.

katmsft:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1202

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
+++ b/sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
@@ -782,7 +782,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
       Azure::Core::Nullable<std::string> CopyProgress;
       Azure::Core::Nullable<std::string> CopySource;
       Azure::Core::Nullable<CopyStatusType> CopyStatus;
-      Azure::Core::Nullable<bool> IsServerEncrypted;
+      bool IsServerEncrypted = bool();
       std::string FileAttributes;
       Core::DateTime FileCreatedOn;
       Core::DateTime FileLastWrittenOn;
@@ -809,7 +809,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
       Azure::Core::Nullable<std::string> CopyProgress;
       Azure::Core::Nullable<std::string> CopySource;
       Azure::Core::Nullable<CopyStatusType> CopyStatus;
-      Azure::Core::Nullable<bool> IsServerEncrypted;
+      bool IsServerEncrypted = bool();
       std::string FileAttributes;
       Core::DateTime FileCreatedOn;
       Core::DateTime FileLastWrittenOn;
--- a/sdk/storage/azure-storage-files-shares/CHANGELOG.md
+++ b/sdk/storage/azure-storage-files-shares/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Removed `FileRange` and `ClearRange`, they are now represented with `Azure::Core::Http::Range`.
 - Removed `Offset` and `Length` pair in options. They are now represented with `Azure::Core::Http::Range`.
 - Replace scoped enums that don't support bitwise operations with extensible enum.
+- `IsServerEncrypted` member in `DownloadFileToResult`, `UploadFileFromResult`, `FileDownloadResult` and `FileGetPropertiesResult` are no longer nullable.
 
 ## 12.0.0-beta.5 (2020-11-13)
 
--- a/sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
+++ b/sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
@@ -118,7 +118,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares { names
     int64_t ContentLength = 0;
     ShareFileHttpHeaders HttpHeaders;
     Storage::Metadata Metadata;
-    Azure::Core::Nullable<bool> IsServerEncrypted;
+    bool IsServerEncrypted = false;
   };
 
   struct ForceCloseFileHandleResult
@@ -127,7 +127,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares { names
 
   struct UploadFileFromResult
   {
-    Azure::Core::Nullable<bool> IsServerEncrypted;
+    bool IsServerEncrypted = false;
   };
 
 }}}}} // namespace Azure::Storage::Files::Shares::Models


sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
```diff
@@ -782,7 +782,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
       Azure::Core::Nullable<std::string> CopyProgress;
       Azure::Core::Nullable<std::string> CopySource;
       Azure::Core::Nullable<CopyStatusType> CopyStatus;
-      Azure::Core::Nullable<bool> IsServerEncrypted;
+      bool IsServerEncrypted = bool();
```
Jinming-Hu, line 785: is `bool()` going to initialize the value `true` or `false`? Can you share some official source of it being defined behavior?

katmsft: https://www.learncpp.com/cpp-tutorial/boolean-values/

Search for `// default initialize to false`

Jinming-Hu: although I'm 99% confident it's defined somewhere in C++ standard, I don't think leancpp is **official**

Jinming-Hu: I can only find [this question](https://stackoverflow.com/questions/3308028/default-constructor-value-for-bool-type) after some googling. The accepted answer provided a good reference of C++14 draft N4296. I don't know if it's still the case for prior C++ 14 compilers.

Jinming-Hu: I also know that this whole SDK requires at least C++14 compatible compilers for now. But gcc-4.8 is still in our [supported platform matrix](https://azure.github.io/azure-sdk/cpp_introduction.html#linux). I remember that we've had some discussions about whether or not to drop support for GCC 4.8, but as far as I know, we haven't come into any conclusion yet.

So the situation here is that if you cannot find any official source stating this is defined behavior in C++11, and we in the future decide to add support for GCC 4.8, this piece of code will lead to a bug that's very hard for us to find out.

katmsft: I don't think so, as long as we verify on GCC4.8 what the behavior is, there is no problem here.

I remember this topic has already been escalated once, and I don't think using `bool()` is being prohibited after our discussion. Why are we re-raising this topic?

Jinming-Hu: > I don't think so, as long as we verify on GCC4.8 what the behavior is, there is no problem here.

This is wrong. There are a whole bunch of `undefined behavior`s in C++ world. So you define a variable `int x = x;` and print it out. You try some compiler on one platform and it outputs `0`. Can you get the conclusion that it's defined behavior and `int x= x;` always intialises `x` to zero?

Jinming-Hu: > I remember this topic has already been escalated once, and I don't think using bool() is being prohibited after our discussion.

Yeah I also remember that. But does `bool()` not being prohibited justify its being defined behavior?

katmsft: > Yeah I also remember that. But does bool() not being prohibited justify its being defined behavior?

As long as it is not prohibited, I don't think there is a concern about using it. I'll make the change if the guideline specifically states `Do not use bool() to initialize a boolean.`.

Jinming-Hu: okay, let me put it in this way.

What if you want to initialize this value to `true` rather than `false`. How would you write the code?

katmsft: Then I do `bool something = true;`.
Just as for int types, if I want to specify a non-zero value, I do `int32_t something = 8;` instead of `int32_t something = int32_t();` for zeros.

Jinming-Hu: Don't you feel this lacks something like symmetrical beauty?

So when you initialize bool to `true`, you use `bool a = true;`, while you initialize it to `false`, you use `bool a = bool();`.

katmsft: For this type, yes, it lacks symmetrical beauty, but it is way more consistent across types. `type()` is used to initialize most of the types within this SDK.
The only reason that bool is initialized is to prevent compiler warnings, not that the default value means anything. With this said I really don't care if it is defined or undefined behavior for `bool()`. Non-nullable bools are always going to be rewritten with meaningful values anyway.
If `bool a;` did not raise a compiler warning, like other non-primitive types, I won't initialize it and it will be equivalent to `bool a = bool();`.

Jinming-Hu: > type() is used to initialize most of the types within this SDK.

Give me a for instance.

Jinming-Hu: > The only reason that bool is initialized is to prevent compiler warnings, not that the default value means anything. With this said I really don't care if it is defined or undefined behavior for bool(). Non-nullable bools are always going to be rewritten with meaningful values anyway.
If bool a; did not raise a compiler warning, like other non-primitive types, I won't initialize it and it will be equivalent to bool a = bool();.

This makes sense

Jinming-Hu (approved):

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
katmsft: > > type() is used to initialize most of the types within this SDK.
> 
> Give me a for instance.

`int64_t()` is used to initalize `int64_t`, `int32_t()` is used to initialize `int32_t`. Any non-primitive types are using `type a;` which is equivalent of `type a = type();`, e.g. `DateTime`, `Nullable<T>`, extensible enum and so on.

Jinming-Hu: > int64_t() is used to initalize int64_t, int32_t() is used to initialize int32_t

does anybody else write code like this other than you? Because I don't think it's convincing if you prove your viewpoint with your own words.

Jinming-Hu: > Any non-primitive types are using type a; which is equivalent of type a = type();

I don't think you can just jump into this conclusion. https://stackoverflow.com/questions/1051379/is-there-a-difference-between-copy-initialization-and-direct-initialization

PR #1282, opened 2021-01-07
Added support for CreateIfNotExists and DeleteIfExists

katmsft:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
@@ -23,10 +23,6 @@ namespace Azure { namespace Storage {
     constexpr static const char* HttpHeaderContentType = "content-type";
     constexpr static const char* DefaultSasVersion = "2020-02-10";
 
-    // Error Codes.
-    constexpr static const char* ParentNotFound = "ParentNotFound";
-    constexpr static const char* ResourceNotFound = "ResourceNotFound";
-
     constexpr int ReliableStreamRetryCount = 3;
   } // namespace Details
 }} // namespace Azure::Storage
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
@@ -23,6 +23,10 @@ namespace Azure { namespace Storage {
     constexpr static const char* HttpHeaderContentType = "content-type";
     constexpr static const char* DefaultSasVersion = "2020-02-10";
 
+    // Error Codes.
+    constexpr static const char* ParentNotFound = "ParentNotFound";
+    constexpr static const char* ResourceNotFound = "ResourceNotFound";
+
     constexpr int ReliableStreamRetryCount = 3;
   } // namespace Details
 }} // namespace Azure::Storage
--- a/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
+++ b/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
@@ -1,0 +1,12 @@
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+namespace Azure { namespace Storage { namespace Files { namespace DataLake { namespace Details {
+  // Error codes:
+  constexpr static const char* DataLakeFilesystemNotFound = "FilesystemNotFound";
+  constexpr static const char* DataLakePathNotFound = "PathNotFound";
+
+}}}}} // namespace Azure::Storage::Files::DataLake::Details
--- a/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
+++ b/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
@@ -6,7 +6,10 @@
 
 namespace Azure { namespace Storage { namespace Files { namespace DataLake { namespace Details {
   // Error codes:
+  constexpr static const char* ContainerAlreadyExists = "ContainerAlreadyExists";
+  constexpr static const char* ContainerNotFound = "ContainerNotFound";
   constexpr static const char* DataLakeFilesystemNotFound = "FilesystemNotFound";
   constexpr static const char* DataLakePathNotFound = "PathNotFound";
+  constexpr static const char* DataLakePathAlreadyExists = "PathAlreadyExists";
 
 }}}}} // namespace Azure::Storage::Files::DataLake::Details
--- a/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
+++ b/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
@@ -35,7 +35,6 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
     bool Created = true;
     std::string ETag;
     Core::DateTime LastModified;
-    std::string NamespaceEnabled;
   };
 
   struct DeleteFileSystemResult
--- a/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
+++ b/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
@@ -21,7 +21,6 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
 
   // FileSystemClient models:
 
-  using DeleteFileSystemResult = FileSystemDeleteResult;
   using ListPathsResult = FileSystemListPathsResult;
 
   struct GetFileSystemPropertiesResult
@@ -31,12 +30,29 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
     Storage::Metadata Metadata;
   };
 
-  using CreateFileSystemResult = FileSystemCreateResult;
+  struct CreateFileSystemResult
+  {
+    bool Created = true;
+    std::string ETag;
+    Core::DateTime LastModified;
+    std::string NamespaceEnabled;
+  };
+
+  struct DeleteFileSystemResult
+  {
+    bool Deleted = true;
+  };
+
   using SetFileSystemMetadataResult = FileSystemCreateResult;
 
   // PathClient models:
 
-  using DeletePathResult = PathDeleteResult;
+  struct DeletePathResult
+  {
+    bool Deleted = true;
+    Azure::Core::Nullable<std::string> ContinuationToken;
+  };
+
   using AcquirePathLeaseResult = Blobs::Models::AcquireBlobLeaseResult;
   using RenewPathLeaseResult = Blobs::Models::RenewBlobLeaseResult;
   using ReleasePathLeaseResult = Blobs::Models::ReleaseBlobLeaseResult;
@@ -124,6 +140,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
 
   struct CreatePathResult
   {
+    bool Created = true;
     std::string ETag;
     Core::DateTime LastModified;
     Azure::Core::Nullable<int64_t> ContentLength;
@@ -161,6 +178,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
 
   struct DeleteFileResult
   {
+    bool Deleted = true;
   };
 
   struct DownloadFileToResult
@@ -185,6 +203,6 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
 
   using SetDirectoryAccessControlRecursiveResult = PathSetAccessControlRecursiveResult;
   using CreateDirectoryResult = CreatePathResult;
-  using DeleteDirectoryResult = PathDeleteResult;
+  using DeleteDirectoryResult = DeletePathResult;
 
 }}}}} // namespace Azure::Storage::Files::DataLake::Models
--- a/sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
+++ b/sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
@@ -277,7 +277,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     auto result = Details::DataLakeRestClient::Path::Create(
         destinationDfsUri, *m_pipeline, options.Context, protocolLayerOptions);
     // At this point, there is not more exception thrown, meaning the rename is successful.
-    auto ret = Models::RenameFileResult();
+    Models::RenameFileResult ret;
     return Azure::Core::Response<Models::RenameFileResult>(
         std::move(ret), result.ExtractRawResponse());
   }
@@ -289,7 +289,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     deleteOptions.AccessConditions = options.AccessConditions;
     deleteOptions.Context = options.Context;
     auto result = PathClient::Delete(deleteOptions);
-    auto ret = Models::DeleteFileResult();
+    Models::DeleteFileResult ret;
     ret.Deleted = true;
     return Azure::Core::Response<Models::DeleteFileResult>(
         std::move(ret), result.ExtractRawResponse());
@@ -302,7 +302,7 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     deleteOptions.AccessConditions = options.AccessConditions;
     deleteOptions.Context = options.Context;
     auto result = PathClient::DeleteIfExists(deleteOptions);
-    auto ret = Models::DeleteFileResult();
+    Models::DeleteFileResult ret;
     ret.Deleted = result->Deleted;
     return Azure::Core::Response<Models::DeleteFileResult>(
         std::move(ret), result.ExtractRawResponse());
--- a/sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
+++ b/sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
@@ -15,6 +15,7 @@
 #include <azure/storage/common/storage_per_retry_policy.hpp>
 #include <azure/storage/common/storage_retry_policy.hpp>
 
+#include "azure/storage/files/datalake/datalake_constants.hpp"
 #include "azure/storage/files/datalake/datalake_utilities.hpp"
 #include "azure/storage/files/datalake/version.hpp"
 
@@ -284,15 +285,25 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
   Azure::Core::Response<Models::DeleteFileResult> FileClient::Delete(
       const FileDeleteOptions& options) const
   {
-    Details::DataLakeRestClient::Path::DeleteOptions protocolLayerOptions;
-    protocolLayerOptions.LeaseIdOptional = options.AccessConditions.LeaseId;
-    protocolLayerOptions.IfMatch = options.AccessConditions.IfMatch;
-    protocolLayerOptions.IfNoneMatch = options.AccessConditions.IfNoneMatch;
-    protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
-    protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
-    auto result = Details::DataLakeRestClient::Path::Delete(
-        m_dfsUri, *m_pipeline, options.Context, protocolLayerOptions);
+    DeletePathOptions deleteOptions;
+    deleteOptions.AccessConditions = options.AccessConditions;
+    deleteOptions.Context = options.Context;
+    auto result = PathClient::Delete(deleteOptions);
     auto ret = Models::DeleteFileResult();
+    ret.Deleted = true;
+    return Azure::Core::Response<Models::DeleteFileResult>(
+        std::move(ret), result.ExtractRawResponse());
+  }
+
+  Azure::Core::Response<Models::DeleteFileResult> FileClient::DeleteIfExists(
+      const FileDeleteOptions& options) const
+  {
+    DeletePathOptions deleteOptions;
+    deleteOptions.AccessConditions = options.AccessConditions;
+    deleteOptions.Context = options.Context;
+    auto result = PathClient::DeleteIfExists(deleteOptions);
+    auto ret = Models::DeleteFileResult();
+    ret.Deleted = result->Deleted;
     return Azure::Core::Response<Models::DeleteFileResult>(
         std::move(ret), result.ExtractRawResponse());
   }
--- a/sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
+++ b/sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
@@ -12,6 +12,7 @@
 #include <azure/storage/common/storage_per_retry_policy.hpp>
 #include <azure/storage/common/storage_retry_policy.hpp>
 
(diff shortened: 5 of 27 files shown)


sdk/storage/azure-storage-files-shares/src/share_directory_client.cpp
```diff
@@ -167,16 +167,77 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
...
+      const DeleteDirectoryOptions& options) const
+  {
+    try
+    {
+      return Delete(options);
+    }
+    catch (StorageException& e)
+    {
+      if (e.ErrorCode == Details::ShareNotFound || e.ErrorCode == Storage::Details::ParentNotFound
+          || e.ErrorCode == Storage::Details::ResourceNotFound)
```
Jinming-Hu, line 232: When will server return `ParentNotFound` or `ResourceNotFound`?

Jinming-Hu: How are they different from `ShareNotFound`?

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
```diff
@@ -0,0 +1,12 @@
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+namespace Azure { namespace Storage { namespace Files { namespace DataLake { namespace Details {
+  // Error codes:
+  constexpr static const char* DataLakeFilesystemNotFound = "FilesystemNotFound";
+  constexpr static const char* DataLakePathNotFound = "PathNotFound";
```
Jinming-Hu, line 10: I think if you define error code as constant strings, you should also define all other error codes here.

sdk/storage/azure-storage-files-shares/CHANGELOG.md
```diff
@@ -20,6 +20,11 @@
 - Removed `FileRange` and `ClearRange`, they are now represented with `Azure::Core::Http::Range`.
 - Removed `Offset` and `Length` pair in options. They are now represented with `Azure::Core::Http::Range`.
 - Replace scoped enums that don't support bitwise operations with extensible enum.
+- Create APIs for Directory and File now returns `FileShareSmbProperties` that aggregates SMB related properties.
+
+### New Features
+
+- Added support for `CreateIfNotExists` for Share and Directory clients, and `DeleteIfExists` for Share, Directory and File clients.
```
Jinming-Hu, line 27: Does it make this PR not `single-purpose`?

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
```diff
@@ -9,15 +9,49 @@
...
+     * @brief Key of the permission to be set for the directory/file.
+     */
+    Azure::Core::Nullable<std::string> PermissionKey;
+
+    /**
+     * @brief If specified, the provided file attributes shall be set. Default value:
+     * FileAttribute::Archive for file and FileAttribute::Directory for directory.
+     * FileAttribute::None can also be specified as default.
+     */
+    FileAttributes Attributes = static_cast<FileAttributes>(0);
```
Jinming-Hu, line 24: Should we have a `None` attribute?

Jinming-Hu: `static_cast<FileAttributes>(0)`

Is this what our customers are supposed to write code?

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
```diff
@@ -9,15 +9,49 @@
...
+
   // ServiceClient models:
 
   using ListSharesSinglePageResult = ServiceListSharesSinglePageResult;
   using SetServicePropertiesResult = ServiceSetPropertiesResult;
   using GetServicePropertiesResult = StorageServiceProperties;
 
   // ShareClient models:
-  using CreateShareResult = ShareCreateResult;
-  using DeleteShareResult = ShareDeleteResult;
```
Jinming-Hu, line 20: I got a questions here. `ShareCreateResult` is in `Models` namespace. But it looks to me is an implementation detail and shouldn't be exposed to our customers. So should we move these return types to `Details` namespace?

sdk/storage/azure-storage-files-shares/src/share_directory_client.cpp
katmsft: `ParentNotFound` when parent directory is not there, and `ResourceNotFound` when the directory to be deleted is not there. Either the share does not exist, or the parent directory does not exist or the resource does not exist should not be reporting error here.

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
katmsft: It's in the `Details` namespace. I don't think there is a hard requirement on defining them all.

sdk/storage/azure-storage-files-shares/CHANGELOG.md
katmsft: Sure, I'll remove the change.

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
katmsft: Yes, but due to the `single-purpose` limitation, I think I won't address this issue in this PR, feel free to open an issue to track it.

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/1287

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
katmsft: Revisited what .NET does, we do have `None` at the moment but we shouldn't be initializing it to this value. I have made the member to `Azure::Core::Nullable<FileAttributes> Attributes;` to resolve this issue.

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_constants.hpp
Jinming-Hu: Well, I don't feel strong about it. So suit yourself

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
Jinming-Hu: so when `Attributes` will be null?

Jinming-Hu: is this `FileShareSmbProperties` struct used in return  types or options?

katmsft: As discussed with Ted, it is used in both return types and options. Customer doesn't have to specify it in options, but I think in return types it will always not be null.

Jinming-Hu: Well this is not good. If it's always non-null, it should not be nullable.

Jinming-Hu: Here's another question, when it's used in options, how will our customer specify `None` to clear all attributes, and `preserve`, and `inherit`?

katmsft: https://github.com/Azure/azure-sdk-for-net/blob/4f191fb32a70446efa6b1efc729692600c0e1587/sdk/storage/Azure.Storage.Files.Shares/src/ShareFileClient.cs#L716
In the REST layer it is required, but .NET makes it nullable and clears what the server stores if it is null just like other properties.

Jinming-Hu: what if the user wants  to use `preserve` or `inherit`?

katmsft: I think it is a feature missing in .NET SDK.
`inherit` is not available for the parameter so one thing we can do in C++ is that null means `preserve`. But we have the risk of future adding `inherit` to this support matrix. What do you think?

Jinming-Hu: I don't know

katmsft: Email sent, I'm proposing to add `preserve` as an attribute enum, let's see what others think.

katmsft: this will be resolved in another PR as it is not related here.

sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
```diff
@@ -23,6 +23,10 @@ namespace Azure { namespace Storage {
     constexpr static const char* HttpHeaderContentType = "content-type";
     constexpr static const char* DefaultSasVersion = "2020-02-10";
 
+    // Error Codes.
+    constexpr static const char* ParentNotFound = "ParentNotFound";
+    constexpr static const char* ResourceNotFound = "ResourceNotFound";
```
Jinming-Hu, line 28: Why are these two so special that they should go in common `constants.hpp` rather than constants in each service?

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
```diff
@@ -31,12 +30,29 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
     Storage::Metadata Metadata;
   };
 
-  using CreateFileSystemResult = FileSystemCreateResult;
+  struct CreateFileSystemResult
+  {
+    bool Created = true;
+    std::string ETag;
+    Core::DateTime LastModified;
+    std::string NamespaceEnabled;
```
Jinming-Hu, line 38: what's the possible value of this field?

sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
```diff
@@ -284,15 +285,25 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
...
-    protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
-    protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
-    auto result = Details::DataLakeRestClient::Path::Delete(
-        m_dfsUri, *m_pipeline, options.Context, protocolLayerOptions);
+    DeletePathOptions deleteOptions;
+    deleteOptions.AccessConditions = options.AccessConditions;
+    deleteOptions.Context = options.Context;
+    auto result = PathClient::Delete(deleteOptions);
+    auto ret = Models::DeleteFileResult();
+    ret.Deleted = true;
```
Jinming-Hu, line 293: Isn't this `true` already?

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
```diff
@@ -194,11 +195,32 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     Models::CreateFileSystemResult ret;
     ret.ETag = std::move(result->ETag);
     ret.LastModified = std::move(result->LastModified);
+    ret.Created = true;
```
Jinming-Hu, line 198: Isn't this `true` already?

sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
```diff
@@ -284,15 +285,25 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
...
-    protocolLayerOptions.IfNoneMatch = options.AccessConditions.IfNoneMatch;
-    protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
-    protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
-    auto result = Details::DataLakeRestClient::Path::Delete(
-        m_dfsUri, *m_pipeline, options.Context, protocolLayerOptions);
+    DeletePathOptions deleteOptions;
+    deleteOptions.AccessConditions = options.AccessConditions;
+    deleteOptions.Context = options.Context;
+    auto result = PathClient::Delete(deleteOptions);
+    auto ret = Models::DeleteFileResult();
```
Jinming-Hu, line 292: You use `Models::CreateFileSystemResult ret;` in another file. Be consistent.

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
```diff
@@ -194,11 +195,32 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
...
+      if (e.ErrorCode == Details::ContainerAlreadyExists)
+      {
+        Models::CreateFileSystemResult ret;
+        ret.Created = false;
+        return Azure::Core::Response<Models::CreateFileSystemResult>(
+            std::move(ret), std::move(e.RawResponse));
+      }
+      throw;
+    }
+  }
```
Jinming-Hu, line 221: why not call `BlobContainerClient::CreateIfNotExists`?

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
```diff
@@ -207,11 +229,31 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     blobOptions.AccessConditions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
     blobOptions.AccessConditions.LeaseId = options.AccessConditions.LeaseId;
     auto result = m_blobContainerClient.Delete(blobOptions);
-    Models::FileSystemDeleteResult ret;
-    return Azure::Core::Response<Models::FileSystemDeleteResult>(
+    Models::DeleteFileSystemResult ret;
+    ret.Deleted = true;
```
Jinming-Hu, line 233: Isn't this `true` already?

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
```diff
@@ -207,11 +229,31 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
...
+    {
+      if (e.ErrorCode == Details::ContainerNotFound)
+      {
+        Models::DeleteFileSystemResult ret;
+        ret.Deleted = false;
+        return Azure::Core::Response<Models::DeleteFileSystemResult>(ret, std::move(e.RawResponse));
+      }
+      throw;
+    }
+  }
```
Jinming-Hu, line 255: why not call `BlobContainerClient::DeleteIfExists`?

sdk/storage/azure-storage-files-datalake/src/datalake_path_client.cpp
```diff
@@ -267,8 +291,34 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
     protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
     protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
     protocolLayerOptions.RecursiveOptional = options.Recursive;
-    return Details::DataLakeRestClient::Path::Delete(
+    auto result = Details::DataLakeRestClient::Path::Delete(
         m_dfsUri, *m_pipeline, options.Context, protocolLayerOptions);
+    auto ret = Models::DeletePathResult();
```
Jinming-Hu, line 296: Be consistent.

sdk/storage/azure-storage-files-datalake/test/datalake_file_client_test.cpp
```diff
@@ -104,6 +104,30 @@ namespace Azure { namespace Storage { namespace Test {
     }
   }
 
+  TEST_F(DataLakeFileClientTest, CreateDeleteIfExistsFiles)
+  {
+    {
+      auto client = m_fileSystemClient->GetFileClient(LowercaseRandomString());
```
Jinming-Hu, line 110: Does filename have to be lowercase string?

sdk/storage/azure-storage-files-datalake/test/datalake_file_client_test.cpp
```diff
@@ -104,6 +104,30 @@ namespace Azure { namespace Storage { namespace Test {
...
+      EXPECT_NO_THROW(created = client.CreateIfNotExists()->Created);
+      EXPECT_FALSE(created);
+      EXPECT_NO_THROW(deleted = client.Delete()->Deleted);
+      EXPECT_TRUE(deleted);
+      EXPECT_NO_THROW(deleted = client.DeleteIfExists()->Deleted);
+      EXPECT_FALSE(deleted);
+    }
+    {
+      auto client = Files::DataLake::FileClient::CreateFromConnectionString(
+          AdlsGen2ConnectionString(), LowercaseRandomString(), LowercaseRandomString());
```
Jinming-Hu, line 124: ditto

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
Jinming-Hu: Create an issue to track that.

sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
katmsft: Those two might be used by other services. There is no specific resource related keyword.

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
katmsft: True/False I suppose, but I haven't tested it yet.

sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
katmsft: Yes, but not relying on the default initialized value is a good practice as far as I'm concerned.

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
katmsft: Yes, but not relying on the default initialized value is a good practice as far as I'm concerned.

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
katmsft: It does not have `Created` in the response.

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
katmsft: Yes, but not relying on the default initialized value is a good practice as far as I'm concerned.

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
katmsft: It does not have `Deleted` in the response.

sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
Jinming-Hu: For example, they're used by which and which service?

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
Jinming-Hu: Why isn't it a `bool`?

sdk/storage/azure-storage-files-datalake/src/datalake_file_system_client.cpp
Jinming-Hu: I will add that, but I don't think that will affect your impl here. You'll need to instantiate a `DeleteFilesystemResult` anyway.

sdk/storage/azure-storage-files-datalake/src/datalake_file_client.cpp
Jinming-Hu: Well, I don't agree but fine.

sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
katmsft: I don't have evidence it is never to be used in other services, now or in the future.

Jinming-Hu: If you don't have evidence, is it reasonable that it won't be used by multiple services?

Because if the other way around, we can add anything to common package.

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
katmsft: Actually, I'm going to remove this, since blob is not supporting this return value.

Jinming-Hu: Sure, removing it is fine.

My only concern is that you have several TODO items that don't have linked issues.

katmsft: I'll do it in this PR since it is raised as a comment in this PR. I think it justifies 'related to this PR'.

Jinming-Hu: Nah, I don't think it justifies the relevance. I'm going to raise more comments. Like some const variables still have `c_` prefix.

sdk/storage/azure-storage-common/inc/azure/storage/common/constants.hpp
katmsft: Moved to file service.

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_responses.hpp
katmsft: https://github.com/Azure/azure-sdk-for-cpp/issues/1305

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
katmsft: Then I think it's best to create issues to track those.

Jinming-Hu (approved):

sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_responses.hpp
Jinming-Hu: Then create an issue

Issue #1098, opened 2020-12-08
Figure out how to re-use the same session handle when using WinHTTP across all the requests within the entire application


ahsonkhan:
Currently, every request creates a new session.
This also requires figuring out the appropriate time to do the session handle cleanup.

See https://github.com/Azure/azure-sdk-for-cpp/pull/897#discussion_r517812289

mikeharder:
## Overview
Currently, each HTTP request sent via WinHTTP creates a new TCP connection, which has two major impacts:

1.  Performance is significantly reduced, over 10x for small requests.  Creating a new TCP connection (and completing the HTTPS handshake) is very expensive compared to sending a request over an existing connection.

2.  After each request, the TCP connection is closed, and the client port is left in state TIME_WAIT for 2 minutes.  On Windows, there are 63475 available client ports by default.  So if an application makes 63475 requests in under 2 minutes, all future TCP connections will be blocked until the TIME_WAIT on each port expires, effectively blocking any network activity from the machine.

## Repro Steps
You can repro this using the Storage Blob performance tests, and use a tool like WireShark to monitor the TCP traffic.

1. Create a storage account
    A. Performance: Premium
    B. Premium account type: Block blobs
    C. Redundancy: Locally-redundant storage (LRS)
2. Compile azure-storage-blobs-perf.exe per instructions at https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/perf/README.md
3. Set env var `STORAGE_CONNECTION_STRING` pointing to your storage account
4. Run perf test to verify it's working.  The exact throughput doesn't matter, just verify the test passes with some reasonable throughput.

```
> azure-storage-blobs-perf.exe DownloadBlob --warmup 0 --duration 5 --size 1024

=== Results ===
Completed 115 operations in a weighted-average of 5s (22.834333 ops/s, 0.0437937 s/op)
```
5. Install a network monitoring tool like WireShark.
6. Open WireShark, select "Ethernet" interface to start capturing
7. Run perf test again
8. In WireShark, click "Capture -> Stop".
9. Find the IP address of your storage account, by running "nslookup your-storage-account.blob.core.windows.net".
10. In WireShark, add a filter with text `ip.addr==your.storage.ip.address`.
11. You will see that a new TCP connection is used for each HTTPS request
    A. The start and end of TCP sessions are shown with a dark gray background, and will contain `SYN` for the start and `FIN` for the end of a session.
[...]

PR #1285, opened 2021-01-08
docs for the HTTP transport adapter

vhvb1989:
fixes: #902


doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
+# HTTP Transport Adapter
+The Azure SDK for C++ creates and sends HTTP request using an HTTP Transport Adapter. This adapter provides a general abstraction for an HTTP stack (like libcurl, WinHTTP, etc). It defines the interface that the Azure SDK for C++ will call to put the HTTP request into the wire.
```
ahsonkhan, line 2:
```suggestion
# HTTP Transport Adapter

The Azure SDK for C++ creates and sends HTTP requests using an HTTP Transport Adapter. This adapter provides a general abstraction for an HTTP stack (like libcurl, WinHTTP, etc) by defining the interface it needs to call that anyone can implement to customize which HTTP stack gets used.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Default HTTP Transport Adapter
+The Azure SDK for C++ provides two HTTP transport adapters as part of the azure-core package:
+- The libcurl transport adapter
+- WinHttp transport adapter
+
+Azure SDK for C++ users can create and use their own transport adapter as well [see below](#Building-a-custom-HTTP-Transport-Adapter).
```
ahsonkhan, line 9:
```suggestion
Azure SDK for C++ users can create and use their own transport adapter as well [by following the guidance below](#Building-a-custom-HTTP-Transport-Adapter).
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+For the simplest case, when there was not any specific configuration set for the Azure SDK for C++, the default transport adapter will be selected based on the OS. The WinHTTP transport adapter will be used for Windows-based systems (making libcurl an optional dependency). The libcurl transport adapter will be selected for any non-windows-based system.
```
ahsonkhan, line 11: Great!

```suggestion
For the simplest case, when no specific configuration is set, the default transport adapter will be selected based on the OS. The WinHTTP transport adapter will be used for Windows-based systems (making libcurl an optional dependency). The libcurl transport adapter will be selected for any non-windows-based system.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Building HTTP transport adapter
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required for each HTTP transport adapter.
+
+Multiple HTTP transport adapters can be built as part of the project. For example, if you want to send HTTP request with the libcurl transport adapter for some cases and the WinHTTP transport adapter for another cases.
```
ahsonkhan, line 18:
```suggestion
Multiple HTTP transport adapters can be built as part of the project. This is to support scenarios where you want to send HTTP request with the libcurl transport adapter for some cases and the WinHTTP transport adapter for others.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+Note that combining WinHTTP transport adapter with another transport adapter is only possible when using Windows.
+
+Another example is if you want to create your own HTTP transport adapter for some other C++ HTTP stack. In this case, you can build your own HTTP transport adapter and also the libcurl HTTP transport adapter and run tests to compare the correctness of both.
```
ahsonkhan, line 21:
```suggestion
Another example is if you want to create your own HTTP transport adapter for some other C++ HTTP stack. In this case, you can build your own HTTP transport adapter and test it by comparing it to the behavior of the libcurl HTTP transport adapter.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required for each HTTP transport adapter.
+
+Multiple HTTP transport adapters can be built as part of the project. For example, if you want to send HTTP request with the libcurl transport adapter for some cases and the WinHTTP transport adapter for another cases.
+Note that combining WinHTTP transport adapter with another transport adapter is only possible when using Windows.
+
+Another example is if you want to create your own HTTP transport adapter for some other C++ HTTP stack. In this case, you can build your own HTTP transport adapter and also the libcurl HTTP transport adapter and run tests to compare the correctness of both. 
+
+## Using the HTTP transport adapter
+The HTTP transport adapter is set up when a service SDK client is init. For example, when creating an Azure Storage SDK client. The HTTP transport adapter is part of the client options argument that can be used to init the client. See the next example.
```
ahsonkhan, line 24:
```suggestion
The HTTP transport adapter is set up when a service SDK client is initialized, for example, when creating an Azure Storage SDK client. The HTTP transport adapter is part of the client options argument that can be used when constructing the client. See the following code snippet as an example:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+The Azure SDK for C++ provides two HTTP transport adapters as part of the azure-core package:
+- The libcurl transport adapter
+- WinHttp transport adapter
+
+Azure SDK for C++ users can create and use their own transport adapter as well [see below](#Building-a-custom-HTTP-Transport-Adapter).
+
+For the simplest case, when there was not any specific configuration set for the Azure SDK for C++, the default transport adapter will be selected based on the OS. The WinHTTP transport adapter will be used for Windows-based systems (making libcurl an optional dependency). The libcurl transport adapter will be selected for any non-windows-based system.
+
+## Building HTTP transport adapter
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
```
ahsonkhan, line 14: Same here and elsewhere. Add new lines between title and body.
```suggestion
## Building HTTP transport adapter

The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+Another example is if you want to create your own HTTP transport adapter for some other C++ HTTP stack. In this case, you can build your own HTTP transport adapter and also the libcurl HTTP transport adapter and run tests to compare the correctness of both. 
+
+## Using the HTTP transport adapter
+The HTTP transport adapter is set up when a service SDK client is init. For example, when creating an Azure Storage SDK client. The HTTP transport adapter is part of the client options argument that can be used to init the client. See the next example.
+```cpp
+  /*
+  * Using the HTTP transport adapter.
+  */
+
+  // When no options are provided, the HTTP transport adapter is set to the default one.
```
ahsonkhan, line 30:
```suggestion
  // When no options are provided, the HTTP transport adapter is set to the default one based on the compiler options passed in to CMake.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  auto defaultStorageClient = BlobServiceClient(url, credential);
+
+  // Use client options to set an HTTP transport adapter.
```
ahsonkhan, line 33:
```suggestion
  // Use client options to set an HTTP transport adapter, which needs to be a shared_ptr.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = std::make_shared<Azure::Core::Http::CurlTransport>();
+  auto storageClient = BlobServiceClient(url, credential, options);
+```
+
+### Re-use the HTTP transport adapter 
+Note that the HTTP transport adapter is a `shared_ptr`. This is because you can re-use the same HTTP transport adapter for multiple clients. There are two ways of doing this. The first one is by using the same HTTP transport adapter to init a second client. The second one is when the client can create a new client directly witch inherits the same configuration. See the next example:
```
ahsonkhan, line 40:
```suggestion
Note that the HTTP transport adapter is a `shared_ptr`. This is because you can re-use the same HTTP transport adapter for multiple clients. There are two ways of doing this. The first one is by using the same HTTP transport adapter when creating another client. The second one is when you get a specific child client from an already created parent client directly, in which case the child client inherits the same configuration. See the next example:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+```cpp
+  /* 
+  *  Option 1. Init a new client with the same HTTP transport adapter.
+  */
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>();
+  BlobClientOptions optionsA;
+  optionsA.TransportPolicyOptions.Transport = curlTransportAdapter;
+  auto storageClientA = BlobServiceClient(url, credential, options);
```
ahsonkhan, line 49:
```suggestion
  auto storageClientA = BlobServiceClient(url, credential, optionsA);
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  // The second client.
+  BlobClientOptions optionsB;
+  optionsB.TransportPolicyOptions.Transport = curlTransportAdapter;
+  auto storageClientB = BlobServiceClient(url, credential, options);
```
ahsonkhan, line 53:
```suggestion
  auto storageClientB = BlobServiceClient(url, credential, optionsB);
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+  /* 
+  *  Option 2. Create new client from a parent client.
+  */
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>();
+  BlobClientOptions optionsA;
+  optionsA.TransportPolicyOptions.Transport = curlTransportAdapter;
```
ahsonkhan, line 60:
```suggestion
  BlobClientOptions options;
  options.TransportPolicyOptions.Transport = curlTransportAdapter;
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  auto storageClientA = BlobServiceClient(url, credential, options);
+  // Create clientB from clientA.
+  auto clientB = storageClientA.GetBlobContainerClient(containerName)
```
ahsonkhan, line 63:
```suggestion
  auto blobContainerClient = storageClient.GetBlobContainerClient(containerName)
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  auto storageClientB = BlobServiceClient(url, credential, options);
+
+  /* 
+  *  Option 2. Create new client from a parent client.
+  */
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>();
+  BlobClientOptions optionsA;
+  optionsA.TransportPolicyOptions.Transport = curlTransportAdapter;
+  auto storageClientA = BlobServiceClient(url, credential, options);
+  // Create clientB from clientA.
```
ahsonkhan, line 62:
```suggestion
  // Create specific child client from the parent service client.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  optionsB.TransportPolicyOptions.Transport = curlTransportAdapter;
+  auto storageClientB = BlobServiceClient(url, credential, options);
+
+  /* 
+  *  Option 2. Create new client from a parent client.
+  */
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>();
+  BlobClientOptions optionsA;
+  optionsA.TransportPolicyOptions.Transport = curlTransportAdapter;
+  auto storageClientA = BlobServiceClient(url, credential, options);
```
ahsonkhan, line 61:
```suggestion
  auto storageClient = BlobServiceClient(url, credential, options);
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  // Create clientB from clientA.
+  auto clientB = storageClientA.GetBlobContainerClient(containerName)
+```
+
+Note that the second option could be not supported for some Azure SDK clients if it doesn't make sense for the service behavior (like Key Vault client).
```
ahsonkhan, line 66:
```suggestion
Note that the second option may not supported for some Azure SDK clients if it is not applicable depending on the service behavior (for example, the Key Vault client).
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+The Azure SDK for C++ client `will hold an internal copy of the options` passed in the arguments. This means that the options object used to init the client can be disposed. However, since the HTTP transport adapter is a shared ptr, it won't be disposed until any client using it is disposed first.
```
ahsonkhan, line 68: Repeating `Azure SDK for C++` as a prefix too much hurts the readability and understanding of the concept at hand. It's OK to use it in some places, but not everywhere.

```suggestion
Each client holds an internal copy of the options passed in the arguments. This means that the options object used to create the client can be disposed. However, since the HTTP transport adapter is a shared pointer, it won't be disposed until all clients using it are disposed first.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+## HTTP Transport Adapter Options
+The libcurl and WinHTTP transport adapters can also be init with specif options that control features from the HTTP stack underneath it.
```
ahsonkhan, line 71:
```suggestion
The libcurl and WinHTTP transport adapters can also be initialized with specific options that expose control of features from the HTTP stack underneath it. For example, you can set a proxy or override the default certificate authority (CA) path for libcurl. See the following example:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+For example, you can set a PROXY or override the default CA path for libcurl. See next example.
```
ahsonkhan, line 72: Avoid separating a single paragraph into a new line explicitly, since it has no impact on how the markdown is rendered. Suggested a change to the wording and merge into line 71 above.
```suggestion
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+```cpp
+  /*
+  * Customize libcurl http transport adapter.
```
ahsonkhan, line 75:
```suggestion
  * Customize libcurl HTTP transport adapter.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  */
+
+  // Create HTTP transport adapter options and set options.
```
ahsonkhan, line 78:
```suggestion
  // Create the HTTP transport adapter options and set them when initializing the adapter.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+  Azure::Core::Http::CurlTransportOptions curlTransportOption;
+  curlTransportOption.Proxy = "https://my.proxy.com";
+  // Create HTTP transport adapter with options.
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>(curlTransportOption);
```
ahsonkhan, line 82: nit:
```suggestion
  Azure::Core::Http::CurlTransportOptions curlTransportOptions;
  curlTransportOptions.Proxy = "https://my.proxy.com";
  // Create HTTP transport adapter with options.
  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>(curlTransportOptions);
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = curlTransportAdapter; 
+  auto storageClient = BlobServiceClient(url, credential, options);
+```
+
+The HTTP transport adapter are specific for each adapter and are optional. This means that if you are creating your own HTTP transport adapter, you don't need to create options for it if you don't want to.
```
ahsonkhan, line 89:
```suggestion
The HTTP transport adapter options are specific for each adapter and are optional. This means that if you are creating your own HTTP transport adapter, you don't need to create options for it if you don't want to.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## Building a custom HTTP Transport Adapter
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required to be set when building your own HTTP transport adapter.
+
+Follow next steps to implement your own HTTP Transport Adapter.
+
+1. Implement interface. The HTTP transport adapter interface is define in `Azure::Core::HTTP::HttpTransport` from `inc/azure/core/http/transport.hpp`.
```
ahsonkhan, line 99:
```suggestion
1. Implement the interface. The HTTP transport adapter interface is define in `Azure::Core::HTTP::HttpTransport` from `inc/azure/core/http/transport.hpp`.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+The HTTP transport adapter are specific for each adapter and are optional. This means that if you are creating your own HTTP transport adapter, you don't need to create options for it if you don't want to.
+
+
+## Building a custom HTTP Transport Adapter
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required to be set when building your own HTTP transport adapter.
+
+Follow next steps to implement your own HTTP Transport Adapter.
+
```
ahsonkhan, line 98: There seems to be just one step. Remove the numbered list?

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,118 @@
...
+class CurlTransport : public Azure::Core::Http::HttpTransport {
+  public:
+    // Override the Send method
+    std::unique_ptr<Azure::Core::Http::RawResponse> Send(
+        Azure::Core::Context const& context,
+        Azure::Core::Http::Request& request) override;
+};
+```
+
+The implementation needs to take care of using the data from the argument `request` containing all the HTTP request information to write the HTTP request using some C++ HTTP stack. Then it needs to create the raw response from the response and return it. The `context` argument can be optionally used to check if the request is cancelled before performing I/O operations (like writing to the network).
```
ahsonkhan, line 118: We may want to call out the need to return the response as streaming data and wrapping that in a bodystream and provide some example.

ahsonkhan, line 118:
```suggestion
The implementation needs to take care of using the members from the `Request` argument which contains all the HTTP request information to write the HTTP request using some C++ HTTP stack. Then it needs to create the `RawResponse` from the response and return it. The `context` argument can be optionally used to check if the request is cancelled before performing I/O operations (like writing to the network). 
```

ahsonkhan (commented):
Great document. Some suggestions and fixups.

CONTRIBUTING.md
```diff
@@ -165,6 +165,9 @@ ctest -N
 # Use -R to use a regular exp for what to run
 ctest -R Http # runs only Http tests
 ```
+
+The Azure SDK for C++ supports libcurl and WinHTTP libraries as HTTP stacks and provides an HTTP transport adapter for each one. It can also support an external, `customer-implemented` *HTTP transport adapter*. To learn more about this feature, [click here](LinkForTransportAdapterDocs).
```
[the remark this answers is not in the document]
ahsonkhan: https://dev.azure.com/azure-sdk/public/_build/results?buildId=678382&view=logs&j=a129effc-2dd1-54d1-fb5a-ad7bdc0e851d&t=ccad100f-2904-5913-2083-c525ad929a03&l=52
Looks like link verification is causing errors:
> Summary of broken links:
'file:///D:/a/1/s/sdk/core/azure-core/README.md' has 1 broken link(s):
  LinkForTransportAdapterDocs
'file:///D:/a/1/s/CONTRIBUTING.md' has 1 broken link(s):
  LinkForTransportAdapterDocs
'file:///D:/a/1/s/doc/HttpTransportAdapter.md' has 1 broken link(s):
  #Building-a-custom-HTTP-Transport-Adapter

Also two warnings:
> ##[warning]Please lower case your anchor tags (i.e. anything after '#' in your link '#Building-a-custom-HTTP-Transport-Adapter'. Check here for more information: https://aka.ms/azsdk/guideline/links

> ##[warning]DO NOT use relative link LinkForTransportAdapterDocs. Please use absolute link instead. Check here for more information: https://aka.ms/azsdk/guideline/links

doc/HttpTransportAdapter.md
vhvb1989: Yes, I though about it but I considered that we need to create a new document for how the bodyStream and the responses work and then link to this section.  But yes, you are right. that needs to be said.

vhvb1989:
> Great document. Some suggestions and fixups.

Thank you very much for you detailed review!  It was awesome!  I have applied your suggested changes.

I will work later on the pending comments and will upted the broken links that are causing the CI to fail

CONTRIBUTING.md
```diff
@@ -165,6 +165,9 @@ ctest -N
 # Use -R to use a regular exp for what to run
 ctest -R Http # runs only Http tests
 ```
+
+Out of the box, the Azure SDK for C++ supports the libcurl and WinHTTP libraries as HTTP stacks for communicating with Azure services over the network and provide an HTTP transport adapter for each one. It can also support an external, `customer-implemented` *HTTP transport adapter* based on other HTTP libraries. To learn more about this feature, [click here](LinkForTransportAdapterDocs).
```
RickWinter, line 169:
```suggestion
Out of the box, the Azure SDK for C++ supports the libcurl and WinHTTP libraries as HTTP stacks for communicating with Azure services over the network.  The SDK also provides a mechanism for `customer-implemented` *HTTP transport adapter*. To learn more about this feature, [click here](LinkForTransportAdapterDocs).
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,119 @@
...
+For the simplest case, when no specific configuration is set, the default transport adapter will be selected based on the OS. The WinHTTP transport adapter will be used for Windows-based systems (making libcurl an optional dependency). The libcurl transport adapter will be selected for any non-windows-based system.
+
+## Building HTTP transport adapter
+
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required for each HTTP transport adapter.
+
+Multiple HTTP transport adapters can be built as part of the project. This is to support scenarios where you want to send HTTP request with the libcurl transport adapter for some cases and the WinHTTP transport adapter for others.
+Note that combining WinHTTP transport adapter with another transport adapter is only possible when using Windows.
```
RickWinter, line 21:
```suggestion
Note WinHTTP transport adapter is only supported on Windows.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,119 @@
...
+
+Another example is if you want to create your own HTTP transport adapter for some other C++ HTTP stack. In this case, you can build your own HTTP transport adapter and test it by comparing it to the behavior of the libcurl HTTP transport adapter.
+
+## Using the HTTP transport adapter
+The HTTP transport adapter is set up when a service SDK client is initialized, for example, when creating an Azure Storage SDK client. The HTTP transport adapter is part of the client options argument that can be used when constructing the client. See the following code snippet as an example:
```
RickWinter, line 26:
```suggestion
The HTTP transport adapter is set up during service SDK client initialization, for example, when creating an Azure Storage SDK client. The HTTP transport adapter can be specified at client initialization via the client options argument. See the following code snippet as an example:
```

RickWinter (approved):

doc/HttpTransportAdapter.md
vhvb1989: Agree.
The only benefit I see is when people copy a direct link to some paragraph tittle in the middle of the page like:
https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/core/Azure.Core#configuring-service-clients-using-clientoptions

In those cases, it could be good when the `Azure SDK for XX` is mentioned on every tittle to improve cases like this.

But I don't really care too much about it anyway.  I am really happy we have @ahsonkhan  in the team to help to improve the docs redaction and quality!

doc/HttpTransportAdapter.md
vhvb1989: Added the missing steps.
Somehow I forgot to continue and finish with all the steps I was hoping to add initially. I think I got lost on step 1 and forgot about it hahahah

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
+# HTTP Transport Adapter
+
+The Azure SDK for C++ creates and sends HTTP requests using an HTTP Transport Adapter. This adapter provides a general abstraction for an HTTP stack (like libcurl, WinHTTP, etc) by defining the interface it needs to call that anyone can implement to customize which HTTP stack gets used.
+
+## Default HTTP Transport Adapter
+The Azure SDK for C++ provides two HTTP transport adapters as part of the azure-core package:
+- The libcurl transport adapter
```
JeffreyRichter, line 8: Can you please add on platforms these work. For example: libcurl is for Windows, Linux, Mac; WinHTTP is for Windows

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+- WinHttp transport adapter
```
JeffreyRichter, line 9: Be consistent with casing: WinHTTP throughout this doc

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+Azure SDK for C++ users can create and use their own transport adapter as well [by following the guidance below](#building-a-custom-http-transport-adapter).
```
JeffreyRichter, line 11: delete "as well"

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+For the simplest case, when no specific configuration is set, the default transport adapter will be selected based on the OS. The WinHTTP transport adapter will be used for Windows-based systems (making libcurl an optional dependency). The libcurl transport adapter will be selected for any non-windows-based system.
```
JeffreyRichter, line 13: Windows needs an uppercase W: non-Windows-based

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Building HTTP transport adapter
```
JeffreyRichter, line 15: Building an HTTP Transport Adapter

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required for each HTTP transport adapter.
+
+Multiple HTTP transport adapters can be built as part of the project. This is to support scenarios where you want to send HTTP request with the libcurl transport adapter for some cases and the WinHTTP transport adapter for others.
+Note WinHTTP transport adapter is only supported on Windows.
```
JeffreyRichter, line 22: You can remove this note when you add what I suggest earlier

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+Another example is if you want to create your own HTTP transport adapter for some other C++ HTTP stack. In this case, you can build your own HTTP transport adapter and test it by comparing it to the behavior of the libcurl HTTP transport adapter.
```
JeffreyRichter, line 24: delete "for some other C++ HTTP stack"

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Using the HTTP transport adapter
```
JeffreyRichter, line 26: Fix case: Transport Adapter

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+The HTTP transport adapter is set up during service SDK client initialization, for example, when creating an Azure Storage SDK client. The HTTP transport adapter can be specified at client initialization via the client options argument. See the following code snippet as an example:
```
JeffreyRichter, line 28: This example below show how to use the default HTTP transport adapter when constructing a client:

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+```cpp
+  /*
+  * Using the HTTP transport adapter.
+  */
+
+  // When no options are provided, the HTTP transport adapter is set to the default one based on the compiler options passed in to CMake.
+  auto defaultStorageClient = BlobServiceClient(url, credential);
+
+  // Use client options to set an HTTP transport adapter, which needs to be a shared_ptr.
```
JeffreyRichter, line 37: Break the example into two and separate with this text:
This example below shows how to override the default HTTP transport adapter when constructing a client:

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = std::make_shared<Azure::Core::Http::CurlTransport>();
```
JeffreyRichter, line 39: I think you should add a comment above this line indicating: 
// This shows using the CurlTrasnport but your custom transport can be specified here

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+  auto storageClient = BlobServiceClient(url, credential, options);
+```
+
+### Re-use the HTTP transport adapter
```
JeffreyRichter, line 43: Transport Adapter

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+  auto defaultStorageClient = BlobServiceClient(url, credential);
+
+  // Use client options to set an HTTP transport adapter, which needs to be a shared_ptr.
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = std::make_shared<Azure::Core::Http::CurlTransport>();
+  auto storageClient = BlobServiceClient(url, credential, options);
+```
+
+### Re-use the HTTP transport adapter 
+Note that the HTTP transport adapter is a `shared_ptr`. This is because you can re-use the same HTTP transport adapter for multiple clients. There are two ways of doing this. The first one is by using the same HTTP transport adapter when creating another client. The second one is when you get a specific child client from an already created parent client directly, in which case the child client inherits the same configuration. See the next example:
```
JeffreyRichter, line 45: Change "configuration" to "options"

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+  */
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>();
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = curlTransportAdapter;
+  auto storageClient = BlobServiceClient(url, credential, options);
+  // Create specific child client from the parent service client.
+  auto blobContainerClient = storageClient.GetBlobContainerClient(containerName)
+```
+
+Note that the second option may not supported for some Azure SDK clients if it is not applicable depending on the service behavior (for example, the Key Vault client).
```
JeffreyRichter, line 71: Just delete this sentence as I think it's obvious if you can't do it, you can't do it.

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+  curlTransportOptions.Proxy = "https://my.proxy.com";
+  // Create HTTP transport adapter with options.
+  auto curlTransportAdapter = std::make_shared<Azure::Core::Http::CurlTransport>(curlTransportOptions);
+
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = curlTransportAdapter; 
+  auto storageClient = BlobServiceClient(url, credential, options);
+```
+
+The HTTP transport adapter options are specific for each adapter and are optional. This means that if you are creating your own HTTP transport adapter, you don't need to create options for it if you don't want to.
```
JeffreyRichter, line 94: Delete this sentence

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+## Building a custom HTTP Transport Adapter
```
JeffreyRichter, line 97: Custom

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required to be set when building your own HTTP transport adapter.
+
+Follow next steps to implement your own HTTP Transport Adapter.
```
JeffreyRichter, line 103: Follow these steps to implement your own HTTP transport adapter:

doc/HttpTransportAdapter.md
vhvb1989: Curious. Why ?

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+class CustomTransportAdapter : public Azure::Core::Http::HttpTransport {
+  public:
+    // Override the Send method
+    std::unique_ptr<Azure::Core::Http::RawResponse> Send(
+        Azure::Core::Context const& context,
+        Azure::Core::Http::Request& request) override;
+};
+```
+
+> The implementation needs to take care of using the members from the `Request` argument which contains all the HTTP request information to write the HTTP request using some C++ HTTP stack. Then it needs to create the `RawResponse` from the response and return it. The `context` argument can be optionally used to check if the request is cancelled before performing I/O operations (like writing to the network).
```
JeffreyRichter, line 126: It's optional because the Transport policy does the check, right?
If so, then it useful only to support cancellation during a synchronous I/O request if the HTTP stack supports this.

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+> If instead of having a session created for every call to Send(), we would place the state directly into the HTTP transport adapter class, any new call to Send() would corrupt the current state when sharing the same HTTP transport adapter from multiple clients.
+
+3. Use the adapter.
+
+At this point you have all you need to use your custom HTTP transport adapter. Refer to [Using the HTTP transport adapter](#using-the-http-transport-adapter) for how to set it upon creating the SDK client.
+
+4. Set as default (optional).
+
+You can optionally set your custom HTTP transport adapter as the default case. This can be helpful if you want to always use this HTTP transport adapter and you would like to avoid setting up the SDK client options every time you create a new SDK client.
```
JeffreyRichter, line 181: You can optionally set your custom HTTP transport adapter as the default to avoid explicitly setting the option every time you create a new client.

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,192 @@
...
+
+```cpp
+/*
+* Setting an HTTP transport adapter as default.
+*/
+
+// From one of the cpp files in your application:
+std::shared_ptr<Azure::Core::Http::HttpTransport> ::AzureSdkGetCustomHttpTransport() {
+  // Create and return a shared ptr for your custom HTTP transport adapter.
+  return std::make_shared<CustomHttpTransportAdapterSession>();
```
JeffreyRichter, line 195: If AzureSdkGetCustomHttpTransport is called multiple times, it keeps returning the pointer to a single CustomHttpTransportAdapter, right?
We don't want to create a new adapter with each call.

Also, CustomHttpTransportAdapterSession should be called CustomHttpTransportAdapter, right?

JeffreyRichter (approved):
Victor, fantastic job on this!
Very clear, well organized, and complete.
I see book writing in your future :)

vhvb1989:
> 
> 
> Victor, fantastic job on this!
> Very clear, well organized, and complete.
> I see book writing in your future :)

hahahaha,, I can take credits for the organization, but @ahsonkhan  and @RickWinter  have already improved the redaction for the time you saw it :).  Thank you

JeffreyRichter:
Well in that case, I'll give "a great job" to @RickWinter  & @ahsonkhan too!
The 3 of you can cowrite a book together then.

doc/HttpTransportAdapter.md
vhvb1989: We are currently creating a new adapter on every call.
The same happen with curl and winHttp.  When a new SDK client is created with default options, a new transport adapter is created.
The benefit of this is that the transport adapter is destroyed as soon as no SDK client is using it. Otherwise, we would be leaving in memory a global adapter to use for default calls.

We can review this. I will create a new issue to open the discussion on it. It is a good point

doc/HttpTransportAdapter.md
ahsonkhan: Markdown linter rule `MD022`:
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines

sdk/core/azure-core/README.md
```diff
@@ -18,6 +18,10 @@ The main shared concepts of Azure::Core include:
 - Exceptions for reporting errors from service requests in a consistent fashion (`RequestFailedException`)
 - Abstractions for Azure SDK Credentials (`TokenCredential`)
 
+### HTTP Transport adapter
+
+Out of the box, the Azure SDK for C++ supports the libcurl and WinHTTP libraries as HTTP stacks for communicating with Azure services over the network.  The SDK also provides a mechanism for `customer-implemented` *HTTP transport adapter*.
```
ahsonkhan, line 23:
```suggestion
Out of the box, the Azure SDK for C++ supports the libcurl and WinHTTP libraries as HTTP stacks for communicating with Azure services over the network. The SDK also provides a mechanism for `customer-implemented` *HTTP transport adapter*.
```

ahsonkhan (changes requested):
Temporarily blocking so I can go through a final read of the doc before merging.

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+- The libcurl Transport Adapter
+
+<center>
+
+| Supported Platforms  |
+| --- |
+| Linux |
+| Windows |
+| MacOS |
+| [Others](https://daniel.haxx.se/blog/wp-content/uploads/2020/02/slide-operating-systems.jpg) |
```
ahsonkhan, line 17: Given the C++ SDK doesn't officially support most of these, I don't think it is worth mentioning this link, even if libcurl itself does work on them.
```suggestion
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
+# HTTP Transport Adapter
+
+The Azure SDK for C++ creates and sends HTTP requests using an HTTP Transport Adapter. This adapter provides a general abstraction for an HTTP stack (like libcurl, WinHTTP, etc) by defining the interface it needs to call that anyone can implement to customize which HTTP stack gets used.
```
ahsonkhan, line 3: nit:
```suggestion
The Azure SDK for C++ creates and sends HTTP requests using an HTTP Transport Adapter. This adapter provides a general abstraction for an HTTP stack (like libcurl, WinHTTP, etc.) by defining the interface it needs to call that anyone can implement to customize which HTTP stack gets used.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Default HTTP Transport Adapter
+
+The Azure SDK for C++ provides two HTTP transport adapters as part of the azure-core package:
+- The libcurl Transport Adapter
+
+<center>
```
ahsonkhan, line 10: Is this html-like tag necessary? It's fine to keep if it is important, but I am curious what happens if you don't add that?

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+| Supported Platforms  |
+| --- |
+| Windows |
+
+</center>
+
+Azure SDK for C++ users can create and use their own transport adapter [by following the guidance below](#building-a-custom-http-transport-adapter).
```
ahsonkhan, line 31:
```suggestion
Users can create and use their own transport adapter [by following the guidance below](#building-a-custom-http-transport-adapter).
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+For the simplest case, when no specific configuration is set, the default transport adapter will be selected based on the OS. The WinHTTP transport adapter will be used for Windows-based systems (making libcurl an optional dependency). The libcurl transport adapter will be selected for any non-Windows-based system.
+
+## Building HTTP Transport Adapter
```
ahsonkhan, line 35:
```suggestion
## Building the HTTP Transport Adapter
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+The Azure SDK for C++ uses CMake options to define what HTTP transport adapters to build.
+
+You can see [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to learn about all the supported options and what is the specific option required for each HTTP transport adapter.
```
ahsonkhan, line 39: Note to self, we need to add `BUILD_TRANSPORT_WINHTTP` to the CMake options table.

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+Multiple HTTP transport adapters can be built as part of the project. This is to support scenarios where you want to send HTTP request with the libcurl transport adapter for some cases and the WinHTTP transport adapter for others.
+
+Another example is if you want to create your own HTTP transport adapter. In this case, you can build your own HTTP transport adapter and test it by comparing it to the behavior of the libcurl HTTP transport adapter.
+
+## Using the HTTP Transport Adapter
+
+The HTTP transport adapter is set up during service SDK client initialization, for example, when creating an Azure Storage SDK client. The HTTP transport adapter can be specified at client initialization via the client options argument. The next example below shows how to use the default HTTP transport adapter when constructing an SDK client:
```
ahsonkhan, line 47:
```suggestion
The HTTP transport adapter is set up during service client initialization, for example, when creating an Azure Storage client. The HTTP transport adapter can be specified at client initialization via the client options argument. The example below shows how to use the default HTTP transport adapter when constructing an SDK client:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+```cpp
+  /*
+  * Using the default HTTP transport adapter.
+  */
+
+  // When no options are provided, the HTTP transport adapter is set to the default one based on the compiler options passed in to CMake.
```
ahsonkhan, line 53:
```suggestion
  // When no options are provided, the HTTP transport adapter is set to the default one
  // based on the compiler options passed in to CMake.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+  auto defaultStorageClient = BlobServiceClient(url, credential);
+```
+
+The next example below shows how to override the default HTTP transport adapter when constructing an SDK client:
```
ahsonkhan, line 57:
```suggestion
The example below shows how to override the default HTTP transport adapter when constructing an SDK client:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+  /*
+  * Override the default HTTP transport adapter.
+  */
+
+  // Use client options to set an HTTP transport adapter, which needs to be a shared_ptr.
+  BlobClientOptions options;
+  options.TransportPolicyOptions.Transport = 
+  // Setting the libcurl transport adapter as an example
+  // Any HTTP transport adapter can be specified here.
+  std::make_shared<Azure::Core::Http::CurlTransport>();
```
ahsonkhan, line 69: Move comment above the line rather than in the middle of assignment.

```suggestion
  BlobClientOptions options;
    // Setting the libcurl transport adapter as an example.
  // Any HTTP transport adapter can be specified here.
  options.TransportPolicyOptions.Transport = std::make_shared<Azure::Core::Http::CurlTransport>();
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+  auto storageClient = BlobServiceClient(url, credential, options);
+  // Create specific child client from the parent service client.
+  auto blobContainerClient = storageClient.GetBlobContainerClient(containerName)
+```
+
+Each client holds an internal copy of the options passed in the arguments. This means that the options object used to create the client can be disposed. However, since the HTTP transport adapter is a shared pointer, it won't be disposed until all clients using it are disposed first.
+
+## HTTP Transport Adapter Options
+
+The libcurl and WinHTTP transport adapters can also be initialized with specific options that expose control of features from the HTTP stack underneath it. For example, you can set a proxy or override the default certificate authority (CA) path for libcurl. See the following example:
```
ahsonkhan, line 106:
```suggestion
The libcurl and WinHTTP transport adapters can also be initialized with specific options that expose control of features from the HTTP stack underneath them. For example, you can set a proxy or override the default certificate authority (CA) path for libcurl. See the following example:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+    // Expose any functionality required. For example, this could
+    // be the main method to produce the HTTP raw response.
+    std::unique_ptr<Azure::Core::Http::RawResponse> Perform(
+      Azure::Core::Context const& context,
+      Azure::Core::Http::Request& request
+    );
+}
+```
+
+Then, within the Send() method implementation.
```
ahsonkhan, line 180:
```suggestion
Then, within the `Send()` method implementation:
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+          // This will ensure that multiple clients can
+          // call Send() at the same time.
+          CustomHttpTransportAdapterSession session;
+
+          // Use any functionally from your created session.
+          // For this example, we could just return like this:
+          return session.Perform();
+        }
+
```
ahsonkhan, line 196: Fix the indentation of this sample by removing the extra spacing in the beginning of each line.

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+> If instead of having a session created for every call to Send(), we would place the state directly into the HTTP transport adapter class, any new call to Send() would corrupt the current state when sharing the same HTTP transport adapter from multiple clients.
```
ahsonkhan, line 199: I believe that isn't necessarily true, especially with WinHTTP, once we address this: https://github.com/Azure/azure-sdk-for-cpp/issues/1098

Maybe remove this for now?

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+3. Use the adapter.
+
+At this point you have all you need to use your custom HTTP transport adapter. Refer to [Using the HTTP transport adapter](#using-the-http-transport-adapter) for how to set it upon creating the SDK client.
+
+4. Set as default (optional).
+
+You can optionally set your custom HTTP transport adapter as the default to avoid explicitly setting the option every time you create a new SDK client.
+
+The first step is to set the required CMake compile option that would configure the build for using a custom HTTP transport adapter. Refer to the [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to find out the required option fot this.
```
ahsonkhan, line 209:
```suggestion
The first step is to set the required CMake compile option that would configure the build for using a custom HTTP transport adapter. Refer to the [CMake options](https://github.com/Azure/azure-sdk-for-cpp/blob/master/CONTRIBUTING.md#cmake-build-options) to find out the required option for this.
```

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,223 @@
...
+
+The second step is to implement the method showed below in the global unnamed namespace.
```
ahsonkhan, line 211:
```suggestion
The second step is to implement the method shown below in the global unnamed namespace.
```

ahsonkhan (approved):
Some typos and fix ups. Looks good otherwise.

doc/HttpTransportAdapter.md
vhvb1989: This makes the table to be centered like:
image

instead of:
image

doc/HttpTransportAdapter.md
vhvb1989: #1098 is a different issue.

This note if to mention what happen if you don't design the adapter in a multi-thread shared way, where it would be corrupted from a second call

doc/HttpTransportAdapter.md
ahsonkhan: That's not what I see here:
https://github.com/Azure/azure-sdk-for-cpp/blob/8238c6de7ab413461d79134f4baed3d4380c2654/doc/HttpTransportAdapter.md#default-http-transport-adapter
image

doc/HttpTransportAdapter.md
ahsonkhan: I think overloading the term "session" here is what's causing the confusion. In WinHTTP, I may not create a **session** (i.e. session handle) for every call to `Send()`.The way it is written, it makes it seem like if I don't create a new session on each call to Send(), I am corrupting some state. How is that always true?

doc/HttpTransportAdapter.md
```diff
@@ -0,0 +1,222 @@
...
+The example below shows how to override the default HTTP transport adapter when constructing an SDK client:
+
+```cpp
+  /*
+  * Override the default HTTP transport adapter.
+  */
+
+  // Use client options to set an HTTP transport adapter, which needs to be a shared_ptr.
+  BlobClientOptions options;
+    // Setting the libcurl transport adapter as an example.
```
ahsonkhan, line 66:
```suggestion
  // Setting the libcurl transport adapter as an example.
```

doc/HttpTransportAdapter.md
vhvb1989: oh that's a shame!  It looks like github doesn't support it.
Still think it's nice to have it for any customer opening the md file in some better reader which supports the tag

doc/HttpTransportAdapter.md
vhvb1989: #1300 assigned to you

doc/HttpTransportAdapter.md
vhvb1989: Correct. Session happens to be the same name as what WinHTTP main component uses.

For WinHTTP, the equivalent would be the `handleManager`. You create a new handleManager  per each Send() call and init the handlers inside it.
If you would have the handlers directly into the transport adapter class, a second call to Send() would interfere collapse if there is currently one Send() happening because it would try to use the same handlers.

Re-using a connection handler is possible, yes. That's what #1098 is all about. However, the connection can't be re-used while it is on use. The same handler can be use again only when the Send() method is finished and there's a new Send() call that happen to have the same configuration used before from a previous call to Send().

doc/HttpTransportAdapter.md
ahsonkhan: The markdown linter doesn't like the use of raw HTML tags for the opposite reason. Some markdown readers don't work if you use non-pure elements:
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md033---inline-html

Issue #1301, opened 2021-01-10
Consider removing Base64EncodeText internal method and call the public Base64Encode method directly


ahsonkhan:
Remove this method and update callers to use the public `vector<uint8_t>` `Base64Encode` method:
https://github.com/Azure/azure-sdk-for-cpp/blob/fb33402844be8aaeedcbcec645b8be9c69fa6242/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp#L14-L19

Callers should use:
https://github.com/Azure/azure-sdk-for-cpp/blob/fb33402844be8aaeedcbcec645b8be9c69fa6242/sdk/core/azure-core/inc/azure/core/base64.hpp#L17-L23

For example, change this:
https://github.com/Azure/azure-sdk-for-cpp/blob/fb33402844be8aaeedcbcec645b8be9c69fa6242/sdk/storage/azure-storage-files-datalake/src/datalake_utilities.cpp#L42

In some cases, we may want to change the code to avoid creating `std::string` when the variable is binary data and can be saved as `vector<uint8_t>`. Base64 encoding is binary data to text, so passing in an `std::string` is an odd implementation choice.

For example, does this `Storage::Metadata` map represent a string key with a binary data value? If so, it should be changed to `std::string, vector<uint8_t>`:
https://github.com/Azure/azure-sdk-for-cpp/blob/b212d287abafcc4e9cf06d78ac57dd87286227c7/sdk/storage/azure-storage-common/inc/azure/storage/common/storage_common.hpp#L68

For more context, see https://github.com/Azure/azure-sdk-for-cpp/pull/1298#discussion_r554285035

cc @Jinming-Hu, @katmsft

PR #1307, opened 2021-01-11
remove Base64EncodeText

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

closes https://github.com/Azure/azure-sdk-for-cpp/issues/1301

### DIFF
--- a/sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
@@ -139,7 +139,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       constexpr std::size_t BlockIdLength = 64;
       std::string blockId = std::to_string(id);
       blockId = std::string(BlockIdLength - blockId.length(), '0') + blockId;
-      return Internal::Base64EncodeText(blockId);
+      return Azure::Core::Base64Encode(std::vector<uint8_t>(blockId.begin(), blockId.end()));
     };
 
     auto uploadBlockFunc = [&](int64_t offset, int64_t length, int64_t chunkId, int64_t numChunks) {
@@ -220,7 +220,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       constexpr std::size_t BlockIdLength = 64;
       std::string blockId = std::to_string(id);
       blockId = std::string(BlockIdLength - blockId.length(), '0') + blockId;
-      return Internal::Base64EncodeText(blockId);
+      return Azure::Core::Base64Encode(std::vector<uint8_t>(blockId.begin(), blockId.end()));
     };
 
     auto uploadBlockFunc = [&](int64_t offset, int64_t length, int64_t chunkId, int64_t numChunks) {
--- a/sdk/storage/azure-storage-files-datalake/src/datalake_utilities.cpp
+++ b/sdk/storage/azure-storage-files-datalake/src/datalake_utilities.cpp
@@ -39,7 +39,10 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
     std::string result;
     for (const auto& pair : dataLakePropertiesMap)
     {
-      result.append(pair.first + "=" + Internal::Base64EncodeText(pair.second) + ",");
+      result.append(
+          pair.first + "="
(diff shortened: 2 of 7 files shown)


katmsft (approved):

sdk/storage/azure-storage-files-datalake/src/datalake_utilities.cpp
```diff
@@ -39,7 +39,10 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake { nam
     std::string result;
     for (const auto& pair : dataLakePropertiesMap)
     {
-      result.append(pair.first + "=" + Internal::Base64EncodeText(pair.second) + ",");
+      result.append(
+          pair.first + "="
+          + Azure::Core::Base64Encode(std::vector<uint8_t>(pair.second.begin(), pair.second.end()))
```
ahsonkhan, line 44: Does `pair.second` need to be a `std::string`?

From the issue:
> For example, does this `Storage::Metadata` map represent a string key with a binary data value? If so, it should be changed to `std::string, vector<uint8_t>`:
https://github.com/Azure/azure-sdk-for-cpp/blob/b212d287abafcc4e9cf06d78ac57dd87286227c7/sdk/storage/azure-storage-common/inc/azure/storage/common/storage_common.hpp#L68

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -139,7 +139,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       constexpr std::size_t BlockIdLength = 64;
       std::string blockId = std::to_string(id);
       blockId = std::string(BlockIdLength - blockId.length(), '0') + blockId;
-      return Internal::Base64EncodeText(blockId);
+      return Azure::Core::Base64Encode(std::vector<uint8_t>(blockId.begin(), blockId.end()));
```
ahsonkhan, line 142: We should try to re-write this in terms of `vector<uint8_t>` and not have `blockId` be `std::string`. That way, you don't need to turn the string to vector of bytes when calling `Base64Encode`.

Jinming-Hu: Nah, `std::string` supports all kinds of operations like concatenation, to int, from int.
Ho can we convert `std::vectort<uint8_t>` between int in an easy way?

sdk/storage/azure-storage-files-datalake/src/datalake_utilities.cpp
Jinming-Hu: Yes, in most of the cases our users will be using `std::string` for metadata. Plus it aligns with all other SDK

ahsonkhan: > Yes, in most of the cases our users will be using `std::string` for metadata. Plus it aligns with all other SDK

I am not familiar with this concept too much, can you share how it aligns with other SDK?

Is the `Metadata` here in our SDK identical to the concept in C#? I ask because I don't see them encoding string to Base64 string. So either their dictionary already contains the base64 encoded string, in which case we may want to do something similar, or change the map to `vector<uint8_t>`

https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-properties-metadata?tabs=dotnet#set-and-retrieve-metadata
https://raw.githubusercontent.com/Azure/azure-sdk-for-net/master/sdk/storage/Azure.Storage.Blobs/src/Generated/BlobRestClient.cs
```C#
if (metadata != null) {
    foreach (System.Collections.Generic.KeyValuePair<string, string> _pair in metadata)
    {
        _request.Headers.SetValue("x-ms-meta-" + _pair.Key, _pair.Value);
    }
}
```

Jinming-Hu: There're two ways to send metadata over the wire. One used by blob service is just like in the .Net code you shared. The other one used by DataLake is putting all metadata in a single header. See https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/filesystem/setproperties#request-headers

Issue #1127, opened 2020-12-10
Change return type for DeleteIfExists and CreateIfNotExists


Jinming-Hu:
```c  
Azure::Storage::Blobs::BlobClient.DeleteIfExists:const_Azure::Storage::Blobs::DeleteBlobOptions_&::Azure::Core::Response<Models::DeleteBlobResult>
```
#
Response&lt;DeleteBlobIfExistResult{bool, DeleteBlobResult}&gt; or DeleteBlobResult having a bool suggesting if the blob is deleted.

PR #1313, opened 2021-01-11
add a bool indicating whether the oepration succeeds for CreateIfNotEx…

Jinming-Hu:
…ists/DeleteIfExists

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

closes https://github.com/Azure/azure-sdk-for-cpp/issues/1127

### DIFF
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
@@ -119,8 +119,8 @@ namespace Azure { namespace Storage { namespace Blobs {
      * exists.
      *
      * @param options Optional parameters to execute this function.
-     * @return A CreateAppendBlobResult describing the newly created append blob. Null if the blob
-     * already exists.
+     * @return A CreateAppendBlobResult describing the newly created append blob.
+     * CreateAppendBlobResult.Created is false if the blob already exists.
      */
     Azure::Core::Response<Models::CreateAppendBlobResult> CreateIfNotExists(
         const CreateAppendBlobOptions& options = CreateAppendBlobOptions()) const;
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp
@@ -297,6 +297,7 @@ namespace Azure { namespace Storage { namespace Blobs {
 
     struct CreateAppendBlobResult
     {
+      bool Created = true;
       std::string ETag;
       Azure::Core::DateTime LastModified;
       Azure::Core::Nullable<std::string> VersionId;
@@ -307,6 +308,7 @@ namespace Azure { namespace Storage { namespace Blobs {
 
     struct CreateBlobContainerResult
     {
+      bool Created = true;
       std::string ETag;
       Azure::Core::DateTime LastModified;
     }; // struct CreateBlobContainerResult
@@ -324,6 +326,7 @@ namespace Azure { namespace Storage { namespace Blobs {
 
     struct CreatePageBlobResult
     {
+      bool Created = true;
       std::string ETag;
       Azure::Core::DateTime LastModified;
       Azure::Core::Nullable<std::string> VersionId;
@@ -335,10 +338,12 @@ namespace Azure { namespace Storage { namespace Blobs {
 
     struct DeleteBlobContainerResult
     {
+      bool Deleted = true;
     }; // struct DeleteBlobContainerResult
 
     struct DeleteBlobResult
     {
+      bool Deleted = true;
     }; // struct DeleteBlobResult
 
     class DeleteSnapshotsOption {
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
@@ -278,7 +278,8 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @brief Marks the specified blob or snapshot for deletion if it exists.
      *
      * @param options Optional parameters to execute this function.
-     * @return A DeleteBlobResult on successfully deleting. Null if the blob doesn't exist.
+     * @return A DeleteBlobResult on successfully deleting. DeleteBlobResult.Deleted is false if the
+     * blob doesn't exist.
      */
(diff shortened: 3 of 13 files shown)


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
```diff
@@ -119,8 +119,8 @@ namespace Azure { namespace Storage { namespace Blobs {
      * exists.
      *
      * @param options Optional parameters to execute this function.
-     * @return A CreateAppendBlobResult describing the newly created append blob. Null if the blob
-     * already exists.
+     * @return A CreateAppendBlobResult describing the newly created append blob.
+     * CreateAppendBlobResult.Created is false if the blob already exists.
```
vhvb1989, line 123: nit.
```suggestion
     * The field `CreateAppendBlobResult.Created` is false if the blob already exists.
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp
```diff
@@ -297,6 +297,7 @@ namespace Azure { namespace Storage { namespace Blobs {
 
     struct CreateAppendBlobResult
     {
+      bool Created = true;
```
vhvb1989, line 300: Isn't it better to expect `false` as the worst case (error case)?   
By using `true` we need to be more careful with this value during any exception

vhvb1989 (approved):
If `Created` is false, how do I know why? 

As an end user, do I need to know what fields I can access when `Created` is true vs what I can access when it is false?   Or do I get access to all fields and errors if something I access is empty for the current `Created` value?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp
Jinming-Hu: I remember it was suggested to be `true` during arch board review meeting.

We set this value `false` on exception. So it's always correctly set, we don't have to worry about it.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
Jinming-Hu: Does doxygen generated stuff support markdown syntax?

Jinming-Hu:
> If `Created` is false, how do I know why?

It's always because the container already exists, which has been explained in doc. Exceptions will be thrown if it fails due to other reasons.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
vhvb1989: Some of it.
At least the " ` " works.  Take look to some sample like context::ValueBase :
https://azuresdkdocs.blob.core.windows.net/$web/cpp/azure-core/1.0.0-beta.3/struct_azure_1_1_core_1_1_value_base.html

Something like:
image

Resolves to:
image

Jinming-Hu: @vhvb1989 Good to know that. I think I'm going to update all docs.

Jinming-Hu: @vhvb1989 BTW, do you know how to reference another class, or member of another class? Reference means it's a link in the generated doc.

vhvb1989: There are a few other nice things we can do to improve the doc's looking.  I know this other one with the " \p " for args like

this:
image

Resolves to this:
image

The best guy for this is @ahsonkhan  :)

vhvb1989: >  do you know how to reference another class

I have seen others to use `#` for that. But I think that only works if the reference is globally known. Otherwise, maybe we need to add the full link like `#Azur::Core::Context` and not only `#Context`.  
I am not sure. I will wait to see if @ahsonkhan knows it already, otherwise, I can learn it some day and share hahah..  We have some broken references like:

image

See how it just displays the " # "  .. The expectation would be a link.   We might even have an issue for this somewhere in the backlog :)

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/1327

vhvb1989: Just tested with full path and worked:

This:
image

Generates:
image

vhvb1989: Btw, you can generate the documentation locally by doing:

```cpp
cmake -DBUILD_DOCUMENTATION=ON ..
cmake --build .
```

And in case you are running from a VM with linux using ssh, you can start an HTTP server from `azure-sdk-for-cpp/build/sdk` and browse to `localhost:8000` for instance, to see how each package's docs looks like

katmsft (approved):

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/append_blob_client.hpp
ahsonkhan: > Just tested with full path and worked:

Great! That's what I expected :)

ahsonkhan: > Btw, you can generate the documentation locally by doing:

We should add that to the README or contributing guide (currently, we only mention it in the cmake options table).

ahsonkhan, line 123: This looks like a behavioral change, worth adding a change log entry, maybe under "other changes"?

Jinming-Hu: Yes, it is a behavioral change. But we just added this API in the same version, without shipping a release.

PR #1320, opened 2021-01-11
DateTime: better integration with chrono, support year 0001+

### LINKED ISSUE #1132: Add TimeSpan in Core SDK
```c  
Azure::Storage::Blobs::BreakBlobContainerLeaseOptions.BreakPeriod
```
#
Option 1: Add duration type.
Option 2: Add `InSeconds` suffix.

### LINKED ISSUE #1174: DateTime: Change epoch to be year 1, allow printing and parsing from year 1 to 1601, in addition to 1601 to 9999

antkmsft:
Closes #1174.
Closes #1132.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/datetime.hpp
+++ b/sdk/core/azure-core/inc/azure/core/datetime.hpp
@@ -104,7 +104,8 @@ namespace Azure { namespace Core {
      * @param systemTime A value of `std::chrono::system_clock::time_point`.
      */
     DateTime(std::chrono::system_clock::time_point const& systemTime)
-        : DateTime(SystemClockEpoch + systemTime.time_since_epoch())
+        : DateTime(
+            SystemClockEpoch + std::chrono::duration_cast<duration>(systemTime.time_since_epoch()))
     {
     }
 
--- a/sdk/core/azure-core/inc/azure/core/datetime.hpp
+++ b/sdk/core/azure-core/inc/azure/core/datetime.hpp
@@ -12,16 +12,90 @@
 #include <string>
 
 namespace Azure { namespace Core {
+  namespace Details {
+    class Clock {
+    public:
+      using rep = int64_t;
+      using period = std::ratio<1, 10000000>;
+      using duration = std::chrono::duration<rep, period>;
+      using time_point = std::chrono::time_point<Clock>;
+      static constexpr bool is_steady = true;
+
+      static time_point now() noexcept;
+    };
+  } // namespace Details
+
   /**
    * @brief Manages date and time in standardized string formats.
+   * FIXME: update, put more details
+   * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    */
-  class DateTime {
+  class DateTime : public Details::Clock::time_point {
+    static DateTime const SystemClockEpoch;
+
   public:
     /**
-     * @brief Units of measurement the difference between instances of @DateTime.
+     * @brief Construct a default instance of @DateTime (00:00:00.0000000 on Janualy 1st, 0001).
      */
-    // 1 == 100 ns (1 / 10,000,000 of a second, 7 fractional digits).
-    typedef std::chrono::duration<int64_t, std::ratio<1, 10000000>> Duration;
+    constexpr DateTime() : time_point() {}
+
+  private:
+    DateTime(
+        int16_t year,
+        int8_t month,
+        int8_t day,
+        int8_t hour,
+        int8_t minute,
+        int8_t second,
+        int32_t fracSec,
+        int8_t dayOfWeek,
+        int8_t localDiffHours,
+        int8_t localDiffMinutes,
+        bool roundFracSecUp = false);
+
+  public:
+    /**
+     * @brief Construct an instance of @DateTime.
+     *
+     * @param year Year.
+     * @param month Month.
+     * @param day Day.
+     * @param hour Hour.
+     * @param minute Minute.
+     * @param second Seconds.
+     *
+     * @throw std::invalid_argument If any parameter is invalid.
+     */
+    explicit DateTime(
+        int16_t year,
+        int8_t month = 1,
+        int8_t day = 1,
+        int8_t hour = 0,
+        int8_t minute = 0,
+        int8_t second = 0)
+        : DateTime(year, month, day, hour, minute, second, 0, -1, 0, 0)
+    {
+    }
+
+    /**
+     * @brief Construct an instance of @DateTime from base class.
+     */
+    constexpr DateTime(time_point const& timePoint) : time_point(timePoint) {}
+
+    /**
+     * @brief Construct an instance of @DateTime from `std::chrono::system_clock::time_point`.
+     * @param systemTime A value of `std::chrono::system_clock::time_point`.
+     */
+    DateTime(std::chrono::system_clock::time_point const& systemTime)
+        : DateTime(SystemClockEpoch + systemTime.time_since_epoch())
+    {
+    }
+
+    /**
+     * @brief Convert an instance of @DateTime to `std::chrono::system_clock::time_point`.
+     * @throw std::invalid_argument if @DateTime is outside of the range that can be represented.
+     */
+    explicit operator std::chrono::system_clock::time_point() const;
 
     /**
      * @brief Defines the format applied to the fraction part of any @DateTime.
@@ -50,31 +124,6 @@ namespace Azure { namespace Core {
       /// RFC 3339.
       Rfc3339,
     };
-
-    /**
-     * @brief Get the current UTC time.
-     */
-    static DateTime Now();
-
-    /**
-     * @brief Construct an instance of @DateTime.
-     *
-     * @param year Year.
-     * @param month Month.
-     * @param day Day.
-     * @param hour Hour.
-     * @param minute Minute.
-     * @param second Seconds.
-     *
-     * @throw std::invalid_argument If any parameter is invalid.
-     */
-    explicit DateTime(
-        int16_t year,
-        int8_t month = 1,
-        int8_t day = 1,
-        int8_t hour = 0,
-        int8_t minute = 0,
-        int8_t second = 0);
 
     /**
      * @brief Create @DateTime from a string representing time in UTC in the specified format.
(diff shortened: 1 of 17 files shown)


JeffreyRichter (approved):

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
+    /**
+     * @brief Construct an instance of @DateTime from base class.
+     */
+    constexpr DateTime(time_point const& timePoint) : time_point(timePoint) {}
+
+    /**
+     * @brief Construct an instance of @DateTime from `std::chrono::system_clock::time_point`.
+     * @param systemTime A value of `std::chrono::system_clock::time_point`.
+     */
+    DateTime(std::chrono::system_clock::time_point const& systemTime)
```
Jinming-Hu, line 106: explicit

Jinming-Hu:
@antkmsft Because you made changes to storage code in this PR, we'd appreciate if you can wait for storage team's approval before merging it next time.

antkmsft:
@Jinming-Hu , sorry, I'll wait next time. It is basically a syntactic update - DateTime no longer has Now() method.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: We discussed it with Jeffrey and decided to make it implicit. Reasons:
1. `DateTime dt = std::chrono::system_clock::now()` - convenient for customers
2. No harm in converting from system_clock - the system_clock's datetime is, realistically, always representable in terms of date range (between year 0001 and 9999; on POSIX system_clock has 1ns precision, while DateTime's is 100ns, but benefits (item 1) seem to outweigh costs.

Jinming-Hu: Sure. I brought this up only because we ever got a conclusion in guidelines meeting that constructors should be explicit unless specifically permitted.

antkmsft: Thank you, good point, let me to bring this up during our next meeting.
I believe I have justification, since this whole effort with deriving from std::chrono::time_point<Clock> is so that our DateTime mimics chrono as close as possible, so this class stands in the intersection of standard library and Azure SDK integrating them and providing syntactic convenience for our users. Let's see what people say, I hope they agree, but it will be good to get formal approval and confirmation from broader audience, thank you for bringing this up.

Jinming-Hu: I also agree it's better to be implicit than explicit for this one, and I see Jeff signed off. So I think it's okay.

antkmsft:
@Jinming-Hu, just checking, did you relied on default-constructed `DateTime` being `1601-01-01`? It is `0001-01-01` now.
Formally, for the previous version, we deliberately never documented what the default-constructed DateTime's date might be, however it is hard to keep that a secret. We do document now, BTW, we added in doxygen comment in this PR.

Jinming-Hu:
> @Jinming-Hu, just checking, did you relied on default-constructed `DateTime` being `1601-01-01`? It is `0001-01-01` now.
> Formally, for the previous version, we deliberately never documented what the default-constructed DateTime's date might be, however it is hard to keep that a secret. We do document now, BTW, we added in doxygen comment in this PR.

No, we don't reply on that.

Actually I tend to make a default-constructed `DateTime` to be undefined value, or uninitialized state. This aligns with `utility::datetime` impl in cpprest sdk.

Jinming-Hu:
Well I feel really bad after I saw the explicit conversion to `std:chrono::time_point`, when I was trying to figure out a bug in comparison operators. So question: when is the explicit conversion function is used?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
    * @brief Manages date and time in standardized string formats.
+   * @detail Supports date range from year 0001 to end of year 9999 with 100ns (7 decimal places for
+   * fractional second) precision.
+   * @remark `std::chrono::system_clock::time_point` can't be used, because there is no guarantees
+   * for the date range and precision.
+   * @remark This class is supposed to be able to handle a DateTime that comes over the wire.
    */
-  class DateTime {
+  class DateTime : public Details::Clock::time_point {
+    static DateTime const SystemClockEpoch;
```
ahsonkhan, line 51: Looks like this PR/change is causing vcpkg CI build to fail:
https://github.com/microsoft/vcpkg/pull/15635

@antkmsft If we can't fix it easily (and if it requires more investigation), should we revert this change fully and try a different approach?

sdk/core/azure-core/test/ut/datetime.cpp
```diff
@@ -501,12 +504,12 @@ TEST(DateTime, ParseTimeInvalid2)
       "1971-01-01T00:60:00Z", // minute too big
       "1971-01-01T00:00:70Z", // second too big
       "1971-01-01T00:00:61Z",
-      "1601-01-01T00:00:00+00:01", // time zone underflow
+      "0001-01-01T00:00:00+00:01", // time zone underflow
       // "1970-01-01T00:00:00.Z", // accepted as invalid timezone above
-      "1970-01-01T00:00:00+24:00", // bad tzoffsets
-      "1970-01-01T00:00:00-30:00",
+      //"1970-01-01T00:00:00+24:00", // bad tzoffsets
```
ahsonkhan, line 509: Why did we comment out these test cases?

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -4,869 +4,832 @@
...
-constexpr char const monthNames[] = "Jan\0Feb\0Mar\0Apr\0May\0Jun\0Jul\0Aug\0Sep\0Oct\0Nov\0Dec";
-
-} // namespace
+constexpr int8_t GetDayOfWeek(int16_t year, int8_t month, int8_t day)
+{
+  return DaySinceEpoch(year, month, day) % 7;
+}
 
-std::string DateTime::GetString(DateFormat format, TimeFractionFormat fractionFormat) const
+constexpr int8_t WeekDayMonthDayOfYear(int8_t* month, int8_t* day, int16_t year, int16_t dayOfYear)
```
ahsonkhan, line 115: nit:
```suggestion
constexpr int8_t WeekDayAndMonthDayOfYear(int8_t* month, int8_t* day, int16_t year, int16_t dayOfYear)
```

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -4,869 +4,832 @@
...
-  }
-
-  int64_t const epochAdjusted = m_since1601.count();
-  int64_t const secondsSince1601 = epochAdjusted / Duration::period::den; // convert to seconds
-  int const fracSec = static_cast<int>(epochAdjusted % Duration::period::den);
-
-  auto const yearData = ComputeYear(secondsSince1601);
-  int const year = yearData.Year + MinYear;
-  int const yearDay = yearData.SecondsLeftThisYear / SecondsInDay;
+    int8_t const MonthDays = MaxDaysPerMonth[i - 1] - ((i == 2 && !IsLeapYear(year)) ? 1 : 0);
```
ahsonkhan, line 120: This logic is difficult to parse and understand. Add a comment explaining what you are trying to do here.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -4,869 +4,832 @@
...
 
-namespace {
-constexpr bool StringStartsWith(char const* haystack, char const* needle)
+static std::string const DayNames[7] = {"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
+static std::string const MonthNames[12]
+    = {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
+
+constexpr int32_t DivideAndUpdateRemainder(int64_t* remainder, int64_t value)
+{
+  auto retval = static_cast<int32_t>(*remainder / value);
```
ahsonkhan, line 236: Why is this `static_cast` safe to do and not lossy?

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -4,869 +4,832 @@
...
   }
-
-  return result;
 }
 } // namespace
 
-/*
-https://tools.ietf.org/html/rfc822
-https://tools.ietf.org/html/rfc1123
+DateTime const DateTime::SystemClockEpoch = GetSystemClockEpoch();
```
ahsonkhan, line 364: I don't quite understand what the purpose of this is. What does `GetSystemClockEpoch` do and why is it needed?

ahsonkhan (commented):
Partial review. Will continue reviewing the rest of the implementation later.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
+      // and it may go backwards in time after adjustments, or jump forward faster than the actual
+      // time passes, if you catch the moment before and after syncing the clock.
+      // Steady clock would be good for measuring elapsed time without reboots (or hybernation?).
+      // Steady clock's epoch = boot time, and it would only go forward in steady fashion, after the
+      // system has started.
+      // Using this clock in combination with system_clock is common scenario.
+      // It would not be possible to base this clock on steady_clock and provide an implementation
+      // that universally works in any context in predictable manner. However, it does not mean that
+      // implementation can't use steady_clock in conjunction with this clock: an author can get a
+      // duration beteen two time_points of this clock (or between system_clock::time point ant this
```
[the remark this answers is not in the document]
antkmsft: I'll sneak this in to one of my other PRs.

sdk/core/azure-core/test/ut/datetime.cpp
antkmsft: Because it is valid offset, per spec, and on .NET

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
   /**
    * @brief Manages date and time in standardized string formats.
+   * @detail Supports date range from year 0001 to end of year 9999 with 100ns (7 decimal places for
+   * fractional second) precision.
+   * @remark `std::chrono::system_clock::time_point` can't be used, because there is no guarantees
+   * for the date range and precision.
+   * @remark This class is supposed to be able to handle a DateTime that comes over the wire.
    */
-  class DateTime {
+  class DateTime : public Details::Clock::time_point {
```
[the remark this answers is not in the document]
antkmsft: It is okay to inherit or typedef an internal type. It is good that the clock is not cluttering the public namespace - it is useless for the customer. Anything that is available via DateTime class is open to customer, including via inherited type, but it is not ok to go and consume the type via `::Details`. We talked about that. And I think you agreed
If you disagree, please explain, what benefit for the customer would be to have the clock as public class (not via DateTime::clock, but as Azure::Core::Clock)

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
+        int8_t month,
+        int8_t day,
+        int8_t hour,
+        int8_t minute,
+        int8_t second,
+        int32_t fracSec,
+        int8_t dayOfWeek,
+        int8_t localDiffHours,
+        int8_t localDiffMinutes,
+        bool roundFracSecUp = false);
```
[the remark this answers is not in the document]
antkmsft: It is private detali

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
+     * @throw std::invalid_argument If any parameter is invalid.
+     */
+    explicit DateTime(
+        int16_t year,
+        int8_t month = 1,
+        int8_t day = 1,
+        int8_t hour = 0,
+        int8_t minute = 0,
+        int8_t second = 0)
+        : DateTime(year, month, day, hour, minute, second, 0, -1, 0, 0)
```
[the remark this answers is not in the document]
antkmsft: it is private detail

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
+    {
+    }
+
+    /**
+     * @brief Construct an instance of @DateTime from base class.
+     */
+    constexpr DateTime(time_point const& timePoint) : time_point(timePoint) {}
```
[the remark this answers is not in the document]
antkmsft: It is `DateTime::time_point`, which is typedeffed to `std::chrono::time_point<Azure::Core::Details::Clock>`.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -12,16 +12,108 @@
...
+        : DateTime(
+            SystemClockEpoch + std::chrono::duration_cast<duration>(systemTime.time_since_epoch()))
+    {
+    }
+
+    /**
+     * @brief Convert an instance of @DateTime to `std::chrono::system_clock::time_point`.
+     * @throw std::invalid_argument if @DateTime is outside of the range that can be represented.
+     */
+    explicit operator std::chrono::system_clock::time_point() const;
```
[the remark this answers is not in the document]
antkmsft: year 0000 or 10000.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -124,114 +191,70 @@ namespace Azure { namespace Core {
...
-     * @brief Add \p duration to this @DateTime.
-     * @param duration @Duration to add.
-     * @return Reference to this @DateTime.
-     */
-    DateTime& operator+=(Duration const& duration)
-    {
-      m_since1601 += duration;
-      return *this;
-    }
+  inline Details::Clock::time_point Details::Clock::now() noexcept
```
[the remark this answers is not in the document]
antkmsft: it is not a declaration, but implementation. It does not place the declared name outside of the Details namespace.

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -4,869 +4,832 @@
...
+    T value,
+    decltype(value) minValue,
+    decltype(value) maxValue,
+    std::string const& valueName)
+{
+  auto outOfRange = 0;
 
-  struct timeval time = {};
-  if (gettimeofday(&time, nullptr) != 0)
+  if (value < minValue)
```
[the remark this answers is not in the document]
antkmsft: I don't think it is complicated, and I think it has value.

sdk/core/azure-core/src/datetime.cpp
antkmsft: Purely from usage / points of invocation we know that. This is a helper function that is carefully used within this class.
If there are problems, unit tests will find, by testing broader behavior.

sdk/core/azure-core/src/datetime.cpp
antkmsft: It is initialization for static class member.
System clock's zero could be 1970, or 1601. Standard does not specify. Which is why GetSystemClockEpoch() finds it out, and returns it in terms of Azure::Core::DateTime, where it is certainly not zero.
Static private class member SystemClockEpoch is needed so that conversion from `std::chrono::system_clock::time_point` works, and is inlineable:
```
DateTime(std::chrono::system_clock::time_point const& systemTime)
        : DateTime(
            SystemClockEpoch + std::chrono::duration_cast<duration>(systemTime.time_since_epoch()))
    {
    }
```

PR #1326, opened 2021-01-12
fix comparison bug introduced in #1320

Jinming-Hu:


vhvb1989 (approved):
Please add some tests whenever possible

Jinming-Hu:
Hi @antkmsft , I created this PR to fix the bug that's blocking us. Can you add test cases for these comparison operators?

antkmsft:
Thank you @Jinming-Hu  for the fix! I added unit tests - https://github.com/Azure/azure-sdk-for-cpp/pull/1332.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -240,21 +240,21 @@ namespace Azure { namespace Core {
...
   inline bool operator<(std::chrono::system_clock::time_point const& tp, DateTime const& dt)
   {
-    return !(dt >= tp);
+    return (dt > tp);
   }
 
   inline bool operator<=(std::chrono::system_clock::time_point const& tp, DateTime const& dt)
   {
-    return !(dt > tp);
+    return (dt >= tp);
```
ahsonkhan, line 248: The way this code is written is a bit confusing. We have implementation of operators like `<=` that are using the opposite operator in the implementation, i.e. `>=` and the subtle reason why that might work is because we swapped the arguments.

Rather than implementing 4 different ways of doing this (one for each operator), we should consider having one implementation of `>` and `<` and right `>=` based on top of `>`, i.e. `>=` is the negation of `<`.

Also, I am having a hard time (by just reading the code without an IDE) figuring out how these operator implementations aren't causing a cycle. For example `<` is implemented in terms of `>` and `>` is implemented in terms of `<`. 

Where is the `time_point` to `DateTime` comparison operator implemented?
Is it the ones above, i.e. `inline bool operator>(DateTime const& dt, std::chrono::system_clock::time_point const& tp)`

antkmsft: Operations do reuse one another - look at the lines 201-229 above.
The next set of operations, lines 231-259 should not be reusing in the usual sense, Jiming is right.
What is different about them: look at the type of the argument on the left and on the right (first and second parameter).

ahsonkhan: I suppose its fine, especially now that we have more unit tests. Operators like these have subtle implementation details (like order in which the argument appears around the operator) which increases the chances of such bugs. A comment here would help highlight that: `//Note: This is calling the other operator implementation which accepts `DateTime` first but with the arguments reversed.`

PR #1332, opened 2021-01-12
DateTime: Add unit tests for comparison operators

antkmsft:
Add unit tests for the https://github.com/Azure/azure-sdk-for-cpp/pull/1326 fix.

### DIFF
--- a/sdk/core/azure-core/test/ut/datetime.cpp
+++ b/sdk/core/azure-core/test/ut/datetime.cpp
@@ -5,6 +5,7 @@
 
 #include <azure/core/datetime.hpp>
 
+#include <chrono>
 #include <limits>
 
 using namespace Azure::Core;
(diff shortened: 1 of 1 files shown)


Jinming-Hu (approved):

sdk/core/azure-core/test/ut/datetime.cpp
```diff
@@ -594,3 +595,42 @@ TEST(DateTime, DefaultConstructible)
   DateTime dt;
   EXPECT_EQ(0, dt.time_since_epoch().count());
 }
+
+TEST(DateTime, ComparisonOperators)
```
ahsonkhan, line 599: Please verify that this test will fail if we didn't have this change:
https://github.com/Azure/azure-sdk-for-cpp/pull/1326

antkmsft: I did

Issue #1056, opened 2020-12-02
Storage API version enum


Jinming-Hu:
Originally brought up by @ahsonkhan in guidelines meeting.

Plans to expose an API version enum for storage services which change the API version sent over the wire (as a “best-effort”) to enable talking to different clouds and private azure stacks, similar to: https://github.com/Azure/azure-sdk-for-net/blob/1637f4b9ed998c40d9ae3fe054de3382b4769150/sdk/storage/Azure.Storage.Blobs/src/BlobClientOptions.cs#L22-L55

Jinming-Hu:
Hi @ahsonkhan , this was originally brought up by you. Storage team wants to know if this is a feature requested by a customer, so that we can properly prioritize it.

Storage team doesn't think of it as a blocker of GA. We can do it after GA and it's not a breaking change.

RickWinter:
@JeffreyRichter @tg-msft can you comment on if the archboard requires this language consistency feature for GA.

JeffreyRichter:
I would say NO as it is non-breaking to add it later and even if we had it today, the enum would have just 1 symbol associated with it so the customer has no real choice anyway.

tg-msft:
Having a good multi service-version story is important across languages.   Even though you're relying on users compiling in the source, I still think you'll hit diamond dependency problems and need to support multiple service-versions from the latest client-version of your library.

You only need to support the latest service-version when you GA.  Storage tends to release pretty frequently though so I would expect you'd need support for another service-version within a few months of GA.  I would recommend you still ship a `ServiceVersion` (or equivalent) enum at GA with a single value that gets defaulted for customers to the latest.  This will allow customers who care about running against a specific service-version to pin if required so they won't be moved forward automatically by a future client-version.

JeffreyRichter:
Ted makes a good point about the pinning - I forgot about that. So, I agree that it would best to have this before GA but I personally would be OK if it was not there by GA. I know there is lots of work to do and time is runnign out - this is lower priority than some other things.

tg-msft:
I agree this isn't a ship blocker and could be added with only a remote possibility of minor semantic breaks.  It would also be a no-op as implemented today (i.e., just add the enum, add the param, and then you're done) so it's more about deciding on the pattern than anything.  You can wait to plumb this value through to every method when you add a second version.

Jinming-Hu:
blocked by https://github.com/Azure/azure-sdk-for-cpp/issues/1393

PR #1349, opened 2021-01-13
Service version const strings

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/1056

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
@@ -124,6 +124,17 @@ namespace Azure { namespace Storage { namespace Blobs {
   struct BlobClientOptions
   {
     /**
+     * Versioning for Azure Storage Services.
+     */
+    struct ServiceVersion
+    {
+      /**
+       * 2020-02-10 service version.
+       */
+      constexpr static const char* V2020_02_10 = "2020-02-10";
+    };
+
+    /**
      * @brief Transport pipeline policies for authentication, additional HTTP headers, etc., that
      * are applied to every request.
      */
--- a/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_options.hpp
+++ b/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_options.hpp
@@ -26,6 +26,17 @@ namespace Azure { namespace Storage { namespace Files { namespace DataLake {
   struct DataLakeClientOptions
   {
     /**
+     * Versioning for Azure Storage Services.
+     */
+    struct ServiceVersion
+    {
+      /**
+       * 2020-02-10 service version.
+       */
+      constexpr static const char* V2020_02_10 = "2020-02-10";
+    };
+
+    /**
      * @brief Transport pipeline policies for authentication, additional HTTP headers, etc., that
      * are applied to every request.
      */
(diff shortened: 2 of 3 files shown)


Jinming-Hu:
@msftbot merge this PR after @katmsft approves

katmsft (approved):

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -123,6 +123,17 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    /**
+     * Versioning for Azure Storage Services.
+     */
+    struct ServiceVersion
+    {
+      /**
+       * 2020-02-10 service version.
+       */
+      constexpr static const char* V2020_02_10 = "2020-02-10";
+    };
```
ahsonkhan, line 135: Shouldn't this be an enum/extensible enum?

In which case, should this be a class with `const static` values?

cc @JeffreyRichter

Jinming-Hu: Extensible enum is used when client side needs to parse a value returned from server side. I don't see when this SDK or customer needs to do that. Actually I don't even know when this service version string will be used.

ahsonkhan: I thought this version was going to be passed as the API version header to the service, but might be mistaken. @tg-msft, are we supposed to plumb the `ServiceVersion` string to any of the storage service API calls?

JeffreyRichter: Jinming is correct; this should not be an extensible enum.
I think we just want a string field in the XxxClient options struct called ApiVersion. By default, we set this to the api-version that corresponds to the latest version of the service and this version is what we test & support.
We want to allow the customer to pass their own string if they want to throw caution to the wind and just try an arbitrary api-version. Usually, the customers experience will be fine if they do this but, we need to say that this is not a tested & supported scenario. This enables scenarios like Azure Stack Hub and other non-public clouds to work that do not have the latest version of the service available.

ahsonkhan: Aside from the `ApiVersion` string field, why not help such a customer by providing them with the list of valid api version strings to choose from, rather than let them pick any arbitrary free-form string?

This way, let's say we ship storage v2, we will have V2020_02_10 and say V2021_12_31, and then the user can set `ApiVersion` to either one of those (still best effort), rather than the user having to fish out from storage service docs the version strings.

JeffreyRichter: Because we don't support this. It feels odd to me to here are a bunch of possible values and we don't guarantee any of them will work for you.

Jinming-Hu: Hi @JeffreyRichter , are you suggesting that I should remove what I've added in this PR and add an extra `std::string` field into `BlobClientOptions`, the default value for this `std::string` should be the latest API version we support, while we still allow our customers to change this `std::string` value if they want to try another API version?

JeffreyRichter: Yes, exactly.

Jinming-Hu: @JeffreyRichter Thanks for your clarification!

PR #1428, opened 2021-01-21
Add support for HTTP Etag to Azure::Core

RickWinter:
Add ETag class
- Supports Weak and Strong ETags
- Supports comparison between ETags


sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+namespace Azure { namespace Core {
+
+  /**
+   * @brief Represents an HTTP ETag.
+   */
+  class ETag {
+  private:
+    Nullable<std::string> m_value;
+
+  private:
```
vhvb1989, line 24:
```suggestion
```

private: is already defined before

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+
+      // If both values are empty then we consider the ETag equal
+      if (!left.m_value && !right.m_value)
+        return true;
+
+      if (!left.m_value || !right.m_value)
+        return false;
+
+      // Previous comparison checked that both are Weak or both are strong
```
vhvb1989, line 68: How?  Didn't prev comparison checked that both are empty or at least one is empty only?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+      //  Thuse we can decide which comparison to do based on one of them
+      if (!left.IsWeak())
+      {
+        // Strong ETag so comparison is that the values match character for character
+        return left.m_value.GetValue().compare(right.m_value.GetValue()) == 0;
+      }
+      else
```
vhvb1989, line 75: Remove this `else`.  The previous if() is returning, so the else is not required here.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+      {
+        // Left side is weak ETag
+        if (right.IsWeak())
+        {
+          return left.m_value.GetValue().compare(right.m_value.GetValue()) == 0;
+        }
+        // Trim the weak tag
+        auto leftLength = left.m_value.GetValue().length();
+
+        if (leftLength - 2 != right.m_value.GetValue().length())
```
vhvb1989, line 85: Why `-2` ?   
Can this end up as negative or underflow?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+        {
+          return false;
+        }
+        return left.m_value.GetValue().compare(2, leftLength - 2, right.m_value.GetValue()) == 0;
+      }
+      return false;
```
vhvb1989, line 91: I think once you remove the else {}, this return would be removed

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+  protected:
+    /**
+     * @brief Construct an empty (null) #ETag.
```
vhvb1989, line 96: Maybe explain why this is protected?  what do we expect from it as protected?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+    /*
+     * @brief Return the contained string value
+     * @return #std::string
+     */
+    const std::string& ToString()
+    {
+      if (!m_value.HasValue())
+      {
+        abort();
```
vhvb1989, line 125:
```suggestion
        std::terminate();
```

https://stackoverflow.com/questions/2820285/abort-terminate-or-exit

or maybe just `throw;`

I would avoid `abort` because we are not doing pure `C`, there might be multi-threads, etc

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+      // Shortest valid weak etag is 4
+      //  W/""
+      if (val.length() < 4)
+      {
+        return false;
+      }
+
+      // Valid format is W/""
+      //   Must end with a " in the string
+      if (val[0] == 'W' && val[1] == '/' && val[2] == '"' && val[val.size() - 1] == '"')
```
vhvb1989, line 167: Can we have more '"' in the middle ?   like W/"this is a "nice" morning"

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * @brief @ETag representing everything
+     * @notes The any ETag is *, (unquoted).  It is NOT the same as "*"
+     */
+    static const ETag& Any()
+    {
+      static ETag Any = ETag("*");
+      return Any;
```
vhvb1989, line 182:
```suggestion
      static ETag any = ETag("*");
      return any;
```

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * @brief @ETag representing no @ETag present
+     */
+    static const ETag& Null()
+    {
+      static ETag Null = ETag();
+      return Null;
```
vhvb1989, line 191:
```suggestion
      static ETag null = ETag();
      return null;
```

vhvb1989 (commented):
Added some comments.

Also, we should implement a better way to parse an eTag and throw/terminate if the format is wrong. We are doing that in general for the expected header formatting in the Http response, so we should also parse eTag following:
```
 ETag       = entity-tag
     entity-tag = [ weak ] opaque-tag
     weak       = %x57.2F ; "W/", case-sensitive
     opaque-tag = DQUOTE *etagc DQUOTE
     etagc      = %x21 / %x23-7E / obs-text
                ; VCHAR except double quotes, plus obs-text
```

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+        return left.m_value.GetValue().compare(2, leftLength - 2, right.m_value.GetValue()) == 0;
+      }
+      return false;
+    }
+
+  protected:
+    /**
+     * @brief Construct an empty (null) #ETag.
+     */
+    ETag() {}
```
CaseyCarter, line 98: Why is this constructor protected, i.e., accessible only to derived classes? Also, this should be explicitly defaulted as `ETag() = default;` so folks aren't left wondering why we've gone out of our way to make the default constructor non-trivial.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+    // etag:"*"                         //This means the etag is value '*'
+    // etag:"some value"                //This means the etag is value 'some value'
+    // etag:/W""                        //Weak eTag
+    // etag:*                           //This is special, means any etag
+    // If-Match header can do this
+    // If-Match:"value1","value2","value3"  // Do this if any of these match
+    //
+
+    */
+    bool ETagComparison(const ETag& left, const ETag& right) const
```
CaseyCarter, line 58: It's unclear to me how this function relates to the discussion in the prior comment; should this be named `StrongComparison` or `WeakComparison` for clarity?

Revisiting after reading the rest of the function: this computes `StrongComparison` if `left` is not a weak tag, and `WeakComparison` if `left` is a weak tag. Is the asymmetry intended?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+      {
+        // Strong ETag so comparison is that the values match character for character
+        return left.m_value.GetValue().compare(right.m_value.GetValue()) == 0;
+      }
+      else
+      {
+        // Left side is weak ETag
+        if (right.IsWeak())
+        {
+          return left.m_value.GetValue().compare(right.m_value.GetValue()) == 0;
```
CaseyCarter, line 80: I'd probably use `return left.m_value.GetValue() == right.m_value.GetValue();` here as well, although I now see that you're probably using this style for consistency with line 89 that performs an substring comparison and truly needs the `compare` syntax.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+  public:
+    /**
+     * @brief Construct a #ETag.
+     */
+    explicit ETag(std::string etag) : m_value(std::move(etag)) {}
+
+    /**
+     * @brief Copy assignment.
+     */
+    ETag& operator=(const ETag&) = default;
```
CaseyCarter, line 109: It's weird that we're explicitly defaulting copy assignment - which suppresses the implicit definition of the move constructor and move assignment operator, which would otherwise be fine - and allowing the copy constructor to be implicitly defined.

The guidelines recommend following the "rule of 3/5/0" (https://azure.github.io/azure-sdk/cpp_implementation.html#type-safety-recommendations) and I observe that defining zero (of the destructor and copy/move constructor/assignments) does the Right Thing for this class, so this assignment operator declaration should probably be omitted.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+      }
+      return m_value.GetValue();
+    }
+
+    /**
+     * @brief Compare with \p other @ETag for equality.
+     * @param other Other @ETag to compare with.
+     * @return `true` if @ETag instances are equal, `false` otherwise.
+     */
+    bool operator==(const ETag& other) const { return ETagComparison(*this, other); }
```
CaseyCarter, line 135: This is the only callsite for `ETagComparison`; consider simply inlining the definition here to reduce the total number of function definitions.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * @brief Compare with \p other @ETag for inequality.
+     * @param other Other @ETag to compare with.
+     * @return `true` if @ETag instances are not equal, `false` otherwise.
+     */
+    bool operator!=(const ETag& other) const { return !(*this == other); }
+
+    /**
+     * @brief Compare with \p other @ETag for inequality.
```
CaseyCarter, line 145: Nope.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+      }
+
+      // Valid format is W/""
+      //   Must end with a " in the string
+      if (val[0] == 'W' && val[1] == '/' && val[2] == '"' && val[val.size() - 1] == '"')
+      {
+        return true;
+      }
+
+      return false;
```
CaseyCarter, line 172: Pure style again: `if (boolean_expression) { return true; } return false;` could as well be `return boolean_expression;`.

sdk/core/azure-core/inc/azure/core/etag.hpp
JeffreyRichter: If it were me, I'd have this method take another parameter, like "comparisonType" which is an enum with Weak and Strong symbols. This way, the caller decides what kind of comparison they want and doesn't depend on our magic algorithm. Then, you could overload the ==/!= operators to have them always do a string comparison.

What I describe is similar to https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=net-5.0#System_String_Equals_System_String_System_String_System_StringComparison_ and then String overridig the ==/!= operators.

sdk/core/azure-core/inc/azure/core/etag.hpp
JeffreyRichter: Not if you take my suggestion

sdk/core/azure-core/inc/azure/core/etag.hpp
JeffreyRichter: No

sdk/core/azure-core/inc/azure/core/etag.hpp
JeffreyRichter: I'm not sure we need this member. Do you have a scenario where it is required?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * @brief Represents an HTTP ETag.
+   */
+  class ETag {
+  private:
+    Nullable<std::string> m_value;
+
+  private:
+    /*
+    2.3.2.  Comparison
```
ahsonkhan, line 26: Note to self, this is from https://tools.ietf.org/html/rfc7232#section-2.3.2

sdk/core/azure-core/inc/azure/core/etag.hpp
RickWinter: No b/c to get here we are guaranteed the length is at least 4

sdk/core/azure-core/inc/azure/core/etag.hpp
RickWinter: I didn't validate the ETag on construction.  Although a client could send/receive an improperly formated ETag, its up to the service what they do with them.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,194 @@
...
+    */
+    bool ETagComparison(const ETag& left, const ETag& right) const
+    {
+
+      // If both values are empty then we consider the ETag equal
+      if (!left.m_value && !right.m_value)
+        return true;
+
+      if (!left.m_value || !right.m_value)
+        return false;
```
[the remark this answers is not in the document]
vhvb1989: but I think we always prefer the using of `{ } `  even for one only stataments ifs

```
if (left || right) 
{
 return aaa;
}

File: sdk/core/azure-core/inc/azure/core/etag.hpp (hunk 7 above)
Comment (Jinming-Hu) [reply sdk/core/azure-core/inc/azure/core/etag.hpp:125]: it says abort in guidelines

https://azure.github.io/azure-sdk/cpp_introduction.html#cpp-design-logical-errorhandling-actofgod-crash

File: sdk/core/azure-core/inc/azure/core/etag.hpp (hunk 9 above)
Comment (Jinming-Hu) [reply sdk/core/azure-core/inc/azure/core/etag.hpp:182]: Can we just define a static member? Like

```C++
static ETag Any;
static ETag Null;
```

sdk/core/azure-core/inc/azure/core/etag.hpp
RickWinter: It's for the case where a model type has an ETag field but the response had no ETag header.  Allows the model to initialize the field representing `ETag = `

sdk/core/azure-core/inc/azure/core/etag.hpp
CaseyCarter: > Can we just define a static member?

We _can_ do so, but users then have to understand the dangers of C++'s hated [static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof). The order of initialization of variables with static storage duration and dynamic initialization is unspecified, so referring to such variables during dynamic initialization becomes a footgun. This pattern (known as the Meyers' Singleton after being popularized by Scott Meyers) avoids the fiasco via function-local statics which use a different mechanism to ensure they are initialized on the first call to the function.

sdk/core/azure-core/inc/azure/core/etag.hpp
RickWinter: Comment adjusted

sdk/core/azure-core/inc/azure/core/etag.hpp
ahsonkhan: What about other uses of static within the SDK, for example, in extensible enums or Datetime? Do they have a similar issue, and if not, why not?

https://github.com/Azure/azure-sdk-for-cpp/blob/5af134b1032b4df1066ac0da7bb4b79e61e9fb79/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/protocol/datalake_rest_client.hpp#L147-L149

https://github.com/Azure/azure-sdk-for-cpp/blob/5af134b1032b4df1066ac0da7bb4b79e61e9fb79/sdk/core/azure-core/inc/azure/core/datetime.hpp#L53

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * @brief Construct an empty (null) #ETag.
+     */
+    ETag() = default;
+
+  public:
+    /**
+     * @brief Construct a #ETag.
+     */
+    explicit ETag(std::string etag) : m_value(std::move(etag)) {}
```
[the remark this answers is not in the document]
ahsonkhan: @Jinming-Hu, @katmsft would storage need/benefit from a validating internal `Parse` method similar to .NET that rejects strings that would be considered invalid `Etag` values?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+    // etag:"*"                         //This means the etag is value '*'
+    // etag:"some value"                //This means the etag is value 'some value'
+    // etag:/W""                        //Weak eTag
+    // etag:*                           //This is special, means any etag
+    // If-Match header can do this
+    // If-Match:"value1","value2","value3"  // Do this if any of these match
+    //
+
+    */
+    static bool Compare(const ETag& left, const ETag& right, const ETagComparison comparisonKind = ETagComparison::Strong)
```
ahsonkhan, line 64: Public doxygen docs missing

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  protected:
+    /**
+     * @brief Construct an empty (null) #ETag.
+     */
+    ETag() = default;
+
+  public:
+    /**
+     * @brief Construct a #ETag.
```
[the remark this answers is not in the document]
ahsonkhan: Also docs missing the `@param`

sdk/core/azure-core/inc/azure/core/etag.hpp
CaseyCarter: Yes, they do. I suppose the simplest way to address this is to forbid using the library before `main` is called and then we can use all of the dynamic initializers we like.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+    // etag:*                           //This is special, means any etag
+    // If-Match header can do this
+    // If-Match:"value1","value2","value3"  // Do this if any of these match
+    //
+
+    */
+    static bool Compare(const ETag& left, const ETag& right, const ETagComparison comparisonKind = ETagComparison::Strong)
+    {
+      // ETags are != if one of the values is null
+      if (!left.m_value || !right.m_value)
```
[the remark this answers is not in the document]
RickWinter: Those two statements do the same thing.  Directly accessing the m_value avoids a potential function call to ETag::HasValue() which itself just does m_value.HasValue().
Note, Nullable overloads the bool operator `!m_value` is implicitly doing `!m_value.HasValue()`

sdk/core/azure-core/inc/azure/core/etag.hpp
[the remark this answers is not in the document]
RickWinter: We could consider adding this in the future.  .NET does not validate the value when it constructs it

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+     * @param other Other @ETag to compare with.
+     * @return `true` if @ETag instances are not equal, `false` otherwise.
+     */
+    bool operator!=(const ETag& other) const { return !(*this == other); }
+
+    /**
+     * @brief Specifies whether the @ETag is strong or weak
+     * @return `true` if @ETag is a weak validator, `false` otherwise.
+     */
+    bool IsWeak() const
```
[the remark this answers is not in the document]
RickWinter: We could add it later if needed.  Its rare that customers will use the IsWeak() and doing !IsWeak() seems reasonable

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+    {
+      // Null ETag is considered Strong
+      // Shortest valid weak etag has length of 4
+      //  W/""
+      // Valid weak format must start with W/"
+      //   Must end with a /"
+      const bool weak = m_value && (m_value.GetValue().length() >= 4)
```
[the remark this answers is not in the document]
RickWinter: `bool x = m_value` is the same as `bool x = m_value.HasValue()`

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+#pragma once
+
+#include "azure/core/nullable.hpp"
+
+#include <string>
+
+namespace Azure { namespace Core {
+
+  /**
+   * @brief Represents an HTTP validator.
```
ahsonkhan, line 18: validator?

```suggestion
   * @brief Represents an HTTP entity tag.
```

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+    // etag:/W""                        //Weak eTag
+    // etag:*                           //This is special, means any etag
+    // If-Match header can do this
+    // If-Match:"value1","value2","value3"  // Do this if any of these match
+    //
+
+    */
+
+    /*
+    * @brief Indicates whether two @ETag values are equal
```
ahsonkhan, line 69:
```suggestion
    * @brief Indicates whether two @ETag values are equal.
```

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+    * @param left @ETag to compare
+    * @param right @ETag to compare
+    * @param comparisonKind Determines what @ETagComparison to perform, default is #ETagComparison Strong
```
ahsonkhan, line 72:
```suggestion
    * @param left @ETag to compare.
    * @param right @ETag to compare.
    * @param comparisonKind Determines what @ETagComparison to perform, default is #ETagComparison Strong.
```

sdk/core/azure-core/inc/azure/core/etag.hpp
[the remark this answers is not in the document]
ahsonkhan: > Note, Nullable overloads the bool operator `!m_value` is implicitly doing `!m_value.HasValue()`

Ah I see, that's what I was missing. Interesting.

sdk/core/azure-core/inc/azure/core/etag.hpp
RickWinter: An ETag is our implementation of the validator used by HTTP conditional requests.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+    /*
+     * @brief Returns the resource metadata represented as a string
+     * @return #std::string
+     */
+    const std::string& ToString() const
+    {
+      if (!m_value.HasValue())
+      {
+        abort();
```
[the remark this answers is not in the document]
ahsonkhan: Escaped the angle brackets above to clarify.

sdk/core/azure-core/inc/azure/core/etag.hpp
ahsonkhan: Isn't validators a bit overloaded and describe different concepts:
- the comparison mechanisms (strong vs weak)
- the generic term describing any version to compare against, including the last-modified date
 
ETags are a specific type of validator, and they use two types of validation.

How about this:
```suggestion
   * @brief Represents an entity tag which is an opaque string, uniquely identifying each version, and used an HTTP validator.
```

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,200 @@
...
+       +--------+--------+-------------------+-----------------+
+
+       // etag:                            //This is possible and means no etag is present
+    // etag:""
+    // etag:"*"                         //This means the etag is value '*'
+    // etag:"some value"                //This means the etag is value 'some value'
+    // etag:/W""                        //Weak eTag
+    // etag:*                           //This is special, means any etag
+    // If-Match header can do this
+    // If-Match:"value1","value2","value3"  // Do this if any of these match
```
ahsonkhan, line 64: nit: inconsistent indentation

sdk/core/azure-core/test/ut/simplified_header.cpp
```diff
@@ -17,13 +17,14 @@ class DllExportTest {
   AZ_CORE_DLLEXPORT static const bool DllExportHIncluded;
 };
 
-TEST(Logging, simplifiedHeader)
+TEST(SimplifiedHeader, core)
```
ahsonkhan, line 20: Good catch :)

sdk/core/azure-core/inc/azure/core/etag.hpp
ahsonkhan: Is there an alternative design for extensible enums that doesn't rely on statics? This is how they are defined in the guidelines:


> I suppose the simplest way to address this is to forbid using the library before `main` is called and then we can use all of the dynamic initializers we like.

How would you document this to let the customers know what not to do? What types of runtime errors or bugs would they see if they did use the library before `main` (if we can share some examples, that will help users recognize the issue)?

sdk/core/azure-core/inc/azure/core/etag.hpp
RickWinter: removed

sdk/core/azure-core/test/ut/etag.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+#include <gtest/gtest.h>
+
+#include <azure/core/etag.hpp>
+
+#include <limits>
+
+using namespace Azure::Core;
+
+TEST(ETag, Equality)
```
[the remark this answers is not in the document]
RickWinter: We don't validate ETags (similar to other languages).  Perhaps we will in the future

sdk/core/azure-core/inc/azure/core/etag.hpp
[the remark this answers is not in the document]
Jinming-Hu: No, we don't expect anybody to pass an invalid string here. I cannot see any benefits of doing validation.

sdk/core/azure-core/inc/azure/core/etag.hpp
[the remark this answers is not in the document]
Jinming-Hu: I don't think weak etag is used anywhere. So it's totally fine if we just support strong etag. We can add weak etag later if needed (I assume it will never be needed).

sdk/core/azure-core/inc/azure/core/etag.hpp
Jinming-Hu: I know the static initialization order fiasco, but I don't know what to do here. There are many many static members in storage SDK, I really don't want to make all of them a static function. Actually I don't expect anybody to use these static members during program loading. So maybe it's not that big a deal.

sdk/core/azure-core/inc/azure/core/etag.hpp
ahsonkhan:
```suggestion
    // etag:                            //This is possible and means no etag is present
    // etag:""
    // etag:"*"                         //This means the etag is value '*'
    // etag:"some value"                //This means the etag is value 'some value'
    // etag:/W""                        //Weak eTag
    // etag:*                           //This is special, means any etag
    // If-Match header can do this
    // If-Match:"value1","value2","value3"  // Do this if any of these match
```

ahsonkhan (approved):
Other than settling on the ToString behavior for an empty/null ETag, looks good.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -0,0 +1,196 @@
...
+          auto rightVal = right.m_value.GetValue();
+
+          // Compare is lengths are equal
+          return ((leftVal.length() - leftStart) == (rightVal.length() - rightStart))
+              && (leftVal.compare(leftStart, leftVal.length() - leftStart, &rightVal[rightStart])
+                  == 0);
+          break;
+      }
+      // Unknown comparison
+      abort();
```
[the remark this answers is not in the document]
ahsonkhan: No, it doesn't:
http://www.cplusplus.com/reference/cstdlib/abort/

Should we consider a standardized way to do this, similar to preconditions, maybe using `exit()` with a specific status integer that we can document, so users have some idea of what the type of error is, when debugging?
http://www.cplusplus.com/reference/cstdlib/exit/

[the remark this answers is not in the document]
Jinming-Hu: I think `abort()` here is better than `exit()`. gdb and VS debugger will trigger a breakpoint for `abort()`. But I don't think they will trigger breakpoint for `exit()`.

sdk/core/azure-core/inc/azure/core/etag.hpp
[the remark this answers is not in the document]
RickWinter: What if the users etag is the string `<null>`.  .NET is saying that is the empty ETag, but according to RFC 7232 it is a valid ETag.  How would I distinguish an empty ETag of `<null>` from the user specifically setting an etag of `<null>`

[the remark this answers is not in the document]
ahsonkhan: From my reading of the RFC, `<null>`  is **not** a valid etag string, so you wouldn't need to distinguish. A valid etag needs to be surrounded within quotes. Therefore, using `<null>` is fine as a sentinel placeholder.
> An entity-tag consists of an opaque quoted string, possibly
   prefixed by a weakness indicator.


```text
ETag       = entity-tag

     entity-tag = [ weak ] opaque-tag
     weak       = %x57.2F ; "W/", case-sensitive
     opaque-tag = DQUOTE *etagc DQUOTE
     etagc      = %x21 / %x23-7E / obs-text
                ; VCHAR except double quotes, plus obs-text
```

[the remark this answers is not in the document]
RickWinter: Yes, `"\"<null>\""` is different than `"<null>"`.

PR #1460, opened 2021-01-26
Update clang format validate script instructions when it fails to be accurate

ahsonkhan:
Syncing the instructions based on changes from https://github.com/Azure/azure-sdk-for-cpp/pull/1438

Currently, the instructions suggest running a command which may not work on all OSes:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=701256&view=logs&j=0585f5d6-0937-5c15-894c-06e189e06847&t=a45688da-a69e-5817-f832-3aa6246959b7
```text
find sdk/ -regex '.*\.\(cpp\|hpp\)' -exec clang-format -style=file -i {} \;
```


vhvb1989 (approved):
<3

Issue #291, opened 2020-07-14
[Storage Blob Service] Wait Copy Operation to Finish Support in storage SDK


Jinming-Hu:

JinmingHu-MSFT:
blocked by base poller class that is going to be provided by Core sdk

PR #1461, opened 2021-01-26
Support poll operations in blob service

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/291

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/operation.hpp
+++ b/sdk/core/azure-core/inc/azure/core/operation.hpp
@@ -34,6 +34,8 @@ namespace Azure { namespace Core {
     OperationStatus m_status = OperationStatus::NotStarted;
 
   public:
+    virtual ~Operation() {}
+
     /**
      * @brief Final reuslt of the long-running operation.
      *
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
@@ -5,6 +5,7 @@
 
 #include <cstdint>
 #include <string>
+#include <thread>
 #include <vector>
 
 #include <azure/core/operation.hpp>
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
@@ -72,8 +72,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       std::string LeaseId;
     };
 
-    class StartCopyBlobResult : public Azure::Core::Operation<GetBlobPropertiesResult>
-    {
+    class StartCopyBlobResult : public Azure::Core::Operation<GetBlobPropertiesResult> {
     public:
       std::string RequestId;
       std::string ETag;
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
@@ -72,7 +72,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       std::string LeaseId;
     };
 
-    struct StartCopyBlobResult : public Azure::Core::Operation<GetBlobPropertiesResult>
+    class StartCopyBlobResult : public Azure::Core::Operation<GetBlobPropertiesResult>
     {
     public:
       std::string RequestId;
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
@@ -7,61 +7,103 @@
 #include <string>
 #include <vector>
 
+#include <azure/core/operation.hpp>
+
 #include "azure/storage/blobs/protocol/blob_rest_client.hpp"
 
-namespace Azure { namespace Storage { namespace Blobs { namespace Models {
+namespace Azure { namespace Storage { namespace Blobs {
 
-  struct DownloadBlobToResult
-  {
-    std::string ETag;
-    Azure::Core::DateTime LastModified;
-    int64_t ContentLength = 0;
-    BlobHttpHeaders HttpHeaders;
-    Storage::Metadata Metadata;
-    Models::BlobType BlobType;
-    bool IsServerEncrypted = false;
-    Azure::Core::Nullable<std::vector<uint8_t>> EncryptionKeySha256;
-  };
+  class BlobClient;
+  class PageBlobClient;
 
-  using UploadBlockBlobFromResult = UploadBlockBlobResult;
+  namespace Models {
 
-  struct AcquireBlobLeaseResult
-  {
-    std::string RequestId;
-    std::string ETag;
-    Azure::Core::DateTime LastModified;
-    std::string LeaseId;
-  };
+    struct DownloadBlobToResult
+    {
+      std::string ETag;
+      Azure::Core::DateTime LastModified;
+      int64_t ContentLength = 0;
+      BlobHttpHeaders HttpHeaders;
+      Storage::Metadata Metadata;
+      Models::BlobType BlobType;
+      bool IsServerEncrypted = false;
+      Azure::Core::Nullable<std::vector<uint8_t>> EncryptionKeySha256;
+    };
 
-  struct BreakBlobLeaseResult
-  {
-    std::string RequestId;
-    std::string ETag;
-    Azure::Core::DateTime LastModified;
-    int32_t LeaseTime = 0;
-  };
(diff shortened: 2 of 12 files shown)


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
```diff
@@ -5,63 +5,108 @@
...
+
+    struct RenewBlobLeaseResult
+    {
+      std::string RequestId;
+      std::string ETag;
+      Azure::Core::DateTime LastModified;
+      std::string LeaseId;
+    };
+
+    struct StartCopyBlobResult : public Azure::Core::Operation<GetBlobPropertiesResult>
```
katmsft, line 75: Should be class instead of struct?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
```diff
@@ -5,63 +5,108 @@
...
+        // Not supported
+        std::abort();
+      }
+
+      std::unique_ptr<Azure::Core::Http::RawResponse> PollInternal(
+          Azure::Core::Context& context) override;
+
+      Azure::Core::Response<GetBlobPropertiesResult> PollUntilDoneInternal(
+          Azure::Core::Context& context,
+          std::chrono::milliseconds period) override;
```
katmsft, line 102:
```suggestion
          std::chrono::milliseconds pollInterval) override;
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
Jinming-Hu: I don't think there is any essential difference between `struct ` and `class`. But other return types are `struct`, so

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
Jinming-Hu: Nah, this is an override function.  You should create issue for core sdk team.

sdk/storage/azure-storage-blobs/test/append_blob_client_test.cpp
```diff
@@ -327,19 +327,21 @@ namespace Azure { namespace Storage { namespace Test {
     Blobs::StartCopyBlobFromUriOptions copyOptions;
     copyOptions.ShouldSealDestination = false;
     auto copyResult = blobClient2.StartCopyFromUri(blobClient.GetUrl() + GetSas(), copyOptions);
-    // TODO: poller wait here
-    getPropertiesResult = blobClient2.GetProperties();
+    getPropertiesResult = copyResult->PollUntilDone(std::chrono::seconds(1));
```
katmsft, line 330: What's the SLA for this operation?

sdk/storage/azure-storage-blobs/test/page_blob_client_test.cpp
```diff
@@ -157,6 +157,9 @@ namespace Azure { namespace Storage { namespace Test {
     EXPECT_FALSE(copyInfo->CopyStatus.Get().empty());
     EXPECT_TRUE(copyInfo->VersionId.HasValue());
     EXPECT_FALSE(copyInfo->VersionId.GetValue().empty());
+    auto getPropertiesResult = copyInfo->PollUntilDone(std::chrono::seconds(1));
```
katmsft, line 160: What's the SLA for this operation?

Jinming-Hu:
@RickWinter I added a virtual destructor to avoid some build error https://github.com/Azure/azure-sdk-for-cpp/blob/06f5ad771619224e308bb2657e151dce490ad2ac/sdk/core/azure-core/inc/azure/core/operation.hpp#L37.

I have to merge this PR today since it conflicts with another one of my PR. We can still discuss if you have any concern about the virtual destructor, or anything else in this PR.

sdk/storage/azure-storage-blobs/test/page_blob_client_test.cpp
Jinming-Hu: duplicate

sdk/storage/azure-storage-blobs/test/append_blob_client_test.cpp
Jinming-Hu: Why do we need an SLA? SLA for what?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
katmsft: https://github.com/Azure/azure-sdk/issues/1974
Topic: Struct vs Class
When should code be in a struct vs in a class?
Conclusion:
•	Structs are for passive objects that carry data;  everything else is a class
o	Invariants between fields are disallowed
o	Constructors/Destructors are allowed on structs but must not enforce any invariant
•	When in doubt make it a class.

We did mention that for non-passive objects, it should not be structs.

sdk/storage/azure-storage-blobs/test/append_blob_client_test.cpp
katmsft: You are waiting for an LRO to finish in test code, there should be a guaranteed ETA so that this won't take more than a certain time period that the whole test can tolerate.

Jinming-Hu: Is there is an SLA, the SLA should be guaranteed by server side, not client side. If the customer becomes impatient, that's when `context` comes into use.

I don't see anything that should be done by storage client code.

katmsft: Not in product code, but definitely in test code right?
What if this call took 1 hour to finish, do you think it is a good practice to wait 1 hour in this line?

Jinming-Hu: > What if this call took 1 hour to finish, do you think it is a good practice to wait 1 hour in this line?

If it did happen, I'd definitely do something. But I don't think it's gonna happen ever.

katmsft: So how long is the maximum execution time of this line of code? To put it in another way:
>What's the SLA for this operation?

katmsft (approved):
Still have a concern on the maximum execution time for the LRO in test code, however approve to unblock.

sdk/storage/azure-storage-blobs/test/append_blob_client_test.cpp
Jinming-Hu: I don't know, this question should be answered by server guys. I only work on client side.

ahsonkhan:
> Still have a concern on the maximum execution time for the LRO in test code

Me too. We shouldn't rely on timing in unit tests since it introduces flakiness and causes false positive test failures, since they end up being a non-deterministic factor.

>  I added a virtual destructor to avoid some build error

What was the error @Jinming-Hu  and why did adding a virtual destructor solve it?

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -34,6 +34,8 @@ namespace Azure { namespace Core {
     OperationStatus m_status = OperationStatus::NotStarted;
 
   public:
+    virtual ~Operation() {}
```
ahsonkhan, line 37: Probably want a CL update.

sdk/storage/azure-storage-blobs/src/blob_responses.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "azure/storage/blobs/blob_responses.hpp"
+
+#include "azure/storage/blobs/blob_client.hpp"
+
+namespace Azure { namespace Storage { namespace Blobs { namespace Models {
+
+  std::unique_ptr<Azure::Core::Http::RawResponse> StartCopyBlobResult::PollInternal(
+      Azure::Core::Context& context)
+  {
+    unused(context);
```
ahsonkhan, line 13: @vhvb1989 why did we avoid using `unused()` and instead went with `(void)` in core.

Looks like storage's reliable stream does the same. We may want to be consistent here:
https://github.com/Azure/azure-sdk-for-cpp/blob/50a4186d9615d327c4cf20d673cec47f0d7e6802/sdk/storage/azure-storage-common/src/reliable_stream.cpp#L15

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/1483

Issue #1139, opened 2020-12-10
StorageException inherits from RequestFailedException


Jinming-Hu:
```C  
struct StorageException : public std::runtime_error
```
#
Why aren't all of these things on RequestFailedException?

PR #1463, opened 2021-01-26
throw exception Azure::Core::RequestFailedException or those inherit from that

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/1139

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -374,7 +374,14 @@ namespace Azure { namespace Storage { namespace Blobs {
     /**
      * @brief Optional conditions that must be met to perform this operation.
      */
-    BlobContainerAccessConditions AccessConditions;
+    struct : public LeaseAccessConditions
```
katmsft, line 377: Not single purpose, and please add change log.

Jinming-Hu: It is single purpose. The main purpose of this PR is to correct storage exception types. In the past, we throw exception on unsupported access condition. Now that we removed the unsupported access condition, we don't need to throw any more

Jinming-Hu: Plus I don't think this need to be reflected in changelog.

katmsft: This option used to have `IfModifiedSince` and `IfUnmodifiedSince`, but now only `IfModifiedSince` right? Why is that not something needs to be reflected in changelog?

Jinming-Hu: `IfUnmodifiedSince` never works. It throws in the past, now it crashes. Why would anybody care about it?

katmsft: No, it does not crash now, it is a compiler error now. Are you saying that the precondition check changed to compiler error is not breaking change?

Jinming-Hu: I may be able to accept adding it into changelog? But is it a breaking change? I don't think so.

katmsft: I believe it is either a breaking change, or we shouldn't mention it at all.
Consider if the customer assigns a `BlobContainerAccessConditions` everywhere in the options, it used to work if `IfUnmodifiedSince` is null, but not now for this API.

Jinming-Hu: well, okay

katmsft (approved):
The only concern is if the change to the option is a breaking change. Leave it to the author to decide.

PR #1515, opened 2021-01-28
Nosignal libcurl option

vhvb1989:
fixes: #1514


vhvb1989:
Won't go for this path for now

PR #1535, opened 2021-02-01
Update changelog for the 'retry-after' fix

### LINKED ISSUE #1526: Add a bug fix changelog entry for retry policy with the casing fix for retry-after
@antkmsft we would also need changelog entry for this

_Originally posted by @vhvb1989 in https://github.com/Azure/azure-sdk-for-cpp/issues/1522#issuecomment-769974212_

antkmsft:
Closes #1526

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Bug Fixes
 
 - Fixed the parsing of the last chunk of a chunked response when using the curl transport adapter.
+- Fixed reading the value from `retry-after` header in `RetryPolicy`.
 
 ## 1.0.0-beta.4 (2021-01-13)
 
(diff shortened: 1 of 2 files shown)


RickWinter (approved):
This part looks good.  The bug mentioned adding test cases around the casing logic as well.  Only close the bug if the test coverage is checked in as well.

antkmsft:
@RickWinter , not sure yet how to test that RetryPolicy does really sleep for the expected value. In C repo, we can inject mocks, not sure about here. If we want this soon, for the release, there may not be an option ready. Let's discuss later today.

ahsonkhan (approved):

RickWinter:
At a simple level we should have tests that parse off the value of the retry header and do a retry.  You could create a custom pipeline with the retrypolicy and your response generator.  Would allow you to simulate the policy parsing the value and calling back to your response generator which can also double as validation on the retry happening.

antkmsft:
@RickWinter , I extracted unit test update to a separate issue, #1570. I'll close #1526, because it only talks about the changelog.

Issue #1525, opened 2021-01-29
Use case insensitive map for our request response headers - move the comparator from storage to core.


ahsonkhan:
Move the one from storage into azure core internal.

Add a case insensitive comparer in our request/response header map, which will make it resilient to such issues going forward, similar to storage SDK's `Metadata`:
https://github.com/Azure/azure-sdk-for-cpp/blob/5e2e84cdbcf49497ac67b77f642c2c7d0f2f0278/sdk/storage/azure-storage-common/inc/azure/storage/common/storage_common.hpp#L54-L64

_Originally posted by @ahsonkhan in https://github.com/Azure/azure-sdk-for-cpp/issues/1522#issuecomment-770043030_

PR #1536, opened 2021-02-01
Use case-insensitive maps for Request and RawResponse headers

antkmsft:
There was idea about this in our recent discussion.
Do we want this? If so, when?
Is there a work item for this?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/http/http.hpp
+++ b/sdk/core/azure-core/inc/azure/core/http/http.hpp
@@ -11,6 +11,7 @@
 #include "azure/core/exception.hpp"
 #include "azure/core/http/body_stream.hpp"
 #include "azure/core/internal/contract.hpp"
+#include "azure/core/internal/strings.hpp"
 #include "azure/core/nullable.hpp"
 
 #include <algorithm>
@@ -45,7 +46,10 @@ namespace Azure { namespace Core { namespace Http {
      * @throw if \p headerName is invalid.
      */
     void InsertHeaderWithValidation(
-        std::map<std::string, std::string>& headers,
+        std::map<
+            std::string,
+            std::string,
+            Azure::Core::Internal::Strings::CaseInsensitiveComparator>& headers,
         std::string const& headerName,
         std::string const& headerValue);
   } // namespace Details
@@ -454,12 +458,16 @@ namespace Azure { namespace Core { namespace Http {
     friend class Azure::Core::Test::TestHttp_getters_Test;
     friend class Azure::Core::Test::TestHttp_query_parameter_Test;
 #endif
+  public:
+    typedef std::
+        map<std::string, std::string, Azure::Core::Internal::Strings::CaseInsensitiveComparator>
+            HeaderMap;
 
   private:
     HttpMethod m_method;
     Url m_url;
-    std::map<std::string, std::string> m_headers;
-    std::map<std::string, std::string> m_retryHeaders;
+    HeaderMap m_headers;
+    HeaderMap m_retryHeaders;
 
     BodyStream* m_bodyStream;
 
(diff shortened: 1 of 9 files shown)


sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -454,12 +458,16 @@ namespace Azure { namespace Core { namespace Http {
     friend class Azure::Core::Test::TestHttp_getters_Test;
     friend class Azure::Core::Test::TestHttp_query_parameter_Test;
 #endif
+  public:
```
RickWinter, line 461: This type is general http.  Move the typedef into Azure::Core::Http and then share it

ahsonkhan:
> Do we want this? If so, when?
> Is there a work item for this?

@antkmsft here's the issue: https://github.com/Azure/azure-sdk-for-cpp/issues/1525

@JeffreyRichter - what are your thoughts on using a case-insensitive map for our request/response headers similar to storage metadata? It makes our API less error prone, but there is a minor drawback, that we now expose "custom" types rather than plain-old `map<string, string>`

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -604,13 +612,17 @@ namespace Azure { namespace Core { namespace Http {
    * @brief Raw HTTP response.
    */
   class RawResponse {
+  public:
+    typedef std::
+        map<std::string, std::string, Azure::Core::Internal::Strings::CaseInsensitiveComparator>
+            HeaderMap;
```
ahsonkhan, line 618: Is it OK for a public typedef to expose an internal type? Doesn't that break our access rules. Are we allowed to rename or  change the `CaseInsensitiveComparator` if we expose it in the type definition like this?

JeffreyRichter:
I'd be OK with our own headers type that is a case-insensitive map.

ahsonkhan:
Would you be OK with the alternative, where rather than exposing our own headers type, or returning types as `map<string, string>`, we made the custom comparator public, and changed the return type to be:
`map<string, string, Azure::Core::CaseInsensitiveComparator>`?

The benefit of that is that we can apply that to other places where we need a case-insensitive map and don't need custom types for each instance.

JeffreyRichter:
I'm less OK with this. How about our own type that embeds the map<string, string, Azure::Core::CaseInsensitiveComparator> as a private field and then we replicate all the map functions forwarding each to the inner map?

ahsonkhan:
I want to understand the distinction between your two suggestions. Can you please flesh out what you mean by option 1? By Option 1 are you suggesting a typedef (like Anton has done in this PR)?

**Option 1:**
> I'd be OK with our own headers type that is a case-insensitive map.

**Option 2:**
> How about our own type that embeds the map<string, string, Azure::Core::CaseInsensitiveComparator> as a private field and then we replicate all the map functions forwarding each to the inner map?

I understand Option 2 as following the wrapper/composition pattern.

JeffreyRichter:
I thought of option #1 as asking me a conceptual question: Am I OK if the SDK has its own header type that is basically a case-insensitive map? Yes, I am. I prefer not to implement this by making the case-insensitivi comparer function public; I prefer to implement this by way of option 2 - wrapping/composition.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -45,7 +46,10 @@ namespace Azure { namespace Core { namespace Http {
      * @throw if \p headerName is invalid.
      */
     void InsertHeaderWithValidation(
-        std::map<std::string, std::string>& headers,
+        std::map<
+            std::string,
+            std::string,
+            Azure::Core::Internal::Strings::CaseInsensitiveComparator>& headers,
```
Jinming-Hu, line 52: You wanna use a typedef for this type? because we're gonna use it in a lot of places.

like

```
using CaseInsensitiveMap = std::map<std::string, std::string, Azure::Core::Internal::Strings::CaseInsensitiveComparator>;
```

PR #1610, opened 2021-02-05
Add WinHttp cmake option to the contributing guide

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1300

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,6 +143,11 @@ The following CMake options are available for adding/removing project features.
 <td>OFF</td>
 </tr>
 <tr>
+<td>BUILD_TRANSPORT_WINHTTP</td>
+<td>Build the WinHTTP transport adapter. When building on Windows systems, if no other transport adapter is built, this option will be automatically turned ON. This option is not supported on non-Windows OSes.</td>
+<td>OFF</td>
+</tr>
+<tr>
 <td>BUILD_DOCUMENTATION</td>
 <td>Build Doxygen documentation</td>
 <td>OFF</td>


CONTRIBUTING.md
```diff
@@ -143,6 +143,11 @@ The following CMake options are available for adding/removing project features.
 <td>OFF</td>
 </tr>
 <tr>
+<td>BUILD_TRANSPORT_WINHTTP</td>
+<td>Build the WinHTTP transport adapter. When building on Windows systems, if no other transport adapter is built, this option will be automatically turned ON. This option is not supported on non-Windows OSes.</td>
```
RickWinter, line 147:
```suggestion
<td>Build the WinHTTP transport adapter. When building on Windows systems, if no other transport adapter is built, this option is ON by default. This option is not supported on non-Windows OSes.</td>
```

RickWinter (approved):

CONTRIBUTING.md
RickWinter: Its a NIT, Up to you if you want to use the suggested wording.

ahsonkhan: The option isn't actually ON by default. It is turned on by one of our cmake modues, only under certain conditions.
This is similar wording to the curl option above.
https://github.com/Azure/azure-sdk-for-cpp/blob/9ed77d8b0648588432ca3e3c26db911c9041c396/CMakeLists.txt#L13
https://github.com/Azure/azure-sdk-for-cpp/blob/9ed77d8b0648588432ca3e3c26db911c9041c396/cmake-modules/AzureTransportAdapters.cmake#L25-L27

PR #1614, opened 2021-02-05
Use Url::Encode() instead of the own implementation

### LINKED ISSUE #738: credentials.cpp still uses its own UrlEncode method
https://github.com/Azure/azure-sdk-for-cpp/blob/6ae0c0daac1542e655e2fd2147d88710a132186c/sdk/core/azure-core/src/credentials/credentials.cpp#L18

antkmsft:
Closes #738.

### DIFF
--- a/sdk/identity/azure-identity/src/client_secret_credential.cpp
+++ b/sdk/identity/azure-identity/src/client_secret_credential.cpp
@@ -29,7 +29,6 @@ Azure::Core::AccessToken ClientSecretCredential::GetToken(
     url.AppendPath("oauth2/v2.0/token");
 
     std::ostringstream body;
-    // TODO: Use encoding from Http::Url::Encode once it becomes public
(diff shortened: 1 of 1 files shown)


sdk/identity/azure-identity/src/client_secret_credential.cpp
```diff
@@ -55,14 +30,14 @@ Azure::Core::AccessToken ClientSecretCredential::GetToken(
 
     std::ostringstream body;
     // TODO: Use encoding from Http::Url::Encode once it becomes public
```
RickWinter, line 32: Remove
```suggestion
```

RickWinter (approved):

RickWinter:
@antkmsft Remove the comment that no longer makes sense with the fix.  LG

ahsonkhan (approved):

PR #1615, opened 2021-02-05
Disable compiler extensions, fix Clang options to have effect, fix warnings

### LINKED ISSUE #1063: Disable language extensions
In order for us to compile on more compilers, let's use compiler flags to disable compiler-specific language extensions.
We currently should be pretty compliant anyways, but, for instance, if we explicitly disable language extensions via compiler flag for MSVC, then, if you develop on a Windows machine and use a compiler extension unwittingly, there is a shorter loop for you to see the error, before it fails in CI on Mac or Linux.

antkmsft:
Closes #1063


sdk/core/azure-core/src/datetime.cpp
```diff
@@ -53,7 +53,7 @@ void ValidateDateElementRange(
   {
     outOfRange = -1;
   }
-  else if (value > maxValue)
+  else if (value > maxValue and true)
```
antkmsft, line 56:
```suggestion
  else if (value > maxValue)
```

sdk/core/azure-core/inc/azure/core/internal/json.hpp
```diff
@@ -2161,14 +2161,14 @@ _az_JSON_HEDLEY_DIAGNOSTIC_PUSH
     - @ref out_of_range for exceptions indicating access out of the defined range
     - @ref other_error for exceptions indicating other library errors
 
-    @internal
+    \@internal
```
Jinming-Hu, line 2164: Maybe it's none of my business. But you've made a lot of changes to a third-party dependency file. What if in the future we decide to upgrade to a new version?

sdk/storage/azure-storage-common/src/xml_wrapper.cpp
```diff
@@ -127,18 +127,25 @@ namespace Azure { namespace Storage { namespace Details {
     xmlBufferFree(static_cast<xmlBufferPtr>(m_buffer));
   }
 
+  namespace {
+    template <typename T> inline constexpr xmlChar* BadCast(T const* x)
+    {
+      return const_cast<xmlChar*>(reinterpret_cast<xmlChar const*>(x));
+    }
+  } // namespace
+
```
Jinming-Hu, line 136: Did you do this to avoid some kind of compiler warning?

sdk/core/azure-core/inc/azure/core/internal/json.hpp
antkmsft: It won't be a first change to this file - we already had to rename macros (tens of them) - we prefixed them with `_az_`. And the file itself is artificially crafted from multiple files - cpps an headers. So, if we pull a new version, we'll have to go through the same process - putting multiple files into one, removing include guards, pulling the list of macros, renaming macros, putting `#undef`s for them at the end, removing `and`s, `or`s, and `not`s (compiler will remind with failure), fixing the doxygen tags (compiler will remind with failure. IMO, this is laborious, but pretty monotonic.

sdk/storage/azure-storage-common/src/xml_wrapper.cpp
antkmsft: Yes, to avoid c-style cast. The macro that was used, `BAD_CAST` from libxml2, extends to a c-style cast, and I enabled a warning for the compiler to warn when it is used, so we use one of the proper C++-style casts.

Jinming-Hu: Do we need template here? Isn't it always const char*?

Jinming-Hu: You don't have to `inline` a template function. But if you mean it, it's okay.

sdk/core/azure-core/inc/azure/core/internal/json.hpp
antkmsft: And absolutely, thank you for the question, this is a good question, and a valid one, please always ask if you see something, I really appreciate it, and if you have concerns please continue conversation, I value all the feedback, I don't consider anything in our codebase as "not your business".

sdk/storage/azure-storage-common/src/xml_wrapper.cpp
Jinming-Hu: We use `const Type*` rather than `Type const*`, be consistent

antkmsft: I rather meant to inline, to come to macro as close as possible. It will be a no-op in the generated code anyways, just a cast to satisfy the compiler. LMK if you'd like to see it non-inlined. But I'd recommend to explicitly specify it as inline.

antkmsft: Interestingly, if you make it non-template, GCC complains it can't be a constexpr. Must be a bug, but still, I'd rather have guaranteed constexpr.

Jinming-Hu: Why are you so obsessed with `constexpr` here? `reinrepret_cast` and `const_cast` are almost always no-op when translated to assembly code. So I think it doesn't matter to be `constexpr` or not.

Jinming-Hu: https://stackoverflow.com/questions/26200827/getting-around-the-reinterpret-cast-limitation-with-constexpr

I found a discussion here. I don't know whether it's defined behavior to do `reinterpret_cast` in `constexpr` function. But I would avoid doing that if I'm not absolutely sure it's safe.

RickWinter (approved):

vhvb1989 (approved):
Awesome job Anton!  Thank you very much!

I just added a few suggestions to make doxygen links to work

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -74,7 +74,7 @@ namespace Azure { namespace Core {
 
   public:
     /**
-     * @brief Construct an instance of @DateTime.
+     * @brief Construct an instance of #DateTime.
```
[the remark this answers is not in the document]
RickWinter: Does doxygen fail to resolve the type?  If so, we probably have this same problem across other parts of the code.  In that case I'd suggest fixing the other issues in a seperate PR as this PR is growing beyond single purpose.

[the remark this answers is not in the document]
vhvb1989: >Does doxygen fail to resolve the type?

We need to document the namespace if we want auto-resolve:
https://stackoverflow.com/questions/46845369/doxygen-wont-resolve-reference-to-typedef-in-namespace

I am happier with full reference. But that's just personal preference.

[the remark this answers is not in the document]
vhvb1989: > I'd suggest fixing the other issues in a seperate PR 

Sure, that's why I approved the PR 👍  I don't consider it as relevant such for blocking the PR

sdk/storage/azure-storage-common/src/xml_wrapper.cpp
Jinming-Hu: My point is: we use template when we need template, we don't use template to get around some compiler error caused by combining `reinterpret_cast` and `constexpr`.

I just checked, we use all those `BAD_CAST` for `const char*`, so I don't think we need template here at all.

RickWinter:
@antkmsft Please address the feedback and then commit this PR.

RickWinter:
@antkmsft Please address feedback and check in this PR.

cmake-modules/AzureGlobalCompileOptions.cmake
```diff
@@ -12,25 +12,25 @@ if(MSVC)
   string(REGEX REPLACE "/W3" "/W4" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
   #https://stackoverflow.com/questions/37527946/warning-unreferenced-inline-function-has-been-removed
-  add_compile_options(/W4 ${WARNINGS_AS_ERRORS_FLAG} /wd5031 /wd4668 /wd4820 /wd4255 /wd4710)
-  
+  add_compile_options(/permissive- /W4 ${WARNINGS_AS_ERRORS_FLAG} /wd5031 /wd4668 /wd4820 /wd4255 /wd4710)
```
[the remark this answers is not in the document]
antkmsft: CMake does add `/std:c++14` for us, because we do specify C++ standard in the cmake files.

sdk/core/azure-core/inc/azure/core/datetime.hpp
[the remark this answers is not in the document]
antkmsft: I've found all instances of unqualiied names in the doxygen documentation and fixed all of them in this PR.

PR #1618, opened 2021-02-06
Remove Storage::Md5 and replace it with the one from Azure::Core.

ahsonkhan:
Follow up from https://github.com/Azure/azure-sdk-for-cpp/pull/1617 and https://github.com/Azure/azure-sdk-for-cpp/pull/1632

~This is the relevant commit to review: https://github.com/Azure/azure-sdk-for-cpp/commit/4eaabdb90d520a64e2e3f265ea810d52434a9f0d~


sdk/core/azure-core/inc/azure/core/md5.hpp
```diff
@@ -0,0 +1,63 @@
...
+namespace Azure { namespace Core {
+
+  /**
+   * @brief Represents the class for the MD5 hash function which maps binary data of an arbitrary
+   * length to small binary data of a fixed length.
+   */
+  class Md5 {
+  public:
+    /**
+     * @brief Construct a default instance of @Md5.
```
vhvb1989, line 23: Does `@Md5` works? Don't we have to use `#Azure::Core::Md5` ?

sdk/core/azure-core/inc/azure/core/md5.hpp
```diff
@@ -0,0 +1,63 @@
...
+     * @param data The input vector that contains the current block of binary data that is used for
+     * hash calculation.
+     * @remark Once all the data has been added, call #Digest() to get the computed hash value.
+     */
+    void Update(const std::vector<uint8_t>& data);
+
+    /**
+     * @brief Computes the hash value of all the binary input data appended to the instance so far.
+     * @return The computed MD5 hash value corresponding to the input provided.
+     * @remark Use #Update() to add more partial data before calling this function.
```
vhvb1989, line 43: Make sure that `#Udpate()` works without the full path to it like `#Azure::Core::Md5::Update`

sdk/core/azure-core/test/ut/md5.cpp
```diff
@@ -0,0 +1,85 @@
...
+  std::size_t length = 0;
+  while (length < data.size())
+  {
+    std::size_t s = static_cast<std::size_t>(RandomInt(0, 4194304));
+    s = std::min(s, data.size() - length);
+    auto begin = data.begin() + length;
+    auto end = begin + s;
+    md5Instance.Update(std::vector<uint8_t>(begin, end));
+    md5Instance.Update(std::vector<uint8_t>(begin, begin));
+    length += s;
```
vhvb1989, line 82: I would much appreciate a comment about what's going on here :)

vhvb1989 (approved):

sdk/core/azure-core/test/ut/md5.cpp
ahsonkhan: Since this is tests, I didn't bother too much (especially because it was borrowed from the existing storage test). If you understand the APIs being used, it would become a bit more clearer what it is doing. That said, I will add a comment regardless, since you brought it up :)

There are two ways to get the hash value, a "single-shot" static API called `Hash()` and one where you can stream partial data blocks with multiple calls to `Update` and then once you are done, call `Digest` to calculate the hash of the whole set of data blocks.

What this test is saying is, split up a 16MB block into many 0-4MB chunks, and compare the  single shot computed hash value with the streaming approach, and validate they are equal.

ahsonkhan:
@vhvb1989 can you approve https://github.com/Azure/azure-sdk-for-cpp/pull/1617 first, since that one needs to go in before this one.

sdk/core/azure-core/inc/azure/core/md5.hpp
ahsonkhan: Do we have documented steps on how to generate docs locally, that you can share?

sdk/storage/azure-storage-files-shares/test/share_file_client_test.cpp
```diff
@@ -600,6 +601,13 @@ namespace Azure { namespace Storage { namespace Test {
     }
   }
 
+  static std::vector<uint8_t> Hash(const std::string& data)
+  {
+    const uint8_t* ptr = reinterpret_cast<const uint8_t*>(data.data());
+    std::vector<uint8_t> v(ptr, ptr + data.length());
+    return Azure::Core::Md5::Hash(v);
```
Jinming-Hu, line 608:
```suggestion
    return Azure::Core::Md5::Hash(reinterpret_cast<const uint8_t*>(data.data()), data.size());
```

Jinming-Hu: nit: avoid copy

Jinming-Hu (approved):

sdk/core/azure-core/inc/azure/core/md5.hpp
vhvb1989: Just enable the BUILD_DOCUMENTATION option for generating and make sure to have doxygen ver 1.8.18 at least

I don't think we have other but the option mentioned in the docs.  I can add some docs in the future if you create an issue :)
Otherwise, we can chat and I can help

ahsonkhan: > Make sure that `#Udpate()` works without the full path to it like `#Azure::Core::Md5::Update`

It works fine:
image

sdk/core/azure-core/inc/azure/core/md5.hpp
ahsonkhan: Fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/1647

PR #1646, opened 2021-02-09
Renamed `Crc64` to `Crc64Hash` and change it to derive from the `Azure::Core::Cryptography::Hash` class.

ahsonkhan:
Follow-up from https://github.com/Azure/azure-sdk-for-cpp/pull/1632

Also partially addresses https://github.com/Azure/azure-sdk-for-cpp/issues/1619 for Crc64 as well, similar to Md5 (topic 1 and 2).

I will leave it up to @Jinming-Hu and storage folks to help decide whether they want this change in this release, or wait for the next one.

### DIFF
--- a/sdk/storage/azure-storage-blobs/test/page_blob_client_test.cpp
+++ b/sdk/storage/azure-storage-blobs/test/page_blob_client_test.cpp
@@ -280,7 +280,11 @@ namespace Azure { namespace Storage { namespace Test {
     Blobs::UploadPageBlobPagesOptions options;
     ContentHash hash;
     hash.Algorithm = HashAlgorithm::Crc64;
-    hash.Value = Crc64::Hash(blobContent.data(), blobContent.size());
+
+    {
+      Crc64Hash instance;
+      hash.Value = instance.Final(blobContent.data(), blobContent.size());
+    }
     options.TransactionalContentHash = hash;
     EXPECT_NO_THROW(pageBlobClient.UploadPages(0, &pageContent, options));
 
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
@@ -16,6 +16,8 @@ namespace Azure { namespace Storage {
   public:
     void Concatenate(const Crc64Hash& other);
 
+    ~Crc64Hash() override = default;
+
   private:
     uint64_t m_context = 0ULL;
     uint64_t m_length = 0ULL;
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
@@ -7,32 +7,21 @@
 #include <string>
 #include <vector>
 
+#include "azure/core/cryptography/hash.hpp"
 #include <azure/core/base64.hpp>
 
 namespace Azure { namespace Storage {
 
-  class Crc64 {
+  class Crc64Hash : public Azure::Core::Cryptography::Hash {
   public:
-    void Update(const uint8_t* data, std::size_t length);
-    void Concatenate(const Crc64& other);
-
-    std::vector<uint8_t> Digest() const;
-
-    static std::vector<uint8_t> Hash(const uint8_t* data, std::size_t length)
-    {
-      Crc64 instance;
-      instance.Update(data, length);
-      return instance.Digest();
-    }
-
-    static std::vector<uint8_t> Hash(const std::string& data)
-    {
-      return Hash(reinterpret_cast<const uint8_t*>(data.data()), data.length());
-    }
+    void Concatenate(const Crc64Hash& other);
 
   private:
     uint64_t m_context = 0ULL;
     uint64_t m_length = 0ULL;
+
+    void OnAppend(const uint8_t* data, std::size_t length) override;
+    std::vector<uint8_t> OnFinal(const uint8_t* data, std::size_t length) override;
   };
 
   namespace Details {
--- a/sdk/storage/azure-storage-blobs/test/block_blob_client_test.cpp
+++ b/sdk/storage/azure-storage-blobs/test/block_blob_client_test.cpp
@@ -131,9 +131,12 @@ namespace Azure { namespace Storage { namespace Test {
     res = m_blockBlobClient->Download(options);
     ASSERT_TRUE(res->TransactionalContentHash.HasValue());
     EXPECT_EQ(res->TransactionalContentHash.GetValue().Algorithm, HashAlgorithm::Crc64);
-    EXPECT_EQ(
-        res->TransactionalContentHash.GetValue().Value,
-        Crc64::Hash(m_blobContent.data(), downloadLength));
+    {
+      Crc64Hash instance;
+      EXPECT_EQ(
+          res->TransactionalContentHash.GetValue().Value,
+          instance.Final(m_blobContent.data(), downloadLength));
+    }
   }
 
   TEST_F(BlockBlobClientTest, DISABLED_LastAccessTime)
--- a/sdk/storage/azure-storage-common/CHANGELOG.md
+++ b/sdk/storage/azure-storage-common/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Breaking Changes
 
-- Remove the `Azure::Storage::Md5` class from `crypt.hpp`. Use the type from `Azure::Core::Cryptography` namespace instead, from `azure/core/cryptography/hash.hpp`.
+- Removed the `Azure::Storage::Md5` class from `crypt.hpp`. Use the type from `Azure::Core::Cryptography` namespace instead, from `azure/core/cryptography/hash.hpp`.
+- Renamed `Crc64` to `Crc64Hash` and change it to derive from the `Azure::Core::Cryptography::Hash` class.
 
 ## 12.0.0-beta.7 (2021-02-03)
 
--- a/sdk/storage/azure-storage-common/src/crypt.cpp
+++ b/sdk/storage/azure-storage-common/src/crypt.cpp
@@ -910,7 +910,7 @@ namespace Azure { namespace Storage {
     return vr[0] ^ vr[1];
   }
 
-  void Crc64::Update(const uint8_t* data, std::size_t length)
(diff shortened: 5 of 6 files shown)


sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
```diff
@@ -7,32 +7,21 @@
 #include <string>
 #include <vector>
 
+#include "azure/core/cryptography/hash.hpp"
 #include <azure/core/base64.hpp>
 
 namespace Azure { namespace Storage {
```
ahsonkhan, line 13: @JeffreyRichter, @RickWinter should this be in a Cryptography sub-namespace, i.e. `Azure::Storage::Cryptography`?

RickWinter: Yes, I suggested Crypto b/c that is what some other languages use.  I'm open to any good name.  I suspect the set grows and thus my push for them to be in a namespace.

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
```diff
@@ -7,32 +7,21 @@
...
 
-  class Crc64 {
+  class Crc64Hash : public Azure::Core::Cryptography::Hash {
```
JeffreyRichter, line 15: I think Jinming has a reason why this class will NOT derive from Core's Hash.

JeffreyRichter (approved):

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
ahsonkhan: >  If it's just because it needs Concatenate but it ALSO works/is useful without Concatenate, then deriving seems good to me.

I think that was his reasoning, but having a concrete class add additional specific APIs is fine, from https://github.com/Azure/azure-sdk-for-cpp/pull/1632#issuecomment-775643215, unless there are other reasons.
> From @Jinming-Hu:
> Plus, I'm not going to make Crc64 class inherit from this hash base. As Crc64 supports concatenate operation, while your base class doesn't.

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
ahsonkhan: Right, @RickWinter, that makes sense for `Azure::Core::Cryptography` and the base `Hash` class (which we already did). Does that apply to services that derive from the base as well, such as Crc64 in Storage?

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
ahsonkhan: At least one benefit for deriving like this, is that we get consistent and well-defined/documented behavior for the edge cases for free such as what happens when you call `Final()` multiple times or `Append()` after a call to `Final()`. The base class enforces the right behavior.

Jinming-Hu (changes requested):
I'm gonna take a look today, wait for storage's approval

ahsonkhan:
> I'm gonna take a look today, wait for storage's approval

I wouldn't merge without it 😊

sdk/storage/azure-storage-blobs/test/page_blob_client_test.cpp
```diff
@@ -280,7 +280,11 @@ namespace Azure { namespace Storage { namespace Test {
     Blobs::UploadPageBlobPagesOptions options;
     ContentHash hash;
     hash.Algorithm = HashAlgorithm::Crc64;
-    hash.Value = Crc64::Hash(blobContent.data(), blobContent.size());
+
+    {
+      Crc64Hash instance;
+      hash.Value = instance.Final(blobContent.data(), blobContent.size());
```
Jinming-Hu, line 286: Nit: can we add a static method `Hash`?

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
```diff
@@ -7,32 +7,21 @@
...
-    {
-      Crc64 instance;
-      instance.Update(data, length);
-      return instance.Digest();
-    }
-
-    static std::vector<uint8_t> Hash(const std::string& data)
-    {
-      return Hash(reinterpret_cast<const uint8_t*>(data.data()), data.length());
-    }
```
Jinming-Hu, line 31: Can we keep this?

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
```diff
@@ -7,32 +7,21 @@
...
+    void Concatenate(const Crc64Hash& other);
```
Jinming-Hu, line 18: need override destructor

ahsonkhan: What will we do in the destructor? It wasn't there in the previous implementation.

sdk/storage/azure-storage-blobs/test/page_blob_client_test.cpp
ahsonkhan: We can add it back later if we get customer feedback, and @JeffreyRichter recommended I remove it for now (same with `Md5Hash`).

How commonly do you think customers would call the `single-shot` hashing API, rather than the streaming one?

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
ahsonkhan: Will folks typically try to compute Crc64 hash for text string data? Or is it similar to other hashing (like md5) which is meant for files/blobs that tend to binary/bytes?

These are convenience methods which we can easily add later, and so can the end-customer if they really want to.

It's not too bad atm:
```C++
CrcHash instance;
auto hash = instance.Final(data, length);
```
versus
```C++
auto hash = CrcHash::Hash(data, length);
```

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
Jinming-Hu: nothing, just 

```C++
Crc64Hash() override = default;
```

to enforce the base class has a virtual destructor.

ahsonkhan: Seems odd to do. The base class already has a virtual destructor, why would the concrete derived class need to enforce such a thing?

Jinming-Hu: Let's say some customer cares a lot about data integrity and always wants to use Md5 or Crc64 or whatever checksum servicer supports. He also wants to be able to switch different hash algorithm with an option, possible value of which can be MD5, CRC64, and others. To avoid duplicate logic, he will use a pointer to the hash base class `HashBase*` and instantiate a concrete hash instance `Md5Hash` or `Crc64Hash` and assign the point to `HashBase*`. Now he can just call `HashBase->Append` to append data and `HashBase->Final` to get final hash value, without caring about the hash algorithm. This helps a lot in reducing code redundancy. Once the operation is done, he will deallocate the hash instance by `delete HashBase`. If the destructor of `HashBase` is not virtual, it will cause potential resource leak because the destructor of concrete hash class isn't executed. Now to avoid this, we will use `override` on the destructor of concrete class. It's always a good habit if some object is possible to be deleted via base pointer.

> The base class already has a virtual destructor, why would the concrete derived class need to enforce such a thing? What's the benefit?

same question goes for [OnFinal](https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp#L136) and [OnAppend](https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp#L146). We don't need `override` to be grammatically correct.

ahsonkhan: > If the destructor of `HashBase` is not virtual, it will cause potential resource leak because the destructor of concrete hash class isn't executed.

But it is virtual, so why safeguard against something that is already true?

> same question goes for [OnFinal](https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp#L136) and [OnAppend](https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp#L146). We don't need `override` to be grammatically correct. I just think it's a good habit

But in this case, we are actually overriding the behavior/logic of those functions.

The `HashBase` already has an implementation of the virtual destructor as `default`, and if `Crc64Hash` is not changing the implementation or doing something meaningful, why override it prematurely? I agree that **IF** the destructor needs to be overridden, we'd use the `override` keyword. I am asking what's the benefit of overriding it today?

All that said, I am fine with adding it.

Jinming-Hu: > But in this case, we are actually overriding the behavior/logic of those functions.

Without `override`, you're also overriding the behavior.

> The HashBase already has an implementation of the virtual destructor as default, and if Crc64Hash is not changing the implementation or doing something meaningful, why override it prematurely?

As you say, it's just a safeguard. Who's to say nobody will change the impl/interface of `HashBase` and forget to make the destructor virtual. If we use `override` in derived destructor, the compiler can remind us.

ahsonkhan: > Without `override`, you're also overriding the behavior.

But currently I am not providing a destructor for `Crc64Hash` at all. How am I overriding anything? I understand **IF** I did provide one, the `override` keyword is helpful, but if I don't add a destructor at all, what's the issue?

Are you suggesting I **add** a destructor?

ahsonkhan:
Looking for PR approval here to avoid making this storage breaking change after the Feb release.

Jinming-Hu (approved):

PR #1647, opened 2021-02-09
Address implementation level feedback for Md5Hash and delete ctors.

ahsonkhan:
Addressing feedback that came up in https://github.com/Azure/azure-sdk-for-cpp/pull/1632#discussion_r572724322 after merge.


sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -125,6 +125,11 @@ namespace Azure { namespace Core { namespace Cryptography {
   private:
     void* m_md5Context;
 
+    // Delete the copy and move constructors, along with the assignment operator.
+    Md5Hash(Md5Hash const&) = delete;
+    Md5Hash(Md5Hash const&&) = delete;
+    void operator=(Md5Hash const&) = delete;
```
ahsonkhan, line 131: @CaseyCarter I would like to understand better the implications of explicitly deleting such ctors/operators, and when I need to use `const&&` vs just `&&`.

Also, are these "deletions" typically private?

CaseyCarter: IIRC https://www.youtube.com/watch?v=vLinb2fgkHk&feature=emb_logo&ab_channel=InsideBloomberg is a good overview of move semantics and how explicit deletions suppress various of the special member functions (SMFs).

Deleted SMFs are typically not private, but with deletions it makes no difference. ("Code outside this class may not call this function which is uncallable in any case.")

`Md5Hash const&&` is a slightly more permissive parameter type than `Md5Hash &&` since the former can bind to `const Md5Hash` rvalues which the latter cannot. That fact also makes no appreciable difference in this case: if the move constructor parameter were plain `Md5Hash&&`, calls with such arguments invoke the deleted copy constructor instead of the deleted move constructor.

It's peculiar that this class explicitly deletes its move constructor but allows the move assignment operator to be deleted implicitly (since there's a user-declared copy assignment operator, the move assignment operator is implicitly deleted.) I'd expect to see either the maximally explicit "delete all four move/copy constructors/assignment operators" or the more terse "delete the copy operations and let the moves be implicitly deleted."

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -97,12 +97,12 @@ namespace Azure { namespace Core { namespace Cryptography {
     std::vector<uint8_t> Final() { return Final(nullptr, 0); };
 
     /**
-     * @brief Cleanup any state when destroying the instance of @Hash.
+     * @brief Cleanup any state when destroying the instance of #Azure::Core::Cryptography::Hash.
      */
     virtual ~Hash(){};
```
CaseyCarter, line 102:
```suggestion
    virtual ~Hash() = default;
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -113,18 +113,23 @@ namespace Azure { namespace Core { namespace Cryptography {
...
-     * @brief Construct a default instance of @Md5Hash.
+     * @brief Construct a default instance of #Azure::Core::Cryptography::Md5Hash.
      */
     explicit Md5Hash();
 
     /**
-     * @brief Cleanup any state when destroying the instance of @Md5Hash.
+     * @brief Cleanup any state when destroying the instance of #Azure::Core::Cryptography::Md5Hash.
      */
     ~Md5Hash();
```
CaseyCarter, line 123:
```suggestion
    ~Md5Hash() override;
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -97,12 +97,12 @@ namespace Azure { namespace Core { namespace Cryptography {
     std::vector<uint8_t> Final() { return Final(nullptr, 0); };
```
CaseyCarter, line 97: We included `<cstdint>`, which is guaranteed to declare `std::uint8_t` and _may_possibly_ declare `::uint8_t`, as opposed to `<stdint.h>` which _may_possibly_ declare `std::uint8_t` and is guaranteed to declare `::uint8_t`. To be consistent, we should either `std::`-qualify every `uint8_t` in this file, or change the include from `<cstdint>` to `<stdint.h>`.

CaseyCarter, line 97:
```suggestion
    std::vector<uint8_t> Final() { return Final(nullptr, 0); }
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
ahsonkhan: >  I'd expect to see either the maximally explicit "delete all four move/copy constructors/assignment operators" or the more terse "delete the copy operations and let the moves be implicitly deleted."

So, if I am understanding this correctly, are you suggesting to apply one of these two options:
1) Either add `void operator=(Md5Hash const&&) = delete;` OR
2) Remove `Md5Hash(Md5Hash const&&) = delete;`

Also, since this class has a destructor, I can't follow the rule of 0, but I don't know how to follow the rule of 3 here:
> A copy constructor, a copy assignment operator, no move constructor, no move assignment operator, and a destructor.

Should I be adding a copy ctor/assignment instead of deleting it to adhere to the guideline?

CaseyCarter: > > I'd expect to see either the maximally explicit "delete all four move/copy constructors/assignment operators" or the more terse "delete the copy operations and let the moves be implicitly deleted."
> 
> So, if I am understanding this correctly, are you suggesting to apply one of these two options:
> 
> 1. Either add `void operator=(Md5Hash const&&) = delete;` OR
> 2. Remove `Md5Hash(Md5Hash const&&) = delete;`

Yes, that's the general idea. I'd prefer the removal, since `void operator=` (vs. `Md5Hash& operator=`) and `(meow const&&)` (vs. `meow&&`) are both weird enough to make someone scrutinize the code to figure out why they're written that way.

> Also, since this class has a destructor, I can't follow the rule of 0, but I don't know how to follow the rule of 3 here:
> 
> > A copy constructor, a copy assignment operator, no move constructor, no move assignment operator, and a destructor.
> 
> Should I be adding a copy ctor/assignment instead of deleting it to adhere to the guideline?

I think the guideline's "should implement" is intended to admit `= default` and `= delete` as "implementation". In this case, we'd follow the rule of 3 by defining a destructor, and defining deleted copy constructor and copy assignment operations. Actually, having read through most of the file now I think it would make more sense to delete `Hash`'s copies so all classes that derive from `Hash` are non-copyable by default.

ahsonkhan: > I think the guideline's "should implement" is intended to admit `= default` and `= delete` as "implementation". In this case, we'd follow the rule of 3 by defining a destructor, and defining deleted copy constructor and copy assignment operations. 

That is good to know. @RickWinter we should add this as a clarification to the guidelines, and include an example so folks reading it in the future can benefit from it as well.

> Actually, having read through most of the file now I think it would make more sense to delete `Hash`'s copies so all classes that derive from `Hash` are non-copyable by default.

👍

Jinming-Hu: why not just put the `std::string` and handle in primate members?

Of course we still need these overload/definition of constructor and assignment operator.

ahsonkhan: I don't follow what you mean @Jinming-Hu .

Jinming-Hu: instead of 

```
private:
    void* m_md5Context;
```

```
private:
    std::string buffer;
    BCRYPT_HASH_HANDLE hashHandle;
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -113,18 +113,23 @@ namespace Azure { namespace Core { namespace Cryptography {
...
      */
     ~Md5Hash();
 
   private:
     void* m_md5Context;
 
+    // Delete the copy and move constructors, along with the assignment operator.
+    Md5Hash(Md5Hash const&) = delete;
+    Md5Hash(Md5Hash const&&) = delete;
+    void operator=(Md5Hash const&) = delete;
```
Jinming-Hu, line 131:
```suggestion
    Md5Hash(Md5Hash const&) = delete;
    Md5Hash(Md5Hash&&) = delete;
    Md5Hash& operator=(Md5Hash const&) = delete;
    Md5Hash& operator=(Md5Hash&&) = delete;
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
ahsonkhan: I would like to understand these subtle differences better. Why do we want to do this over the existing?

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
Jinming-Hu, line 123:
```suggestion
    ~Md5Hash() override;
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -113,18 +113,23 @@ namespace Azure { namespace Core { namespace Cryptography {
 
   public:
     /**
-     * @brief Construct a default instance of @Md5Hash.
+     * @brief Construct a default instance of #Azure::Core::Cryptography::Md5Hash.
      */
     explicit Md5Hash();
```
Jinming-Hu, line 118:
```suggestion
    Md5Hash();
```

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
ahsonkhan: See discussion above, which suggests removing one rather than adding one: https://github.com/Azure/azure-sdk-for-cpp/pull/1647#discussion_r573231549

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
ahsonkhan: FWIW, we use `<cstdint>` everywhere else in the SDK without using `std::uint8_t`. Do you recommend we change all such instances? Is this something we should put in guidelines?

For example: 
https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/core/azure-core/inc/azure/core/http/body_stream.hpp#L28
https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/core/azure-core/inc/azure/core/http/body_stream.hpp#L87

https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp#L7
https://github.com/Azure/azure-sdk-for-cpp/blob/2d4d766c6d454bfbd2a3627c5a3d8a4d81196480/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp#L114

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
ahsonkhan: Because the member types would be different based on the OS (for example, it's `MD5_CTX` on non-Windows), and I didn't really want to expose bcrypt.h/openssl in the public header.

ahsonkhan: > Actually, having read through most of the file now I think it would make more sense to delete `Hash`'s copies so all classes that derive from `Hash` are non-copyable by default.

@CaseyCarter, would I need to define a default `Hash()` ctor too? If I just have those methods deleted, I get an error on the ctor of the derived type, i.e. `Md5Hash`:
image

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -113,18 +113,22 @@ namespace Azure { namespace Core { namespace Cryptography {
...
      */
-    ~Md5Hash();
+    ~Md5Hash() override;
 
   private:
     void* m_md5Context;
 
+    // Delete the copy constructor, along with the assignment operator.
+    Md5Hash(Md5Hash const&) = delete;
+    void operator=(Md5Hash const&) = delete;
```
vhvb1989, line 130: @CaseyCarter  since the destructor is user declared for this class, do we still need to delete the copy constructors?

image

Isn't it deprecated to be implicitly defaulted?   Just curious about this one here.

vhvb1989 (approved):

Jinming-Hu (approved):

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
CaseyCarter: (1) The `;` in `virtual ~Hash(){};` is extraneous noise.
(2) `= default` is "better" than `{}` in general since it sometimes (though not in this case) produces trivial special members which have performance advantages over non-trivial members. (`= default` is also a nice "this space intentionally left blank" making it explicit that someone didn't forget to fill in the `{}`.) I prefer to teach using `= default` even in cases like this one where there's no technical reason to get folks' fingers in the habit so they do the right thing when it does matter.

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
CaseyCarter: Yes; if we declare any kind of constructor the default constructor is not implicitly declared. So once we declare the deleted copy constructor in `Hash`, we'll need to define the default constructor explicitly (`Hash() = default;`).

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
CaseyCarter: If we apply my suggestion to delete `Hash`'s copies, then we won't need to (explicitly) delete them in `Md5Hash` since they'll be implicitly deleted by virtue of `Md5Hash`'s base class `Hash` having inaccessible-or-deleted copies.

In the status quo, I think they should be explicitly deleted because this "deprecation" is really just WG21 acknowledging that they got the defaults wrong in C++98. The corresponding behavior is unlikely to ever removed (or result in a non-off-by-default warning) due to the enormous amount of code it would break.

PR #1659, opened 2021-02-10
Add Operation<T>, OperationStatus, Exception, and Platform headers to the simplified core header.

ahsonkhan:


vhvb1989 (approved):
Changelog?  This might be a bugfix since those headers were missing.

ahsonkhan:
Not really a bug worth reporting in CL since the customer generally won't be impacted. This is more of a convenience.

RickWinter (approved):

PR #1682, opened 2021-02-13
Remove copyright headers from produced VcPkg port files

### LINKED ISSUE #1664: Automate trimming out the copyright/license headers from vcpkg control, potfile when submitting a PR
https://github.com/Azure/azure-sdk-for-cpp/blob/fcd64f7bd2b983148c0b5099afa8edf3ba05091c/sdk/core/azure-core/vcpkg/CONTROL#L1-L4

Example manual effort: https://github.com/microsoft/vcpkg/pull/16160

cc @antkmsft, @danieljurek

antkmsft:
Closes #1664

### DIFF
--- a/cmake-modules/AzureVcpkg.cmake
+++ b/cmake-modules/AzureVcpkg.cmake
@@ -24,7 +24,10 @@ macro(az_vcpkg_portfile_prep targetName fileName contentToRemove)
 
   # Windows -> Unix line endings
   string(FIND fileContents "\r\n" crLfPos)
-  string(REPLACE "\r\n" "\n" fileContents ${fileContents})
+
+  if (crLfPos GREATER -1)
+    string(REPLACE "\r\n" "\n" fileContents ${fileContents})
+  endif()
 
   # remove comment header
   string(REPLACE "${contentToRemove}" "" fileContents ${fileContents})
@@ -51,6 +54,10 @@ macro(az_vcpkg_portfile_prep targetName fileName contentToRemove)
 endmacro()
 
 macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
+  # CONTROL file has an extra '#' in the comment section, because VcPkg can't handle an empty line at that position,
+  # and without that extra '#' line, the file contents look too crowded.
+  # Ultimately, the lines passed to az_vcpkg_portfile_prep() have to match the header comment that is actually
+  # present in the files, or otherwise nothing will be removed.
   az_vcpkg_portfile_prep(
     "${targetName}"
     "CONTROL"
--- a/cmake-modules/AzureVcpkg.cmake
+++ b/cmake-modules/AzureVcpkg.cmake
@@ -18,13 +18,50 @@ macro(az_vcpkg_integrate)
   endif()
 endmacro()
 
-macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
+macro(az_vcpkg_portfile_prep targetName fileName contentToRemove)
+  # with sdk/<lib>/vcpkg/<fileName>
+  file(READ "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/${fileName}" fileContents)
+
+  # Windows -> Unix line endings
+  string(FIND fileContents "\r\n" crLfPos)
+  string(REPLACE "\r\n" "\n" fileContents ${fileContents})
+
+  # remove comment header
+  string(REPLACE "${contentToRemove}" "" fileContents ${fileContents})
+
+  # undo Windows -> Unix line endings (if applicable)
+  if (crLfPos GREATER -1)
+    string(REPLACE "\n" "\r\n" fileContents ${fileContents})
+  endif()
+  unset(crLfPos)
+
+  # output to an intermediate location
+  file (WRITE "${CMAKE_BINARY_DIR}/vcpkg_prep/${targetName}/${fileName}" ${fileContents})
+  unset(fileContents)
+
   # Produce the files to help with the VcPkg release.
   # Go to the /out/build/<cfg>/vcpkg directory, and copy (merge) "ports" folder to the vcpkg repo.
   # Then, update portfile.cmake's SHA512 from "1" to the actual hash (a good way to do it is to uninstall a package,
   # clean vcpkg/downloads, vcpkg/buildtrees, run "vcpkg install <pkg>", and get the SHA from the error message).
-  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/CONTROL" "${CMAKE_BINARY_DIR}/vcpkg/ports/${targetName}-cpp/CONTROL" @ONLY)
-  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/portfile.cmake" "${CMAKE_BINARY_DIR}/vcpkg/ports/${targetName}-cpp/portfile.cmake" @ONLY)
+  configure_file(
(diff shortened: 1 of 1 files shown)


cmake-modules/AzureVcpkg.cmake
```diff
@@ -18,13 +18,50 @@ macro(az_vcpkg_integrate)
   endif()
 endmacro()
 
-macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
+macro(az_vcpkg_portfile_prep targetName fileName contentToRemove)
```
ahsonkhan, line 21: Will this step happen **before** the sha512 is calculated?

cmake-modules/AzureVcpkg.cmake
```diff
@@ -18,13 +18,50 @@ macro(az_vcpkg_integrate)
...
+  string(REPLACE "\r\n" "\n" fileContents ${fileContents})
+
+  # remove comment header
+  string(REPLACE "${contentToRemove}" "" fileContents ${fileContents})
+
+  # undo Windows -> Unix line endings (if applicable)
+  if (crLfPos GREATER -1)
+    string(REPLACE "\n" "\r\n" fileContents ${fileContents})
+  endif()
+  unset(crLfPos)
```
ahsonkhan, line 36: Why is this line ending normalizing/reverting necessary?

ahsonkhan: This trimming logic is a bit more complex than I initially expected. I am wondering if its better to just remove the copyright headers from the checked-in file OR maybe leave it as is and ask vcpkg folks for an exception.

This PR already got merged with the copyright headers: https://github.com/microsoft/vcpkg/pull/16183

cmake-modules/AzureVcpkg.cmake
```diff
@@ -18,13 +18,50 @@ macro(az_vcpkg_integrate)
...
+    "${CMAKE_BINARY_DIR}/vcpkg/ports/${targetName}-cpp/${fileName}"
+    @ONLY
+  )
+endmacro()
+
+macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
+  az_vcpkg_portfile_prep(
+    "${targetName}"
+    "CONTROL"
+    "# Copyright (c) Microsoft Corporation. All rights reserved.\n# SPDX-License-Identifier: MIT\n#\n"
```
ahsonkhan, line 57: Why is the last `\n` on a new line here (preceded by `#`) but not below for the portfile?

cmake-modules/AzureVcpkg.cmake
antkmsft: Yes, it is before.
So, our build process produces `out/<buildCfg>/vcpkg/ports/<libname>`. You are free to grab directories from there and put them to your local copy of vcpkg repo to `%VCPKG_ROOT%/ports/`, play with it locally, or even create a PR to VcPkg manually, but you need to update SHA first.
Daniel's automation grabs these files, replaces "SHA 1" with the actual SHA of the release, renerates version SHAs, and pushes the PR to VcPkg repo.
And this is why I think it is the right division of labor between us and EngSys, when it is us who drops the copyrights, because both files live in our repo, and should we update copyring, we would update the script that updates them too.
But everythign related to build release process itself is CentralEngSys, and it even lives in their repo.

cmake-modules/AzureVcpkg.cmake
antkmsft: We find a 3-line string and replace it with ann empty one. Without it, find and replace would fail when Windows line endings are used (git checks out with Windows line ending on Windows, Unix endings on Max and Linux).

cmake-modules/AzureVcpkg.cmake
antkmsft: See any CONTROL file - `#` is there as 3rd line. And it is not removable, in the terms that you cannot replace it with just empty string - VcPkg fails. What we can do is to get rid of that line entirely, but then it becomes too cluttered.

antkmsft: See this file, for example - https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/vcpkg/CONTROL

ahsonkhan: Ah yes, just saw that. The checked-in files have that discrepancy. Thanks for clarifying why its there.

cmake-modules/AzureVcpkg.cmake
antkmsft: Otherwise, we would have to search for `"# Copyright (c) Microsoft Corporation. All rights reserved.\n# SPDX-License-Identifier: MIT\n\n"` (as today) on Linu/Mac, and for `"# Copyright (c) Microsoft Corporation. All rights reserved.\r\n# SPDX-License-Identifier: MIT\r\n\r\n"` on Windows.

cmake-modules/AzureVcpkg.cmake
antkmsft: I think it is ok now. In VcPkg, no other package has these headers.
In our repo, we have them for every file.
This logic removes the exact comment, or does nothing. I deliberately not using any smarter than necessary logic here - not using regexes, but not removing "just first 3 lines" either. In case you don't have any copyright header, nothing happens. In case you have something else in the header, nothing happens, it goes to PR. And in case when it is this exact header, it gets removed. I like it.

RickWinter (approved):
Add comments to this file notating the special behaviors with #.  

Nit, Moving forward we need better output from our CMake code as it performs steps.  It will help diagnose failures if we are outputting status as things progress.

cmake-modules/AzureVcpkg.cmake
```diff
@@ -18,13 +18,50 @@ macro(az_vcpkg_integrate)
...
 endmacro()
 
-macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
+macro(az_vcpkg_portfile_prep targetName fileName contentToRemove)
+  # with sdk/<lib>/vcpkg/<fileName>
+  file(READ "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/${fileName}" fileContents)
+
+  # Windows -> Unix line endings
+  string(FIND fileContents "\r\n" crLfPos)
+  string(REPLACE "\r\n" "\n" fileContents ${fileContents})
```
ahsonkhan, line 27: Put this replace under an `if (crLfPos GREATER -1)` block just like below. If we don't find any "\r\n", we don't need to do a replace.

cmake-modules/AzureVcpkg.cmake
```diff
@@ -18,13 +18,50 @@ macro(az_vcpkg_integrate)
...
+  string(REPLACE "${contentToRemove}" "" fileContents ${fileContents})
+
+  # undo Windows -> Unix line endings (if applicable)
+  if (crLfPos GREATER -1)
+    string(REPLACE "\n" "\r\n" fileContents ${fileContents})
+  endif()
+  unset(crLfPos)
+
+  # output to an intermediate location
+  file (WRITE "${CMAKE_BINARY_DIR}/vcpkg_prep/${targetName}/${fileName}" ${fileContents})
```
ahsonkhan, line 39: Any particular reason to pick this file path as the intermediary?

ahsonkhan (approved):
Other than that, looks good.

cmake-modules/AzureVcpkg.cmake
antkmsft: no specific reason. The normal output is under `/vcpkg`, this intermediate output is at `/vcpkg_prep`.

PR #1711, opened 2021-02-20
Exceptions should be caught as const.

RickWinter:
Add a holder for the registry key to ensure it is always cleaned up when it goes out of scope.

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
+++ b/sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
@@ -3,7 +3,7 @@
 
 /**
  * @file
- * @brief Internal utility functions for strings.
+ * @brief Internal HKEY holder.
  *
  */
 #pragma once
@@ -20,34 +20,37 @@
 
 namespace Azure { namespace Core { namespace Internal {
 
-  class HKeyHolder {
+  /**
+   * @brief HKEYHolder ensures native handle resource is released.
+  */
+  class HKEYHolder {
   private:
-    HKEY m_value;
+    HKEY m_value = nullptr;
 
   public:
-    HKeyHolder() : m_value(NULL) {}
+    explicit HKEYHolder() noexcept : m_value(nullptr){};
 
-    ~HKeyHolder()
+    ~HKEYHolder() noexcept
     {
-      if (m_value != NULL)
+      if (m_value != nullptr)
       {
         ::RegCloseKey(m_value);
       }
     }
 
-    void operator=(HKEY p)
+    void operator=(HKEY p) noexcept
     {
-      if (p != NULL)
+      if (p != nullptr)
       {
         m_value = p;
       }
     }
 
-    operator HKEY() { return m_value; }
+    operator HKEY() noexcept { return m_value; }
 
-    operator HKEY*() { return &m_value; }
+    operator HKEY*() noexcept { return &m_value; }
 
-    HKEY* operator&() { return &m_value; }
+    HKEY* operator&() noexcept { return &m_value; }
   };
 
 }}} // namespace Azure::Core::Internal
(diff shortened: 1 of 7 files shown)


ahsonkhan:
> Exceptions should be caught as const.

Worth adding this to the guidelines?

ahsonkhan:
> Add a holder for the registry key to ensure it is always cleaned up when it goes out of scope.

This seems like the more appropriate PR title, given the extent of the changes :)

vhvb1989 (approved):
Add changelog

sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
```diff
@@ -0,0 +1,55 @@
...
+      {
+        m_value = p;
+      }
+    }
+
+    operator HKEY() { return m_value; }
+
+    operator HKEY*() { return &m_value; }
+
+    HKEY* operator&() { return &m_value; }
```
ahsonkhan, line 50: Do we need all these? I don't see where they are being called. Can we one method called `GetHKey()` or `GetValue() that returns the `HKEY`?

sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
```diff
@@ -0,0 +1,55 @@
...
+      {
+        ::RegCloseKey(m_value);
+      }
+    }
+
+    void operator=(HKEY p)
+    {
+      if (p != NULL)
+      {
+        m_value = p;
```
ahsonkhan, line 42: Should we close the previous `m_value` before overriding it during assignment?

Otherwise, what happens with this:
```C++
{
  Azure::Core::Internal::HKeyHolder key1 = <some registry key 1>;
  Azure::Core::Internal::HKeyHolder key2 = <some registry key 2>;

  key2 = key1; // Who will now close <some registry key 2>
}
// When we leave out of scope, the destructor for key2 gets called, which
// closes <some registry key 1>
```

Jinming-Hu:
> > Exceptions should be caught as const.
> 
> Worth adding this to the guidelines?

This is more of a advisory thing, rather than mandatory, right? Because sometimes we need to change the exception, so it cannot be caught by const.

RickWinter:
> > > Exceptions should be caught as const.
> > 
> > 
> > Worth adding this to the guidelines?
> 
> This is more of a advisory thing, rather than mandatory, right? Because sometimes we need to change the exception, so it cannot be caught by const.

There is a performance benefit to catching const so it is desired.  It would be desired if we always caught const but there are some cases where exceptions are modified with addtional details and rethrown.  I assume your scenario falls into the latter.  @Jinming-Hu  can you add a link to your specific scenario, it will help me when I add the "should catch const" to the guidelines.

sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
RickWinter: These are used by the methods that access the HKEY.  They allow you to interact with the key as if its an HKEY.
For example this usage
```
    Azure::Core::Internal::HKEYHolder regKey;
    if (RegOpenKeyExA(
            HKEY_LOCAL_MACHINE,
            "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion",
            0,
            KEY_READ,
            &regKey)
```
The last parameter is an HKEY.  The methods above allow us to pass an HKEYHolder ref to that function because the operators allow the cast to HKEY to do the right thing.

sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
RickWinter: We could enhance this to not allow assignment or we could make the assignment close the handle.  I can look at that in a future change

PR #1719, opened 2021-02-22
Defend public headers against inclusion of Windows.h

CaseyCarter:
... which defines some nasty function-like macros `min` and `max` when `_NOMINMAX` isn't defined. We prevent expansion as a function-like macro by inserting some token(s) between `min`/`max` and the following `(`. Most commonly that means wrapping the entire qualified-name in `()` a la `(std::min)(x, y)`, but an explicit template argument list (`std::min<int>(x, y)`) works as well.

EDIT: To be perfectly clear, I'm not suggesting introduction of an explicit template argument list as a good way to fix this issue so much as pointing out that existing occurrences with an explicit template argument list - there are a couple in the repo - don't need any additional fix.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -272,7 +272,9 @@ namespace Azure { namespace Core {
         return time_point() + static_cast<std::chrono::milliseconds>(msec);
       }
 
-      explicit ContextSharedState() : CancelAtMsecSinceEpoch(ToMsecSinceEpoch((time_point::max)())) {}
+      explicit ContextSharedState() : CancelAtMsecSinceEpoch(ToMsecSinceEpoch((time_point::max)()))
+      {
+      }
 
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -272,7 +272,7 @@ namespace Azure { namespace Core {
         return time_point() + static_cast<std::chrono::milliseconds>(msec);
       }
 
-      explicit ContextSharedState() : CancelAtMsecSinceEpoch(ToMsecSinceEpoch(time_point::max())) {}
+      explicit ContextSharedState() : CancelAtMsecSinceEpoch(ToMsecSinceEpoch((time_point::max)())) {}
 
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
@@ -330,7 +330,7 @@ namespace Azure { namespace Core {
     Context WithValue(const std::string& key, ContextValue&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
-          m_contextSharedState, time_point::max(), key, std::move(value))};
+          m_contextSharedState, (time_point::max)(), key, std::move(value))};
     }
 
     /**
@@ -389,7 +389,7 @@ namespace Azure { namespace Core {
     void Cancel()
     {
       m_contextSharedState->CancelAtMsecSinceEpoch
-          = ContextSharedState::ToMsecSinceEpoch(time_point::min());
+          = ContextSharedState::ToMsecSinceEpoch((time_point::min)());
     }
 
     /**
--- a/sdk/core/azure-core/test/ut/macro_guard.cpp
+++ b/sdk/core/azure-core/test/ut/macro_guard.cpp
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+// Define `min` and `max` as function-like macros before including all public
+// headers to ensure that uses of those identifiers are defended against
+// expansion as function-like macros. Define `small` as an object-like macro to
+// ensure that identifier isn't used at all. Windows.h is badly behaved and
+// defines similar macros with these names and we want to ensure the SDK headers
+// function even when a naive user includes Windows.h first.
+//
+#define small FAIL><TO][COMPILE)(VERY{{{LOUDLY!!!
+#define max(x, y) small
+#define min(x, y) small
+
+#include <azure/core.hpp>
--- a/sdk/core/azure-core/inc/azure/core/http/body_stream.hpp
+++ b/sdk/core/azure-core/inc/azure/core/http/body_stream.hpp
@@ -263,7 +263,7 @@ namespace Azure { namespace Core { namespace Http {
      * @param max_length Maximum number of bytes to provide to the readers.
      */
     LimitBodyStream(BodyStream* inner, int64_t max_length)
-        : m_inner(inner), m_length(std::min(inner->Length(), max_length))
+        : m_inner(inner), m_length((std::min)(inner->Length(), max_length))
     {
     }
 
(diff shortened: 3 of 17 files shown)


vhvb1989 (approved):
this should be go to guidelines

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -272,7 +272,7 @@ namespace Azure { namespace Core {
         return time_point() + static_cast<std::chrono::milliseconds>(msec);
       }
 
-      explicit ContextSharedState() : CancelAtMsecSinceEpoch(ToMsecSinceEpoch(time_point::max())) {}
+      explicit ContextSharedState() : CancelAtMsecSinceEpoch(ToMsecSinceEpoch((time_point::max)())) {}
```
ahsonkhan, line 275: We should write and check-in a sample that we build in CI which includes `Windows.h` to ensure we stay green going forward.

https://github.com/Azure/azure-sdk-for-cpp/blob/50efbdb4c08ed57e7eedd1bd571dc377d3455c0a/samples/integration/cmake-fetch-content/src/main.cpp#L12

ahsonkhan (approved):

antkmsft (approved):

RickWinter (approved):

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -330,7 +330,7 @@ namespace Azure { namespace Core {
     Context WithValue(const std::string& key, ContextValue&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
```
ahsonkhan, line 332: @CaseyCarter, you run clang format on this file to fix formatting.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=746260&view=logs&j=0585f5d6-0937-5c15-894c-06e189e06847&t=a45688da-a69e-5817-f832-3aa6246959b7&l=24
```text
find ./sdk \( -iname '*.hpp' -o -iname '*.cpp' \) ! -iname 'json.hpp' -exec clang-format -i {} \;

List of files not formatted correctly:
sdk/core/azure-core/inc/azure/core/context.hpp

```

antkmsft:
@CaseyCarter, https://github.com/CaseyCarter/azure-sdk-for-cpp/pull/1

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: Formatting + tests: https://github.com/CaseyCarter/azure-sdk-for-cpp/pull/1

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: Formatting + tests: https://github.com/CaseyCarter/azure-sdk-for-cpp/pull/1

CaseyCarter: I've added a commit with a simple test change for each SDK that just adds a CPP that defines equivalent nasty macros before including the corresponding SDK's public header. Hopefully adding it to `azure-template` will be enough to ensure that future SDKs include coverage as well.

sdk/core/azure-core/inc/azure/core/context.hpp
CaseyCarter: Resolved in e911b2d.

sdk/core/azure-core/test/ut/macro_guard.cpp
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: MIT
+
+// Define `min` and `max` as function-like macros before including all public
+// headers to ensure that uses of those identifiers are defended against
+// expansion as function-like macros. Define `small` as an object-like macro to
+// ensure that identifier isn't used at all. Windows.h is badly behaved and
+// defines similar macros with these names and we want to ensure the SDK headers
+// function even when a naive user includes Windows.h first.
+//
+#define small FAIL><TO][COMPILE)(VERY{{{LOUDLY!!!
```
antkmsft, line 11: Now we can't use `small` as name either, can this be avoided?

CaseyCarter: This is intended: `<Windows.h>` includes `#define small char`.

antkmsft: OMG! I didn't know about that. What were they thinking?

CaseyCarter: > OMG! I didn't know about that. What were they thinking?

¯\_(ツ)_/¯ One of the reasons I'm so militant about all Microsoft projects defending against these macros is that it provides motivation for other teams to complain to Windows. The current state of `<Windows.h>` - defining all-lower-case macros - is user-hostile. `small` in particular cost me nearly a full day of work to learn about; it's not as well-publicized as `min` and `max`.

antkmsft: At least, it is not something like
```
// Happy debugging!
#define this (this && __LINE__ % 10 == 1 ? (this + 1) : this)
```
:)

Jinming-Hu (approved):

CaseyCarter:
Ugh, I've done something terrible to this branch revising history while resolving a merge conflict. I'll fix and force push momentarily.

CaseyCarter:
There we go: I reset to the prior commit and rebased on master. Apologies for resetting incremental review: what's here is much the same as before except that `macro_guard.cpp` moved from `test` into `test/ut` for the SDKs that made that change on master.

RickWinter (approved):
@CaseyCarter Change looks good.  Our team owns the KeyVault portion and has signed off. 
 Please submit

CaseyCarter:
> @CaseyCarter Change looks good. Our team owns the KeyVault portion and has signed off.
> Please submit

I don't have merge rights:

image

Not that I mind, but someone else will have to push the big red button.

antkmsft:
@CaseyCarter, if you join the same orgs that I have on my profile, you should be able to merge, would you like to do that? I think it makes sense. Otherwise, I can merge your PR.

ahsonkhan:
Do we need to make similar fixes to the source cpp files as well?
https://github.com/Azure/azure-sdk-for-cpp/blob/50efbdb4c08ed57e7eedd1bd571dc377d3455c0a/sdk/core/azure-core/src/http/curl/curl.cpp#L690
https://github.com/Azure/azure-sdk-for-cpp/blob/bf347d24c24d8a46d0013e2b4a34197c424ecff2/sdk/core/azure-core/src/http/body_stream.cpp#L76
https://github.com/Azure/azure-sdk-for-cpp/blob/249327a3fa8002a8e25804cd6ebe139cb8612cd3/sdk/storage/azure-storage-blobs/src/blob_client.cpp#L212

CaseyCarter:
> Do we need to make similar fixes to the source cpp files as well?
> https://github.com/Azure/azure-sdk-for-cpp/blob/50efbdb4c08ed57e7eedd1bd571dc377d3455c0a/sdk/core/azure-core/src/http/curl/curl.cpp#L690
> 
> 
> https://github.com/Azure/azure-sdk-for-cpp/blob/bf347d24c24d8a46d0013e2b4a34197c424ecff2/sdk/core/azure-core/src/http/body_stream.cpp#L76
> 
> 
> https://github.com/Azure/azure-sdk-for-cpp/blob/249327a3fa8002a8e25804cd6ebe139cb8612cd3/sdk/storage/azure-storage-blobs/src/blob_client.cpp#L212

Nope: we control source files, and can simply avoid including `<Windows.h>` and/or using problem identifiers and/or build with `_NOMINMAX` defined if we like. The problem only exists for headers which - thanks to textual inclusion - can be affected by headers a user includes before ours. (Modules can't get here soon enough!)

PR #1723, opened 2021-02-23
Update the root readme to include clear acquisition and installation details.

ahsonkhan:
Things left to do (out-of-scope for this PR):

- Add FAQ section for common errors and potential solutions when trying to build via vcpkg
- Structure the doc with the following groups, if there is not a lot of overlap:
  - Windows
    - Visual Studio
    - VS Code
    - Command Line
  - Linux/Posix/Mac
    - VS Code
    - Command Line
 - Another dimension is `Vcpkg` vs `fetch content from GitHub` vs `build the SDK from source`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,36 +2,138 @@
 
 [![Build Status](https://dev.azure.com/azure-sdk/public/_apis/build/status/cpp/cpp%20-%20client%20-%20ci?branchName=master)](https://dev.azure.com/azure-sdk/public/_build/latest?definitionId=1611&branchName=master)
 
-This repository is for active development of the Azure SDK for C++. For consumers of the SDK we recommend visiting our [public developer docs](https://azure.github.io/azure-sdk-for-cpp) or our versioned [developer docs](https://azure.github.io/azure-sdk-for-cpp).
+This repository is for active development of the Azure SDK for C++. For consumers of the SDK we recommend visiting our [public developer docs](https://azure.github.io/azure-sdk-for-cpp).
 
 ## Getting started
 
-To get started with a library, see the **README.md** file located in the library's project folder. You can find these library folders grouped by service in the `/sdk` directory.
+For the best development experience, we recommend that developers use the [CMake projects in Visual Studio](https://docs.microsoft.com/cpp/build/cmake-projects-in-visual-studio?view=vs-2019) to view and build the source code together with its dependencies. You can also use any other text editor of your choice, such as [VS Code](https://code.visualstudio.com/), along with the command line for building your application with the SDK.
 
-For tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developers](https://azure.github.io/azure-sdk-for-cpp).
+You can find additional information for specific libraries by navigating to the appropriate folder in the `/sdk` directory. See the **README.md** file located in the library's project folder. For example, [here is the doc for the storage client library](https://github.com/Azure/azure-sdk-for-cpp/tree/master/sdk/storage#azure-storage-client-library-for-c).
+
+For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developers](https://azure.github.io/azure-sdk-for-cpp).
+
+### Download & Install the SDK
+
+The easiest way to get the C++ SDK is to leverage [vcpkg](https://github.com/microsoft/vcpkg). You'll need to have [Git](https://git-scm.com/downloads) installed before getting started.
+
+First clone and bootstrap vcpkg itself. You can install it anywhere on your machine, but **make note** of the directory where you clone the vcpkg repo.
+
+```cmd
+> git clone https://github.com/microsoft/vcpkg
+# Windows
+> .\vcpkg\bootstrap-vcpkg.bat
+# Linux
+> ./vcpkg/bootstrap-vcpkg.sh
+```
+
+To install the libraries for your project, run the following, optionally specifying the triplet:
+
+```cmd
+> .\vcpkg\vcpkg install [packages to install]
+# For example, this will install the x64-windows triplet, default is x86-windows:
+> .\vcpkg\vcpkg install azure-storage-blobs-cpp:x64-windows
+```
+
+See the [list of packages](https://github.com/Azure/azure-sdk-for-cpp#vcpkg) available via vcpkg below. All Azure C++ SDK package names start with `azure-`.
+
+You can also search for the libraries you need with the `search` subcommand:
+
+```cmd
+> .\vcpkg\vcpkg search [search term]
+# For example:
+> .\vcpkg\vcpkg search azure-
+```
+
+Once the library is installed, follow the instructions from the console output to include the library in your `CMake` application. For example, to include `azure-storage-blobs-cpp`, add the following to your `CMakeLists.txt` file:
+
+```CMake
+find_package(azure-storage-blobs-cpp CONFIG REQUIRED)
+target_link_libraries(myProject PRIVATE Azure::azure-storage-blobs)
+```
+
+> NOTE: All the Azure client libraries take a dependency on `azure-core-cpp` which provides functionality commonly needed by all Azure clients. When you install any client library via vcpkg, it will bring in all the necessary dependencies as well. You don't need to install those individually to get started.
+
+You can reference this [vcpkg Quick Start](https://github.com/microsoft/vcpkg#quick-start-windows) for more details.
+
+### Building your Application
+
+In order to use the SDK installed via vcpkg with CMake, you can use the toolchain file from vcpkg:
+
+```cmd
+> cmake -B [build directory] -S . -DCMAKE_TOOLCHAIN_FILE=[path to vcpkg repo]/scripts/buildsystems/vcpkg.cmake
+> cmake --build [build directory]
+```
+
+The **entry point** for most scenarios when using the SDK will be a top-level client type corresponding to the Azure service you want to talk to. For example, sending requests to blob storage can be done via the `Azure::Storage::Blobs::BlobClient` API.
+
+All the Azure C++ SDK headers needed to be included are located within the `<azure>` folder, with sub-folders corresponding to each service. Similarly, all types and APIs can be found within the `Azure::` namespace. For example, to use functionality form `Azure::Core`, include the following header at the beginning of your application `#include <azure/core.hpp>`.
+
+Here's an example application to help you get started:
+
+```C++
+#include <azure/core.hpp>
+#include <azure/storage/blobs.hpp>
+
+using namespace Azure::Storage::Blobs;
+
+int main()
+{
+	std::string containerName = "testcontainer";
+	std::string blobName = "sample-blob";
+
+	try
+	{
+		auto client = BlobClient::CreateFromConnectionString(
+            GetConnectionString(), containerName, blobName);
+	}
+	catch (Azure::Core::RequestFailedException& e)
+	{
+		puts(e.what());
+		return 1;
+	}
+	return 0;
+}
+```
+
+#### Visual Studio - CMakeSettings.json
+
+When building your application via Visual Studio, you can create and update a CMakeSettings.json file and include the following properties to let VS know where the package is installed and which triplet needs to be used:
+
+```json
+{
+  "configurations": [
+    {
+        "cmakeToolchain": "<path to vcpkg repo>/vcpkg/scripts/buildsystems/vcpkg.cmake",
+        "variables": [
+        {
+          "name": "VCPKG_TARGET_TRIPLET",
+          "value": "x64-windows",
+          "type": "STRING"
+        }
+      ]
+    }
+  ]
+}
+```
+
+### Azure Requirements
+
+To call Azure services, you must first have an Azure subscription. Sign up for a [free trial](https://azure.microsoft.com/pricing/free-trial/) or use your [MSDN subscriber benefits](https://azure.microsoft.com/pricing/member-offers/msdn-benefits-details/).
 
 ## Packages available
-Each service might have a number of libraries available from each of the following categories:
-* [Client - New Releases](#client-new-releases)
-* [Client - Previous Versions](#client-previous-versions)
 
-### Client: New Releases
+Each service might have a number of libraries available. These libraries follow the [Azure SDK Design Guidelines for C++](https://azure.github.io/azure-sdk/cpp_introduction.html) and share a number of core features such as HTTP retries, logging, transport protocols, authentication protocols, etc., so that once you learn how to use these features in one client library, you will know how to use them in other client libraries. You can learn about these shared features at [Azure::Core](https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/README.md).
 
-New wave of packages that we are announcing as **GA** and several that are currently releasing in **beta**. These libraries follow the [Azure SDK Design Guidelines for C++](https://azure.github.io/azure-sdk/cpp_introduction.html) and share a number of core features such as HTTP retries, logging, transport protocols, authentication protocols, etc., so that once you learn how to use these features in one client library, you will know how to use them in other client libraries. You can learn about these shared features at [Azure::Core](https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/README.md).
-
-These new client libraries can be identified by the naming used for their folder, package, and namespace. Each will start with `azure`, followed by the service category, and then the name of the service. For example `azure-storage-blobs`.
+The client libraries can be identified by the naming used for their folder, package, and namespace. Each will start with `azure`, followed by the service category, and then the name of the service. For example `azure-storage-blobs`.
 
 For a complete list of available packages, please see the [latest available packages](https://azure.github.io/azure-sdk/releases/latest/#c) page.
 
 > NOTE: If you need to ensure your code is ready for production we strongly recommend using one of the stable, non-beta libraries.
 
-### Client: Previous Versions
-
-Last stable versions of packages that are production-ready. These libraries provide similar functionalities to the beta packages, as they allow you to use and consume existing resources and interact with them, for example: upload a storage blob. They might not implement the [guidelines](https://azure.github.io/azure-sdk/cpp_introduction.html) or have the same feature set. They do however offer wider coverage of services.
-
 ### VcPkg
 
-Released SDK library versions are available on [VcPkg](https://github.com/microsoft/vcpkg):
+The following SDK library releases are available on [VcPkg](https://github.com/microsoft/vcpkg):
+
 * `azure-core-cpp`
 * `azure-identity-cpp`
 * `azure-storage-blobs-cpp`
@@ -39,8 +141,7 @@ Released SDK library versions are available on [VcPkg](https://github.com/micros
 * `azure-storage-files-datalake-cpp`
 * `azure-storage-files-shares-cpp`
 
-> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [VcPkg trilet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
-
+> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [VcPkg triplet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
 
 ## Need help
 
@@ -84,6 +185,7 @@ Many people all over the world have helped make this project better.  You'll wan
 Security issues and bugs should be reported privately, via email, to the Microsoft Security Response Center (MSRC) <secure@microsoft.com>. You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your original message. Further information, including the MSRC PGP key, can be found in the [Security TechCenter](https://www.microsoft.com/msrc/faqs-report-an-issue).
 
 ### License
+
 Azure SDK for C++ is licensed under the [MIT](https://github.com/Azure/azure-sdk-for-cpp/blob/master/LICENSE.txt) license.
 
 ![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-cpp%2FREADME.png)
--- a/README.md
+++ b/README.md
@@ -79,10 +79,8 @@ Here's an example application to help you get started:
 
 using namespace Azure::Storage::Blobs;
 
-std::string GetConnectionString()
-{
-  return "get connection string from azure portal/environment";
-}
+// Get the required connection string/key from an environment variable or Azure KeyVault.
+std::string GetConnectionString();
 
 int main()
 {
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ For the best development experience, we recommend that developers use the [CMake
 
 You can find additional information for specific libraries by navigating to the appropriate folder in the `/sdk` directory. See the **README.md** file located in the library's project folder. For example, [here is the doc for the storage client library](https://github.com/Azure/azure-sdk-for-cpp/tree/master/sdk/storage#azure-storage-client-library-for-c).
 
-For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developers](https://azure.github.io/azure-sdk-for-cpp).
+For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developer Docs](https://azure.github.io/azure-sdk-for-cpp).
 
 ### Download & Install the SDK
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,45 +2,162 @@
...
+using namespace Azure::Storage::Blobs;
+
+int main()
+{
+	std::string containerName = "testcontainer";
+	std::string blobName = "sample-blob";
+
+	try
+	{
+		auto client = BlobClient::CreateFromConnectionString(
```
antkmsft, line 102: Something wrong with indentation?

README.md
```diff
@@ -2,45 +2,162 @@
...
+            GetConnectionString(), containerName, blobName);
+	}
+	catch (Azure::Core::RequestFailedException& e)
+	{
+		puts(e.what());
```
antkmsft, line 107: use `std::cout`, the "more C++ way"?

README.md
```diff
@@ -2,45 +2,162 @@
...
+{
+	std::string containerName = "testcontainer";
+	std::string blobName = "sample-blob";
+
+	try
+	{
+		auto client = BlobClient::CreateFromConnectionString(
+            GetConnectionString(), containerName, blobName);
+	}
+	catch (Azure::Core::RequestFailedException& e)
```
antkmsft, line 105: catch by const ref?

README.md
```diff
@@ -2,45 +2,162 @@
...
 * `azure-core-cpp`
 * `azure-identity-cpp`
 * `azure-storage-blobs-cpp`
 * `azure-storage-common-cpp`
 * `azure-storage-files-datalake-cpp`
 * `azure-storage-files-shares-cpp`
 
-> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [VcPkg trilet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
-
+> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [Vcpkg triplet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
```
antkmsft, line 160:
```suggestion
> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [VcPkg triplet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.

```

ahsonkhan:
cc @kyle-patterson

README.md
```diff
@@ -2,45 +2,162 @@
...
+            GetConnectionString(), containerName, blobName);
+	}
+	catch (Azure::Core::RequestFailedException& e)
+	{
+		puts(e.what());
+		return 1;
+	}
+	return 0;
+}
+```
```
ahsonkhan, line 112: I avoided including this in https://github.com/Azure/azure-sdk-for-cpp/tree/UpdatedReadMe since it reveals a bit too much that can just be copy/pasted.

README.md
```diff
@@ -2,45 +2,162 @@
...
+
+int main()
+{
+	std::string containerName = "testcontainer";
+	std::string blobName = "sample-blob";
+
+	try
+	{
+		auto client = BlobClient::CreateFromConnectionString(
+            GetConnectionString(), containerName, blobName);
```
antkmsft, line 103: If GetConnectionString() is not a part of SDK, but is something that user has to define, do the declaration after the include headers and before "usnig namespace": `std::string GetConnectionString();`

README.md
```diff
@@ -2,45 +2,162 @@
...
+> .\vcpkg\bootstrap-vcpkg.bat
+```
+
+On Linux:
+
+```sh
+> ./vcpkg/bootstrap-vcpkg.sh
+```
+
+To install the libraries for your project, run the following, optionally specifying the triplet:
```
antkmsft, line 37: If we go to explain vcpkg basics, I'd specify, what the triplet is: "the triplet is a combination of a, b, and c [link to vcpkg docs]". And in this sentence, "optionally specifying the triplet (such as X, Y etc.)"

README.md
```diff
@@ -2,45 +2,162 @@
...
+
+```cmd
+> .\vcpkg\vcpkg install [packages to install]
+```
+
+For example, this will install the `x64-windows` triplet. On Windows, not specifying a triplet will default to x86-windows:
```
antkmsft, line 43:
```suggestion
For example, this will install packages for the `x64-windows` triplet. On Windows, not specifying a triplet will default to x86-windows:

```

antkmsft, line 43: "On Windows, not specifying a triplet will default to x86-windows" - are we sure?

README.md
```diff
@@ -2,45 +2,162 @@
...
-Each service might have a number of libraries available from each of the following categories:
-* [Client - New Releases](#client-new-releases)
-* [Client - Previous Versions](#client-previous-versions)
+For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developer Docs](https://azure.github.io/azure-sdk-for-cpp).
+
+### Download & Install the SDK
+
+The easiest way to get the C++ SDK is to leverage [vcpkg](https://github.com/microsoft/vcpkg). You'll need to have [Git](https://git-scm.com/downloads) installed before getting started.
+
+First clone and bootstrap vcpkg itself. You can install it anywhere on your machine, but **make note** of the directory where you clone the vcpkg repo.
```
antkmsft, line 19: This is great, but do we want to go to explanation of how VcPkg works? Do they have a doc that maybe we can link to instead?

README.md
```diff
@@ -2,45 +2,162 @@
 
 [![Build Status](https://dev.azure.com/azure-sdk/public/_apis/build/status/cpp/cpp%20-%20client%20-%20ci?branchName=master)](https://dev.azure.com/azure-sdk/public/_build/latest?definitionId=1611&branchName=master)
 
-This repository is for active development of the Azure SDK for C++. For consumers of the SDK we recommend visiting our [public developer docs](https://azure.github.io/azure-sdk-for-cpp) or our versioned [developer docs](https://azure.github.io/azure-sdk-for-cpp).
+This repository is for active development of the Azure SDK for C++. For consumers of the SDK we recommend visiting our [public developer docs](https://azure.github.io/azure-sdk-for-cpp).
```
antkmsft, line 5: Do we reed to say "public developer docs", or can we just say "developer docs"?

antkmsft (approved):

README.md
antkmsft: Ore, maybe, even put a body:
```
std::string GetConnectionString()
{
    return ...
}
```

README.md
```diff
@@ -2,45 +2,162 @@
...
 
-This repository is for active development of the Azure SDK for C++. For consumers of the SDK we recommend visiting our [public developer docs](https://azure.github.io/azure-sdk-for-cpp) or our versioned [developer docs](https://azure.github.io/azure-sdk-for-cpp).
+This repository is for active development of the Azure SDK for C++. For consumers of the SDK we recommend visiting our [public developer docs](https://azure.github.io/azure-sdk-for-cpp).
 
 ## Getting started
 
-To get started with a library, see the **README.md** file located in the library's project folder. You can find these library folders grouped by service in the `/sdk` directory.
+For the best development experience, we recommend that developers use the [CMake projects in Visual Studio](https://docs.microsoft.com/cpp/build/cmake-projects-in-visual-studio?view=vs-2019) to view and build the source code together with its dependencies. You can also use any other text editor of your choice, such as [VS Code](https://code.visualstudio.com/), along with the command line for building your application with the SDK.
```
RickWinter, line 9:
```suggestion
For the best development experience, we recommend developers use [CMake projects in Visual Studio](https://docs.microsoft.com/cpp/build/cmake-projects-in-visual-studio?view=vs-2019) to view and build the source code together with its dependencies. You can also use any other text editor of your choice, such as [VS Code](https://code.visualstudio.com/), along with the command line for building your application with the SDK.

```

README.md
```diff
@@ -2,45 +2,162 @@
...
 
 ## Getting started
 
-To get started with a library, see the **README.md** file located in the library's project folder. You can find these library folders grouped by service in the `/sdk` directory.
+For the best development experience, we recommend that developers use the [CMake projects in Visual Studio](https://docs.microsoft.com/cpp/build/cmake-projects-in-visual-studio?view=vs-2019) to view and build the source code together with its dependencies. You can also use any other text editor of your choice, such as [VS Code](https://code.visualstudio.com/), along with the command line for building your application with the SDK.
 
-For tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developers](https://azure.github.io/azure-sdk-for-cpp).
+You can find additional information for specific libraries by navigating to the appropriate folder in the `/sdk` directory. See the **README.md** file located in the library's project folder. For example, [here is the doc for the storage client library](https://github.com/Azure/azure-sdk-for-cpp/tree/master/sdk/storage#azure-storage-client-library-for-c).
```
RickWinter, line 11:
```suggestion
You can find additional information for specific libraries by navigating to the appropriate folder in the `/sdk` directory. See the **README.md** file located in the library's project folder. For example, [Azure Storage client library](https://github.com/Azure/azure-sdk-for-cpp/tree/master/sdk/storage#azure-storage-client-library-for-c).

```

README.md
```diff
@@ -2,45 +2,162 @@
...
 
-## Packages available
-Each service might have a number of libraries available from each of the following categories:
-* [Client - New Releases](#client-new-releases)
-* [Client - Previous Versions](#client-previous-versions)
+For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure for C++ Developer Docs](https://azure.github.io/azure-sdk-for-cpp).
```
RickWinter, line 13:
```suggestion
For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure SDK for C++ Developer Docs](https://azure.github.io/azure-sdk-for-cpp).

```

README.md
```diff
@@ -2,45 +2,162 @@
...
+
+### Download & Install the SDK
+
+The easiest way to get the C++ SDK is to leverage [vcpkg](https://github.com/microsoft/vcpkg). You'll need to have [Git](https://git-scm.com/downloads) installed before getting started.
```
RickWinter, line 17:
```suggestion
The easiest way to acquire the C++ SDK is leveraging [vcpkg](https://github.com/microsoft/vcpkg). [Git](https://git-scm.com/downloads) is required before getting started.

```

README.md
RickWinter: I'd avoid too much detail and instead show "Install x64 on Windows" followed by steps.  Have 2 or 3 of the basic exact commands.  Then link to vcpkg for those that want a more in depth explanation.  
We should avoid being documentation for vcpkg.  Also you should wait to provide deep dive links to external sites as each time its an opportunity to "lose" them in the journey.

RickWinter (changes requested):

README.md
ahsonkhan: They do, but having the end user link back-forth between various doc locations isn't a good experience. There are only ~3 steps or the customer to follow, so in-lining them here helps them get started easily. I have a link below to the vcpkg repo if they need more info.

README.md
ahsonkhan: Yes. :)

README.md
ahsonkhan: This was an intentional casing change to be consistent across the board, and with what the actual `Vcpkg` repo does. I didn't find this mixed casing used anywhere. For example, see https://github.com/microsoft/vcpkg#vcpkg-with-visual-studio-cmake-projects

README.md
ahsonkhan: Agreed, that was my approach with the docs here.

README.md
ahsonkhan: This is what you are suggesting here, which isn't really a sentence
> For example, Azure Storage client library.`

Maybe I can merge it in the previous sentence:
> See the README.md file located in the library's project folder, for example, the Azure Storage client library.

README.md
antkmsft: Tha's good, then please make sure we don't have occurrences of using "VcPkg" - I think we have paragraph title called that.

README.md
ahsonkhan: I had heard feedback in the past not to use a passive "reference manual" style tone (e.g. x is required) and refer to the reader directly in an active voice (e.g. you need x), which tends to be more engaging.

How about:
> You will need to install Git before getting started?

antkmsft, line 17:
```suggestion
The easiest way to get the C++ SDK is to leverage [Vcpkg](https://github.com/microsoft/vcpkg). You'll need to have [Git](https://git-scm.com/downloads) installed before getting started.

```

README.md
antkmsft, line 19:
```suggestion
First clone and bootstrap Vcpkg itself. You can install it anywhere on your machine, but **make note** of the directory where you clone the Vcpkg repo.

```

README.md
ahsonkhan: Can you be specific if there are any that I missed in this README? I only saw one and fixed it already.

README.md
```diff
@@ -2,45 +2,162 @@
...
+> .\vcpkg\vcpkg install [packages to install]
+```
+
+For example, this will install the `x64-windows` triplet. On Windows, not specifying a triplet will default to x86-windows:
+
+```cmd
+> .\vcpkg\vcpkg install azure-storage-blobs-cpp:x64-windows
+```
+
+See the [list of packages](https://github.com/Azure/azure-sdk-for-cpp#vcpkg) available via vcpkg below. All Azure C++ SDK package names start with `azure-`.
```
antkmsft, line 49:
```suggestion
See the [list of packages](https://github.com/Azure/azure-sdk-for-cpp#vcpkg) available via Vcpkg below. All Azure C++ SDK package names start with `azure-`.

```

README.md
```diff
@@ -2,45 +2,162 @@
...
+```
+
+Once the library is installed, follow the instructions from the console output to include the library in your `CMake` application. For example, to include `azure-storage-blobs-cpp`, add the following to your `CMakeLists.txt` file:
+
+```CMake
+find_package(azure-storage-blobs-cpp CONFIG REQUIRED)
+target_link_libraries(myProject PRIVATE Azure::azure-storage-blobs)
+```
+
+> NOTE: All the Azure client libraries take a dependency on `azure-core-cpp` which provides functionality commonly needed by all Azure clients. When you install any client library via vcpkg, it will bring in all the necessary dependencies as well. You don't need to install those individually to get started.
```
antkmsft, line 70:
```suggestion
> NOTE: All the Azure client libraries take a dependency on `azure-core-cpp` which provides functionality commonly needed by all Azure clients. When you install any client library via Vcpkg, it will bring in all the necessary dependencies as well. You don't need to install those individually to get started.

```

README.md
```diff
@@ -2,45 +2,162 @@
...
 
-### Client: New Releases
+You can reference this [vcpkg Quick Start](https://github.com/microsoft/vcpkg#quick-start-windows) for more details.
```
antkmsft, line 72:
```suggestion
You can reference this [Vcpkg Quick Start](https://github.com/microsoft/vcpkg#quick-start-windows) for more details.

```

README.md
```diff
@@ -2,45 +2,162 @@
...
 
-New wave of packages that we are announcing as **GA** and several that are currently releasing in **beta**. These libraries follow the [Azure SDK Design Guidelines for C++](https://azure.github.io/azure-sdk/cpp_introduction.html) and share a number of core features such as HTTP retries, logging, transport protocols, authentication protocols, etc., so that once you learn how to use these features in one client library, you will know how to use them in other client libraries. You can learn about these shared features at [Azure::Core](https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/README.md).
+### Building your Application
 
-These new client libraries can be identified by the naming used for their folder, package, and namespace. Each will start with `azure`, followed by the service category, and then the name of the service. For example `azure-storage-blobs`.
+In order to use the SDK installed via vcpkg with CMake, you can use the toolchain file from vcpkg:
```
antkmsft, line 76:
```suggestion
In order to use the SDK installed via Vcpkg with CMake, you can use the toolchain file from Vcpkg:

```

README.md
```diff
@@ -2,45 +2,162 @@
...
+
+```cmd
+> cmake -B [build directory] -S . -DCMAKE_TOOLCHAIN_FILE=[path to vcpkg repo]/scripts/buildsystems/vcpkg.cmake
```
antkmsft, line 79:
```suggestion
> cmake -B [build directory] -S . -DCMAKE_TOOLCHAIN_FILE=[path to Vcpkg repo]/scripts/buildsystems/vcpkg.cmake

```

README.md
```diff
@@ -2,45 +2,162 @@
...
+
+#### Visual Studio - CMakeSettings.json
+
+When building your application via Visual Studio, you can create and update a `CMakeSettings.json` file and include the following properties to let Visual Studio know where the packages are installed and which triplet needs to be used:
+
+```json
+{
+  "configurations": [
+    {
+        "cmakeToolchain": "<path to vcpkg repo>/vcpkg/scripts/buildsystems/vcpkg.cmake",
```
antkmsft, line 122:
```suggestion
        "cmakeToolchain": "<path to Vcpkg repo>/vcpkg/scripts/buildsystems/vcpkg.cmake",

```

antkmsft (commented):
Let's also convert all other occurrences of "VcPkg" and some of "vcpkg" to "Vcpkg" in our entire repo.

README.md
ahsonkhan: Following existing vcpkg docs, I don't think we need to capitalize it if isn't in a title or start of a sentence. For example:
https://github.com/microsoft/vcpkg#vcpkg-as-a-submodule
https://vcpkg.readthedocs.io/en/latest/tool-maintainers/layout/
https://en.wikipedia.org/wiki/Vcpkg
https://docs.microsoft.com/en-us/cpp/build/vcpkg?view=msvc-160&viewFallbackFrom=vs-2019

README.md
ahsonkhan: Leaving as is, lower-case unless title or start of sentence.

README.md
antkmsft: I am not questioning you saw it, I meant would this be the case on all customer Windows machines, i.e. is this behavior guaranteed/documented, or is it selected based on some system variables, and/or the default be changed by Vcpkg in the future? Is it possible to rewrite as less certain statement/avoid the statement about Vcpkg defaults?

README.md
antkmsft: Oh, so it looks like the name is lowercase "vcpkg". Then sure, we don't capitalize it unless it starts the sentence. I think we need another set of case fixes then.

README.md
```diff
@@ -2,45 +2,154 @@
...
 
-## Packages available
-Each service might have a number of libraries available from each of the following categories:
-* [Client - New Releases](#client-new-releases)
-* [Client - Previous Versions](#client-previous-versions)
+For API reference docs, tutorials, samples, quick starts, and other documentation, go to [Azure SDK for C++ Developer Docs](https://azure.github.io/azure-sdk-for-cpp).
+
+### Download & Install the SDK
+
+The easiest way to acquire the C++ SDK is leveraging vcpkg package manager. You will need to install [Git](https://git-scm.com/downloads) before getting started.
```
ahsonkhan, line 17: @RickWinter I removed the link to the vcpkg repo from here, since I link it later below and also removed example commands like `> .\vcpkg\vcpkg install [packages to install]`, leaving just the concrete ones.

Let me know if this is what you had in mind, and if you have any other suggestions.

README.md
```diff
@@ -2,45 +2,154 @@
...
 * `azure-core-cpp`
 * `azure-identity-cpp`
 * `azure-storage-blobs-cpp`
 * `azure-storage-common-cpp`
 * `azure-storage-files-datalake-cpp`
 * `azure-storage-files-shares-cpp`
 
-> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [VcPkg trilet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
-
+> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [Vcpkg triplet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
```
antkmsft, line 152:
```suggestion
> NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [vcpkg triplet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.

```

README.md
ahsonkhan: Yes, I had chatted with vcpkg folks and they mentioned the default is x86 which won't be changed.
If you don't have any specific vcpkg environment variables set (i.e. clean Windows machine), yes, that is the default.

I think it is important to call this out as it can be surprising to folks since they get hard to debug errors when trying to find the package and build their application. I don't see much benefit to try to make a less certain statement.

Here's a doc that mentions it:
https://vcpkg.readthedocs.io/en/latest/users/triplets/#additional-remarks

README.md
```diff
@@ -2,45 +2,154 @@
...
+#include <iostream>
+#include <azure/core.hpp>
+#include <azure/storage/blobs.hpp>
+
+using namespace Azure::Storage::Blobs;
+
+std::string GetConnectionString()
+{
+  return "get connection string from azure portal/environment";
+}
```
vhvb1989, line 85: Well, this gives the impression that we are recommending having the connection string in the source code, which is bad.

Since this is an example of how we would do a sample app, can we use `std::getenv()` to recommend always putting connection string on the env?

README.md
```diff
@@ -2,45 +2,154 @@
...
+The **entry point** for most scenarios when using the SDK will be a top-level client type corresponding to the Azure service you want to talk to. For example, sending requests to blob storage can be done via the `Azure::Storage::Blobs::BlobClient` API.
 
-These new client libraries can be identified by the naming used for their folder, package, and namespace. Each will start with `azure`, followed by the service category, and then the name of the service. For example `azure-storage-blobs`.
+All the Azure C++ SDK headers needed to be included are located within the `<azure>` folder, with sub-folders corresponding to each service. Similarly, all types and APIs can be found within the `Azure::` namespace. For example, to use functionality form `Azure::Core`, include the following header at the beginning of your application `#include <azure/core.hpp>`.
+
+Here's an example application to help you get started:
+
+```C++
+#include <iostream>
+#include <azure/core.hpp>
```
vhvb1989, line 77: We can remove this if we are not using anything from core.hpp directly.
If you want to keep it, add something like a body stream or context

README.md
```diff
@@ -2,45 +2,154 @@
...
+    std::cout << e.what() << std::endl;
+    return 1;
+  }
+  return 0;
+}
+```
+
+#### Visual Studio - CMakeSettings.json
+
+When building your application via Visual Studio, you can create and update a `CMakeSettings.json` file and include the following properties to let Visual Studio know where the packages are installed and which triplet needs to be used:
```
vhvb1989, line 108: I would be surprised by people working with the json directly instead of the json viewer from Visual Studio. ( No actions for this comment, just sharing with you)

vhvb1989 (approved):

README.md
ahsonkhan: `RequestFailedException` gets used.

README.md
ahsonkhan: I was considering including a picture like this, later:
image

README.md
ahsonkhan: This isn't really the point of the sample, and I didn't want to encourage use of `getenv()` especially without "is string empty" checks. Doing it properly ends up taking up too much code space. But you bring up a good point.

Maybe I should just leave it without a body and add a comment that the user can read it from an environment variable.

ahsonkhan:
@msftbot do not merge unless @RickWinter approves

ahsonkhan:
@msftbot hold this pr for the next 1 hour

bhardwajs:
@ahsonkhan Is consuming SDK as a submodule supported scenario?

ahsonkhan:
> Is consuming SDK as a submodule supported scenario?

We plan it to be, and it might just work already, but I haven't tried it yet. It is a bit lower in the priority list. Do you know of any particular pitfalls to look out for or document that is unique to consuming as a submodule (which would be different from just building the SDK as source outside of the application repo)?

RickWinter:
> @ahsonkhan Is consuming SDK as a submodule supported scenario?

Consuming as a submodule is not the mainline consumption scenario.  It may work but is not officially supported.  
The reason for that is we do not expect all packages to ship on the same cadence.  i.e. Storage may ship 1.0 at the same time KeyVault ships the first Beta.  The folders within sdk directory are only verified stable at the release tags for each package (i.e. sdk/<folder>)

README.md
RickWinter: Seems reasonable.  We should give enough detail to onboard and unblock the customer while not becoming the vcpkg documentation.

ahsonkhan:
> The reason for that is we do not expect all packages to ship on the same cadence.

Ah yes, good reminder Rick. Even if you submodule for a particular GA tag, it may contain libraries that are still in beta.

RickWinter (approved):

PR #1731, opened 2021-02-24
CaseInsensitiveMap

### LINKED ISSUE #1525: Use case insensitive map for our request response headers - move the comparator from storage to core.

### LINKED ISSUE #1626: The `Metadata` type from Storage is exposing a type that is meant to be private in public surface area via type alias
Public APIs that accept `Metadata` are leaking a type from the `Details::` namespace, specifically `CaseInsensitiveComparator` because of the type alias. Given `Metadata` is an alias for a custom templated map, the user ends up relying on a private implementation detail, which goes against the guidelines, which means we wouldn't be able to rename or otherwise break the type.
https://github.com/Azure/azure-sdk-for-cpp/blob/76db013528d60e3f54919628747dd990a9c36600/sdk/storage/azure-storage-common/inc/azure/storage/common/storage_common.hpp#L69

For example, see:
https://github.com/Azure/azure-sdk-for-cpp/blob/90430040d5b446ade88caf2400f344040be16cb3/sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_directory_client.hpp#L163-L165
https://github.com/Azure/azure-sdk-for-cpp/blob/90430040d5b446ade88caf2400f344040be16cb3/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_file_system_client.hpp#L143-L146
https://github.com/Azure/azure-sdk-for-cpp/blob/90430040d5b446ade88caf2400f344040be16cb3/sdk/storage/azure-storage-files-datalake/inc/azure/storage/files/datalake/datalake_options.hpp#L110-L128

cc @Jinming-Hu, @JeffreyRichter

### LINKED ISSUE #1640: Create Azure::Core::(<?>::)CaseInsensitiveMap?
Inspired by https://github.com/Azure/azure-sdk-for-cpp/pull/1536 and PR comments/proposals

* Should it be `<string, T>`, or `<string, string>`?
* We do lowercase on set, not only compare case insensitively. Right?
* Do we lowercase values?
* Which namespaceshould it be in?
* Bring up a "Follow up on Review Comments - Case-insensitive Metadata Lookup" email thread from December 2020.

antkmsft:
Closes #1640.
Closes #1626 (exposing comparator is not a concern - users should not refer to it as `Azure::Core::Strings::Detsils::CaseInsensitiveComparator`, but as `Azure::Core::CaseInsensitiveMap<T>::key_compare`; and even then there should be no need for that).
Closes #1525.


antkmsft:
@Jinming-Hu, you've been active in these discussions, please review.

sdk/core/azure-core/inc/azure/core/case_insensitive_map.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <string>
+
+namespace Azure { namespace Core {
+
+  /**
+   * @brief A type alias of `std::map<std::string, T>` with case-insensitive key comparison.
+   *
+   * @tparam T Type of values being stored in the map.
+   */
+  template <typename T>
```
Jinming-Hu, line 23: when T is not `std::string`?

Jinming-Hu (approved):

sdk/core/azure-core/inc/azure/core/case_insensitive_map.hpp
antkmsft: Never yet. I am fine to drop it, let me know. We always can reintrocuce it, and make it to be a type alias of `CaseInsensitiveMap<T>`. But what would we call this map?

antkmsft: Hmm, ok, I think I know. Let's remove T, and if we need it later, we will do `template<typename T = std::string>` to not break the existing code.

antkmsft: @Jinming-Hu, thank you, good question, check it out now.

Jinming-Hu (approved):

JeffreyRichter (approved):

sdk/core/azure-core/inc/azure/core/case_insensitive_map.hpp
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ahsonkhan, line 1: Add this to the simplified core.hpp header + test.

sdk/core/azure-core/inc/azure/core/case_insensitive_map.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <string>
+
+namespace Azure { namespace Core {
+
+  /**
+   * @brief A type alias of `std::map<std::string, std::string>` with case-insensitive key
+   * comparison.
+   */
+  using CaseInsensitiveMap
+      = std::map<std::string, std::string, Internal::Strings::CaseInsensitiveComparator>;
```
ahsonkhan, line 23: Can you please summarize why we decided against the "wrap this in our own type" approach that was discussed before?
https://github.com/Azure/azure-sdk-for-cpp/issues/1626#issuecomment-775353294

antkmsft: This is simpler, we introduce less code, and it is familiar for user and less chance for us to introduce a bug. It satisfies all our requirements, metadata uses it, it is minimum invasive behavior. Jeffrey and Jinming liked it. Jinming still could make Metadata class into something else (storage uses, basically, a typedef), but I don't think he will, he liked the map.

sdk/core/azure-core/inc/azure/core/case_insensitive_map.hpp
antkmsft: I am adding it in #1752.

sdk/core/azure-core/inc/azure/core/case_insensitive_map.hpp
Jinming-Hu: > Can you please summarize why we decided against the "wrap this in our own type" approach that was discussed before?

As I explained in this issue https://github.com/Azure/azure-sdk-for-cpp/issues/1640#issuecomment-776480022, doing so would be a lot of work, which I don't think is worth it.

I'm not saying you cannot do that, I'm just trying to say if you wanna do it, think thoroughly and do it right.

Issue #1537, opened 2021-02-01
Removing type-safety from Context


JeffreyRichter:
I feel strongly that Context should NOT have any type-safety applied to its key/value pairs. The usage pattern does not allow for this and it complicates the API and surface area for no value.

PR #1745, opened 2021-02-25
Remove type safety from context

vhvb1989:
fixes: #1537


Jinming-Hu:
After this change, how should we use this `context` thing? Can you share some usage scenarios or examples?

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -16,69 +16,55 @@ using namespace Azure::Core;
...
-  auto& valueT = c2["key"];
-  auto value = valueT.Get<bool>();
-  EXPECT_TRUE(value == true);
-
-  auto kind = valueT.Alternative();
-  EXPECT_TRUE(kind == ContextValue::ContextValueType::Bool);
+  bool value(true);
+  auto c2 = context.WithValue("key", reinterpret_cast<void*>(&value));
+  auto valueVoidP = c2["key"];
+  auto valueT = *reinterpret_cast<bool*>(valueVoidP);
```
Jinming-Hu, line 32: If we set this value in one function and get this value from another function, and the original value is already destructed on stack, this will crash

Jinming-Hu:
After this PR, if context values are allocated on stack, who is responsible for keeping the values valid from stack unwinding?

If the values are allocated on heap, who is responsible for releasing the heap memory?

vhvb1989:
> After this PR, if context values are allocated on stack, who is responsible for keeping the stack frame valid?
> 
> If the values are allocated on heap, who is responsible for releasing the heap memory?

Yes, I commented this to @JeffreyRichter  some time ago but it using void* was just simpler than all the required machinery to keep the type and maintain the value inside the context.

By this approach, whatever value is pointer by the context needs to be kept alive for sending the request.
It might not be super bad because this is per-operation only.  We should not try to keep one context to be used for all or multiple requests to pass values to the policies. For some scenario like this, the right approach would be to consider creating a policy

vhvb1989:
> if context values are allocated on stack, who is responsible for keeping the stack frame valid?

The right approach would be to let the Operation method from the SDK client (say blobClient.Upload(), for instance) to be responsible for the lifetime of the value that is smuggled to any policy.
Either if the value is stack or heap allocated, the operation method should init the value, get a void* from it, create a context with it and finally use that context to call the protocol layer.

We should never expect to see bad code like a helper method that allocate values in stack and return a ptr to it (compiler might even warn about it). It could allocate on heap (and the operation method should free the memory before returning ( use catch and re-throw to free for exceptions).

I becomes using this feature a little more complicated, but at the end this feature is for advanced scenarios. We don't event have one yet for it.

Jinming-Hu:
> I becomes using this feature a little more complicated, but at the end this feature is for advanced scenarios. We don't event have one yet for it.

I don't know why the original impl doesn't work for Jeff and as far as I can see, the new impl just makes it far less easy for customers to use. It's a very C-style thing, the customer has to consider the lifetime and cast the pointers back and force. I wouldn't want to use this stuff if I were a customer. But luckily I'm not, and as storage developer, I don't use `context` at all. So, it doesn't hurt me anyway.

Jinming-Hu (approved):

RickWinter (changes requested):
I know Jeff wanted this change.  I disagree that its the right thing to do.  The ContextValue being type safe has advantages we should keep.  Its also inline with modern C++.

JeffreyRichter:
I implore you to write the code that is necessary to use this feature successfully with ContextValue and demonstrate that it is useful.

RickWinter:
> I implore you to write the code that is necessary to use this feature successfully with ContextValue and demonstrate that it is useful. 

The benefit is the type safety.  I can be assured the value stored in the slot is the type that I expect it to be.
```C++
  std::string s("Some string value");

  Context context;
  auto c2 = context.WithValue("MyUrlAsString", s);
  auto& myString = c2["MyUrlAsString"];
  if(myString .Alternative() != ContextValue::ContextValueType::StdString) 
  {
     //Oh no something is wrong, that key should have been a string
     //throw exception...
  }

  //Safely grab the value as I'm assured it is a string
  //No casting required on retrieval
  auto value = myString.Get<std::string>();
  
  //Use the string without casting b/c I know it is valid and a string
  cout << value;

```

JeffreyRichter:
Your code does not demonstrate the real life scenario. You need to write the code in the separated scenario where the key/value are set by an end user and the key/value is read by the policy implementor.

First, policies will be looking for these keys and values will almost always be a pointer to a structure. In this case, the type safety is just overhead because it only works for some simple primitive types which is not our mainstream scenario. Furthermore, the policy implementor will define the structure with the fields it supports. The policy implementor will also define the key. In fact, the policy implementor will probably offer a function that accepts a context and the structure and this function will set the key/value and return the new context. You should write all this code so that our whole team understands the real use case. It would be a good model to hand off to client library teams that need to do this. Storage will want this to give end users a per-operation retry override mechanism.

Second, the ContextValue adds memory overhead for what is not our mainstream scenario.

Third, the ContextValue type exists purely for our Context type and so customers will be compelled to learn about it in order to have increased memory consumption and then not find the type safety useful anyway. This is anti value.

Jinming-Hu:
Hi @JeffreyRichter, I'm not sure I fully understand the problem you're trying to solve in this PR. But I'm with @RickWinter on the benefit of type safety.

I think I can come up with a way to support both any-type context-value and type-safety with template. I wrote a demonstration here https://onlinegdb.com/BJw1yBPMO. So maybe we can meet each other halfway.

This sample code is overly simplified to be easy-to-understand. I can still work on memory issue (too many copies) if you think this idea is basically good.

JeffreyRichter:
@Jinming-Hu : I like your implementation a lot! My only concern is the overhead. A Context must have a pointer to a previous Context and a long for expiration time.  Cancellation is the champion scenario for Context. Adding a key/value pair to a Context is a much lower priority feature as this will rarely be used by comparison. So, to support this low priority feature, we're adding >100% overhead to a Context because of these 3 fields: Key, Value and ValueType. In addition, you're already using a reinterpret_cast and so you're manually ensuring type-safety via the typeid check. 

So, I love your programming model, I'd love to get the overhead reduced for the key/value feature, and also, it is very common that a Get passing a key will fail. So, I'd like to NOT call HasKey first and then Get as this walks the linked-list twice. We need a function like C#'s TryGet that returns a bool and the T as an output parameter so the policy code can call it like this:

int int_value;
if (c5.Get<int>("key_i", &int_value)) { ... }

vhvb1989:
@Jinming-Hu  on your example, when we pass the instance as argument, the Context is later making a copy of it.  This will become expensive for big structures ( imagine we want to refer a const JSON ).

One thing I would like to be able to do with the Context is to set a `const&` or a `ValueBase const*const` or just `ValueBase *` . 

My use case would be for an HttpPipeline to create a Context node with a direct reference to itself or to the policies array without creating a copy of it.

If a copy is always allocated with your model (or like the current model), some people would prefer to create the value globally and access to it from their custom policy instead of making a copy of it for every request

Jinming-Hu:
Hi @JeffreyRichter and @vhvb1989 , thanks for your feedback. I'd like to clarify that in that demonstration, I'm just trying to show a way to avoid exposing `void*` and manual memory allocation/deallocation, because usually they're considered C-style things.

The copies of `Context` and the long chain of each context having a pointer to its parent is not what I intended to demonstrate. It's already in the current implementation, and I do not intend to change it.

Jinming-Hu:
@vhvb1989 Now about your other concerns

> @Jinming-Hu on your example, when we pass the instance as argument, the Context is later making a copy of it. This will become expensive for big structures ( imagine we want to refer a const JSON ).

Your this comment applies to the whole C++ world, we can do our best to avoid copies with `std::move` and rvalue refs.

vhvb1989:
> Your this comment applies to the whole C++ world

What about raw pointers and shared_ptr.   Somethings that's all we need, to read memory which is already at some place :)

Jinming-Hu:
> > Your this comment applies to the whole C++ world
> 
> What about raw pointers and shared_ptr. Somethings that's all we need, to read memory which is already at some place :)

if you look at my example, `T value`, `T` can be a pointer or shared pointer

Jinming-Hu:
> My use case would be for an HttpPipeline to create a Context node with a direct reference to itself or to the policies array without creating a copy of it.

I'm okay with this but I don't think it'd be necessary. When a function accepts a `context` as its parameter, we never know if the same `context` is passed to other functions in a multi-thread application. So to avoid race condition, we have to make a copy of the original `context` and and append key/value pair to the copy.

An example of this scenario would be multi-thread upload in storage sdk. Users call `UploadTo` with a `context` as parameter, this `context` will be later used in multiple threads that actually perform HTTP requests.

vhvb1989:
Closing in favor of : https://github.com/Azure/azure-sdk-for-cpp/issues/1815

PR #1750, opened 2021-02-26
Rename add headers

vhvb1989:
fixes: #1746


sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -517,7 +517,7 @@ namespace Azure { namespace Core { namespace Http {
      *
      * @throw if \p name is an invalid header key.
      */
-    void AddHeader(std::string const& name, std::string const& value);
+    void SetHeader(std::string const& name, std::string const& value);
```
ahsonkhan, line 520: The documentation says "Add HTTP header...". Why are we renaming this to Set?

`SetHeader` (or maybe even SetHeaderValue) only makes sense when we are overriding an **existing** header. It doesn't make sense when we want to Add a **new** header.

If I create a request with no headers, it is strange for me to start calling `SetHeader` to add new headers.

ahsonkhan: Rather than renaming AddHeader, we should provide new SetHeader methods, if we want the "override" functionality.

ahsonkhan: image

ahsonkhan (changes requested):
Do not rename AddHeader (keep it as is, for adding new headers), instead **add** a SetHeader method that does overriding. And update the API docs to make the distinction clear.

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: Our default behavior is to override headers when we call AddHeader().  We don't need the single AddHeader() which won't override if the key is already there (we don't want to throw or ignore the AddHeader() ).
We can either just document that our AddHeader will do override as if was SetHeader()  or we can document that SetHeader behaves similar to the `[]` operator from a map and it will add the key if it is not there or everride the value.

ahsonkhan: @vhvb1989 What method do I call to add a **new** header to the request object?

cc @tg-msft

vhvb1989: >  which throws for existing keys

That's shooting ourselves in the foot =P 
A common case (probably the most important) is one Http policy that adds a header.  The policy doesn't care if the header is already there or not. It just want to override the value of it. So, it want to either add it or replace its value.

Right now, the policy just calls `AddHeader()` and if the header was already there, it gets overriden.

If we have both methods, the policy will need to check first if the header is there to know what method to call..  So, it making things complicated for us..  We only need one behavior, whichever the name is for the method.

I think SetHeader() make it clearer about the override.  But I would also be happy with `AddOrSetHeader()` or `AddOrOverrideHeader()`..  etc etc

ahsonkhan: Both these options are good:
- `AddHeader()` + clear documentation that it will override existing ones OR 
- `AddOrSetHeader()`

The concern with just having `AddHeader`/`AddOrSetHeader`, is what happens if someone just wants to add a header, but doesn't want to override if the key already exists.

ahsonkhan (commented):
Based on offline discussion, `SetHeader` is fine to mean add if the key doesn't exist, or replace if it does.

**Waiting for a API doc update before approving, to make the details clear.**


This is similar to some other language SDKs like Java:
https://github.com/Azure/azure-sdk-for-java/blob/f8cdccb633543d59015d65df3706c27169e13d05/sdk/core/azure-core/src/main/java/com/azure/core/http/HttpRequest.java#L146-L154

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -515,14 +515,17 @@ namespace Azure { namespace Core { namespace Http {
     explicit Request(HttpMethod httpMethod, Url url);
 
     /**
-     * @brief Add HTTP header to the #Azure::Core::Http::Request.
+     * @brief Set an  HTTP header to the #Azure::Core::Http::Request.
      *
-     * @param name The name for the header to be added.
-     * @param value The value for the header to be added.
+     * @remark If the header key does not exists, it is added.
```
ahsonkhan, line 520:
```suggestion
     * @remark If the header key does not exist, it is added. Otherwise, the value is overridden.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -662,7 +665,7 @@ namespace Azure { namespace Core { namespace Http {
      *
      * @throw if \p name contains invalid characters.
      */
-    void AddHeader(std::string const& name, std::string const& value);
+    void SetHeader(std::string const& name, std::string const& value);
```
ahsonkhan, line 668: Update the docs for this and other `SetHeader` overloads as well, similar to `Request::SetHeader()` above, to remark about existing/new header names and what happens in those cases.

Also update to use `Set` instead of `Add` in all the doc comments for consistency.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -674,7 +677,7 @@ namespace Azure { namespace Core { namespace Http {
      *
      * @throw if \p header has an invalid header name or if the delimiter is missing.
      */
-    void AddHeader(std::string const& header);
+    void SetHeader(std::string const& header);
```
ahsonkhan, line 680: Same here, update docs, talking about what happens if the header key already exists.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -690,7 +693,7 @@ namespace Azure { namespace Core { namespace Http {
      * @throw if the string referenced by \p first and \p last contains an invalid header name or if
      * the delimiter is missing.
      */
-    void AddHeader(uint8_t const* const first, uint8_t const* const last);
+    void SetHeader(uint8_t const* const first, uint8_t const* const last);
```
ahsonkhan, line 696: Same here, update docs.

ahsonkhan (approved):
Other than updating the docs for all `SetHeader` methods, looks good.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -660,19 +663,19 @@ namespace Azure { namespace Core { namespace Http {
     // ===== Methods used to build HTTP response =====
 
     /**
-     * @brief Add HTTP header to the #Azure::Core::Http::RawResponse.
+     * @brief Set an HTTP header to the #Azure::Core::Http::Request.
```
ahsonkhan, line 666: Isn't this for the `RawResponse` type? Please fix.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -660,19 +663,19 @@ namespace Azure { namespace Core { namespace Http {
...
+     * @param value The value for the header to be set or added.
      *
      * @throw if \p name contains invalid characters.
      */
-    void AddHeader(std::string const& name, std::string const& value);
+    void SetHeader(std::string const& name, std::string const& value);
 
     /**
-     * @brief Add HTTP header to the #Azure::Core::Http::RawResponse.
+     * @brief Set an HTTP header to the #Azure::Core::Http::Request.
```
ahsonkhan, line 678: Same here and below. This isn't the `Request`, but rathe the `RawResponse`.

vhvb1989: I should take care more attention to the doxygen comments. :(
Here's the fix: https://github.com/Azure/azure-sdk-for-cpp/pull/1783

Thank you for checking Ahson!  soo cool! 👍

PR #1752, opened 2021-02-26
API Review Feedback: Logging and DateTime updates

### LINKED ISSUE #1639: Logging: review once again
* Default log level, what it should be -  `Verbose`, or Warning, or something else?
* Do we like free functions inside the `Azure::Core::Logging::` namespace, or should we make a class?
* `std::function`, or something else for callback?
* Add `string componentId`? (other proposal was string serviceId) Or don't? Or not a string, but some enum?
* LogLevels order - ascending or descending? - https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-5.0
* Do we want more log levels? Any problems if we add more in the future?

Victor had cool proposal for the LoggingOptions, so logging callback+logLevel would behave something like HttpTransport - you don't have to specify it, or use different transport, but if you want, it is overridable.

* Go and instrument more code with logging? One person does it, or everyone does their part? Storage does not currently have logging at all. Let's also come up with pattern, which log levels we use. Create a separate work item for that?

* Use atomic instead of mutex? I'm pretty sure no one would disagree, but what if would require us to not use `std::function` in the user-facing API? (#1638)

(Just in case: https://github.com/Azure/azure-sdk-for-cpp/pull/1568)

### LINKED ISSUE #1722: Add allowlist for headers
Logging should only log the whitelist of headers.

antkmsft:
This takes the API review feedback, and I reviewed the design with Jeffrey, took his feedback.

Closes #1639.
Closes #1722.

* If `AZURE_LOG_LEVEL` is set in the environment to a corresponding level, init the default log listener with console logger (see https://github.com/Azure/azure-sdk-for-java/wiki/Logging-with-Azure-SDK) || unless a build flag is specified, so the env-listening console logger functionality won't be built in.
* Created Logger class, moved the public contents of Azure::Core::Logging there.
* `Write()` now uses `std::shared_timed_mutex`, which is more effective than the plain `std::mutex` in our case - we get shared reads for cheaper price. Atomic is not possible with `std::function`, and we do want to use `std::function`, Jeffrey agrees.
* `ShouldWrite()` is 100% lock-free. It is an inlined atomic bool variable check, then, sometimes, another atomic variable check.
* Created LogOptions, which has 3 whitelists for request and response headers, and for query parameters, in response to API review feedback. Use them in the code to not output anything that is not there.
* Added CaseInsensitiveSet, which is used in LoggingPolicyOptions.
* RetryPolicy stores retry codes in `std::set` instead of `std::vector`.


ahsonkhan:
> Tweak logging

Please update the PR title to be more descriptive with the details of the change and intent of the PR

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,84 @@
...
+    static void SetListener(Listener listener);
+
+    /**
+     * @brief Sets the #Azure::Core::Logger::Level an application is interested in receiving.
+     *
+     * @param level Maximum log level.
+     */
+    static void SetLevel(LogLevel level);
+
+  private:
```
vhvb1989, line 75: is there a reason why to do:
class Name {
public:
...
private:
...
public:
...
private:
...
}

???

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,84 @@
...
+    Logger() = delete;
+    ~Logger() = delete;
+
+    friend bool Internal::ShouldLog(LogLevel);
+    friend void Internal::Log(LogLevel, std::string const&);
```
vhvb1989, line 80: I know someone who won't be happy about this 🤣

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -30,22 +30,69 @@ std::string TruncateIfLengthy(std::string const& s)
...
+    auto qm = url.find('?');
+    if (qm != std::string::npos)
+    {
+      url = url.substr(0, qm);
+    }
+  }
+
+  {
+    auto separ = '?';
+    for (auto qparam : request.GetUrl().GetQueryParameters())
```
vhvb1989, line 51:
```suggestion
    for (auto& qparam : request.GetUrl().GetQueryParameters())
```

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -30,22 +30,69 @@ std::string TruncateIfLengthy(std::string const& s)
...
-std::string GetRequestLogMessage(Request const& request)
+std::string GetRequestLogMessage(
+    Azure::Core::Http::LoggingPolicyOptions const& options,
+    Request const& request)
 {
   std::ostringstream log;
 
-  log << "HTTP Request : " << HttpMethodToString(request.GetMethod()) << " "
-      << request.GetUrl().GetAbsoluteUrl();
+  std::string url = request.GetUrl().GetAbsoluteUrl();
```
vhvb1989, line 39: Since you are calling GetUrl() more than once in this function, consider doing
auto const& url = request.GetUrl();  here and then using it below

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -30,22 +30,69 @@ std::string TruncateIfLengthy(std::string const& s)
...
+  if (!url.empty())
```
vhvb1989, line 40: consider returning if url is empty.  We won't find ? or anything else below if the AbsoluteUrl is empty.

```cpp
if (url.empty()) {
  return {}; // empty string or any placeholder you want here.
}
```

This might be, however, a very unlikely scenario to reach the logging policy with an empty Url.  I would expect that to be thrown much earlier from sdk clients

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -30,22 +30,69 @@ std::string TruncateIfLengthy(std::string const& s)
...
+  {
+    auto qm = url.find('?');
+    if (qm != std::string::npos)
+    {
+      url = url.substr(0, qm);
```
vhvb1989, line 45: Feel free to update the Url class and add a `HasQueryParameters()` method which can save you from calling `find()` here.

I would also recommend to add a `GetShortUrl()` or something which would return only the url up to the query parameters start (if any). Then you can just call GetQueryParameters.

We can use the Url for our own benefit here

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -30,22 +30,69 @@ std::string TruncateIfLengthy(std::string const& s)
...
+  {
+    auto separ = '?';
+    for (auto qparam : request.GetUrl().GetQueryParameters())
+    {
+      url += separ + qparam.first + '=';
+
+      if (options.AllowedHttpQueryParameters.find(qparam.first)
+          != options.AllowedHttpQueryParameters.end())
+      {
+        url += TruncateIfLengthy(qparam.second);
```
vhvb1989, line 58: Consider that query parameters will be encoded, it might be easier to digest for humans if we call Decode() on them, unless we want to show exactly what we sent as verbatin.

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -30,22 +30,69 @@ std::string TruncateIfLengthy(std::string const& s)
...
+        url += "[hidden]";
+      }
+
+      separ = '&';
+    }
+  }
+
+  log << "HTTP Request : " << HttpMethodToString(request.GetMethod()) << " " << url;
 
   for (auto header : request.GetHeaders())
```
vhvb1989, line 71:
```suggestion
  for (auto& header : request.GetHeaders())
```

sdk/core/azure-core/src/http/logging_policy.cpp
```diff
@@ -60,13 +107,23 @@ std::string GetResponseLogMessage(
   for (auto header : response.GetHeaders())
```
vhvb1989, line 107:
```suggestion
  for (auto& header : response.GetHeaders())
```

sdk/core/azure-core/test/ut/logger.cpp
```diff
@@ -220,7 +221,7 @@ TEST(Logging, Message)
   }
   catch (...)
   {
-    SetLogListener(nullptr);
+    Logger::SetListener(nullptr);
```
vhvb1989, line 224: Might be worth it to test Http Request and Response samples
- Request with allowed and hidden query and headers
- Response with allowed and hidden headers
- Request with no query parameters or headers
- Request/Response with no headers

vhvb1989 (commented):
Adding few comments

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -251,7 +251,7 @@ namespace Azure { namespace Core { namespace Http {
     uint16_t m_port{0};
     std::string m_encodedPath;
     // query parameters are all encoded
-    std::map<std::string, std::string> m_encodedQueryParameters;
+    CaseInsensitiveMap m_encodedQueryParameters;
```
Jinming-Hu, line 254: Keys of query parameters are case-sensitive according to RFC

sdk/core/azure-core/inc/azure/core/logger.hpp
antkmsft: It is now gone, but previously yes - if a private section is using something that needs to be defined in public (typedef), and then public section needs to use something that is defined in private (inline function), then this is what you do, compiler would read the file from top to bottom.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -359,18 +360,35 @@ namespace Azure { namespace Core { namespace Http {
...
    */
   class LoggingPolicy : public HttpPolicy {
+    LoggingPolicyOptions m_options;
+
   public:
     /**
      * @brief Constructs HTTP logging policy.
      */
-    explicit LoggingPolicy() {}
+    explicit LoggingPolicy(LoggingPolicyOptions options) : m_options(std::move(options)) {}
```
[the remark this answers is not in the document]
antkmsft: Policies are being created by clients. We want clients to have logging options. I don't want them to forget to pass the client options here, then the issue being found, and then releasing a fix to pass the options here. It is not really that user-facing functionality for us to optimize the convenience of using the defaults. I'd say, if you pass the default options here and not the LoggingPolicyOptions from the client, it is likely a mistake, and if your client really wants this, is should do so explicitly.

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: I am not sure if we'd still want to distinguish between them (other than preserving the case), but at the very least this can be changed later without breaking the compile-time compatibility. I reverted this change.

sdk/core/azure-core/inc/azure/core/logger.hpp
antkmsft: Ok, I removed friend declarations. cc @RickWinter

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
vhvb1989: > Policies are being created by clients. 

Working on letting client to add only custom policies not the default ones like Logging. Azure Core will be adding this one.

>  I don't want them to forget to pass the client options here

That's the same approach we are using for the other policies, like retry, if nothing is set on the options, it is the signal for using whatever is the default.  The same `forgetting` issue is happening right now to the other policies.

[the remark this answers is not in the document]
antkmsft: RetryPolicy does not have the default value in the constructor either.

sdk/core/azure-core/src/http/logging_policy.cpp
antkmsft: I don't think any special logic needs to be there. If the URL is empty, it should be caught by clients. Logger's job is to log everything as it sees it, so that diagnostics can be done. Plus, if at some point empty URL has some meaning (useful in oversimplified unit tests?), you don't need to update this place. I only need logging to not blow up if URL is empty, and then if it is, we log it as is, without thinking out of our scope what this might mean. I think it is reasonable.

sdk/core/azure-core/test/ut/logger.cpp
antkmsft: Tests are covered by #271, and will be added later soon.

sdk/core/azure-core/src/http/logging_policy.cpp
vhvb1989: > I don't think any special logic needs to be there

That's what I meant. We don't need to check if url is empty() here.
We don't need to manually parse the url to check if it contains queryParams or not and brake it into url without query.  Just update our Url class to return what you need directly.  The Url knows if there are query params or not already.

sdk/core/azure-core/src/http/logging_policy.cpp
antkmsft: I think that since we show in the URL form, we should not decode, but show what gets sent. However, AllowedHttpQueryParameters shouldn't be encoded, I'll fix that.

antkmsft: On the other hand, that would require us to decode it every time we log, so for performance reasons it makes sense to store it as encoded. I'll update the naming.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -359,18 +360,42 @@ namespace Azure { namespace Core { namespace Http {
         NextHttpPolicy policy) const override;
   };
 
+  /**
+   * @brief Options for Azure::Core::Http::LoggingPolicy.
+   */
+  struct LoggingPolicyOptions
```
JeffreyRichter, line 366: Rename to LogOptions

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -359,18 +360,42 @@ namespace Azure { namespace Core { namespace Http {
...
+
   /**
    * @brief Logs every HTTP request.
    *
-   * @details Logs every HTTP request, response, or retry attempt.
-   * @remark See #logging.hpp
+   * @details Logs every HTTP request and response.
+   * @remark See Azure::Core::Logger.
    */
   class LoggingPolicy : public HttpPolicy {
```
JeffreyRichter, line 391: Renaem to LogPolicy

sdk/core/azure-core/inc/azure/core/internal/log.hpp
```diff
@@ -3,9 +3,49 @@
...
+      return g_isLoggingEnabled && static_cast<LogLevelInt>(level) <= g_logLevel
+          ? GetLogListener()
+          : Logger::Listener(nullptr);
+    }
+
+    Log() = delete;
+    ~Log() = delete;
+
+  public:
+    static bool ShouldWrite(Logger::Level level) { return GetLogListener(level) != nullptr; }
```
JeffreyRichter, line 38: This should not be public - customers can't call it

sdk/core/azure-core/inc/azure/core/internal/log.hpp
```diff
@@ -3,9 +3,49 @@
...
+
+    static void Write(Logger::Level level, std::string const& message)
```
JeffreyRichter, line 40: This should not be public - customers can't call it

sdk/core/azure-core/inc/azure/core/internal/log.hpp
```diff
@@ -3,9 +3,49 @@
...
+    {
+      if (auto listener = GetLogListener(level))
+      {
+        listener(level, message);
+      }
+    }
+
+    static void ToggleLogging(bool isEnabled);
```
JeffreyRichter, line 48: Why do we need this?

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,67 @@
...
+#pragma once
+
+#include <functional>
+#include <string>
+
+namespace Azure { namespace Core {
+  /**
+   * @brief Log message handler.
+   */
+  class Logger {
```
JeffreyRichter, line 19: Something went wack y here. Why do we have both Log and Logger classes? We should NOT have these 2 things - this is so confusing!
I thought our plan was to have a Log class with private methods (ShouldWrite & Write) for our internal stuff and public methods for customer-callable stuff (SetLevel & SetListener). Then, an internal header file that made the 2 internal functions (ShouldWrite & Write) callable using just global functions that were freinds of the class.

This seems to have gotten overly complex.

sdk/core/azure-core/inc/azure/core/internal/log.hpp
antkmsft: This is in the class that is in `Internal` namespace, customers should not be calling it.

sdk/core/azure-core/inc/azure/core/internal/log.hpp
antkmsft: This is in the class that is in `Internal` namespace, customers should not be calling it.

sdk/core/azure-core/inc/azure/core/internal/log.hpp
antkmsft: This is used internally by `Logger::SetListener()`:
```
void Logger::SetListener(Logger::Listener listener)
{
  std::unique_lock<std::shared_timed_mutex> loggerLock(g_logListenerMutex);
  g_logListener = std::move(listener);
  Log::ToggleLogging(g_logListener != nullptr);
}
```

ToggleLogging is inline function that sets an atomic boolean flag:
```
static AZ_CORE_DLLEXPORT std::atomic<bool> g_isLoggingEnabled;
static AZ_CORE_DLLEXPORT std::atomic<LogLevelInt> g_logLevel;
```

That boolean flag ends up being used by `Log::ShouldWrite()` inline:
```
static bool ShouldWrite(Logger::Level level) { return GetLogListener(level) != nullptr; }
```

`GetLogListener(Logger::Level level)` is inline and does lock-free check, based on that atomic boolean flag, and atomic logLevel.

```
static Logger::Listener GetLogListener(Logger::Level level)
    {
      return g_isLoggingEnabled && static_cast<LogLevelInt>(level) <= g_logLevel
          ? GetLogListener()
          : Logger::Listener(nullptr);
    }
```

Finally, `GetLogListener()` (without arguments) uses mutex and is non-inline.
```
Logger::Listener Log::GetLogListener()
{
  std::unique_lock<std::shared_timed_mutex> loggerLock(g_logListenerMutex);
  return g_logListener;
}
```

sdk/core/azure-core/inc/azure/core/logger.hpp
antkmsft: Log class is for internal functions - `ShouldWrite()` and `Write()`, the same way as Logger is for `SetListener()` and `SetLevel()`.

Our team really dislikes friend class declarations. This is why there are no private methods in the `Logger` class.

[Other side effect, if we have `Write()` and `ShouldWrite()` as free functions in Internal namespace, `Level` cannot be declared inside `Logger` class, it has to be outside. That's because to declare a function from another namespace, it has to be forward-declared before the class declaration. And in order to even forwarddeclare `ShouldLog()`, `Logger::Level` needs to be known at the point of forward-declaration (because `ShouldLog()` takes it as argument), which is not possible - circular dependecny. But it is always possible to forward-declare `Internal::Log` class, and make it a friend.]

sdk/core/azure-core/inc/azure/core/http/policy.hpp
ahsonkhan: Why do we prefer the name `LogPolicy` over `LoggingPolicy`?

.Net: [LoggingPolicy](https://github.com/Azure/azure-sdk-for-net/blob/49d9b2c3be3d48fc6d8f7044ae74c20c2cfc34c0/sdk/core/Azure.Core/src/Pipeline/Internal/LoggingPolicy.cs#L16)
Java: [HttpLoggingPolicy](https://github.com/Azure/azure-sdk-for-java/blob/d7bcc73f3866dbfb007a6ae8512b1fb01ac58ada/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLoggingPolicy.java#L39)
Python: [HttpLoggingPolicy](https://github.com/Azure/azure-sdk-for-python/blob/4f1c4a7bff4cf9a50f5d882e1f8adf8663adcadb/sdk/core/azure-core/azure/core/pipeline/policies/_universal.py#L347)
JS: [LogPolicy](https://github.com/Azure/azure-sdk-for-js/blob/50c5a5447454c2ae1d2b4fa5b02470671f41dd77/sdk/core/core-http/src/policies/logPolicy.ts#L45)

JeffreyRichter:
I'm looking for naming consistency. If we have a log policy with log options and a log class for setting listeners/level. Or, a logging policy, with logging options, and a logging class. Or, a logger policy with logger options and a logger class.

Consistent naming helps the user understand that there's a relationship between all these things.

This is why I also don't like having both a log class and a logger class.

sdk/core/azure-core/inc/azure/core/logger.hpp
antkmsft: @JeffreyRichter, internal functions look better when inside a static class, especially that now it has data members and typedefs ().

I don't think that names are that confusing, given that they do live in different namespaces, and SDK code does not need to interact with the `Logger` class - only with the `Internal::Log` class.

`(Internal::)Log::ShouldWrite()` and `(Internal::)Log::Write()` looks readable to me.
`Logger::SetListener()`, `Logger::SetLevel()`, `Logger::Level`, `Logger::Level::Error` - readable as well.
We write to log, users listen to loggers.

sdk/core/azure-core/src/logger.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace {
+static std::shared_timed_mutex g_logListenerMutex;
+static Logger::Listener g_logListener(nullptr);
+} // namespace
+
+std::atomic<bool> Log::g_isLoggingEnabled(false);
+std::atomic<Log::LogLevelInt> Log::g_logLevel(static_cast<LogLevelInt>(Logger::Level::Warning));
+
+inline void Log::ToggleLogging(bool isEnabled) { g_isLoggingEnabled = isEnabled; }
```
JeffreyRichter, line 21: If we keep this (and I need to undersatnd the usage of it), then it should be renamed sisnce it doesn't actually Toggle; it sets, so EnableLogging(bool isEnabled)

sdk/core/azure-core/inc/azure/core/internal/log.hpp
JeffreyRichter: I don't understand. How do customers call SetLogLevel which is also in this Internal class?

sdk/core/azure-core/inc/azure/core/internal/log.hpp
JeffreyRichter: I still don't understand. If you're already using the read/writer lock, then why is toggling necessary? Just get the shared lock when getting a copy of the value and use the exclusive lock when setting the value - this is all you need.

sdk/core/azure-core/src/logger.cpp
antkmsft: I agree with the naming, it is probably better (I was thinking of it as "toggling on" and "toggling off", but "enable on"/"enable off" is less confusing).

This is needed, because it sets an atomic boolean variable, which is lock free.
The moment when our code tries to determine whether it should log:
1. It "calls" an **inline** function `ShouldWrite(Logger::Level logLevel)`
2. `ShouldLog(Level)` "calls" an **inline** function `GetLogListener(Logger::Level logLevel)`, and compares it to nullptr.
3. `GetLogListener(Logger::Level logLevel)` checks an atomic and lock free boolean static variable, `g_isLoggingEnabled`. If it is false, returns nullptr.
4. If `g_isLoggingEnabled` is true, it checks an atomic and lock free variable `g_logLevel`: logLevel of the message that we are trying to log should be `<= g_logLevel`.

As you can see, `ShouldWrite()` check is lock-free. Plus, it gets inlined, so the code ends up being as simple as checking a boolean variable, which is as fast as you can get. Maximum, you have two checks, when `g_isLoggingEnabled == true`. A second check is integer `=<` comparison, and is also inlined.

`g_isLoggingEnabled` gets set by `Logger::SetListener(listener)` : if (listener) is nullptr, `g_isLoggingEnabled` gets set to false, otherwise it is true. `g_isLoggingEnabled` can be read as "is g_logListener a non-null pointer"? Only that it is lock-free (we can't check `g_logListener` for being null without a lock. Even though it is an inexpensive lock, especially now, because we now use shared lock when reading, it is still not as fast as lock-free inline comparison (probably something like `CMP g_isLoggingEnabled, 0; JE :should_not_write; CMP EAX, g_logLevel; JG :should_not_write`). @ahsonkhan was suggesting that in WinHttp, logging may get responsible for 80% slowdown due to mutex.

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ option(RUN_LONG_UNIT_TESTS "Tests that takes more than 5 minutes to complete. No
 option(BUILD_STORAGE_SAMPLES "Build sample application for Azure Storage clients" OFF)
 option(BUILD_PERFORMANCE_TESTS "Build the performance test library" OFF)
 option(MSVC_USE_STATIC_CRT "(MSVC only) Set to ON to link SDK with static CRT (/MT or /MTd switch)." OFF)
+option(BUILTIN_ENV_LOGGER "Build suport for enabling logging to console when AZURE_LOG_LEVEL environment variable is set." ON)
```
vhvb1989, line 22: We use `BUILD_XXX_XXX` for all other optional components.
Can you rename to `BUILD_AZURE_ENV_LOGGER` or `BUILD_ENV_LOGGER` ?

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  return envLogLevelPtr;
+}
+
+// Log level textual representation, including space padding, matches slf4j and log4net.
+std::string const ErrorText = "ERROR";
+std::string const WarningText = "WARN ";
+std::string const InformationalText = "INFO ";
+std::string const VerboseText = "DEBUG";
+std::string const UnknownText = "?????";
```
vhvb1989, line 99: what if we move this as `static constexpr char[]`

sdk/core/azure-core/src/http/url.cpp
```diff
@@ -196,25 +215,7 @@ std::string Url::GetRelativeUrl() const
...
-    full_url += ":" + std::to_string(m_port);
-  }
-  if (!m_encodedPath.empty())
-  {
-    full_url += "/";
-  }
-  full_url += GetRelativeUrl();
-  return full_url;
-}
+std::string Url::GetAbsoluteUrl() const { return GetUrlWithoutQuery() + GetRelativeUrl(); }
```
vhvb1989, line 218: ❤️

sdk/core/azure-core/test/ut/logger.cpp
vhvb1989: yes, I am not asking to test the other policies, I am asking to test the logger when the message is an Http request and response.  But I think I got what you said

sdk/core/azure-core/src/http/url.cpp
```diff
@@ -177,6 +177,25 @@ void Url::AppendQueryParameters(const std::string& query)
...
+  }
+  full_url += m_host;
+  if (m_port != 0)
+  {
+    full_url += ":" + std::to_string(m_port);
+  }
+  if (!m_encodedPath.empty())
+  {
+    full_url += "/";
+  }
```
vhvb1989, line 195: Do we need to add `/` depending if the url has or not a path?
This integrates fine with the current implementation for calling GetUrlWithoutQuery() + GetRelativeUrl() for the absolute URL.
However, if you are calling `GetUrlWithoutQuery()` for logging, are expecting show the path as well?  
Right now, the path won't be displayed and only `/` when it has a path or nothing for no path.

Is this what you expect for logging?

For example, a request to `http://host.com/resource/foo` would be logged as `http://host.com/`
and
`http://host.com` is logged as `http://host.com`

sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -30,22 +30,63 @@ std::string TruncateIfLengthy(std::string const& s)
...
+      url += separ + qparam.first + '=';
+
+      if (options.AllowedHttpQueryParameters.find(qparam.first)
+          != options.AllowedHttpQueryParameters.end())
+      {
+        url += TruncateIfLengthy(qparam.second);
+      }
+      else
+      {
+        url += "[hidden]";
```
vhvb1989, line 56: Are other languages adding this `hidden` placeholder?  or are they just not logging the header at all?  We could try to follow the same approach here.

sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -30,22 +30,63 @@ std::string TruncateIfLengthy(std::string const& s)
...
+    else if (
+        options.AllowedHttpRequestHeaders.find(header.first)
+        != options.AllowedHttpRequestHeaders.end())
+    {
+      log << " : " << TruncateIfLengthy(header.second);
+    }
+    else
+    {
+      log << " [hidden]";
+    }
```
vhvb1989, line 82: Nit:  This code is duplicated 3 times, 
- Request -> query
- Request -> headers
- Response -> headers

Consider creating a helper method which takes the map, and the allowed values set and updates the ofstream.
Then if we want to change the `[hidden]` value or anything in the logic it would be just one place

vhvb1989 (commented):
Adding comments and questions.
I would like to understand about the expectation of logging a URL before giving my full approval. I will be pending to your reply

sdk/core/azure-core/src/environment_log_level_listener.cpp
antkmsft: Then we can't return `std::string const&` from the inline functin below - because it would be cunstructing strings on every call, and would be returning references to temporary objects.

sdk/core/azure-core/src/http/log_policy.cpp
antkmsft: Thank you, I'll unify it.
.NET says "REDACTED" - https://github.com/Azure/azure-sdk-for-net/blob/41a7cf2cca7badc4be320ce57aa8b47adcdff137/sdk/core/Azure.Core/src/Shared/HttpMessageSanitizer.cs

sdk/core/azure-core/src/http/url.cpp
antkmsft: If that's the behavior, this is not what I expect. I'll fix.

sdk/core/azure-core/src/http/log_policy.cpp
antkmsft: Thank you, I will try to unify.

vhvb1989 (approved):

sdk/core/azure-core/src/logger.cpp
ahsonkhan: > was suggesting that in WinHttp, logging may get responsible for 80% slowdown due to mutex.

Just to clarify, I think @vhvb1989 mentioned something about the perf framework showing a regression when we use `Context` (see https://github.com/Azure/azure-sdk-for-cpp/issues/1589). I didn't measure with winhttp yet (it doesn't log anything atm anyway), so I am not sure where that 80% came from.

CMakeLists.txt
ahsonkhan: Following what other SDKs do, logging should be off be default.
```suggestion
option(BUILD_AZURE_CONSOLE_LOGGER "Build support for enabling logging to console when AZURE_LOG_LEVEL environment variable is set." OFF)
```

sdk/core/azure-core/test/ut/logging.cpp
```diff
@@ -1,226 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
ahsonkhan, line 1: nit: leave this test file name as is.

this will also make it easier to review any test changes.

sdk/core/azure-core/test/ut/simplified_header.cpp
```diff
@@ -28,6 +28,8 @@ class DllExportTest {
 
 TEST(SimplifiedHeader, core)
 {
+  EXPECT_NO_THROW(Azure::Core::CaseInsensitiveMap imap);
```
ahsonkhan, line 31: Add test for the logging header in core.hpp.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -390,22 +391,46 @@ namespace Azure { namespace Core { namespace Http {
...
+  /**
+   * @brief Options for Azure::Core::Http::LogPolicy.
+   */
+  struct LogOptions
+  {
+    /**
+     * @brief HTTP query parameters that are allowed to be logged.
+     * @remark URL-encoded parameter names are expected.
+     */
+    std::set<std::string> AllowedHttpQueryParameters;
```
ahsonkhan, line 403: Do we have an example code where some client library or unit tests would set these and what the outcome would be?

Presumably we want to follow the similar defaults for storage and other client libraries like .NET, once storage starts adding logging.

How would the end user customize which headers to log?

ahsonkhan: Should we turn these headers on by default?

https://github.com/Azure/azure-sdk-for-net/blob/d9a83c970d456c4d93c308869d44a84b531f44a2/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L24-L49

@tg-msft what do you think?

sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -30,22 +30,63 @@ std::string TruncateIfLengthy(std::string const& s)
...
+
+  log << "HTTP Request : " << HttpMethodToString(request.GetMethod()) << " " << url;
+
+  for (auto const& header : request.GetHeaders())
+  {
+    log << "\n\t" << header.first;
+
+    if (header.second.empty())
+    {
+      log << " [empty]";
```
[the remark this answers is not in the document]
vhvb1989: yes, things like `headerKey;` is fine.  Headers can be used as flags for the server only with no value.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -390,22 +391,46 @@ namespace Azure { namespace Core { namespace Http {
...
+
+    /**
+     * @brief HTTP request headers that are allowed to be logged.
+     */
+    Azure::Core::CaseInsensitiveSet AllowedHttpRequestHeaders;
```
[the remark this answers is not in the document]
antkmsft: Because we treat headers as case-insensitive. They are stored and returned as `CaseInsensitiveMap`.

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -59,15 +63,15 @@ set(
     inc/azure/core/internal/null_body_stream.hpp
     inc/azure/core/internal/strings.hpp
     inc/azure/core/io/body_stream.hpp
-    inc/azure/core/logging/logging.hpp
     inc/azure/core/base64.hpp
-    inc/azure/core/case_insensitive_map.hpp
+    inc/azure/core/case_insensitive_containers.hpp
```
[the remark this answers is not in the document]
antkmsft: We end up using `CaseInsensitiveSet` in this PR.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 - Renamed `TelemetryPolicyOptions` to `TelemetryOptions`.
 - Renamed `ValuePolicyOptions` to `ValueOptions`.
 - Removed `StartTry()` from `Azure::Core::Http::Request`.
+- Changed type of `Azure::Core::Http::RetryOptions::StatusCodes` from `std::vector` to `std::set`.
+- Renamed `Azure::Core::Http::LoggingPolicy` to `LogPolicy`.
+- Introduced `Azure::Core::Http::LogOptions`, a mandatory parameter for `LogPolicy` construction. Entities that are not specified in the allow lists are hidden in the log.
```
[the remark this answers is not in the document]
antkmsft: Because we don't want the clients to accidentally skip initializing the policy. If you are not providing options, it is likely a mistake. At least, if you don't want to have log options in your client, which is unlikely, initialize it with with default explicitly. RetryPolicy doesn't have default parameter either.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
ahsonkhan: I get that, but why?

sdk/core/azure-core/src/http/log_policy.cpp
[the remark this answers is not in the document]
ahsonkhan: In that case, @antkmsft, don't output "[empty]" but rather output nothing for the value, just `key:` is fine.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
vhvb1989: Well, we might want to log only headers we are setting, like the requestID, but not the one the server returns back, or vise versa. Having allow list per req and res help to those scenarios.
But those cases could be just very few; then having one single Set for all headers would be enough and easier to set up.

A more complex solution would be to create a class `LogHttpElement { name, asRequestHeader, asResponseHeader, asQuery }`  , and make a Set<LogHttpElement>.  Then each record can be set up to support even the case where there is a query and header with the same name and we want to log it.  But might not be worth it such a complexity to be learn by customers.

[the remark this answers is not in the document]
vhvb1989: And for the another question @ahsonkhan  , using `std::set<string>` would require to always perform a ToLower() to the key while trying to find it in the Set. (Or saving all values as toLower and forgeting about the original value).
The CaseInsensitiveMap keeps the original value plus the toLower version for comparing

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 - Renamed `TelemetryPolicyOptions` to `TelemetryOptions`.
 - Renamed `ValuePolicyOptions` to `ValueOptions`.
 - Removed `StartTry()` from `Azure::Core::Http::Request`.
+- Changed type of `Azure::Core::Http::RetryOptions::StatusCodes` from `std::vector` to `std::set`.
```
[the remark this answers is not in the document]
antkmsft: That's because repetition doesn't do anything, and order does not matter. Also, lookup is faster.
This is a tiny change, and I touched related code/neighboring lines.

CMakeLists.txt
antkmsft: This is not about the logging being off, it is about having the code to listen to the environment variable `AZURE_LOG_LEVEL` on start up, and enabling console logging IF `AZURE_LOG_LEVEL` env variable is set in the environment.
Java has that functionality ON by default - https://github.com/Azure/azure-sdk-for-java/blob/5672b860b15aa32d1dd5b559f47c3cb1058f0166/sdk/core/azure-core/src/main/java/com/azure/core/implementation/logging/DefaultLogger.java#L66 - in fact, I don't think you can even compile out that functionality from the Java SDK.

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -87,7 +91,7 @@ set(
     src/cryptography/md5.cpp
     src/http/bearer_token_authentication_policy.cpp
     src/http/http.cpp
-    src/http/logging_policy.cpp
+    src/http/log_policy.cpp
```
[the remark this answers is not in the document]
antkmsft: I like the name that Jeffrey proposed more. Also, JS SDK uses LogPolicy.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -390,22 +391,46 @@ namespace Azure { namespace Core { namespace Http {
...
   /**
    * @brief Logs every HTTP request.
    *
-   * @details Logs every HTTP request, response, or retry attempt.
-   * @remark See #logging.hpp
+   * @details Logs every HTTP request and response.
+   * @remark See Azure::Core::Logger.
    */
-  class LoggingPolicy : public HttpPolicy {
+  class LogPolicy : public HttpPolicy {
```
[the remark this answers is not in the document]
antkmsft: Jeffrey asked for it, and I also like it more.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -390,22 +391,46 @@ namespace Azure { namespace Core { namespace Http {
...
+    LogOptions m_options;
+
   public:
     /**
      * @brief Constructs HTTP logging policy.
      */
-    explicit LoggingPolicy() {}
+    explicit LogPolicy(LogOptions options) : m_options(std::move(options)) {}
```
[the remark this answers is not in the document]
antkmsft: I should not be optional.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -390,22 +391,46 @@ namespace Azure { namespace Core { namespace Http {
...
   };
 
+  /**
+   * @brief Options for Azure::Core::Http::LogPolicy.
+   */
+  struct LogOptions
+  {
+    /**
+     * @brief HTTP query parameters that are allowed to be logged.
+     * @remark URL-encoded parameter names are expected.
```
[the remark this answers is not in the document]
antkmsft: Good question, I'll check. Could be a good argument for keeping them non-encoded.
Encoded requirement was due to perf considerations.

sdk/core/azure-core/inc/azure/core/case_insensitive_containers.hpp
```diff
@@ -11,6 +11,7 @@
 #include "azure/core/internal/strings.hpp"
 
 #include <map>
+#include <set>
```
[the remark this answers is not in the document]
antkmsft: We never released this header.

sdk/core/azure-core/inc/azure/core/case_insensitive_containers.hpp
```diff
@@ -22,4 +23,9 @@ namespace Azure { namespace Core {
   using CaseInsensitiveMap
       = std::map<std::string, std::string, Internal::Strings::CaseInsensitiveComparator>;
 
+  /**
+   * @brief A type alias of `std::set<std::string>` with case-insensitive key comparison.
+   */
+  using CaseInsensitiveSet = std::set<std::string, Internal::Strings::CaseInsensitiveComparator>;
```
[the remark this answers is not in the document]
antkmsft: Yes, we need it for LogOptions.

sdk/core/azure-core/inc/azure/core.hpp
```diff
@@ -38,6 +39,3 @@
 
 // azure/core/io
 #include "azure/core/io/body_stream.hpp"
-
-// azure/core/logging
-#include "azure/core/logging/logging.hpp"
```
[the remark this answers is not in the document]
antkmsft: Jeffrey asked for it, and I also prefer Log over Logging.

sdk/core/azure-core/inc/azure/core/internal/client_options.hpp
```diff
@@ -81,6 +81,12 @@ namespace Azure { namespace Core { namespace Internal {
      * @brief Telemetry options.
      */
     Azure::Core::Http::TelemetryOptions Telemetry;
+
+    /**
+     * @brief Define the information to be used for logging.
+     *
+     */
+    Azure::Core::Http::LogOptions Log;
```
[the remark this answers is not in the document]
antkmsft: Jeffrey asked for it, and I also prefer Log over Logging.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * @brief This header defines the types and functions your application uses to be notified of Azure
+ * SDK client library log messages.
+ */
+
+#pragma once
+
+#include <functional>
+#include <string>
+
+namespace Azure { namespace Core {
```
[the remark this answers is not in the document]
antkmsft: We moved everything from that namespace into a "static class" Logger, this was one of the first changes we made with Jeffrey when we started this refactoring.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * @brief Log message handler.
+   */
+  class Logger {
+  public:
+    /**
+     * @brief Log message level.
+     */
+    // https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/LogLevel.java
```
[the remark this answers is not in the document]
antkmsft: I agree, but this is not a doxygen comment, but a dev comment.
I find it useful to have it here.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+    enum class Level : int
```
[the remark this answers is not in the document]
antkmsft: This enum is in "static class" called `Logger`, so it is `Logger::Level`. That's why it is `Logger::SetLevel()` and not `Logger::SetLogLevel()`. The same for listener.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    /**
+     * @brief Defines the signature of the callback function that application developers must write
+     * in order to receive Azure SDK log messages.
+     *
+     * @param level The log message level.
+     * @param message The log message.
+     */
+    typedef std::function<void(Level level, std::string const& message)> Listener;
```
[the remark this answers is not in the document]
antkmsft: See comment about `LogLevel`. This was also one of the first thing that we did with Jeffrey after we moved `Logging` namespace into a static class, we dropped "Log" name prefixes as unnecessary in that context.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @brief Set the function that will be invoked to report an SDK log message.
+     *
+     * @param listener An #Azure::Core::Logger::Listener function that will be invoked when
+     * the SDK reports a log message. If `nullptr`, no function will be invoked.
+     */
+    static void SetListener(Listener listener);
```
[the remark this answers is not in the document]
antkmsft: This rename was intentional, after working together with Jeffrey.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @brief Sets the #Azure::Core::Logger::Level an application is interested in receiving.
+     *
+     * @param level Maximum log level.
+     */
+    static void SetLevel(Level level);
```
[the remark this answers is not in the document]
antkmsft: This rename was intentional, after working together with Jeffrey.

sdk/core/azure-core/inc/azure/core/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+  private:
+    Logger() = delete;
+    ~Logger() = delete;
```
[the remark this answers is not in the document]
antkmsft: == "static class".
So that users won't be able to create instances of this class, they would be useless (`auto logger = Azure::Core::Logger()` that does not have any non-static class members).

sdk/core/azure-core/inc/azure/core/internal/log.hpp
```diff
@@ -3,9 +3,37 @@
...
+
+namespace Azure { namespace Core { namespace Internal {
+  class Log {
+    using LogLevelInt = std::underlying_type<Logger::Level>::type;
+
+    static_assert(
+        std::is_same<int, LogLevelInt>::value == true && ATOMIC_INT_LOCK_FREE == 2,
+        "Logger::Level values must be representable as lock-free");
+
+    static_assert(ATOMIC_BOOL_LOCK_FREE == 2, "atomic<bool> must be lock-free");
```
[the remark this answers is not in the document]
antkmsft: A message supplied to `static_assert()` clarifies, and there is documentation for `ATOMIC_BOOL_LOCK_FREE`.
The only thing I coulddo would be to link to a C++ documentation for `ATOMIC_BOOL_LOCK_FREE`, but then why do it for `ATOMIC_BOOL_LOCK_FREE`, and not for `std::strcmp`, for example.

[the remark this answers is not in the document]
antkmsft: If you have some specific comment in mind, please let me know.

sdk/core/azure-core/inc/azure/core/internal/log.hpp
```diff
@@ -3,9 +3,37 @@
...
 
-namespace Azure { namespace Core { namespace Logging { namespace Internal {
-  bool ShouldLog(LogLevel level);
-  void Log(LogLevel level, std::string const& message);
-}}}} // namespace Azure::Core::Logging::Internal
+#include <atomic>
+#include <type_traits>
+
+namespace Azure { namespace Core { namespace Internal {
+  class Log {
```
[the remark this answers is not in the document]
antkmsft: So it can have private members and typedefs:
* `LogLevelInt`
* `g_isLoggingEnabled`
* `g_logLevel`

sdk/core/azure-core/src/environment_log_level_listener_private.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+  public:
+    static Logger::Level GetLogLevel(Logger::Level defaultValue);
+    static Logger::Listener GetLogListener();
+  };
+
+#if defined(AZ_NO_ENV_LOGGER) \
+    || (defined(WINAPI_PARTITION_DESKTOP) \
+        && !WINAPI_PARTITION_DESKTOP) // See azure/core/platform.hpp for explanation.
+  inline Logger::Level EnvironmentLogLevelListener::GetLogLevel(Logger::Level defaultValue)
```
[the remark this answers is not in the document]
antkmsft: There is no environment on UWP, at all.

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+#pragma warning(disable : 4996)
+#endif
+    auto envVar = std::getenv("AZURE_LOG_LEVEL");
+#if defined(_MSC_VER)
+#pragma warning(pop)
+#endif
+
+    if (envVar)
+    {
+      auto const logLevelStr = Internal::Strings::ToLower(envVar);
```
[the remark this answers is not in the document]
antkmsft: LoLower() is safe for empty strings, why add 3 more lines and extra point of return to optimize a few instructions that only get executed once on startup?

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+      return WarningText;
+
+    case Logger::Level::Informational:
+      return InformationalText;
+
+    case Logger::Level::Verbose:
+      return VerboseText;
+
+    default:
+      return UnknownText;
```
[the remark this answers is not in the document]
antkmsft: Logging should not throw.

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+      else if (
+          logLevelStr == "informational" || logLevelStr == "info" || logLevelStr == "information"
+          || logLevelStr == "2")
+      {
+        envLogLevel = Logger::Level::Informational;
+      }
+      else if (logLevelStr == "verbose" || logLevelStr == "debug" || logLevelStr == "1")
+      {
+        envLogLevel = Logger::Level::Verbose;
+      }
```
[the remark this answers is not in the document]
antkmsft: I think it is useful to mimic Java here, since we're using the same env variable name.

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  return envLogLevelPtr;
+}
+
+// Log level textual representation, including space padding, matches slf4j and log4net.
+std::string const ErrorText = "ERROR";
+std::string const WarningText = "WARN ";
+std::string const InformationalText = "INFO ";
+std::string const VerboseText = "DEBUG";
```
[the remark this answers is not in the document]
antkmsft: I am mimicking slf4j and log4net - both names and padding.

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+        envLogLevelPtr = nullptr;
+      }
+    }
+  }
+
+  return envLogLevelPtr;
+}
+
+// Log level textual representation, including space padding, matches slf4j and log4net.
+std::string const ErrorText = "ERROR";
```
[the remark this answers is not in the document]
antkmsft: We would have it as
```
[2021-01-01T14:51:57.9327840Z] DEBUG : [CURL Transport Adapter]: Creating a new session.
[2021-01-01T14:52:34.5490580Z] INFO  : HTTP status code 500 will be retried.
[2021-01-01T14:52:34.5490580Z] INFO  : HTTP Retry attempt #2 will be made in 4000ms.
[2021-01-01T14:52:38.0124670Z] WARN  : HTTP Transport error: ...
```
I do not see how the proposed format is functionally different other than a personal preference.
I, for instance, like the fact that the log messages are aligned.

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+std::string const WarningText = "WARN ";
+std::string const InformationalText = "INFO ";
```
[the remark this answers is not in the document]
antkmsft: I like that the log messages are evenly aligned.

sdk/core/azure-core/src/environment_log_level_listener.cpp
```diff
@@ -0,0 +1,145 @@
...
+Logger::Listener EnvironmentLogLevelListener::GetLogListener()
+{
+  bool const isEnvLogLevelSet = GetEnvironmentLogLevel() != nullptr;
+  if (!isEnvLogLevelSet)
+  {
+    return nullptr;
+  }
+
+  static Logger::Listener const consoleLogger = [](auto level, auto message) {
+    std::cerr << '[' << Azure::Core::DateTime(std::chrono::system_clock::now()) << "] "
```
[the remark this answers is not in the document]
antkmsft: They are capable of doing both:
* https://stackify.com/log4net-guide-dotnet-logging/
* https://examples.javacodegeeks.com/enterprise-java/slf4j/slf4j-tutorial-beginners/

Given that our default logger is non-configurable, but is a way to diagnose without recompiling, I think it is better to utilize  StdErr. Leave StdOut to customer, they are very likely using std::printfs and std::couts. They can still see their output on the console, but redirect logs to a file `CustomerTool.exe 2>AzSdkLog.txt`. They also can redirect their output and our logs to different files: `CustomerTool.exe > customerOutput.txt 2>AzSdkLog.txt`
Also, another advantage of StdErr vs StdOut by default is that it is non-buffered.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/stdin.html

The name "Err" must be confusing, but POSIX intended it for _diagnostic_ output, not necessarily errors:
"At program start-up, three streams shall be predefined and need not be opened explicitly: standard input (for reading conventional input), standard output (for writing conventional output), and standard error (for writing diagnostic output). When opened, the standard error stream is not fully buffered; the standard input and standard output streams are fully buffered if and only if the stream can be determined not to refer to an interactive device."

sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -30,22 +30,63 @@ std::string TruncateIfLengthy(std::string const& s)
...
-    log << "\n\t" << header.first << " : " << TruncateIfLengthy(header.second);
+    auto separ = '?';
+    for (auto const& qparam : requestUrl.GetQueryParameters())
+    {
+      url += separ + qparam.first + '=';
+
+      if (options.AllowedHttpQueryParameters.find(qparam.first)
+          != options.AllowedHttpQueryParameters.end())
+      {
+        url += TruncateIfLengthy(qparam.second);
```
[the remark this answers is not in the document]
antkmsft: I'll remove it.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
tg-msft: Yes, and Storage should add https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/src/BlobClientOptions.cs#L128 which we worked carefully with their team on (and used across all the other languages).

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
vhvb1989: People can use our function `Encode()` if they need for adding the value here.
That's what they need to do for adding headers to the Request.  

Maybe just add a note about that in the comments, about our public Encode and Decode helpers methods

sdk/core/azure-core/inc/azure/core/http/policy.hpp
ahsonkhan: @tg-msft do you expect users to pass in query parameter strings for the allow list that already URL-encoded? The docs don't specify.
https://github.com/Azure/azure-sdk-for-net/blob/d9a83c970d456c4d93c308869d44a84b531f44a2/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L85-L88

And why do you ignore casing when comparing headers and query parameter names?

https://github.com/Azure/azure-sdk-for-net/blob/675631c134f1ae01fe170bb3815f29cec96ea1fc/sdk/core/Azure.Core/src/Shared/HttpMessageSanitizer.cs#L27
https://github.com/Azure/azure-sdk-for-net/blob/675631c134f1ae01fe170bb3815f29cec96ea1fc/sdk/core/Azure.Core/src/Shared/HttpMessageSanitizer.cs#L100

tg-msft: I don't know if Azure has any query parameter names that need to be URL encoded so I'm not sure if there's a difference in practice.  I'm guessing our default is unencoded but @pakrym could correct me if I'm wrong.  HTTP header names are case insensitive.  Query parameter names are case sensitive, but I'm guessing we erred on the side of overabundant caution so nothing gets leaked if customers accidentally specified the wrong case (and again, I don't know of any distinct Azure query parameter names that are the same modulo case so I don't think it's a problem in practice).

pakrym: I never really thought about urlencoding the query parameter names so both the UrlBuilder and sanitizer expect already pre-encoded names. 

>Query parameter names are case sensitive, but I'm guessing we erred on the side of overabundant caution so nothing gets leaked if customers accidentally specified the wrong case (and again, I don't know of any distinct Azure query parameter names that are the same modulo case so I don't think it's a problem in practice).

Yep, we are yet to see a scenario where you would like to make an opposite decision for headers that differ only in the casing. Missing a header during the sanitization feels like a more common case.

antkmsft:
Per agreement with the team, merging together all the API review changes otherwise extractable to several PRs, to win us time on namespace/file moving changes. If there's any unaddressed feedback left, will be fixed in subsequent PRs. I addressed all the feedback that I took to my best knowledge, if something is not reflected in the code changes, I'll reply with the comment mentioning the fix was not taken intentionally.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
ahsonkhan: > Well, we might want to log only headers we are setting, like the requestID, but not the one the server returns back, or vise versa. Having allow list per req and res help to those scenarios.
But those cases could be just very few; then having one single Set for all headers would be enough and easier to set up.

Is that an actual scenario though? Decided which header to log isn't up to us, but rather the end customer. Why would they differentiate request vs response? .NET for example doesn't differentiate it.

@tg-msft and @pakrym should we have separate allow lists for request headers  and response headers, or just one list for any headers (in either direction)?

sdk/core/azure-core/inc/azure/core/http/policy.hpp
ahsonkhan: > HTTP header names are case insensitive. 

OK, that makes sense for headers.

> but I'm guessing we erred on the side of overabundant caution so nothing gets leaked if customers accidentally specified the wrong case (and again, I don't know of any distinct Azure query parameter names that are the same modulo case so I don't think it's a problem in practice).

Regarding query parameters, I would have expected the opposite, if we wanted to avoid accidental leaks. Wouldn't having strict matching in the allow list ensure that some other parameters don't get leaked that weren't explicitly allowed? With case insensitive matching, we are allowing more today.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
pakrym: >@tg-msft and @pakrym should we have separate allow lists for request headers and response headers, or just one list for any headers (in either direction)?

I've never heard a scenario where somebody wanted to log the header in one direction but not the other.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
tg-msft: You're right, sorry, these are allow lists rather than deny lists.  I'm guessing we just did that for usability then.  I'm not particularly worried about it though because I've never seen distinct query parameter names in Azure that only differ by case.  The REST API review board would never allow them through.

CMakeLists.txt
ahsonkhan: That 's a bit confusing. Do we really need this env var? Why don't we always look for `AZURE_LOG_LEVEL` and if it is set, we honor it/turn logging on.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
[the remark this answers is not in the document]
ahsonkhan: Pavel also shared below what they do in .NET, which is to expect the user to encode them (if needed), and Azure service query parameters typically don't need to be url encoded:
https://github.com/Azure/azure-sdk-for-cpp/pull/1752#discussion_r588585510

> Maybe just add a note about that in the comments, about our public Encode and Decode helpers methods

Good idea.

sdk/core/azure-core/inc/azure/core/logger.hpp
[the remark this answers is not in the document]
ahsonkhan: That makes sense now, thanks.
nit: consider adding that as a dev comment

PR #1753, opened 2021-02-26
Reserve Claims in TokenRequestOptions to be used when CAE support is implemented

### LINKED ISSUE #1601: Implement Continuous Access Evaluation
Add a Claims field of type String to support Continuous Access feature.

antkmsft:
Both Java and .NET have Claims as string:
* Java - https://github.com/Azure/azure-sdk-for-java/blob/a4f73634f6044e479fd19c0fa18514f2bd7f0128/sdk/core/azure-core/src/main/java/com/azure/core/credential/TokenRequestContext.java#L16
* .NET - https://github.com/Azure/azure-sdk-for-net/blob/e3d0298d92883afd722d400de65f1cfe7134ea11/sdk/core/Azure.Core/src/TokenRequestContext.cs#L49

We currently don't have Continuous Access support, this field will be used when it gets implemented.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/http/policy.hpp
+++ b/sdk/core/azure-core/inc/azure/core/http/policy.hpp
@@ -318,6 +318,11 @@ namespace Azure { namespace Core { namespace Http {
      * @brief Authentication scopes.
      */
     std::vector<std::string> Scopes;
+
+    /**
+     * @brief Authentication claims.
+     */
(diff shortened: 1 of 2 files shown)


vhvb1989 (commented):
changelog and tests?

RickWinter (changes requested):
Can you elaborate on why this must be added before its supported?

PR #1756, opened 2021-02-26
Fix UWP build

### LINKED ISSUE #1728: UWP build is broken
```
hkeyholder.hpp(37): error C2039: 'RegCloseKey': is not a member of '`global namespace'' (compiling source file telemetry_policy.cpp)
hkeyholder.hpp(37): error C3861: 'RegCloseKey': identifier not found (compiling source file telemetry_policy.cpp)
```

(It is HKeyHolder)

antkmsft:
Closes #1728.

The failure was found via private build on VcPkg, the fix was verified via private build on VcPkg - all builds (arm-uwp)

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
+++ b/sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
@@ -8,6 +8,8 @@
  */
 #pragma once
 
+#include "azure/core/platform.hpp"
+
 #if defined(AZ_PLATFORM_WINDOWS)
 #if !defined(WIN32_LEAN_AND_MEAN)
 #define WIN32_LEAN_AND_MEAN
@@ -17,6 +19,9 @@
 #endif
 
 #include <windows.h>
+
+#if !defined(WINAPI_PARTITION_DESKTOP) \
+    || WINAPI_PARTITION_DESKTOP // See azure/core/platform.hpp for explanation.
 
 namespace Azure { namespace Core { namespace Internal {
 
@@ -54,5 +59,7 @@ namespace Azure { namespace Core { namespace Internal {
   };
 
 }}} // namespace Azure::Core::Internal
+
+#endif
 
 #endif


sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
```diff
@@ -18,6 +20,9 @@
 
 #include <windows.h>
 
+#if !defined(WINAPI_PARTITION_DESKTOP) \
+    || WINAPI_PARTITION_DESKTOP // See azure/core/platform.hpp for explanation.
```
ahsonkhan, line 24: What will happen on UWP if the whole type is under this if/def?

antkmsft: t won't be available, unknown identifier. The same as with Http Transport Adapters.

ahsonkhan: Correct me if I am mistaken, but does that mean we just fixed the UWP build issue, but would anyone running our SDK on UWP would end up with some runtime error or wrong behavior? If folks can't run it on that platform, do we need to support it?

antkmsft: It won't compile, no runtime errors. Which is fine - UWP doesn't have registry, doesn't even have environment variables, so what would you use that class from, it would be useless on that platform, and that compilation issue is a good sign that you went into a code that you need to `#ifdef` for non-UWP case. Which is fine, it is exactly the same if you attempt to use this class on Linux, which also doesn't have registry.

RickWinter (approved):

ahsonkhan (approved):

sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp
ahsonkhan: This is our telemetry output for UWP atm:
https://github.com/Azure/azure-sdk-for-cpp/blob/ae575b0c0838f73ef666270260c3c0385d097e78/sdk/core/azure-core/src/http/telemetry_policy.cpp#L67

Issue #1742, opened 2021-02-25
Azure::Core APIs that accept Context should take it as the last parameter (but required)


ahsonkhan:
This will be consistent with the client library APIs, which continue to have it be the last parameter, but optional.

PR #1779, opened 2021-03-05
For consistency within the SDK, move Azure::Core::Context to be the last parameter for various Azure::Core APIs.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1742


sdk/core/azure-core/inc/azure/core/credentials.hpp
```diff
@@ -47,12 +47,13 @@ namespace Azure { namespace Core {
...
-     * @param context #Azure::Core::Context so that operation can be cancelled.
      * @param tokenRequestOptions Options to get the token.
+     * @param context #Azure::Core::Context so that operation can be cancelled.
+     *
      */
     virtual AccessToken GetToken(
-        Context const& context,
-        Http::TokenRequestOptions const& tokenRequestOptions) const = 0;
+        Http::TokenRequestOptions const& tokenRequestOptions,
+        Context const& context) const = 0;
```
ahsonkhan, line 56: I am not sure about this one, since it is being moved after options. Should these options be optional and defaulted?

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -21,6 +22,12 @@
 - Renamed `TelemetryPolicyOptions` to `TelemetryOptions`.
 - Renamed `ValuePolicyOptions` to `ValueOptions`.
 - Removed `StartTry()` from `Azure::Core::Http::Request`.
+- Move `Azure::Core::Context` to be the last parameter for consistency, instead of first in various azure-core types. For example:
+  - `BodyStream::Read(uint8_t* buffer, int64_t count, Azure::Core::Context const& context)`
+  - `BodyStream::ReadToEnd(BodyStream& body, Azure::Core::Context const& context)`
+  - `HttpPolicy::Send(Request& request, NextHttpPolicy policy, Azure::Core::Context const& context)`
+  - `Operation<T>::PollUntilDone(std::chrono::milliseconds period, Azure::Core::Context& context)`
+  - `TokenCredential::GetToken(Http::TokenRequestOptions const& tokenRequestOptions, Azure::Core::Context const& context)`
```
ahsonkhan, line 30: Let me know if you find any that I might have missed.
I did a search for `"(Context"` and variant of partial namespace `"(Azure::Core::Context"`

sdk/core/azure-core/test/ut/curl_session_test.cpp
```diff
@@ -28,7 +28,7 @@ namespace Azure { namespace Core { namespace Test {
     EXPECT_CALL(*curlMock, SendBuffer(_, _, _)).WillOnce(Return(CURLE_OK));
     EXPECT_CALL(*curlMock, ReadFromSocket(_, _, _))
         .WillOnce(DoAll(
-            SetArrayArgument<1>(response.data(), response.data() + response.size()),
+            SetArrayArgument<0>(response.data(), response.data() + response.size()),
```
ahsonkhan, line 31: @vhvb1989 please verify this mocking framework update makes sense.

sdk/keyvault/azure-security-keyvault-common/inc/azure/keyvault/common/internal/keyvault_pipeline.hpp
```diff
@@ -148,7 +148,7 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Common { n
     {
       auto request = CreateRequest(method, path);
       // Use the core pipeline directly to avoid checking the response code.
-      return m_pipeline.Send(context, request);
+      return m_pipeline.Send(request, context);
```
ahsonkhan, line 151: Anything else in keyvault need changing, @vhvb1989? Any implementation detail or APIs where context is first?

sdk/core/azure-core/inc/azure/core/credentials.hpp
ahsonkhan: cc @antkmsft

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
```diff
@@ -1141,7 +1141,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
                     std::to_string(setPropertiesOptions.Timeout.GetValue())));
           }
           request.AddHeader(Details::HeaderVersion, setPropertiesOptions.ApiVersionParameter);
-          return SetPropertiesParseResult(context, pipeline.Send(context, request));
+          return SetPropertiesParseResult(context, pipeline.Send(request, context));
```
ahsonkhan, line 1144: FYI @Jinming-Hu, I left all these "Result" methods untouched across all of storage, but they all take context as the first parameter. I only focused on common public types from core.

Are they meant to be public?

Consider changing the auto-gen to move the context parameter to be last.

For example, we may want to have it be as follows:
```C++
return SetPropertiesParseResult(pipeline.Send(request, context), context);
```

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
```diff
@@ -1167,7 +1167,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
                     std::to_string(getPropertiesOptions.Timeout.GetValue())));
           }
           request.AddHeader(Details::HeaderVersion, getPropertiesOptions.ApiVersionParameter);
-          return GetPropertiesParseResult(context, pipeline.Send(context, request));
+          return GetPropertiesParseResult(context, pipeline.Send(request, context));
```
ahsonkhan, line 1170: Same here, and elsewhere. If you search for `"Result(context"`, you'll get ~50 results.

sdk/core/azure-core/inc/azure/core/credentials.hpp
ahsonkhan: @JeffreyRichter - let me know what you think here around `TokenCredential::GetToken`
```C++
TokenCredential::GetToken(Http::TokenRequestOptions const& tokenRequestOptions, Azure::Core::Context const& context)
```

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
Jinming-Hu: @ahsonkhan , I believe they're not meant to be public. But auto-gen of file and datalake services are maintained by @katmsft

sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/protocol/share_rest_client.hpp
Jinming-Hu: @katmsft

Jinming-Hu (approved):
Changes to storage sdk LGTM

vhvb1989:
Well, I don't like this, if possible, I would like to re-consider it.
It is fine for all services to take the Context at the end because it is considered an optional parameter.
But that's not the case for Azure Core. When it comes to Azure Core, we want it to be seen as important

sdk/core/azure-core/test/ut/curl_session_test.cpp
vhvb1989: looks good, but I still don't like this change to Core.
If @RickWinter  and @JeffreyRichter  want it, that's fine anyway.
Consider that keeping the Context as the last argument for all these APIs will require adding any new parameter in the middle

sdk/keyvault/azure-security-keyvault-common/inc/azure/keyvault/common/internal/keyvault_pipeline.hpp
vhvb1989: All API's uses this common function to send a request. :)

ahsonkhan:
Victor you were present in the meetings where this was discussed and agreed upon :)

Also from arch board review:
image
image

vhvb1989:
I remember the meeting, and also that I mentioned the reasons why I don't think it's a good idea for Azure Core.
There isn't any mechanism which can help us to always keep the context at the end. We will depend on manual supervision (PR and Api review inspection).  Otherwise, we can end up having inconsistency.

RickWinter (approved):

RickWinter:
> I remember the meeting, and also that I mentioned the reasons why I don't think it's a good idea for Azure Core.
> There isn't any mechanism which can help us to always keep the context at the end. We will depend on manual supervision (PR and Api review inspection). Otherwise, we can end up having inconsistency.

The same arguement could be said for always making it the first parameter.  Code reviews are required to enforce that as well.

sdk/core/azure-core/inc/azure/core/credentials.hpp
ahsonkhan: @schaabs and @antkmsft - can you please take a look at this particular change. If we make `TokenRequestOptions` optional, like most options, context would move to be the first parameter.

sdk/keyvault/azure-security-keyvault-common/inc/azure/keyvault/common/internal/keyvault_pipeline.hpp
ahsonkhan: I meant other than pipeline.Send. Are there any other KeyVault APIs that I need to update?

sdk/core/azure-core/inc/azure/core/credentials.hpp
antkmsft: There can be no default TokenRequestOptions, for the reason they all look different depending on the client, and no empty value would make any sense, you are writing an erroneous code if you put empty scopes here. This was discussed in https://github.com/Azure/azure-sdk-for-cpp/pull/1527#discussion_r568968684

Let's rename the struct from `*Options` to `TokenRequestContext` or anything of that sort - it seems the problem truly is between the name suffix and optionality; But let's not make this parameter to have default value, a bad and problematic one, to follow the pattern.

sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -273,14 +273,14 @@ namespace Azure { namespace Core { namespace Http {
...
         Request& request,
-        NextHttpPolicy nextHttpPolicy) const override
+        NextHttpPolicy nextHttpPolicy,
+        Context const& ctx) const override
     {
       auto uuid = Uuid::CreateUuid().ToString();
 
       request.SetHeader(RequestIdHeader, uuid);
-      return nextHttpPolicy.Send(ctx, request);
+      return nextHttpPolicy.Send(request, ctx);
```
antkmsft, line 283: nit: I've seen that `ctx` from the old times we started the repo, we started to use `context` later on. This might be a good opportunity to update all occurrences.

sdk/core/azure-core/inc/azure/core/credentials.hpp
ahsonkhan: Alright, I will leave it as I have it then for now.

Though unrelated to this PR, I would love to hear Scott's thoughts on that conversation you linked to Anton. What do other languages do here?

> Let's rename the struct

Can you please file an issue and we can settle on a name. `TokenRequestContext` is problematic given we have a Context type already and that's the other parameter being passed in here.

antkmsft (approved):
LGTM; I think that if there is something missing, compiler should catch it, but once it compiles, it should be good to go, it would mean you didn't miss any renames.
Maybe search for `"(Core::Context"` as well.
And I suggest to rename `ctx`s to `context`s, this is really good opportunity, and it will look nice in git History. But I am fine if you chose not to do this.

ahsonkhan:
> this is really good opportunity, and it will look nice in git History. But I am fine if you chose not to do this.

Since it isn't urgent, I will do that separately.

PR #1827, opened 2021-03-10
Remove BUILD_ENV_LOGGER cmake option and always honor AZURE_LOG_LEVEL environment variable, if it is set.

ahsonkhan:
I don't currently see a customer scenario to turn building the environment logger off. To reduce breaking changes, I propose we remove this capability for now and add it back once it is explicitly requested.

If we were to add this in the future, we'll need a CI leg which tests building with this option turned off.


antkmsft (changes requested):
Let's keep it. I would not want it in my programs. Some env variable is set, and now my app starts printing to console. You would not want it in any serious desktop app (Photoshop, MS Office, games). Why go after this build option, we have many others, never held them to the same standard of questioning whether their existence is justified.

antkmsft (approved):

vhvb1989 (approved):

PR #1849, opened 2021-03-10
DefaultAllowedHttpHeaders: Make static default const

antkmsft:


sdk/core/azure-core/inc/azure/core/http/policy.hpp
```diff
@@ -393,7 +393,7 @@ namespace Azure { namespace Core { namespace Http {
   };
 
   namespace _detail {
-    AZ_CORE_DLLEXPORT extern Azure::Core::CaseInsensitiveSet g_defaultAllowedHttpHeaders;
+    AZ_CORE_DLLEXPORT extern Azure::Core::CaseInsensitiveSet const g_defaultAllowedHttpHeaders;
```
ahsonkhan, line 396: Can you please explain why we need to do this.

ahsonkhan (approved):

sdk/core/azure-core/inc/azure/core/http/policy.hpp
ahsonkhan: What's the bug exactly, and the customer impact if we don't do this?

antkmsft: Potentially, someone (even our code) can go and modify the set of default headers. Even if our code does, that would be an "unintended feature". I never intended this to be modifiable, forgot the const specifier in the original PR. We create the list of default values once, and then never modify it.

PR #1854, opened 2021-03-11
Revert checked-in strings and delete the resources.

ahsonkhan:
- keys rotated and the resources are now deleted as well


vhvb1989 (approved):
here's a taco 🌮

PR #1864, opened 2021-03-11
Fix build on VS2017

antkmsft:


antkmsft:
@ahsonkhan, not sure how it happened, but it looks like https://github.com/Azure/azure-sdk-for-cpp/pull/1863 did not update the unit tests, and now they're failing to compile.
```
  D:\a\_work\1\s\sdk\core\azure-core\test\ut\bodystream.cpp(72,50): error C2660: 'Azure::Core::IO::BodyStream::ReadToEnd': function does not take 2 arguments [D:\a\_work\1\s\build\sdk\core\azure-core\test\ut\azure-core-test.vcxproj]
         D:\a\_work\1\s\sdk\core\azure-core\test\ut\bodystream.cpp(74,3): error C3536: 'readResult': cannot be used before it is initialized [D:\a\_work\1\s\build\sdk\core\azure-core\test\ut\azure-core-test.vcxproj]
         D:\a\_work\1\s\sdk\core\azure-core\test\ut\bodystream.cpp(74,3): error C2660: 'testing::internal::EqHelper::Compare': function does not take 3 arguments [D:\a\_work\1\s\build\sdk\core\azure-core\test\ut\azure-core-test.vcxproj]
         D:\a\_work\1\s\sdk\core\azure-core\test\ut\bodystream.cpp(74,3): error C2512: 'testing::AssertionResult': no appropriate default constructor available [D:\a\_work\1\s\build\sdk\core\azure-core\test\ut\azure-core-test.vcxproj]
```

ahsonkhan:
I saw that. I am surprised local testing and CI didn't catch that, maybe a change went in between. Let me fix it.

ahsonkhan:
Fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/1866

RickWinter (approved):

ahsonkhan (approved):
Interesting, why would this fail only on VS 2017? `Url` lives in `Azure::Core`, so how do other compilers know where to find it without this `using directive`?

antkmsft:
@ahsonkhan I don't know who is right, I am pretty sure that VS 2017 is wrong and does not support the standard fully.
But the nature of the problem is that
```
namespace A
{
    class B;
    void f(B const& b);
}

//void A::f(
//    all but VS 2017 consider '(' as being 'inside' the context of A::f() and, therefore, the type names being resolved
//    go from the point of view of being located in 'namespace A', so you don't have to explicitly qualify 'B' as 'A::B'.
//   );

// i.e.:
// all compilers:
void A::f(B const& b)
{
    ...
}

// VS 2017 requires 'B' to still be qualified:
void A::f(A::B const& b)
{
    ...
}
// and therefore, in VS2017, we need an additional 'using namespace A' in the .cpp file.
```

ahsonkhan:
If you have it, can you share the build error message you see without this fix, from VS 2017.

antkmsft:
@ahsonkhan , yes, that was not expected that CI didn't catch it, that was wierd, and it should not happen. CI should be catching these issues for us. Thank you for finding the reason why that happened and fixing the tests!

ahsonkhan:
@antkmsft I recommend pulling in master into your branch.

antkmsft:
> If you have it, can you share the build error message you see without this fix, from VS 2017.

I didn't savbe the logs, unfortunately. But it was along the lines that 'Url' is unknown, type name is expected. The same error as if you would try to use 'MyType' instead of `Azure::Url` without declaring what 'MyType' is.

PR #1955, opened 2021-03-22
Update relative links to sections in the contributing guide doc to fix link verification

ahsonkhan:


vhvb1989 (approved):

sdk/core/azure-core/src/context.cpp
```diff
@@ -8,8 +8,8 @@ using time_point = std::chrono::system_clock::time_point;
 
 Context& Azure::Core::Context::GetApplicationContext()
 {
-  static Context ctx;
-  return ctx;
+  static Context context;
```
ahsonkhan, line 11: Make some source code change modifications to trigger GenerateRelease CI steps (which is the one that has link verification).

@sima-zhu, @weshaggard - README updates like the following PR (https://github.com/Azure/azure-sdk-for-cpp/pull/1921) don't trigger the link verification step in CI means we end up checking-in "bad" links and other, and then unrelated PRs then start failing (because they are the ones actually running that step). Can we fix this so any markdown updates run the link verification step (while keeping some of the CI optimization where we don't run other build/test legs)?

cc @danieljurek

CONTRIBUTING.md
```diff
@@ -48,13 +48,13 @@ Codespaces is new technology that allows you to use a container as your developm
 ### GitHub Codespaces
 
 1. From the Azure SDK GitHub repo, click on the "Code -> Open with Codespaces" button.
-1. Open a Terminal. The development environment will be ready for you. Continue to [Building and Testing](#building-and-Testing).
```
ahsonkhan, line 51: @sima-zhu are warnings like these intended to fail the build? If so, why is it a warning and not an error?

https://dev.azure.com/azure-sdk/public/_build/results?buildId=796379&view=logs&j=a129effc-2dd1-54d1-fb5a-ad7bdc0e851d&t=617ceecd-bf48-5ce3-83e2-7468b1632a13

```text
##[warning]Please lower case your anchor tags (i.e. anything after '#' in your link '#building-and-Testing'. Check here for more information: https://aka.ms/azsdk/guideline/links 
Found 27 links on page file:///D:/a/_work/1/s/README.md 
Found 12 links on page file:///D:/a/_work/1/s/SECURITY.md 
Summary of broken links: 
'file:///D:/a/_work/1/s/CONTRIBUTING.md' has 1 broken link(s): 
 #building-and-Testing
```

PR #1970, opened 2021-03-24
Fix the date in the changelog for storage beta.6 packages using the correct year to avoid auto version bump PR re-ordering it

ahsonkhan:

PR #1971, opened 2021-03-24
Update sdk/storage codeowners list to use the up-to-date GitHub alias

ahsonkhan:
Now the codeowner list will add @Jinming-Hu as PR reviewer for changes under sdk/storage:

It wasn't working before. For example: https://github.com/Azure/azure-sdk-for-cpp/pull/1970

The previous GitHub alias 404s since it was renamed: https://github.com/JinmingHu-MSFT


Jinming-Hu (approved):
Thanks

PR #1989, opened 2021-03-27
Introduce Context::Key with pointer comparison speeds && Context::Key does not have public constructor

antkmsft:
If you say that in addition to what we have in #1988, we want users to not create their own `Context::Key`s, here is the solution.

It is now not possible for users to create `Context::Key` - they have to get it from somewhere. Constructor is private.
A function to create them is internal.
SDK can create them - we now use `char*` (i.e. pointers) as unique IDs, but they have to have external linkage, i.e. you can't just create it with dynamic string - it should be a compile-time string. A slight benefit is that `char*` is visible in debugger as a string.

We still compare pointers and not the actual values.


Jinming-Hu (commented):
It doesn't make sense to forbid users from creating their own keys.

Storage sdk is a user of core sdk and we do create our own key in storage sdk.

PR #1998, opened 2021-03-30
remove-http-response-apis

vhvb1989:
fixes: https://github.com/Azure/azure-sdk-for-cpp/issues/1997


sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -141,6 +141,27 @@ void WinSocketSetBuffSize(curl_socket_t socket)
...
+
+void static inline SetHeader(Azure::Core::Http::RawResponse& response, std::string const& header)
+{
+  return Azure::Core::Http::_detail::RawResponse::SetHeader(
+      response,
+      reinterpret_cast<uint8_t const*>(header.data()),
+      reinterpret_cast<uint8_t const*>(header.data() + header.size()));
+}
+
+static inline std::string GetHTTPMessagePreBody(Azure::Core::Http::Request const& request)
```
ahsonkhan, line 153: Keep the previous comment
```suggestion
// Writes an HTTP request with RFC 7230 without the body (head line and headers)
// https://tools.ietf.org/html/rfc7230#section-3.1.1
static inline std::string GetHTTPMessagePreBody(Azure::Core::Http::Request const& request)
```

sdk/core/azure-core/inc/azure/core/http/win_http_transport.hpp
```diff
@@ -21,7 +21,6 @@ namespace Azure { namespace Core { namespace Http {
   namespace _detail {
 
     constexpr static int64_t DefaultUploadChunkSize = 1024 * 64;
-    constexpr static int64_t MaximumUploadChunkSize = 1024 * 1024;
```
ahsonkhan, line 24: We still want to keep this.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -267,28 +269,14 @@ void WinHttpTransport::CreateRequestHandle(std::unique_ptr<_detail::HandleManage
...
-    }
-    else
-    {
-      uploadChunkSize = _detail::DefaultUploadChunkSize;
-    }
-  }
-
-  auto unique_buffer = std::make_unique<uint8_t[]>(static_cast<size_t>(uploadChunkSize));
+  auto unique_buffer
+      = std::make_unique<uint8_t[]>(static_cast<size_t>(_detail::DefaultUploadChunkSize));
```
ahsonkhan, line 274: Only use `DefaultUploadChunkSize` if `streamLength` is > `MaximumUploadChunkSize`, otherwise use `MaximumUploadChunkSize`.

sdk/core/azure-core/test/ut/http.cpp
```diff
@@ -105,12 +105,8 @@ namespace Azure { namespace Core { namespace Test {
         response.GetHeaders(),
         expected2);
 
-    // Response SetHeader overload method to add from string
-    EXPECT_THROW(response.SetHeader("inv(): header"), std::invalid_argument);
```
ahsonkhan, line 109: Keep this test with the existing `SetHeader` overload, passing in `inv()` for the key.

ahsonkhan (changes requested):
Don't make changes to yml, keep winhttp logic as is, and fixup struct -> class and naming.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -56,15 +56,15 @@ jobs:
         CMAKE_GENERATOR: 'Visual Studio 16 2019'
         CMAKE_GENERATOR_PLATFORM: Win32
         CmakeArgs: ' -DBUILD_TRANSPORT_CURL=ON' #Leaving curl on here to explicitly test what the default behavior is on windows.
-        BuildArgs: '-v --parallel 8'
+        BuildArgs: '--parallel 8'
```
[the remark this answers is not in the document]
vhvb1989: why?

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -528,4 +468,53 @@ namespace Azure { namespace Core { namespace Http {
     std::vector<uint8_t> const& GetBody() const { return this->m_body; }
   };
 
+  namespace _detail {
+    struct RawResponse
```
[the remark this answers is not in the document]
vhvb1989: why?  its the same thing

sdk/core/azure-core/inc/azure/core/http/win_http_transport.hpp
vhvb1989: what's the benefit? we are adding an extra check which we don't need

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
vhvb1989: why?
We just need the default now.  Since it won't be customizable anymore.

sdk/core/azure-core/test/ut/http.cpp
vhvb1989: that overload was removed

eng/pipelines/templates/jobs/archetype-sdk-client.yml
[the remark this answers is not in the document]
ahsonkhan: 1) This change is unrelated to this PR, and
2) Why would we not want verbose build output in the CI, which might provide more diagnostic info?

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
ahsonkhan: We decided that `struct` would only be used for holders of "properties", and any object that has methods and ctors would be a `class`. Let me see if that ended up in the C++ guidelines.

Regarding the name, I don't think we should use the identical name as the public type (`RawResponse`), as it adds confusion at the call-site.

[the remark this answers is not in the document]
ahsonkhan: https://azure.github.io/azure-sdk/cpp_implementation.html#cpp-design-naming-classes-no-struct-keyword

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
ahsonkhan: We only want to split the buffer into 64k chunks if the entire buffer is >= 1 MB. When it's smaller, we just want to do it in one-shot.

Removing the `MaximumUploadChunkSize` is changing the logic here, unrelated to the customization of defaults. Please keep that logic as is, especially in this PR.

sdk/core/azure-core/test/ut/http.cpp
ahsonkhan: I understand. We still have the overload that takes two string `key, name` parameters, correct? Do we have a test already where we pass in `inv()` for the key?

`response.SetHeader("inv()", "header");`

eng/pipelines/templates/jobs/archetype-sdk-client.yml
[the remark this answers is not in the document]
vhvb1989: I can create one unique PR for this.
The current verbosity on Windows CI is hiding the actual bugs. It is not human readable!  I wasn't been able to easily find the MSVC compiler error from CI logs, I need to go and run on VS to see the error.
The -v is adding too much info which ends up hiding whats relevant

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
vhvb1989: https://github.com/Azure/azure-sdk-for-cpp/pull/2002

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
vhvb1989: Then why not setting the default to 1MB?  
Why do we prefer to handle 2 possible flows (we are not even testing it).  We can always just use 1MB,

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
ahsonkhan: It has public member functions. Structs are for types that only have **public data members**.

That means options, model types, and things like `HttpRange`/`MatchConditions` should be struct, and everything else becomes class.

We have a couple of private/implementation detail types within the repo that currently don't follow this, but we should fix those.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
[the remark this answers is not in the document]
vhvb1989: https://github.com/Azure/azure-sdk-for-cpp/pull/2002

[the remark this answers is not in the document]
ahsonkhan: Do you mean like this?
https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/811744/logs/118

The error shows up at the bottom, seems pretty clear to me:
image

And if you search for `error ` in the logs, you can see the build errors/failures right away:
image

Up to you. The verbosity isn't too concerning to me from looking at where the build error occurred and is actually helpful in some of the more weird build bugs (such as intermittent failures due to with gtest).

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
vhvb1989: There's no benefit at all from such a guideline,  we should not have something like that.
The only difference from struct and class is the default access for it. 
Specially if we are disabling having methods directly under a namespace like ::_detail::Foo() and we expect to always use a Container for the methods like ::_detail::Container::Foo() , then all methods will be public, and class is just weird..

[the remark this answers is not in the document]
ahsonkhan: > class is just weird..

Why?

> There's no benefit at all from such a guideline, we should not have something like that.

Well, if you have concerns about the guidelines, let's have that guideline discussion separately. @RickWinter should we change our guideline on when we use `struct` vs `class`?

Up to you what you want to do here, it's your PR. Please rename the type though.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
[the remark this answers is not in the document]
vhvb1989: Yes, that's the only output I would like to see. 
In the past I updated the CI legs to enable verbose, but it didn't really add more info about the error, it just add info which I don't need

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
ahsonkhan: 1) I don't think we should be discussing or changing implementation detail of a transport adapter as part of this change. Let's keep the change limited to what we actually decided. There is a buffering heuristic in place, which was done intentionally. Other changes should be done separately.
2) The purpose of the 64k vs 1 MB separation was to avoid creating a large 1MB vector when the stream is actually small. So the heuristic is, if the stream is less than 1 MB, use the actual stream size (don't create a 1 MB vector when the stream is 5 KB). If the stream is larger, use 64 KB chunks. I agree with you that it is reasonable to say, use 1 MB chunk rather than 64KB. for that (and I want to make this change after doing before/after perf comparison).

I expect this PR to keep the following code:
```C++
  // Consider using `MaximumUploadChunkSize` here, after some perf measurements
  int64_t uploadChunkSize = _detail::DefaultUploadChunkSize;
  if (streamLength < _detail::MaximumUploadChunkSize)
  {
    uploadChunkSize = streamLength;
  }
  auto unique_buffer = std::make_unique<uint8_t[]>(static_cast<size_t>(uploadChunkSize));
```

eng/pipelines/templates/jobs/archetype-sdk-client.yml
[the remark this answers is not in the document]
ahsonkhan: > Yes, that's the only output I would like to see.

The link above I showed you is **with** `-v` on. It's from your first commit: https://github.com/Azure/azure-sdk-for-cpp/pull/1998/commits/aacdfca3d25a2ad18ce24214f3a92a8858cf9980

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
ahsonkhan: > won't give you nothing but an extra line of code.

I really wish we stop revisiting closed discussions in PRs like this (especially relitigating using the exact same reasons that were already settled, rather than providing new information/insight to motivate the change). I was assuming @RickWinter had updated the guideline after we discussed and agreed on when we would use `class` and `struct`.

My suggestion of the code change is based on what I understood to be the guideline, if you disagree, ignore my feedback. My intent is not to keep defending the guideline.

ahsonkhan (approved):
Other than reverting the buffering logic change on winhttp, looks good.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
vhvb1989: fine.
Just have in mind that the mentioned heuristic can be the source of bugs.
I wouldn't expect any significant different from allocating 64k vs 1Mb.  The difference will be when it comes to copying the source to the allocated memory for uploading, we basically are expecting that
- One only 1MB allocate, copy and upload will be better than.
- One only 64k allocate and multiple 64k copy and upload operation.

But after copying 1Mb to the allocated memory, you will have the bottle neck on network because ssl will only upload 64k at a time, so the upload operation will do the iteration of read-and-upload until all is done

ahsonkhan: > I wouldn't expect any significant different from allocating 64k vs 1Mb.

Is this allocating the array on the stack? Shouldn't we avoid unnecessary allocation of space that we don't need or use, even if there is now runtime perf cost of allocating a static array of 64 KB vs 1 MB?
`auto unique_buffer = std::make_unique<uint8_t[]>(static_cast<size_t>(uploadChunkSize));`


>  because ssl will only upload 64k at a time

Really?! With WinHTTP? Can you please share some documentation which mentions this limitation because this is news to me.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -269,26 +271,18 @@ void WinHttpTransport::Upload(std::unique_ptr<_detail::HandleManager>& handleMan
...
   }
-
   auto unique_buffer = std::make_unique<uint8_t[]>(static_cast<size_t>(uploadChunkSize));
 
   while (true)
   {
-    auto rawRequestLen
-        = streamBody->Read(unique_buffer.get(), uploadChunkSize, handleManager->m_context);
+    auto rawRequestLen = streamBody->Read(
+        unique_buffer.get(), _detail::DefaultUploadChunkSize, handleManager->m_context);
```
ahsonkhan, line 285:
```suggestion
    auto rawRequestLen = streamBody->Read(
        unique_buffer.get(), uploadChunkSize, handleManager->m_context);
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -269,26 +271,17 @@ void WinHttpTransport::Upload(std::unique_ptr<_detail::HandleManager>& handleMan
...
+    uploadChunkSize = streamLength;
   }
-
   auto unique_buffer = std::make_unique<uint8_t[]>(static_cast<size_t>(uploadChunkSize));
 
   while (true)
   {
-    auto rawRequestLen
-        = streamBody->Read(unique_buffer.get(), uploadChunkSize, handleManager->m_context);
+    auto rawRequestLen = streamBody->Read(unique_buffer.get(), uploadChunkSize, handleManager->m_context);
```
ahsonkhan, line 284: I am guessing this needs to be clang-formatted and line length split.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
 ### Breaking Changes
 
+- Removed from `Azure::Core::Http::Request`:
+  - `SetUploadChunkSize`.
+  - `GetHTTPMessagePreBody`.
+  - `GetUploadChunkSize`.
+- Removed from `Azure::Core::Http::RawResponse`:
+  - `SetHeader(std::string const& header)`
+  - `SetHeader(uint8_t const* const first, uint8_t const* const last)`.
+  - `GetMajorVersion`.
```
ahsonkhan, line 14:
```suggestion
  - `GetMajorVersion()`.
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
+  - `GetMinorVersion`.
```
ahsonkhan, line 15:
```suggestion
  - `GetMinorVersion()`.
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
 
 ### Breaking Changes
 
+- Removed from `Azure::Core::Http::Request`:
+  - `SetUploadChunkSize`.
```
```suggestion
  - `SetUploadChunkSize()`.
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
+  - `GetHTTPMessagePreBody`.
```
ahsonkhan, line 9:
```suggestion
  - `GetHTTPMessagePreBody()`.
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
+  - `GetUploadChunkSize`.
```
ahsonkhan, line 10:
```suggestion
  - `GetUploadChunkSize()`.
```

sdk/core/azure-core/CHANGELOG.md
ahsonkhan: We use `()` to make it clear it's a method that is being removed rather than a type or field.

ahsonkhan (approved):

sdk/core/azure-core/CHANGELOG.md
ahsonkhan: CL feedback fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/2037

sdk/core/azure-core/CHANGELOG.md
ahsonkhan: CL feedback fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/2037

Issue #2018, opened 2021-04-01
Do not require setting the `AZURE_KEYVAULT_HSM_URL` environment variable, for running KeyVault tests locally that don't need it


ahsonkhan:
cc @vhvb1989

PR #2019, opened 2021-04-01
Add error checking around std::getenv for KeyVault tests

ahsonkhan:
TODO: Ideally, only the `HSM` url requiring tests should require setting the `AZURE_KEYVAULT_HSM_URL` environment variable, and all other tests should run and pass locally.
This can be done as a follow up issue: https://github.com/Azure/azure-sdk-for-cpp/issues/2018


vhvb1989 (approved):
<3

PR #2033, opened 2021-04-02
Changes around context expiration naming and types

### LINKED ISSUE #1880: Context: review `CancelAtMsecSinceEpoch(ToMsecSinceEpoch(cancelAt))`
That looks like something that can be rewritten to use DateTime / may not work correctly if it currently uses DateTime

### LINKED ISSUE #2032: Feature Request: make Context::CancelWhen() public and maybe also rename it

antkmsft:
This would unblock #2010.

Closes #2032.
Closes #1880.

I know this comes out of the blue, but it seems to me the right change to do. This is a proposal, let me know what you think.

First, I started with the fact that we need to make `CancelWhen()` public.
Then, I realized that I don't like the `CancelWhen()` naming. `CancelAt()` is better?
But `CancelAt()` sounds like an order to do something, not as getter.
So, it should be named `Get...`. `GetCancelWhen()`? `GetCancelAt()`? Sounds strage.

`GetDeadline()` - not bad, bit it is not that clear, what the deadline is.
And I looked at the `WithDeadline()` method. And in comment, the first line is `@brief Create a context with expiration.`.
"Expiration" sounds better explanation for the essense of the "deadline".

So, I went with `GetExpiration()`. I think "Expiration" is also a better name, because should we want to add the method called "bool IsExpired()", it comes naturally using the existing terminology, sounds better than "`IsPastDeadline()`".

Next thing - if we have "`GetExpiration()`", it is strange to have a method called `Get()`. So, we have `WithExpiration()` and `WithValue()`. So, it sounds like the getter should be called `GetValue()`. I did that rename as well.

Then, I looked at "`With...`" method naming. It is a factory method. If for getters we have `Get`, then for factory methods we should have `Create`. So, I renamed `WithExpiration()` and `WithValue()` to `CreateWithExpiration()` and `CreateWithValue()`.

Then I looked at `Context::time_point` typedef. First, in general, if we can avoid public typedefs, it is better, because we don't need to document them and worry if we broke client code when we change them.
Second, it is strange that we use `Azure::DateTime` everywhere, but not in context.

So, I replaced it with `Azure::DateTime`. It is good because it allows us to drop to/from epoch conversions (#1880), and really all that extra dance we do to cast to representation and back to datetime is the ways to stay within legal type casting limits of compiler type declaration, while in reality there isn't much that is happening in the code: `DateTime` is essentially an `int64`, and with all these conversions to epoch time, then to representation and back, compiler does not generate real code, it still operates with that only `int64`.

Why we cast to `DateTime::rep` and back at all? Context stores it as `atomic`, and I am not questioning that. You can't get `std::atomic<DateTime>` today explicitly, so we "extract" representation (`int64`) and store it as `std::atomic<DateTime::rep>`, which is the same thing as `std::atomic_int64`.

-- Update: --
1. "Expiration" => "Deadline" (Jeffrey's request)
2. Added `Context::HasExpiration()` (response to Jinming) // plus, we do have `HasValue()` so why not have `HasDeadline()`.
3. `WithDeadline()`, `WithValue` => Two overloads of `CreateChildContext()` (my own initiative).

-- Update 2: --
`CreateChildContext()` => `WithDeadline()`, `WithValue`

-- Update 3: --
Removed `HasDeadline()`

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -16,8 +16,8 @@
   - `GetMinorVersion`.
 - Changes to `Azure::Core::Context`:
   - Renamed `Get()` to `GetValue()`.
-  - Renamed `WithValue()` to `CreateWithValue()`.
-  - Renamed `WithDeadline()` to `CreateWithExpiration()`, changed input parameter type to `Azure::DateTime`.
+  - Renamed `WithValue()` to `CreateChildContext()`.
+  - Renamed `WithDeadline()` to `CreateChildContext()`, and changed input parameter type to `Azure::DateTime`.
 - Removed from `Azure::Core::Http::Policies` namespace: `HttpPolicyOrder`, `TransportPolicy`, `RetryPolicy`, `RequestIdPolicy`, `TelemetryPolicy`, `BearerTokenAuthenticationPolicy`, `LogPolicy`.
 - Renamed `Azure::Core::Http::RawResponse::GetBodyStream()` to `ExtractBodyStream()`.
 
@@ -29,7 +29,7 @@
 - Added `Azure::Core::Operation<T>::GetRawResponse()`.
 - Added `Azure::Core::PackageVersion`.
 - Added support for logging to console when `AZURE_LOG_LEVEL` environment variable is set.
-- Added `Azure::Core::Context::GetExpiration()`.
+- Added `Azure::Core::Context::GetDeadline()` and `Azure::Core::Context::HasDeadline()`.
 
 ### Breaking Changes
 
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -14,6 +14,10 @@
   - `SetHeader(uint8_t const* const first, uint8_t const* const last)`.
   - `GetMajorVersion`.
   - `GetMinorVersion`.
+- Changes to `Azure::Core::Context`:
+  - Renamed `Get()` to `GetValue()`.
+  - Renamed `WithValue()` to `CreateWithValue()`.
+  - Renamed `WithDeadline()` to `CreateWithExpiration()`, changed input parameter type to `Azure::DateTime`.
 - Removed from `Azure::Core::Http::Policies` namespace: `HttpPolicyOrder`, `TransportPolicy`, `RetryPolicy`, `RequestIdPolicy`, `TelemetryPolicy`, `BearerTokenAuthenticationPolicy`, `LogPolicy`.
 - Renamed `Azure::Core::Http::RawResponse::GetBodyStream()` to `ExtractBodyStream()`.
 
@@ -25,6 +29,7 @@
 - Added `Azure::Core::Operation<T>::GetRawResponse()`.
 - Added `Azure::Core::PackageVersion`.
 - Added support for logging to console when `AZURE_LOG_LEVEL` environment variable is set.
+- Added `Azure::Core::Context::GetExpiration()`.
 
 ### Breaking Changes
 
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -16,8 +16,7 @@
   - `GetMinorVersion`.
 - Changes to `Azure::Core::Context`:
   - Renamed `Get()` to `GetValue()`.
-  - Renamed `WithValue()` to `CreateChildContext()`.
-  - Renamed `WithDeadline()` to `CreateChildContext()`, and changed input parameter type to `Azure::DateTime`.
+  - Changed input parameter type of `WithDeadline()` to `Azure::DateTime`.
 - Removed from `Azure::Core::Http::Policies` namespace: `HttpPolicyOrder`, `TransportPolicy`, `RetryPolicy`, `RequestIdPolicy`, `TelemetryPolicy`, `BearerTokenAuthenticationPolicy`, `LogPolicy`.
 - Renamed `Azure::Core::Http::RawResponse::GetBodyStream()` to `ExtractBodyStream()`.
 - Introduced `Azure::Core::Context::Key` class which takes place of `std::string` used for `Azure::Core::Context` keys previously.
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -29,7 +29,7 @@
 - Added `Azure::Core::Operation<T>::GetRawResponse()`.
 - Added `Azure::Core::PackageVersion`.
 - Added support for logging to console when `AZURE_LOG_LEVEL` environment variable is set.
-- Added `Azure::Core::Context::GetDeadline()` and `Azure::Core::Context::HasDeadline()`.
+- Added `Azure::Core::Context::GetDeadline()`.
 
 ### Breaking Changes
 
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -20,6 +20,7 @@
   - Renamed `WithDeadline()` to `CreateChildContext()`, and changed input parameter type to `Azure::DateTime`.
 - Removed from `Azure::Core::Http::Policies` namespace: `HttpPolicyOrder`, `TransportPolicy`, `RetryPolicy`, `RequestIdPolicy`, `TelemetryPolicy`, `BearerTokenAuthenticationPolicy`, `LogPolicy`.
 - Renamed `Azure::Core::Http::RawResponse::GetBodyStream()` to `ExtractBodyStream()`.
+- Introduced `Azure::Core::Context::Key` class which takes place of `std::string` used for `Azure::Core::Context` keys previously.
 
 ## 1.0.0-beta.7 (2021-03-11)
 
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -176,7 +176,10 @@ namespace Azure { namespace Core {
      * @return `true` if this context, or the tree branch this context belongs to has a deadline
      * associated with it. `false` otherwise.
      */
-    bool HasDeadline() const { return GetDeadline() != (DateTime::max)(); }
+    bool HasDeadline() const
+    {
+      return static_cast<DateTime::time_point>(GetDeadline()) != (DateTime::max)();
+    }
 
     /**
      * @brief Check whether the context has a key matching \p key parameter in it itself, or in the
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include "azure/core/datetime.hpp"
+
 #include <atomic>
 #include <chrono>
 #include <memory>
@@ -38,42 +40,35 @@ namespace Azure { namespace Core {
    * @brief A context is a node within a tree that represents expiration times and key/value pairs.
    */
   class Context {
-  public:
-    /**
-     * @brief A type used to provide a point in time for context expiration.
-     */
-    using time_point = std::chrono::system_clock::time_point;
-
   private:
     struct ContextSharedState
     {
       std::shared_ptr<ContextSharedState> Parent;
-      std::atomic_int64_t CancelAtMsecSinceEpoch;
+      std::atomic<DateTime::rep> Expiration;
       std::string Key;
       std::shared_ptr<void> Value;
       const std::type_info& ValueType;
 
-      static constexpr int64_t ToMsecSinceEpoch(time_point time)
+      static constexpr DateTime::rep ToDateTimeRepresentation(DateTime const& dateTime)
       {
-        return static_cast<int64_t>(
-            std::chrono::duration_cast<std::chrono::milliseconds>(time - time_point()).count());
+        return dateTime.time_since_epoch().count();
       }
 
-      static constexpr time_point FromMsecSinceEpoch(int64_t msec)
+      static constexpr DateTime FromDateTimeRepresentation(DateTime::rep dtRepresentation)
       {
-        return time_point() + static_cast<std::chrono::milliseconds>(msec);
+        return DateTime(DateTime::time_point(DateTime::duration(dtRepresentation)));
       }
 
       explicit ContextSharedState()
-          : CancelAtMsecSinceEpoch(ToMsecSinceEpoch((time_point::max)())), Value(nullptr),
+          : Expiration(ToDateTimeRepresentation((DateTime::max)())), Value(nullptr),
             ValueType(typeid(std::nullptr_t))
       {
       }
 
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
-          time_point cancelAt)
-          : Parent(parent), CancelAtMsecSinceEpoch(ToMsecSinceEpoch(cancelAt)), Value(nullptr),
+          DateTime const& expiration)
+          : Parent(parent), Expiration(ToDateTimeRepresentation(expiration)), Value(nullptr),
             ValueType(typeid(std::nullptr_t))
       {
       }
@@ -81,10 +76,10 @@ namespace Azure { namespace Core {
       template <class T>
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
-          time_point cancelAt,
+          DateTime const& expiration,
           const std::string& key,
           T value) // NOTE, should this be T&&
-          : Parent(parent), CancelAtMsecSinceEpoch(ToMsecSinceEpoch(cancelAt)), Key(key),
+          : Parent(parent), Expiration(ToDateTimeRepresentation(expiration)), Key(key),
             Value(std::make_shared<T>(std::move(value))), ValueType(typeid(T))
       {
       }
@@ -96,8 +91,6 @@ namespace Azure { namespace Core {
         : m_contextSharedState(std::move(impl))
     {
     }
-
-    time_point CancelWhen() const;
 
   public:
     /**
@@ -113,13 +106,13 @@ namespace Azure { namespace Core {
     /**
      * @brief Create a context with expiration.
      *
-     * @param cancelWhen A point in time after which a context expires.
+     * @param expiration A point in time after which a context expires.
      *
      * @return A child context with expiration.
      */
-    Context WithDeadline(time_point cancelWhen) const
+    Context CreateWithExpiration(DateTime const& expiration) const
     {
-      return Context{std::make_shared<ContextSharedState>(m_contextSharedState, cancelWhen)};
+      return Context{std::make_shared<ContextSharedState>(m_contextSharedState, expiration)};
     }
 
     /**
@@ -131,11 +124,20 @@ namespace Azure { namespace Core {
      *
      * @return A child context with no expiration and the \p key and \p value associated with it.
      */
-    template <class T> Context WithValue(const std::string& key, T&& value) const
+    template <class T> Context CreateWithValue(const std::string& key, T&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
-          m_contextSharedState, (time_point::max)(), key, std::forward<T>(value))};
+          m_contextSharedState, (DateTime::max)(), key, std::forward<T>(value))};
     }
+
+    /**
+     * @brief Get an expiration time point associated with this context or the branch of contexts
+     * this context belongs to.
+     *
+     * @return An expiration associated with the context found; an empty value if a specific value
+     * can't be found.
+     */
+    DateTime GetExpiration() const;
 
     /**
      * @brief Get a value associated with a \p key parameter within this context or the branch of
@@ -146,7 +148,7 @@ namespace Azure { namespace Core {
      * @return A value associated with the context found; an empty value if a specific value can't
      * be found.
      */
-    template <class T> const T& Get(const std::string& key) const
+    template <class T> const T& GetValue(const std::string& key) const
     {
       if (!key.empty())
       {
@@ -197,15 +199,15 @@ namespace Azure { namespace Core {
      */
     void Cancel()
     {
-      m_contextSharedState->CancelAtMsecSinceEpoch
-          = ContextSharedState::ToMsecSinceEpoch((time_point::min)());
+      m_contextSharedState->Expiration
+          = ContextSharedState::ToDateTimeRepresentation((DateTime::min)());
     }
 
     /**
      * @brief Check if the context is cancelled.
      * @return `true` if this context is cancelled, `false` otherwise.
      */
-    bool IsCancelled() const { return CancelWhen() < std::chrono::system_clock::now(); }
+    bool IsCancelled() const { return GetExpiration() < std::chrono::system_clock::now(); }
 
     /**
      * @brief Throw an exception if the context was cancelled.
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -40,12 +40,30 @@ namespace Azure { namespace Core {
    * @brief A context is a node within a tree that represents deadline times and key/value pairs.
    */
   class Context {
+  public:
+    /**
+     * @brief A context key.
+     */
+    class Key final {
+      Key const* m_uniqueAddress;
+
+    public:
+      Key() : m_uniqueAddress(this) {}
+
+      bool operator==(Key const& other) const
+      {
+        return this->m_uniqueAddress == other.m_uniqueAddress;
+      }
+
+      bool operator!=(Key const& other) const { return !(*this == other); }
+    };
+    
   private:
     struct ContextSharedState
     {
       std::shared_ptr<ContextSharedState> Parent;
       std::atomic<DateTime::rep> Deadline;
-      std::string Key;
+      Context::Key Key;
       std::shared_ptr<void> Value;
       const std::type_info& ValueType;
 
@@ -77,7 +95,7 @@ namespace Azure { namespace Core {
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
           DateTime const& deadline,
-          const std::string& key,
+          Context::Key const& key,
           T value) // NOTE, should this be T&&
           : Parent(parent), Deadline(ToDateTimeRepresentation(deadline)), Key(key),
             Value(std::make_shared<T>(std::move(value))), ValueType(typeid(T))
@@ -123,7 +141,7 @@ namespace Azure { namespace Core {
      *
      * @return A child context with no deadline and the \p key and \p value associated with it.
      */
-    template <class T> Context CreateChildContext(const std::string& key, T&& value) const
+    template <class T> Context WithValue(Key const& key, T&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
           m_contextSharedState, (DateTime::max)(), key, std::forward<T>(value))};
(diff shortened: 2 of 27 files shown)


sdk/core/azure-core/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
   - `SetHeader(uint8_t const* const first, uint8_t const* const last)`.
   - `GetMajorVersion`.
   - `GetMinorVersion`.
+- Changes to `Azure::Core::Context`:
+  - Renamed `Get()` to `GetValue()`.
+  - Renamed `WithValue()` to `CreateWithValue()`.
+  - Renamed `WithDeadline()` to `CreateWithExpiration()`, changed input parameter type to `Azure::DateTime`.
```
Jinming-Hu, line 20: I feel the old names are better than the new ones.

Because for `Create`, I have the feeling that something is going to be created out of thin air. Like `CreateBlob`, there was nothing, and now we have a blob.

Jinming-Hu:
Do we have an reliable way to check if expiration is set/unset?

Before this PR, we use `timepoint::max()` and convert it to milliseconds. There's a problem with this method. Sub-milliseconds are discarded.

sdk/core/azure-core/CHANGELOG.md
JeffreyRichter: I like GetValue although I'm hoping for a GetValue(key, T valueIfKeyNotFound) function and then the removal of HasValue. Or, I'll take a bool TryGetValue(key, T* valueIfFound)

(Create)WithValue actually does create a Context on the heap "out of thin air". I prefer the old name too but with documentation that explains that a heap allocation occurs.

I also prefer no WithDeadline without Create and without Expiration. Accepting a DateTime is OK with me.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -25,6 +29,7 @@
 - Added `Azure::Core::Operation<T>::GetRawResponse()`.
 - Added `Azure::Core::PackageVersion`.
 - Added support for logging to console when `AZURE_LOG_LEVEL` environment variable is set.
+- Added `Azure::Core::Context::GetExpiration()`.
```
JeffreyRichter, line 32: GetDeadline

sdk/core/azure-core/CHANGELOG.md
JeffreyRichter: I looked-up deadline & expiration in the dictionary. I think deadline is better because of this: https://www.powerthesaurus.org/deadline/definitions

The Context holds "the point in time at which something must be completed".

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -38,53 +40,46 @@ namespace Azure { namespace Core {
...
-     * @brief A type used to provide a point in time for context expiration.
-     */
-    using time_point = std::chrono::system_clock::time_point;
-
   private:
     struct ContextSharedState
     {
       std::shared_ptr<ContextSharedState> Parent;
-      std::atomic_int64_t CancelAtMsecSinceEpoch;
+      std::atomic<DateTime::rep> Expiration;
```
JeffreyRichter, line 47: Deadline

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -38,53 +40,46 @@ namespace Azure { namespace Core {
...
       std::string Key;
       std::shared_ptr<void> Value;
       const std::type_info& ValueType;
 
-      static constexpr int64_t ToMsecSinceEpoch(time_point time)
+      static constexpr DateTime::rep ToDateTimeRepresentation(DateTime const& dateTime)
```
JeffreyRichter, line 52: Do we need these conversion functions now if we're just storing a DateTime?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -38,53 +40,46 @@ namespace Azure { namespace Core {
...
+          : Parent(parent), Expiration(ToDateTimeRepresentation(expiration)), Value(nullptr),
             ValueType(typeid(std::nullptr_t))
       {
       }
 
       template <class T>
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
-          time_point cancelAt,
+          DateTime const& expiration,
```
JeffreyRichter, line 79: deadline

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -131,12 +124,21 @@ namespace Azure { namespace Core {
...
     }
 
+    /**
+     * @brief Get an expiration time point associated with this context or the branch of contexts
+     * this context belongs to.
+     *
+     * @return An expiration associated with the context found; an empty value if a specific value
+     * can't be found.
+     */
+    DateTime GetExpiration() const;
```
JeffreyRichter, line 140: GetDeadline

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -197,15 +199,15 @@ namespace Azure { namespace Core {
...
+      m_contextSharedState->Expiration
+          = ContextSharedState::ToDateTimeRepresentation((DateTime::min)());
     }
 
     /**
      * @brief Check if the context is cancelled.
      * @return `true` if this context is cancelled, `false` otherwise.
      */
-    bool IsCancelled() const { return CancelWhen() < std::chrono::system_clock::now(); }
+    bool IsCancelled() const { return GetExpiration() < std::chrono::system_clock::now(); }
```
JeffreyRichter, line 210: Can we compare against DateTime::now and avoid chrono

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -131,12 +124,21 @@ namespace Azure { namespace Core {
...
+    template <class T> Context CreateWithValue(const std::string& key, T&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
-          m_contextSharedState, (time_point::max)(), key, std::forward<T>(value))};
+          m_contextSharedState, (DateTime::max)(), key, std::forward<T>(value))};
     }
 
+    /**
+     * @brief Get an expiration time point associated with this context or the branch of contexts
+     * this context belongs to.
```
JeffreyRichter, line 135: Maybe add that a value of DateTime::max means that no deadline was ever set.

sdk/core/azure-core/src/context.cpp
```diff
@@ -4,23 +4,22 @@
...
+Azure::DateTime Azure::Core::Context::GetExpiration() const
 {
-  auto result = time_point::max();
+  auto result = DateTime::max();
   for (auto ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
   {
-    auto cancelAt = ContextSharedState::FromMsecSinceEpoch(ptr->CancelAtMsecSinceEpoch);
-    if (result > cancelAt)
+    auto expiration = ContextSharedState::FromDateTimeRepresentation(ptr->Expiration);
+    if (result > expiration)
```
JeffreyRichter, line 20: Please add a comment here to explain the algorithm

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: That's because `ContextSharedState` wants an atomic timestamp. Our DateTime is not explicitly atomic, but it is essentially an `uint64` (I put the explanantion in PR description). int64 can be made atomic, and this is what we do there.

DateTime can be made explicitly supporting atomic, but then it means that we are also committing to customers that `std::atomic<DateTime>` would always compile and work.

Now, whether Context's deadline should be atomic, I am not fully questioning that, but at first it seems it should - so that one thread can cancel the context while other is checking whether it has expired.

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: Our `Azure::DateTime` does not have `Now()`. `std::chrono::system_clock` has, it returns `std::chrono::system_clock::time_point`, which can be implcitly converted to `Azure::DateTime`.

antkmsft:
@JeffreyRichter , I'll rename Expiration to Deadline as you are suggesting.
What do you think if instead of two methods `CreateWithDeadline()` and `CreateWithValue()`, we create two overloads (taking timestamp or key-value) named `CreateChildContext()`?

vhvb1989 (approved):
LGTM, I don't have strong feelings about the naming, so, I would go with any name

antkmsft:
@Jinming-Hu, in response to your question, see `HasExpiration()`. Would it help you?
@JeffreyRichter, I renamed "Expiration" to "Deadline", and put a comment explaining walking the tree / DateTime::max() as you suggested.

Plus, I renamed `WithDeadline()` and `WithValue` to tow overloads of `CreateChildContext()`.

ahsonkhan:
The benefit of having named factory methods, especially using `Deadline` is that it ties the scenario of cancellation to context and makes it clearer that `Context` can be used for that. We even observed this in the usability studies, where the API became self-documenting, and the person, just by method names could infer what they could do with the type.

The `WithDeadLine` and `WithValue` methods also match with golang's prior art:
https://golang.org/pkg/context/#example_WithDeadline

Therefore, we should leave the API name as is, in this case, rather than overloading it with a generic name.

JeffreyRichter:
I agree, please leave the method names as they were.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
   - `SetHeader(uint8_t const* const first, uint8_t const* const last)`.
   - `GetMajorVersion`.
   - `GetMinorVersion`.
+- Changes to `Azure::Core::Context`:
+  - Renamed `Get()` to `GetValue()`.
+  - Renamed `WithValue()` to `CreateChildContext()`.
+  - Renamed `WithDeadline()` to `CreateChildContext()`, and changed input parameter type to `Azure::DateTime`.
```
ahsonkhan, line 20: Revert the rename.

ahsonkhan (changes requested):
Keep `WithValue` and `WithDeadline`

sdk/core/azure-core/CHANGELOG.md
antkmsft, line 20:
```suggestion
  - Changed input parameter type of `WithDeadline()` to `Azure::DateTime`.
```

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -128,32 +122,40 @@ namespace Azure { namespace Core {
...
+     * @brief Create a context with a deadline.
      *
-     * @param cancelWhen A point in time after which a context expires.
+     * @param deadline A point in time after which a context expires.
      *
-     * @return A child context with expiration.
+     * @return A child context with deadline.
      */
-    Context WithDeadline(time_point cancelWhen) const
+    Context CreateChildContext(DateTime const& deadline) const
```
antkmsft, line 131:
```suggestion
    Context WithDeadline(DateTime const& deadline) const
```

sdk/core/azure-core/src/http/retry_policy.cpp
```diff
@@ -161,9 +161,9 @@ Context inline CreateRetryContext(Context const& parent)
   int retryCount = 0;
   if (parent.HasKey(RetryKey))
   {
-    retryCount = parent.Get<int>(RetryKey) + 1;
+    retryCount = parent.GetValue<int>(RetryKey) + 1;
   }
-  return parent.WithValue(RetryKey, retryCount);
+  return parent.CreateChildContext(RetryKey, retryCount);
```
antkmsft, line 166:
```suggestion
  return parent.WithValue(RetryKey, retryCount);
```

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -77,7 +77,7 @@ TEST(Context, IsCancelled)
   auto deadline = std::chrono::system_clock::now() + duration;
 
   Context context;
-  auto c2 = context.WithDeadline(deadline);
+  auto c2 = context.CreateChildContext(deadline);
```
antkmsft, line 80:
```suggestion
  auto c2 = context.WithDeadline(deadline);
```

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -96,7 +96,7 @@ TEST(Context, NestedIsCancelled)
   EXPECT_TRUE(c2.HasKey(key));
   EXPECT_FALSE(context.HasKey(key));
 
-  auto c3 = context.WithDeadline(deadline);
+  auto c3 = context.CreateChildContext(deadline);
```
antkmsft, line 99:
```suggestion
  auto c3 = context.WithDeadline(deadline);
```

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -136,7 +136,7 @@ TEST(Context, ThrowIfCancelled)
   auto deadline = std::chrono::system_clock::now() + duration;
 
   Context context;
-  auto c2 = context.WithDeadline(deadline);
+  auto c2 = context.CreateChildContext(deadline);
```
antkmsft, line 139:
```suggestion
  auto c2 = context.WithDeadline(deadline);
```

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -325,3 +326,51 @@ TEST(Context, KeyComparison)
...
+    EXPECT_EQ(ctx.GetDeadline(), Azure::DateTime::max());
+
+    ctx.Cancel();
+    EXPECT_TRUE(ctx.HasDeadline());
+    EXPECT_EQ(ctx.GetDeadline(), Azure::DateTime::min());
+  }
+
+  {
+    Context ctx;
+    ctx = ctx.CreateChildContext(deadline);
```
antkmsft, line 348:
```suggestion
    ctx = ctx.WithDeadline(deadline);
```

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -325,3 +326,51 @@ TEST(Context, KeyComparison)
...
+    EXPECT_EQ(childCtx.GetDeadline(), deadline);
+  }
+
+  {
+    Context ctx;
+    ctx.Cancel();
+
+    auto childCtx = ctx.CreateChildContext(deadline)
+                        .CreateChildContext(key, "val")
+                        .CreateChildContext(key2, "val2");
```
antkmsft, line 371:
```suggestion
    auto childCtx = ctx.WithDeadline(deadline)
                        .WithValue(key, "val")
                        .WithValue(key2, "val2");
```

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -325,3 +326,51 @@ TEST(Context, KeyComparison)
...
+    EXPECT_TRUE(ctx.HasDeadline());
+    EXPECT_EQ(ctx.GetDeadline(), deadline);
+  }
+
+  {
+    Context ctx;
+
+    auto childCtx = ctx.CreateChildContext(deadline)
+                        .CreateChildContext(key, "val")
+                        .CreateChildContext(key2, "val2");
```
antkmsft, line 359:
```suggestion
    auto childCtx = ctx.WithDeadline(deadline)
                        .WithValue(key, "val")
                        .WithValue(key2, "val2");
```

sdk/keyvault/azure-security-keyvault-keys/test/ut/key_client_delete_test_live.cpp
```diff
@@ -59,7 +59,7 @@ TEST_F(KeyVaultClientTest, DeleteKey)
     // Setting 3 min as timeout just because I like number 3. We just want to prevent test running
     // for so long if something happens and no exception is thrown (paranoid scenario)
     auto duration = std::chrono::system_clock::now() + std::chrono::minutes(3);
-    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
+    auto cancelToken = Azure::Core::Context::GetApplicationContext().CreateChildContext(duration);
```
antkmsft, line 62:
```suggestion
    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
```

sdk/keyvault/azure-security-keyvault-keys/test/ut/key_client_delete_test_live.cpp
```diff
@@ -147,7 +147,7 @@ TEST_F(KeyVaultClientTest, DoubleDelete)
   }
   {
     auto duration = std::chrono::system_clock::now() + std::chrono::minutes(3);
-    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
+    auto cancelToken = Azure::Core::Context::GetApplicationContext().CreateChildContext(duration);
```
antkmsft, line 150:
```suggestion
    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
```

sdk/keyvault/azure-security-keyvault-keys/test/ut/key_client_delete_test_live.cpp
```diff
@@ -221,7 +221,7 @@ TEST_F(KeyVaultClientTest, CreateDeletedKey)
...
```
antkmsft, line 224:
```suggestion
    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
```

sdk/keyvault/azure-security-keyvault-keys/test/ut/key_client_delete_test_live.cpp
```diff
@@ -302,7 +302,7 @@ TEST_F(KeyVaultClientTest, GetDeletedKey)
   {
     // Wait until key is deleted
     auto duration = std::chrono::system_clock::now() + std::chrono::minutes(3);
-    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
+    auto cancelToken = Azure::Core::Context::GetApplicationContext().CreateChildContext(duration);
```
antkmsft, line 305:
```suggestion
    auto cancelToken = Azure::Core::Context::GetApplicationContext().WithDeadline(duration);
```

sdk/storage/azure-storage-common/inc/azure/storage/common/storage_switch_to_secondary_policy.hpp
```diff
@@ -16,7 +16,7 @@ namespace Azure { namespace Storage { namespace _internal {
 
   inline Azure::Core::Context WithReplicaStatus(const Azure::Core::Context& context)
   {
-    return context.WithValue(SecondaryHostReplicaStatusKey, std::make_shared<bool>(true));
+    return context.CreateChildContext(SecondaryHostReplicaStatusKey, std::make_shared<bool>(true));
```
antkmsft, line 19:
```suggestion
    return context.WithValue(SecondaryHostReplicaStatusKey, std::make_shared<bool>(true));
```

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -35,7 +37,7 @@ namespace Azure { namespace Core {
   };
 
   /**
-   * @brief A context is a node within a tree that represents expiration times and key/value pairs.
+   * @brief A context is a node within a tree that represents deadline times and key/value pairs.
```
ahsonkhan, line 40: The use of `deadline times` reads strangely. `expiration` was better.

If you really want to keep it, maybe `represents deadlines and ...`

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -55,53 +57,47 @@ namespace Azure { namespace Core {
...
-     */
-    using time_point = std::chrono::system_clock::time_point;
-
+    
   private:
     struct ContextSharedState
     {
       std::shared_ptr<ContextSharedState> Parent;
-      std::atomic_int64_t CancelAtMsecSinceEpoch;
+      std::atomic<DateTime::rep> Deadline;
```
ahsonkhan, line 65: What's a `::rep`? Is it public?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -114,11 +110,9 @@ namespace Azure { namespace Core {
     {
     }
 
-    time_point CancelWhen() const;
-
   public:
     /**
-     * @brief Construct a new context with no expiration, and no value associated.
+     * @brief Construct a new context with no deadline, and no value associated.
```
ahsonkhan, line 115: Do we have a concept of `max` or `infinite` deadline? Should we use that rather than `no deadline`?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -128,32 +122,40 @@ namespace Azure { namespace Core {
...
      */
     template <class T> Context WithValue(Key const& key, T&& value) const
     {
       return Context{std::make_shared<ContextSharedState>(
-          m_contextSharedState, (time_point::max)(), key, std::forward<T>(value))};
+          m_contextSharedState, (DateTime::max)(), key, std::forward<T>(value))};
     }
 
+    /**
+     * @brief Get a deadline time point associated with this context or the branch of contexts this
```
ahsonkhan, line 151:
```suggestion
     * @brief Get a deadline associated with this context or the branch of contexts this
```

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -182,6 +184,18 @@ namespace Azure { namespace Core {
...
+     * @brief Check whether the context has a deadline in it itself, or in the branch the context
+     * belongs to.
+     *
+     * @return `true` if this context, or the tree branch this context belongs to has a deadline
+     * associated with it. `false` otherwise.
+     */
+    bool HasDeadline() const
+    {
+      return static_cast<DateTime::time_point>(GetDeadline()) != (DateTime::max)();
+    }
```
ahsonkhan, line 197: Can we defer adding this to later, once we see a need for it. @JeffreyRichter is this method really necessary for 1.0?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -128,32 +122,40 @@ namespace Azure { namespace Core {
...
       return Context{std::make_shared<ContextSharedState>(
-          m_contextSharedState, (time_point::max)(), key, std::forward<T>(value))};
+          m_contextSharedState, (DateTime::max)(), key, std::forward<T>(value))};
     }
 
+    /**
+     * @brief Get a deadline time point associated with this context or the branch of contexts this
+     * context belongs to.
+     *
+     * @return A deadline associated with the context found; `Azure::DateTime::max()` value if a
```
ahsonkhan, line 154: What does `Azure::DateTime::max()` mean and return? Is it public because of time_point max? https://en.cppreference.com/w/cpp/chrono/time_point/max

sdk/core/azure-core/test/ut/context.cpp
```diff
@@ -325,3 +325,51 @@ TEST(Context, KeyComparison)
...
+  Context::Key const key2;
+  
+  {
+    Context ctx;
+    EXPECT_FALSE(ctx.HasDeadline());
+    EXPECT_EQ(ctx.GetDeadline(), Azure::DateTime::max());
+
+    ctx.Cancel();
+    EXPECT_TRUE(ctx.HasDeadline());
+    EXPECT_EQ(ctx.GetDeadline(), Azure::DateTime::min());
```
ahsonkhan, line 342: @JeffreyRichter It is odd that we lose information such as the actual deadline time after `ctx.Cancel()` is called. Should we set the deadline to `::now()` when `Cancel()` get's called? This way the caller knows when the context was cancelled.

sdk/core/azure-core/inc/azure/core/context.hpp
JeffreyRichter: No, we don't need this

sdk/core/azure-core/inc/azure/core/context.hpp
JeffreyRichter: I don't think it will matter in real life. So, setting it to now instead of min is fine with me.

sdk/core/azure-core/test/ut/context.cpp
JeffreyRichter: I don't think it will matter in real life. So, setting it to now instead of min is fine with me.

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: yes - it is inherited from `std::chrono_time_point`

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: We treat `DateTime::max()` as infinite deadline. Which, for the user, the better way to think of, as "no deadline". `HasDeadline()` method naming corresponds to that.

sdk/core/azure-core/test/ut/context.cpp
antkmsft: @ahsonkhan , let's track that feature request separately, please open an issue if you think we should preserve the cancelation time. We can store `system_clock::now()` as cancelation but then the context state should have `bool isCanceled`. Currently, all that `Cancel()` does is it sets the deadline to `DateTime::min()`. What I don't want to happen is the context getting resurrected once system clock gets adjusted.

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: exactly

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: @Jinming-Hu asked for it. He would be using it: https://github.com/Azure/azure-sdk-for-cpp/pull/2010/files#diff-736005d9109cc9aa59c6ebc322fdb5dae5158a370c1ebb06ac7098a47561fc53R35-R40

What I can do, is to make an internal function.

JeffreyRichter: Just let @Jinming-Hu compare the deadline time with DateTime::max himself.
We dont' need functions for everything.

antkmsft: ok

antkmsft: @Jinming-Hu, `bool const hasDeadline = (ctx.GetDeadline() != Azure::DateTime::max());`

JeffreyRichter: I was thinking about this more. Why does @Jinming-Hu need this at all?
Most (maybe all) scenarios should just be caring whether the Context has reached deadline or not - the actual deadline time shouldn't matter.
If this is for passing timeout query param to service, then I think the policy should just add some timeout if the deadline is far enough into the future (say 1 week or something). Or, maybe do not add the timeout QP at all if the deadline is > 1 week out. 

I think that checking for the sentinel value is not the best way to go; someone could call WithDeadline(DateTime::max) to be explicit.

sdk/core/azure-core/test/ut/context.cpp
ahsonkhan: >  let's track that feature request separately, please open an issue

https://github.com/Azure/azure-sdk-for-cpp/issues/2038

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -56,52 +58,46 @@ namespace Azure { namespace Core {
...
+          : Parent(parent), Deadline(ToDateTimeRepresentation(deadline)), Value(nullptr),
             ValueType(typeid(std::nullptr_t))
       {
       }
 
       template <class T>
       explicit ContextSharedState(
           const std::shared_ptr<ContextSharedState>& parent,
-          time_point cancelAt,
+          DateTime const& deadline,
```
ahsonkhan, line 97: All these changes related to `ContextSharedState` are all private, correct?

ahsonkhan (approved):
LGTM

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: yes - `ContextSharedState` is `Context`'s private struct.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -26,6 +29,7 @@
 - Added `Azure::Core::Operation<T>::GetRawResponse()`.
 - Added `Azure::Core::PackageVersion`.
 - Added support for logging to console when `AZURE_LOG_LEVEL` environment variable is set.
+- Added `Azure::Core::Context::GetDeadline()`.
```
ahsonkhan, line 32: This is in the wrong release. Fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/2069

PR #2034, opened 2021-04-05
Renamed `Azure::Nullable<T>::GetValue()` to `Value()`.

ahsonkhan:
This rename makes `Nullable<T>` consistent with other Core types like `Operation<T>` and the C++ `Optional<T>`

https://en.cppreference.com/w/cpp/utility/optional/value
https://github.com/Azure/azure-sdk-for-cpp/blob/a7fdea42e65435c354e64c862f8490e410aef0e2/sdk/core/azure-core/inc/azure/core/operation.hpp#L82

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -14,6 +14,7 @@
   - `SetHeader(uint8_t const* const first, uint8_t const* const last)`.
   - `GetMajorVersion`.
   - `GetMinorVersion`.
+- Renamed `Azure::Nullable<T>::GetValue()` to `Value()`.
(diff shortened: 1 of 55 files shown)


JeffreyRichter (approved):

ahsonkhan:
@RickWinter, @antkmsft any feedback or review on https://github.com/Azure/azure-sdk-for-cpp/pull/2034​​​​​​​ before I merge it?

RickWinter (approved):

antkmsft (approved):
I prefer `Get...()` and `Create...()` but I guess our SDK is no longer uniform about that, and we don't have a guideline for that, and I'd prefer if we had one; But as of this change, I am fine with it, given that there's no such naming guideline.

RickWinter:
> I prefer `Get...()` and `Create...()` but I guess our SDK is no longer uniform about that, and we don't have a guideline for that, and I'd prefer if we had one; But as of this change, I am fine with it, given that there's no such naming guideline.

Value is idiomatic to STL in C++.  This change was about standardizing on Value.  If there are placing that aren't standardized we need to fix them.

PR #2050, opened 2021-04-06
remove all usage of AppendQueryParameters

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-common/src/storage_credential.cpp
+++ b/sdk/storage/azure-storage-common/src/storage_credential.cpp
@@ -96,10 +96,18 @@ namespace Azure { namespace Storage { namespace _internal {
     std::string sas = getWithDefault(connectionStringMap, "SharedAccessSignature");
     if (!sas.empty())
     {
-      connectionStringParts.BlobServiceUrl.AppendQueryParameters(sas);
-      connectionStringParts.DataLakeServiceUrl.AppendQueryParameters(sas);
-      connectionStringParts.FileServiceUrl.AppendQueryParameters(sas);
-      connectionStringParts.QueueServiceUrl.AppendQueryParameters(sas);
+      if (sas[0] != '?')
+      {
+        sas = '?' + sas;
+      }
+      connectionStringParts.BlobServiceUrl
+          = Azure::Core::Url(connectionStringParts.BlobServiceUrl.GetAbsoluteUrl() + sas);
+      connectionStringParts.DataLakeServiceUrl
+          = Azure::Core::Url(connectionStringParts.DataLakeServiceUrl.GetAbsoluteUrl() + sas);
+      connectionStringParts.FileServiceUrl
+          = Azure::Core::Url(connectionStringParts.FileServiceUrl.GetAbsoluteUrl() + sas);
+      connectionStringParts.QueueServiceUrl
+          = Azure::Core::Url(connectionStringParts.QueueServiceUrl.GetAbsoluteUrl() + sas);
     }
 
     return connectionStringParts;


sdk/storage/azure-storage-common/src/storage_credential.cpp
```diff
@@ -96,10 +96,18 @@ namespace Azure { namespace Storage { namespace _internal {
     std::string sas = getWithDefault(connectionStringMap, "SharedAccessSignature");
```
ahsonkhan, line 96: Question: Rather than building a map from the connection string and then doing this look up to build `connectionStringParts`, why not build the parts directly at this point?
https://github.com/Azure/azure-sdk-for-cpp/blob/c455f5d6d65eeaf24f103240b4ffeb88b5b12aca/sdk/storage/azure-storage-common/src/storage_credential.cpp#L12-L40

sdk/storage/azure-storage-common/src/storage_credential.cpp
```diff
@@ -96,10 +96,18 @@ namespace Azure { namespace Storage { namespace _internal {
...
     if (!sas.empty())
     {
-      connectionStringParts.BlobServiceUrl.AppendQueryParameters(sas);
-      connectionStringParts.DataLakeServiceUrl.AppendQueryParameters(sas);
-      connectionStringParts.FileServiceUrl.AppendQueryParameters(sas);
-      connectionStringParts.QueueServiceUrl.AppendQueryParameters(sas);
+      if (sas[0] != '?')
+      {
+        sas = '?' + sas;
+      }
```
ahsonkhan, line 102: When would this condition be true?

sdk/storage/azure-storage-common/src/storage_credential.cpp
```diff
@@ -96,10 +96,18 @@ namespace Azure { namespace Storage { namespace _internal {
...
+      connectionStringParts.BlobServiceUrl
+          = Azure::Core::Url(connectionStringParts.BlobServiceUrl.GetAbsoluteUrl() + sas);
+      connectionStringParts.DataLakeServiceUrl
+          = Azure::Core::Url(connectionStringParts.DataLakeServiceUrl.GetAbsoluteUrl() + sas);
+      connectionStringParts.FileServiceUrl
+          = Azure::Core::Url(connectionStringParts.FileServiceUrl.GetAbsoluteUrl() + sas);
+      connectionStringParts.QueueServiceUrl
+          = Azure::Core::Url(connectionStringParts.QueueServiceUrl.GetAbsoluteUrl() + sas);
```
ahsonkhan, line 110: Can you add a comment in source explaining what this is doing exactly?

To help me understand, can you show me with a fake example of a storage service connection string which once passed into the `ParseConnectionString` method, how it changes these `Url` values?

ahsonkhan (approved):
I have some questions to understand the logic a bit better. Looks good otherwise.

sdk/storage/azure-storage-common/src/storage_credential.cpp
Jinming-Hu: image


image

sometimes the SAS token has a leading `?` while sometimes it doesn't

ahsonkhan: Why does it not impact correctness for us to add back a `?` to the generated URL when the SAS token doesn't start with a leading `?`?

Jinming-Hu: There has to be a `?` between path and query parameters in URL. So if there's no `?` in query parameters, we add one. If there's already one, we don't add.

sdk/storage/azure-storage-common/src/storage_credential.cpp
Jinming-Hu: Parsing connection string is kind of complicated. The code is long and not easy to understand, especially when mixed with other code logics.

So I prefer separating the code logics into two parts, first parsing connection string, then building the urls.

Issue #2041, opened 2021-04-05
Consider removing `Url::AppendQueryParameters` and update storage connection string parsing logic.


ahsonkhan:
This method is too bespoke and has custom logic that isn't commonly useful in a `Url` class:
https://github.com/Azure/azure-sdk-for-cpp/blob/9c7d4c61e48dc4b76da937c194840dc90b172894/sdk/core/azure-core/inc/azure/core/url.hpp#L186-L192

Currently, this API only has one usage scenario, which is in parsing the connection string, which is storage specific:
https://github.com/Azure/azure-sdk-for-cpp/blob/9c7d4c61e48dc4b76da937c194840dc90b172894/sdk/storage/azure-storage-common/src/storage_credential.cpp#L99-L102

cc @vhvb1989

ahsonkhan:
This connection parser is doing a lot of repetitive work, by first creating the map an d then parsing it repeatedly using `AppendQueryParameters`. This can be simplified and remove the call-sites.

cc @Jinming-Hu

Jinming-Hu:
@ahsonkhan Sure, I'll remove all usages of this API first, then you can remove the API.

Jinming-Hu:
done in https://github.com/Azure/azure-sdk-for-cpp/pull/2050

PR #2054, opened 2021-04-06
Removed `Url::AppendQueryParameters()` since it is no longer used within the SDK.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2041

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/url.hpp
+++ b/sdk/core/azure-core/inc/azure/core/url.hpp
@@ -58,6 +58,14 @@ namespace Azure { namespace Core {
 
     std::string GetUrlWithoutQuery(bool relative) const;
 
+    /**
+     * @brief Finds the first '?' symbol and parses everything after it as query parameters.
+     * separated by '&'.
+     *
+     * @param encodedQueryParameters String containing one or more query parameters.
+     */
+    void AppendQueryParameters(const std::string& encodedQueryParameters);
+
   public:
     /**
      * @brief Decodes \p value by transforming all escaped characters to it's non-encoded value.
@@ -167,14 +175,6 @@ namespace Azure { namespace Core {
     {
       m_encodedQueryParameters[encodedKey] = encodedValue;
     }
-
-    /**
-     * @brief Finds the first '?' symbol and parses everything after it as query parameters.
-     * separated by '&'.
-     *
-     * @param encodedQueryParameters String containing one or more query parameters.
-     */
-    void AppendQueryParameters(const std::string& encodedQueryParameters);
 
     /**
      * @brief Removes an existing query parameter.
--- a/sdk/core/azure-core/inc/azure/core/url.hpp
+++ b/sdk/core/azure-core/inc/azure/core/url.hpp
@@ -223,17 +223,9 @@ namespace Azure { namespace Core {
     }
 
     /**
-     * @brief Get Scheme, host, and path, without query parameters.
-     * @return Absolute URL without query parameters.
+     * @brief Get the URL scheme.
      */
-    std::string GetUrlWithoutQuery() const { return GetUrlWithoutQuery(false); }
-
-    /**
-     * @brief Get Scheme, host and port.
-     *
-     * @return Url authority.
-     */
-    std::string GetUrlAuthorityWithScheme() const;
+    const std::string& GetScheme() const { return m_scheme; };
 
     /**
      * @brief Get the path and query parameters.
--- a/sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
+++ b/sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
@@ -10,6 +10,35 @@
 
 namespace Azure { namespace Storage { namespace Test {
 
+  static std::string AppendQP(Azure::Core::Url url, std::string qp)
+  {
+    std::string absoluteUrl = url.GetAbsoluteUrl();
+    auto existingQP = url.GetQueryParameters();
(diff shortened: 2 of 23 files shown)


sdk/storage/azure-storage-blobs/test/ut/page_blob_client_test.cpp
```diff
@@ -141,14 +141,44 @@ namespace Azure { namespace Storage { namespace Test {
         0, m_pageBlobClient->GetUrl() + GetSas(), {0, static_cast<int64_t>(m_blobContent.size())});
   }
 
+  static std::string AppendQP(Azure::Core::Url url, std::string qp)
```
ahsonkhan, line 144: @Jinming-Hu where should I put this test helper to avoid duplicating it twice.

sdk/core/azure-core/inc/azure/core/url.hpp
```diff
@@ -58,6 +58,14 @@ namespace Azure { namespace Core {
 
     std::string GetUrlWithoutQuery(bool relative) const;
 
+    /**
+     * @brief Finds the first '?' symbol and parses everything after it as query parameters.
+     * separated by '&'.
+     *
+     * @param encodedQueryParameters String containing one or more query parameters.
+     */
+    void AppendQueryParameters(const std::string& encodedQueryParameters);
```
ahsonkhan, line 67: Interestingly, this API doesn't append the query parameters at the end, but rather in the beginning. I would have expected it to append at the end.

For example, it goes from this:
```text
https://url/path?key1=value1
```
To:
```
https://url/path?se=date&sig=signature&sp=...&spr=...&sr=bv&st=anotherdate&sv=anothervalue&key1=value1
```

@vhvb1989, @Jinming-Hu - Is it generally OK to change the order of query parameters?

antkmsft: Should it be named `AddQueryParameters()` then? Prepending does not sound like "Append".
And for parameters order, I think it orders them alphabetically ignoring case? Anyways, it is Add() and not Append() if it does not puts them to the end.

antkmsft (approved):

sdk/core/azure-core/inc/azure/core/url.hpp
ahsonkhan: Since it is private now, the name is not as important anymore. That said I would be OK with renaming it to `AddQueryParameters`.

Jinming-Hu: It's a `std::map`. Query parameters will be in alphabetical order, right?

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
```diff
@@ -10,6 +10,37 @@
 
 namespace Azure { namespace Storage { namespace Test {
 
+  static std::string AppendQP(Azure::Core::Url url, std::string qp)
```
Jinming-Hu, line 13:
```suggestion
  static std::string AppendQueryParameters(const Azure::Core::Url& url, const std::string& qp)
```

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
```diff
@@ -10,6 +10,37 @@
...
+  {
+    std::string absoluteUrl = url.GetAbsoluteUrl();
+    auto existingQP = url.GetQueryParameters();
```
Jinming-Hu, line 16:
```suggestion
    const auto& existingQP = url.GetQueryParameters();
```

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
```diff
@@ -10,6 +10,37 @@
...
+    bool startWithQuestion = qp[0] == '?';
```
Jinming-Hu, line 17: What if `qp` is empty?

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
```diff
@@ -10,6 +10,37 @@
...
+    if (existingQP.size() == 0)
```
Jinming-Hu, line 18:
```suggestion
    if (existingQP.empty())
```

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
Jinming-Hu, line 13: We should define this function in `test_base.hpp` and `test_base.cpp` if it's used more than once

Jinming-Hu (approved):

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
ahsonkhan: This is in tests. Why would it be empty? If it's empty, don't call the method :)

sdk/storage/azure-storage-blobs/test/ut/blob_sas_test.cpp
ahsonkhan: test_base.hpp is in `storage common`. This is only used in `storage blobs` test. Do you still recommend moving it in there, rather than a new private header within the storage blobs test source?

Jinming-Hu: Yes, who's to say we're not going to use it in another package sometime. It's a general purpose function, it should go in common package.

Issue #2040, opened 2021-04-05
Remove `GetUrlWithoutQuery()` and update the log policy call-site to use the individual URL component getters.


ahsonkhan:
https://github.com/Azure/azure-sdk-for-cpp/blob/c5ddbbc430790fb8718dab7df792633c63f33d19/sdk/core/azure-core/inc/azure/core/url.hpp#L225-L229

cc @vhvb1989

ahsonkhan:
@antkmsft - can you remove the usage of `GetUrlWithoutQuery` within log policy?
https://github.com/Azure/azure-sdk-for-cpp/blob/3b36c9091a6088281ca7fb17faa5335b1b04ffa3/sdk/core/azure-core/src/http/log_policy.cpp#L44-L46

PR #2055, opened 2021-04-06
Removed GetUrlWithoutQuery() and GetUrlAuthorityWithScheme() from Azure::Core::Url and added GetScheme().

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2040

### DIFF
--- a/sdk/core/azure-core/src/http/log_policy.cpp
+++ b/sdk/core/azure-core/src/http/log_policy.cpp
@@ -37,23 +37,21 @@ inline void AppendHeaders(
   }
 }
 
-inline std::string GetUrlWithoutQuery(Url url)
+inline void LogUrlWithoutQuery(std::ostringstream& log, Url const& url)
 {
-  std::string urlString;
   if (!url.GetScheme().empty())
   {
-    urlString += url.GetScheme() + "://";
+    log << url.GetScheme() << "://";
   }
-  urlString += url.GetHost();
+  log << url.GetHost();
   if (url.GetPort() != 0)
   {
-    urlString += ":" + std::to_string(url.GetPort());
+    log << ":" << url.GetPort();
   }
   if (!url.GetPath().empty())
   {
-    urlString += "/" + url.GetPath();
+    log << "/" << url.GetPath();
   }
-  return urlString;
 }
 
 inline std::string GetRequestLogMessage(LogOptions const& options, Request const& request)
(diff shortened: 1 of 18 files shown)


sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -37,13 +37,32 @@ inline void AppendHeaders(
   }
 }
 
+inline std::string GetUrlWithoutQuery(Url url)
```
antkmsft, line 40: This is going to be invoked when logging during each request. Let's use `std::ostringstream` - either take as a parameter (non-const ref), or build in a string stream, and then `return sstream.c_str();`.

sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -37,13 +37,32 @@ inline void AppendHeaders(
...
+}
+
 inline std::string GetRequestLogMessage(LogOptions const& options, Request const& request)
 {
   auto const& requestUrl = request.GetUrl();
 
   std::ostringstream log;
   log << "HTTP Request : " << request.GetMethod().ToString() << " "
-      << requestUrl.GetUrlWithoutQuery();
+      << GetUrlWithoutQuery(requestUrl);
```
antkmsft, line 65: Yep - better accept `ostringstream& log` as a parameter for `GetUrlWithoutQuery()`.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/details/key_serializers.hpp
```diff
@@ -40,11 +40,26 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Keys { nam
         KeyVaultKey& key,
         Azure::Core::Json::_internal::json const& json);
 
+    std::string static inline GetUrlAuthorityWithScheme(Azure::Core::Url url)
```
antkmsft, line 43: static before return type.
inline is not needed - functions defined within a class or structs are automatically inline.

antkmsft (approved):
LGTM, given that you take the feedback

sdk/core/azure-core/src/http/log_policy.cpp
antkmsft, line 40: Take value by const ref. We should avoid creating copies.
```suggestion
inline std::string GetUrlWithoutQuery(Url const& url)
```

sdk/core/azure-core/src/http/log_policy.cpp
ahsonkhan: What's the benefit? Right now the log ostringstream change is all in one place, I don't know if it's a good idea to have that be passed to the `GetUrlWithoutQuery()` method.

Issue #2045, opened 2021-04-05
Remove Request::GetHeadersAsString()


ahsonkhan:
Remove `GetHeadersAsString()`. It is only used by the transport adapters.

https://github.com/Azure/azure-sdk-for-cpp/blob/63b2d21bc65a4348cad5fd69cecbe353f74b7120/sdk/core/azure-core/inc/azure/core/http/http.hpp#L245-L248

cc @vhvb1989

PR #2056, opened 2021-04-06
Remove Azure::Core::Http::Request::GetHeadersAsString().

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2045

### DIFF
--- a/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
+++ b/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
@@ -22,23 +22,13 @@ inline std::wstring HttpMethodToWideString(HttpMethod method)
   // This string should be all uppercase.
   // Many servers treat HTTP verbs as case-sensitive, and the Internet Engineering Task Force (IETF)
   // Requests for Comments (RFCs) spell these verbs using uppercase characters only.
-  switch (method)
-  {
-    case HttpMethod::Get:
-      return L"GET";
-    case HttpMethod::Head:
-      return L"HEAD";
-    case HttpMethod::Post:
-      return L"POST";
-    case HttpMethod::Put:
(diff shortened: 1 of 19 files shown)


sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -292,6 +292,22 @@ void WinHttpTransport::Upload(std::unique_ptr<_detail::HandleManager>& handleMan
   }
 }
 
+static inline std::string GetHeadersAsString(Azure::Core::Http::Request const& request)
```
antkmsft, line 295: Move to unnamed namespace, remove static.
it would be nice if you removed static from `MergeMaps` above in this file as well.

Static functions are visible to linker from outside of the cpp file. It can be linked with, names can collide. Even if it is inline - inline is a hint to compiler, not an order.
Everything defined in the unnamed namespace is not visible outside of the .cpp file.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -292,6 +292,22 @@ void WinHttpTransport::Upload(std::unique_ptr<_detail::HandleManager>& handleMan
...
+{
+  std::string requestHeaderString;
+
+  for (auto const& header : request.GetHeaders())
```
antkmsft, line 299: consider using ostringstream, it should be more performance-effective for building strings

antkmsft (approved):

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
ahsonkhan: Good idea, will do that separately.

PR #2061, opened 2021-04-07
small fix: const Type& to be consistent

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-common/test/test_base.cpp
+++ b/sdk/storage/azure-storage-common/test/test_base.cpp
@@ -141,7 +141,7 @@ namespace Azure { namespace Storage { namespace Test {
     return connectionString;
   }
 
-  std::string AppendQueryParameters(Azure::Core::Url const& url, std::string const& queryParameters)
+  std::string AppendQueryParameters(const Azure::Core::Url& url, const std::string& queryParameters)
   {
     std::string absoluteUrl = url.GetAbsoluteUrl();
     if (queryParameters.empty())
--- a/sdk/storage/azure-storage-common/test/test_base.hpp
+++ b/sdk/storage/azure-storage-common/test/test_base.hpp
@@ -44,8 +44,8 @@ namespace Azure { namespace Storage { namespace Test {
   }
 
   std::string AppendQueryParameters(
-      Azure::Core::Url const& url,
-      std::string const& queryParameters);
+      const Azure::Core::Url& url,
+      const std::string& queryParameters);
 
   const static Azure::ETag DummyETag("0x8D83B58BDF51D75");
   const static Azure::ETag DummyETag2("0x8D812645BFB0CDE");


sdk/storage/azure-storage-common/test/test_base.cpp
```diff
@@ -141,7 +141,7 @@ namespace Azure { namespace Storage { namespace Test {
     return connectionString;
   }
 
-  std::string AppendQueryParameters(Azure::Core::Url const& url, std::string const& queryParameters)
+  std::string AppendQueryParameters(const Azure::Core::Url& url, const std::string& queryParameters)
```
ahsonkhan, line 144: @Jinming-Hu - we are not consistent within storage/core about the ordering of `const`. Should we decide on and settle on one particular ordering and update all instances?

In many places within core/keyvault, we use `type_name const&`.

I saw storage using `const type_name&` in most places, except for when it is overriding base classes from core.
https://github.com/Azure/azure-sdk-for-cpp/blob/4db9a563e75bd7f67ffd3f2a895577ecda122ef1/sdk/storage/azure-storage-common/inc/azure/storage/common/reliable_stream.hpp#L55

https://github.com/Azure/azure-sdk-for-cpp/blob/4db9a563e75bd7f67ffd3f2a895577ecda122ef1/sdk/storage/azure-storage-common/inc/azure/storage/common/shared_key_policy.hpp#L29-L32

ahsonkhan (approved):
Looks good but we should pick a consistent pattern across all the SDK and stick to it. cc @RickWinter, @CaseyCarter - which pattern is more common/idiomatic. We should add it in the guidelines.

ahsonkhan:
@Jinming-Hu Is changing the order of where `const` shows up in the API a breaking change? I would assume it isn't.

Therefore, I don't see this PR as blocking release (especially since it is just test code).

antkmsft (approved):
LGTM, right const is better though :P
https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/

ahsonkhan:
For reference, in the Embedded C SDK, we use `type_name const*`:
https://github.com/Azure/azure-sdk-for-c/blob/6a61fdc180cb43d9bb6e22d06166724bd4156e73/sdk/inc/azure/core/az_json.h#L139

sdk/storage/azure-storage-common/test/test_base.cpp
Jinming-Hu: > we are not consistent within storage/core about the ordering of const. Should we decide on and settle on one particular ordering and update all instances?

Yeah, I know. But it's better if we can keep consistent in each sdk than not even consistent within one sdk, isn't it?

Jinming-Hu:
> LGTM, right const is better though :P
> https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/

This is a very controversial topic. I can also find a lot of arguments for left-const. 

like Linux kernel code uses left-const https://elixir.bootlin.com/linux/latest/source/lib/bootconfig.c

Jinming-Hu:
> @Jinming-Hu Is changing the order of where `const` shows up in the API considered a breaking change? I would assume it isn't.

No, it's not

antkmsft:
Left const certainly isn't bad. IMO, it is actually the second best, right after the right const being the top.

But, if we adopt #2062, we'll decide on the left const, I think: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-const

sdk/storage/azure-storage-common/test/test_base.cpp
ahsonkhan: > Yeah, I know. But it's better if we can keep consistent in each sdk than not even consistent within one sdk, isn't it?

Yes :)

Jinming-Hu:
> Left const certainly isn't bad. IMO, it is actually the second best, right after the right const being the top.
> 
> But, if we adopt #2062, we'll decide on the left const, I think: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-const

By saying `second best`, you mean it's worst? I don't think there is a third option other than left-const and right const.

This kind of debate can go on forever.

antkmsft:
Please forgive my humor :) Worst is the inconsistent one - sometimes on the left, sometimes on the right.
Jokes aside, left const has at least that benefit that it looks less unusual to the eyes of the most customers, and it is more important, because we are source-distributed.
I think I might be able to use some tool to unify it (move to the left) across all our repo.

Re: endless right vs left const debate:
"A student asks a professor with a long beard: 'When you sleep, do you put your beard under or over the blanket?' - 'I don't know. What kind of question is that?'. Next week professor runs to the student in anger: 'I lost my sleep! Before you asked that question, I was sleeping just fine!'"

PR #2092, opened 2021-04-09
pageable for ListBlobs

Jinming-Hu:
I only added Listblobs while didn't remove other APIs. I want this proposal get approved first before starting making other changes.

The example API in this PR is

```C++
Pageable<ListBlobsPageResult> ListBlobs(
        const ListBlobsOptions& options = ListBlobsOptions(),
        const Azure::Core::Context& context = Azure::Core::Context()) const
```

usage examples are:

```C++
// use the same context for all HTTP requests
for (auto& pageResult : client.ListBlobs())
{
    for (auto& blob : pageResult.Items)
    {
        std::cout << blob.name << std::endl;
    }
    // access continuation token
    // pageResult.NextPageToken

    // access raw response
    // pageResult.RawResponse

    // if there was another colletion
    // for (auto& blobPrefix : pageResult.Prefixes) {}
}
```

```C++
// list from a continuation token we got from other place
ListBlobOptions options;
options.ContinuationToken = token_from_other_place;
for (auto& pageResult : client.ListBlobs(options))
{
    for (auto& blob : pageResult.Items) {}
}
```

```C++
// use different context for each requests:
for (auto& pageResult = *client.ListBlobs(options, context).begin(); pageResult.HasMore(); pageResult.NextPage(context))
{
    for (auto& blob : pageResult.Items) {}
}
```

### DIFF
--- a/sdk/storage/azure-storage-blobs/src/blob_container_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_container_client.cpp
@@ -261,6 +261,22 @@ namespace Azure { namespace Storage { namespace Blobs {
     return response;
   }
 
+  Pageable<ListBlobsPageResult> BlobContainerClient::ListBlobs(
+      const ListBlobsOptions& options,
+      const Azure::Core::Context& context) const
+  {
+    ListBlobsPageResult pageResult;
+    pageResult.NextPageToken = options.ContinuationToken.ValueOr(std::string());
+    pageResult.m_blobContainerClient = std::make_shared<BlobContainerClient>(*this);
+    pageResult.m_operationOptions = options;
+    pageResult.m_context = _internal::WithReplicaStatus(context);
+
+    // Populate the first page
+    pageResult.OnNextPage(pageResult.m_context);
+
+    return Pageable<ListBlobsPageResult>(std::move(pageResult));
+  }
+
   Azure::Response<Models::ListBlobsByHierarchySinglePageResult>
   BlobContainerClient::ListBlobsByHierarchySinglePage(
       const std::string& delimiter,
--- a/sdk/storage/azure-storage-blobs/src/blob_responses.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_responses.cpp
@@ -61,6 +61,13 @@ namespace Azure { namespace Storage { namespace Blobs {
 
   void ListBlobsPageResult::OnNextPage(const Azure::Core::Context& context)
   {
+    auto contextRef = std::cref(context);
+    if (&context == &Azure::Core::Context::GetApplicationContext())
+    {
+      // Called via iterator increment, use the same context as the last call.
+      contextRef = std::cref(m_context);
+    }
+
     _detail::BlobRestClient::BlobContainer::ListBlobsSinglePageOptions protocolLayerOptions;
     protocolLayerOptions.Prefix = m_operationOptions.Prefix;
     protocolLayerOptions.ContinuationToken = NextPageToken;
@@ -70,7 +77,7 @@ namespace Azure { namespace Storage { namespace Blobs {
         *m_blobContainerClient->m_pipeline,
         m_blobContainerClient->m_blobContainerUrl,
         protocolLayerOptions,
-        context);
+        contextRef.get());
     for (auto& i : response.Value.Items)
     {
       if (i.Details.AccessTier.HasValue() && !i.Details.IsAccessTierInferred.HasValue())
--- a/sdk/storage/azure-storage-blobs/src/blob_responses.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_responses.cpp
@@ -4,6 +4,7 @@
 #include "azure/storage/blobs/blob_responses.hpp"
 
 #include "azure/storage/blobs/blob_client.hpp"
+#include "azure/storage/blobs/blob_container_client.hpp"
 
 namespace Azure { namespace Storage { namespace Blobs {
 
@@ -56,6 +57,43 @@ namespace Azure { namespace Storage { namespace Blobs {
 
       std::this_thread::sleep_for(period);
     };
+  }
+
+  void ListBlobsPageResult::OnNextPage(const Azure::Core::Context& context)
+  {
+    _detail::BlobRestClient::BlobContainer::ListBlobsSinglePageOptions protocolLayerOptions;
+    protocolLayerOptions.Prefix = m_operationOptions.Prefix;
+    protocolLayerOptions.ContinuationToken = NextPageToken;
+    protocolLayerOptions.MaxResults = m_operationOptions.PageSizeHint;
+    protocolLayerOptions.Include = m_operationOptions.Include;
+    auto response = _detail::BlobRestClient::BlobContainer::ListBlobsSinglePage(
+        *m_blobContainerClient->m_pipeline,
+        m_blobContainerClient->m_blobContainerUrl,
+        protocolLayerOptions,
+        context);
+    for (auto& i : response.Value.Items)
+    {
+      if (i.Details.AccessTier.HasValue() && !i.Details.IsAccessTierInferred.HasValue())
+      {
+        i.Details.IsAccessTierInferred = false;
+      }
+      if (i.VersionId.HasValue() && !i.IsCurrentVersion.HasValue())
+      {
+        i.IsCurrentVersion = false;
+      }
+      if (i.BlobType == Models::BlobType::AppendBlob && !i.Details.IsSealed)
+      {
+        i.Details.IsSealed = false;
+      }
+    }
+
+    ServiceEndpoint = std::move(response.Value.ServiceEndpoint);
+    BlobContainerName = std::move(response.Value.BlobContainerName);
+    Prefix = std::move(response.Value.Prefix);
+    Items = std::move(response.Value.Items);
+
+    NextPageToken = response.Value.ContinuationToken.ValueOr(std::string());
+    RawResponse = std::move(response.RawResponse);
   }
 
 }}} // namespace Azure::Storage::Blobs
--- a/sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
+++ b/sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
@@ -227,6 +227,66 @@ namespace Azure { namespace Storage { namespace Test {
     EXPECT_TRUE(std::includes(listBlobs.begin(), listBlobs.end(), p1Blobs.begin(), p1Blobs.end()));
   }
 
+  TEST_F(BlobContainerClientTest, ListBlobsFlatPageable)
+  {
+    const std::string prefix1 = RandomString();
+    const std::string baseName = "blob";
+
+    std::set<std::string> blobs;
+
+    for (int i = 0; i < 10; ++i)
+    {
+      std::string blobName = prefix1 + baseName + std::to_string(i);
+      auto blobClient = m_blobContainerClient->GetBlockBlobClient(blobName);
+      auto emptyContent = Azure::Core::IO::MemoryBodyStream(nullptr, 0);
+      blobClient.Upload(emptyContent);
+      blobs.insert(blobName);
+    }
+
+    {
+      // General
+      std::set<std::string> listBlobs;
+      Azure::Storage::Blobs::ListBlobsOptions options;
+      options.PageSizeHint = 3;
+      for (auto& pageResult : m_blobContainerClient->ListBlobs(options))
+      {
+        for (auto& b : pageResult.Items)
+        {
+          listBlobs.emplace(b.Name);
+        }
+      }
+      EXPECT_TRUE(std::includes(listBlobs.begin(), listBlobs.end(), blobs.begin(), blobs.end()));
+    }
+
+    {
+      // With continuation token
+      std::set<std::string> listBlobs;
+      Azure::Storage::Blobs::ListBlobsOptions options;
+      options.PageSizeHint = 3;
+      for (auto& pageResult : m_blobContainerClient->ListBlobs(options))
+      {
+        for (auto& b : pageResult.Items)
+        {
+          listBlobs.emplace(b.Name);
+        }
+        ASSERT_TRUE(pageResult.HasMore());
+        if (pageResult.HasMore())
+        {
+          options.ContinuationToken = pageResult.NextPageToken;
+          break;
+        }
+      }
+      for (auto& pageResult : m_blobContainerClient->ListBlobs(options))
+      {
+        for (auto& b : pageResult.Items)
+        {
+          listBlobs.emplace(b.Name);
+        }
+      }
+      EXPECT_TRUE(std::includes(listBlobs.begin(), listBlobs.end(), blobs.begin(), blobs.end()));
+    }
+  }
+
   TEST_F(BlobContainerClientTest, ListBlobsHierarchy)
   {
     const std::string delimiter = "/";
--- a/sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
+++ b/sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
@@ -263,19 +263,15 @@ namespace Azure { namespace Storage { namespace Test {
       std::set<std::string> listBlobs;
       Azure::Storage::Blobs::ListBlobsOptions options;
       options.PageSizeHint = 3;
-      for (auto& pageResult : m_blobContainerClient->ListBlobs(options))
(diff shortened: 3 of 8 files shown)


sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,110 @@
...
+      using iterator_category = std::input_iterator_tag;
+      using difference_type = std::ptrdiff_t;
+      using value_type = T;
+      using pointer = value_type*;
+      using reference = value_type&;
+
+      explicit Iterator(pointer ptr) : m_ptr(ptr) {}
+      Iterator(const Iterator&) = default;
+      Iterator(Iterator&&) = default;
+      ~Iterator() = default;
```
Jinming-Hu, line 69:
```suggestion
```

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+      Iterator& operator=(const Iterator& other) = default;
+      Iterator& operator=(Iterator&& other) = default;
```
Jinming-Hu, line 72:
```suggestion
```

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,105 @@
...
+      ~PageResult() = default;
+      PageResult(PageResult&&) = default;
+      PageResult& operator=(PageResult&&) = default;
+
+      Azure::Core::Context m_context;
+    };
+
+  } // namespace _internal
+
+  template <class T> class Pageable {
```
Jinming-Hu, line 49: Question:

Should we have this `Pageable` at all?
We can move `Iterator` to `PageResult<T>`.

But I think, `Pageable` works like a collection, so naturally it's iterable. `PageResult<T>` corresponds to one page, itself is not iterable, it may contain some fields that are iterable.

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+  namespace _internal {
+    template <class Derived> class PageResult {
+    public:
+      std::unique_ptr<Azure::Core::Http::RawResponse> RawResponse;
+      std::string NextPageToken;
+
+      bool HasMore() const { return !NextPageToken.empty(); }
+
+      void NextPage()
```
Jinming-Hu, line 23: I'm thinking about adding an overload that accepts a `Context`, so that user can use a different context for each HTTP request

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <type_traits>
```
ahsonkhan, line 8: What is this needed for?

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include <azure/core/context.hpp>
+#include <azure/core/http/raw_response.hpp>
+
+namespace Azure { namespace Storage {
```
ahsonkhan, line 13: Keep it in the Azure namespace since that's where it will end up once moved down to core. It helps reduce the impact of the break.
```suggestion
namespace Azure {
```

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+  namespace _internal {
+    template <class Derived> class PageResult {
```
ahsonkhan, line 16: Why make this internal? Could all this be part of `Pageable<T>` instead, and we remove `PageResult`?

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,104 @@
...
+          // NextPage() should NEVER be called if HasMore() is false
+          std::abort(); // or throw std::out_of_range
+        }
+
+        static_assert(
+            std::is_base_of<PageResult, Derived>::value,
+            "The template argument \"Derived\" should derive from PageResult<Derived>.");
+
+        CurrentPageToken = NextPageToken;
+        static_cast<Derived*>(this)->OnNextPage(context);
```
ahsonkhan, line 37: Consider using `dynamic_cast` instead. It helps code auditing sine we minimize static_cast usage. Could we merge the static_assert above if we use dynamic_cast?

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
```diff
@@ -0,0 +1,104 @@
...
+  template <class T> class Pageable {
+  public:
+    explicit Pageable(T value) : m_value(std::move(value))
+    {
+      static_assert(
+          std::is_base_of<_internal::PageResult<T>, T>::value,
+          "The template argument should derive from PageResult");
+    }
+
+    class Iterator {
```
ahsonkhan, line 57: 1) Does the Iterator class have to be public?
2) If we decide to not make Pageable<T> today, can we make it iteratable in the future by adding this back, in a non-breaking way?

If so, maybe we can leave out iterator for now and revisit it later once we get some usability feedback from customers that request it.

sdk/storage/azure-storage-blobs/src/blob_responses.cpp
```diff
@@ -58,4 +59,48 @@ namespace Azure { namespace Storage { namespace Blobs {
...
   }
 
+  void ListBlobsPageResult::OnNextPage(const Azure::Core::Context& context)
+  {
+    auto contextRef = std::cref(context);
+    if (&context == &Azure::Core::Context::GetApplicationContext())
+    {
+      // Called via iterator increment, use the same context as the last call.
+      contextRef = std::cref(m_context);
+    }
```
ahsonkhan, line 69: For v1, we may want to remain explicit and avoid this "magic" of reusing the same context.

sdk/storage/azure-storage-blobs/src/blob_responses.cpp
```diff
@@ -58,4 +59,48 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+      }
+      if (i.VersionId.HasValue() && !i.IsCurrentVersion.HasValue())
+      {
+        i.IsCurrentVersion = false;
+      }
+      if (i.BlobType == Models::BlobType::AppendBlob && !i.Details.IsSealed)
+      {
+        i.Details.IsSealed = false;
+      }
+    }
```
ahsonkhan, line 95: @JeffreyRichter we seem to already be iterating through the list of results returned from the service. Is that expected?

The `ListBlobsSinglePageResult` model type just contains a single `std::vector<BlobItem> Items;`

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
```diff
@@ -227,6 +227,62 @@ namespace Azure { namespace Storage { namespace Test {
     EXPECT_TRUE(std::includes(listBlobs.begin(), listBlobs.end(), p1Blobs.begin(), p1Blobs.end()));
   }
 
+  TEST_F(BlobContainerClientTest, ListBlobsFlatPageable)
```
ahsonkhan, line 230: Can we add a test for hierarchical and not just the flat API?

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
```diff
@@ -227,6 +227,62 @@ namespace Azure { namespace Storage { namespace Test {
...
+    const std::string baseName = "blob";
+
+    std::set<std::string> blobs;
+
+    for (int i = 0; i < 10; ++i)
+    {
+      std::string blobName = prefix1 + baseName + std::to_string(i);
+      auto blobClient = m_blobContainerClient->GetBlockBlobClient(blobName);
+      auto emptyContent = Azure::Core::IO::MemoryBodyStream(nullptr, 0);
+      blobClient.Upload(emptyContent);
```
ahsonkhan, line 242: Wait, why would we allow a nullptr memorystream? What is the expectations when calling `blobClient.Upload` like that?

This issue would add input validation with either abort or throw if the argument was null.
https://github.com/Azure/azure-sdk-for-cpp/issues/1769

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
```diff
@@ -227,6 +227,62 @@ namespace Azure { namespace Storage { namespace Test {
...
+      blobs.insert(blobName);
+    }
+
+    {
+      // General
+      std::set<std::string> listBlobs;
+      Azure::Storage::Blobs::ListBlobsOptions options;
+      options.PageSizeHint = 3;
```
ahsonkhan, line 250: We should assert that at most 3 items were returned per page.

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
```diff
@@ -227,6 +227,62 @@ namespace Azure { namespace Storage { namespace Test {
...
+      auto firstPageResult = std::move(*m_blobContainerClient->ListBlobs(options).begin());
+
+      for (auto& b : firstPageResult.Items)
+      {
+        listBlobs.emplace(b.Name);
+      }
+      ASSERT_TRUE(firstPageResult.HasMore());
```
ahsonkhan, line 272: Assert `listBlobs.size() <= 3`

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
```diff
@@ -227,6 +227,62 @@ namespace Azure { namespace Storage { namespace Test {
...
+      blobClient.Upload(emptyContent);
+      blobs.insert(blobName);
+    }
+
+    {
+      // General
+      std::set<std::string> listBlobs;
+      Azure::Storage::Blobs::ListBlobsOptions options;
+      options.PageSizeHint = 3;
+      for (auto& pageResult : m_blobContainerClient->ListBlobs(options))
```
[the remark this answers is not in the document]
ahsonkhan: @Jinming-Hu I am OK with deferring adding the iterator pattern to v-next, if it can remain feasible. Could we remove this feature for now?

sdk/storage/azure-storage-blobs/src/blob_responses.cpp
```diff
@@ -58,4 +59,48 @@ namespace Azure { namespace Storage { namespace Blobs {
     };
   }
 
+  void ListBlobsPageResult::OnNextPage(const Azure::Core::Context& context)
+  {
+    auto contextRef = std::cref(context);
+    if (&context == &Azure::Core::Context::GetApplicationContext())
+    {
+      // Called via iterator increment, use the same context as the last call.
```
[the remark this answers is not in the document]
ahsonkhan: Care to elaborate Jeff why storing a context field not a good idea and why comparing the ref to the root context is dangerous?

I haven't seen this coding pattern with Context  elsewhere so it stood out to me as well.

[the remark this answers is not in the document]
JeffreyRichter: A context field means that the PageResponse is holding on to a Context so that it can be reused. But, we don't want to reuse them: each I/O operation gets its very own Context tree and this is important because the customer might want to specify that the first page operation should exist at 100pm and then the next page operation should expire at 1:10pm, etc.

If the first Context is used for all operations, then ALL operations must complete by 1pm and this really bad because the customer probably does not know how long processing each page's items will take and this time will be included in the Context's deadline and this is not the semantic meaning of Contexts. A Context is passed to a SINGLE I/O operation to control this operation's deadline.

The reason why comparing to root is dangerous/bad is because not all Contexts have the same root. So, this "magic code" has baked in an assumption that if the application context root is canceled then this operation should be canceled but the customer has never specified this and the behavior is unpredictable to them.

sdk/storage/azure-storage-blobs/src/blob_responses.cpp
JeffreyRichter: This loop is iterating over the single collection to set some default values. It's not allocating any memory so I think this is fine.

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
Jinming-Hu: 1. Does the Iterator class have to be public?

I think so.

2. If we decide to not make Pageable today, can we make it iteratable in the future by adding this back, in a non-breaking way?

If we decide to not make Pageable today, I guess we'll never have it. It'll be a breaking change.

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
Jinming-Hu: Yes, I just want this example get approved first before I start working on the others.

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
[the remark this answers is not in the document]
Jinming-Hu: @JeffreyRichter @ahsonkhan If you say so, but I guess then we'll never have iterator.

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
ahsonkhan: > If we decide to not make Pageable today, can we make it iteratable in the future by adding this back, in a non-breaking way?

Sorry, I meant if we decide to not make `Pageable<T>` **iteratable** today, can we make it iteratable in the future?

Jinming-Hu: > Sorry, I meant if we decide to not make Pageable<T> iteratable today, can we make it iteratable in the future?

I don't think so. To make iterator to work, I have to add the Pageable<T> class. The return type is different from the first prototype by Jeffrey. It's like a breaking  change.

sdk/storage/azure-storage-blobs/src/blob_container_client.cpp
```diff
@@ -261,6 +261,22 @@ namespace Azure { namespace Storage { namespace Blobs {
     return response;
   }
 
+  Pageable<ListBlobsPageResult> BlobContainerClient::ListBlobs(
+      const ListBlobsOptions& options,
+      const Azure::Core::Context& context) const
+  {
+    ListBlobsPageResult pageResult;
+    pageResult.NextPageToken = options.ContinuationToken.ValueOr(std::string());
+    pageResult.m_blobContainerClient = std::make_shared<BlobContainerClient>(*this);
```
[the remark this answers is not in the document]
Jinming-Hu: Just suppose there are two classes A and B and they depend on each other. How are you going to define them, the only way is to

```C++
struct A {
    B* m_b;  // have to be a pointer, because B is not defined yet.
};

struct B {
    A* m_a1;  // can be a pointer or value, becasue A is already defined.
    A m_a2;
};
```

sdk/storage/azure-storage-common/inc/azure/storage/common/paged_response.hpp
Jinming-Hu: If we know the correct type, we should use `static_cast` rather than `dynamic_cast`. `dynamic_cast` has runtime overhead.

sdk/storage/azure-storage-blobs/src/blob_container_client.cpp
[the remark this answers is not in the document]
ahsonkhan: Why does `ListBlobsPageResult` have a  `std::shared_ptr<BlobContainerClient> m_blobContainerClient;`? Why doesn't it take ownership of the client (unique ptr)?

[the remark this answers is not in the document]
Jinming-Hu: I don't think shared_ptr has any difference with unique_ptr in regard of ownership

PR #2101, opened 2021-04-12
Enable retain runs.

mitchdenny:
Enable run retention when tag is applied to repo.


mitchdenny:
Related: https://github.com/Azure/azure-sdk/issues/1059

weshaggard (approved):

PR #2138, opened 2021-04-20
VcPkg: use version-semver

antkmsft:


ahsonkhan:
@antkmsft can you clarify why version-semver is the preferred name to use? Is there a doc somewhere that recommends using it? The command that converted the CONTROL file to vcpkg.json didn't' use it. I am wondering why.

ahsonkhan (approved):
Good change. Thanks!
We follow semver compliant versioning, and this doc recommends using it instead of the free form `version-string`:
https://vcpkg.readthedocs.io/en/latest/specifications/versioning/#2-specifying-package-versions

PR #2145, opened 2021-04-21
Enable retain runs.

mitchdenny:
This PR enables the retain runs template in the C++ repo.


weshaggard (approved):

ahsonkhan:
@mitchdenny what does that let us do?

weshaggard:
@ahsonkhan this just makes sure the build pipelines that we end up release stick around longer so we can revisit logs or artifacts from that build longer.

Issue #1751, opened 2021-02-26
Implementation detail of many storage APIs shows up as public doxygen docs, confusing the end customer reading them


ahsonkhan:
The doxygen API reference docs published show a lot of the implementation detail of the storage APIs which make readers think these are usage samples for them to copy.
This makes the docs filled with a lot of noise that isn't relevant to the end customer.

Instead, example calling code or code snippets that end users can leverage should be written and displayed.

https://azuresdkdocs.blob.core.windows.net/$web/cpp/azure-storage-blobs/12.0.0-beta.8/class_azure_1_1_storage_1_1_blobs_1_1_blob_container_client.html#ad58f17faab40f56b891b3ba442b8bf8a
![image](https://user-images.githubusercontent.com/6527137/109346035-eb325700-7825-11eb-9cb7-49d56c6eeeef.png)

This is a side-effect of having all the implementation of the APIs within header files. There is probably some knob on doxygen to address this concern, in the short term.

cc @Jinming-Hu, @katmsft

Jinming-Hu:
what caused this issue? Is this some kind of bug of doxygen?

Jinming-Hu:
![image](https://user-images.githubusercontent.com/52266153/116175485-f846c600-a742-11eb-9b7a-ef4a76dbbf8d.png)
https://azuresdkdocs.blob.core.windows.net/$web/cpp/azure-core/1.0.0-beta.8/class_azure_1_1_core_1_1_http_1_1_raw_response.html

Jinming-Hu:
This is not a problem specific to storage.

Jinming-Hu:
https://github.com/Azure/azure-sdk-for-cpp/blob/fe9002e9054b04a2375da242e49badb1f6880252/cmake-modules/AzureDoxygen.cmake#L21 Change this value to `NO` will fix the problem

PR #2164, opened 2021-04-28
Update for doxygen inline docs

vhvb1989:
fixes: https://github.com/Azure/azure-sdk-for-cpp/issues/1751


sdk/core/azure-core/test/ut/md5.cpp
```diff
@@ -118,3 +118,33 @@ TEST(Md5Hash, CtorDtor)
...
+  // Each created thread will wait from 0 to 3 milliseconds to start to make threads overlap
+  for (int counter = 0; counter < size; counter++)
+  {
+    pool.emplace_back(std::thread(hashThreadRoutine, counter % 4));
+  }
+  for (int counter = 0; counter < size; counter++)
+  {
+    pool[counter].join();
+  }
+}
```
Jinming-Hu, line 150: Whoa, do it in another PR.

vhvb1989: My bad

vhvb1989: btw, thank you for the fix @Jinming-Hu , I saw you provided in the issue, I just took it and applied

cmake-modules/AzureDoxygen.cmake
```diff
@@ -18,7 +18,7 @@ function(generate_documentation PROJECT_NAME PROJECT_VERSION)
         set(DOXYGEN_LAYOUT_FILE ${CMAKE_SOURCE_DIR}/eng/docs/api/assets/DoxygenLayout.xml)
         set(DOXYGEN_RECURSIVE YES)
         set(DOXYGEN_USE_MDFILE_AS_MAINPAGE ./README.md)
-        set(DOXYGEN_INLINE_SOURCES YES)
+        set(DOXYGEN_INLINE_SOURCES NO)
```
Jinming-Hu, line 21: Do we need to check with the original author to see why it was `YES` before? Are there any side effects?

RickWinter (approved):
This looks fine assuming Enums are still showing up in documentation correctly

vhvb1989:
> This looks fine assuming Enums are still showing up in documentation correctly

It also fixes enums.  Sharing the before and after below:

## Before the change 

### Class

image

### Enum

image


## After

### Class

image


## Enum

image

PR #2165, opened 2021-04-28
FIX BUG: cannot upload 5000MiB blob, introduced in #2162

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
@@ -115,6 +115,16 @@ namespace Azure { namespace Storage { namespace Blobs {
     constexpr int64_t MaxBlockNumber = 50000;
     constexpr int64_t BlockGrainSize = 1 * 1024 * 1024;
 
+    if (bufferSize <= static_cast<std::size_t>(options.TransferOptions.SingleUploadThreshold))
+    {
+      Azure::Core::IO::MemoryBodyStream contentStream(buffer, bufferSize);
+      UploadBlockBlobOptions uploadBlockBlobOptions;
+      uploadBlockBlobOptions.HttpHeaders = options.HttpHeaders;
+      uploadBlockBlobOptions.Metadata = options.Metadata;
+      uploadBlockBlobOptions.AccessTier = options.AccessTier;
+      return Upload(contentStream, uploadBlockBlobOptions, context);
+    }
+
     int64_t chunkSize;
     if (options.TransferOptions.ChunkSize.HasValue())
     {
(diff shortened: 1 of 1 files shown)


katmsft (approved):

vhvb1989 (approved):
Can you add a test case for this? How did you discover it?

There is a build option called `RUN_LONG_UNIT_TESTS` which we could use for tests like these and we can add it to only one of the nightly ci tests (or maybe one in Windows and one in Linux).
But this is something we don't want regressions to happen.

Maybe doing 1Gb only is enough for testing?

Jinming-Hu:
> Can you add a test case for this? How did you discover it?
> 
> There is a build option called `RUN_LONG_UNIT_TESTS` which we could use for tests like these and we can add it to only one of the nightly ci tests (or maybe one in Windows and one in Linux).
> But this is something we don't want regressions to happen.
> 
> Maybe doing 1Gb only is enough for testing?

We do target test.

I don't think the vm used to run Azure pipelines has 5GB memory, so no, we cannot cover this with unit test.

ahsonkhan:
> I don't think the vm used to run Azure pipelines has 5GB memory, so no, we cannot cover this with unit test.

@danieljurek, @weshaggard I would imagine other language SDKs have similar issues when trying to stress test the storage SDK with large payloads. Do you have any recommendations?

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -115,6 +115,16 @@ namespace Azure { namespace Storage { namespace Blobs {
     constexpr int64_t MaxBlockNumber = 50000;
     constexpr int64_t BlockGrainSize = 1 * 1024 * 1024;
 
+    if (bufferSize <= static_cast<std::size_t>(options.TransferOptions.SingleUploadThreshold))
```
ahsonkhan, line 118: We should add a comment here to explain how this fixes the issue and why the check and return needs to be here rather than below.

ahsonkhan: Add a changelog entry please.

weshaggard:
>@danieljurek Daniel Jurek FTE, @weshaggard Wes Haggard FTE I would imagine other language SDKs have similar issues when trying to stress test the storage SDK with large payloads. Do you have any recommendations?

Yes we have actually setup another pool for storage tests in .NET which has 8GB of memory, see https://github.com/Azure/azure-sdk-for-net/pull/20037. That might be a pool worth considering for these tests as well.

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
Jinming-Hu: We don't need a CL entry. The bug was introduced after latest release.

PR #2175, opened 2021-04-29
Add HttpMethod extensible enum tests

### LINKED ISSUE #2051: Add test cases for the use of HttpMethod as extensible enum (using it with non-standard-pre-defined-strings)
One valid test case is the use of `Trace` which doesn't exist as an enum value today.

From https://github.com/Azure/azure-sdk-for-cpp/pull/2048#pullrequestreview-628257664

antkmsft:
Closes #2051

### DIFF
--- a/sdk/core/azure-core/test/ut/http_method.cpp
+++ b/sdk/core/azure-core/test/ut/http_method.cpp
@@ -1,0 +1,288 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <gtest/gtest.h>
+
+#include <azure/core/http/http.hpp>
+
+using namespace Azure::Core::Http;
+
+TEST(HttpMethod, Get)
+{
+  HttpMethod const get = HttpMethod::Get;
+  EXPECT_EQ(get.ToString(), "GET");
+
+  EXPECT_EQ(HttpMethod::Get, HttpMethod::Get);
+  EXPECT_EQ(get, get);
+  EXPECT_EQ(get, HttpMethod("GET"));
+  EXPECT_EQ(HttpMethod("GET"), HttpMethod::Get);
+
+  EXPECT_NE(get, HttpMethod("Get"));
+  EXPECT_NE(HttpMethod("Get"), get);
+
+  EXPECT_NE(get, HttpMethod(std::string()));
+  EXPECT_NE(HttpMethod(std::string()), get);
+
+  EXPECT_EQ(get, HttpMethod::Get);
+  EXPECT_EQ(HttpMethod::Get, get);
+
+  EXPECT_NE(get, HttpMethod::Head);
+  EXPECT_NE(get, HttpMethod::Post);
+  EXPECT_NE(get, HttpMethod::Put);
+  EXPECT_NE(get, HttpMethod::Delete);
+  EXPECT_NE(get, HttpMethod::Patch);
+  EXPECT_NE(get, HttpMethod("TRACE"));
+  EXPECT_NE(get, HttpMethod("HUMUHUMUNUKUNUKUAPUAA"));
+
+  EXPECT_NE(HttpMethod::Head, get);
+  EXPECT_NE(HttpMethod::Post, get);
+  EXPECT_NE(HttpMethod::Put, get);
+  EXPECT_NE(HttpMethod::Delete, get);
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/test/ut/http_method.cpp
```diff
@@ -0,0 +1,288 @@
...
+  EXPECT_NE(HttpMethod::Post, trace);
+  EXPECT_NE(HttpMethod::Put, trace);
+  EXPECT_NE(HttpMethod::Delete, trace);
+  EXPECT_NE(HttpMethod::Patch, trace);
+  EXPECT_NE(HttpMethod("HUMUHUMUNUKUNUKUAPUAA"), trace);
+}
+
+TEST(HttpMethod, HawaiianFish)
+{
+  HttpMethod const humuhumunukunukuapuaa = HttpMethod("HUMUHUMUNUKUNUKUAPUAA");
```
ahsonkhan, line 260: Lol :)

image

sdk/core/azure-core/test/ut/http_method.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+using namespace Azure::Core::Http;
+
+TEST(HttpMethod, Get)
+{
+  HttpMethod const get = HttpMethod::Get;
+  EXPECT_EQ(get.ToString(), "GET");
+
+  EXPECT_EQ(HttpMethod::Get, HttpMethod::Get);
+  EXPECT_EQ(get, get);
```
ahsonkhan, line 16: What is this testing? I don't think we need it.

sdk/core/azure-core/test/ut/http_method.cpp
```diff
@@ -0,0 +1,288 @@
...
+  EXPECT_NE(head, HttpMethod("TRACE"));
+  EXPECT_NE(head, HttpMethod("HUMUHUMUNUKUNUKUAPUAA"));
+
+  EXPECT_NE(HttpMethod::Get, head);
+  EXPECT_NE(HttpMethod::Post, head);
+  EXPECT_NE(HttpMethod::Put, head);
+  EXPECT_NE(HttpMethod::Delete, head);
+  EXPECT_NE(HttpMethod::Patch, head);
+  EXPECT_NE(HttpMethod("TRACE"), head);
+  EXPECT_NE(HttpMethod("HUMUHUMUNUKUNUKUAPUAA"), head);
```
ahsonkhan, line 79: A lot of this is being duplicated for each of the enum values. Could we extract these to a helper function instead to avoid duplication?

ahsonkhan (approved):
Maybe re-write some of the tests to avoid code duplication so that if/when we add new enum values or tests, it doesn't require a lot of code addition.

sdk/core/azure-core/test/ut/http_method.cpp
antkmsft: Yes, a name that is unlikely to ever be picked for HTTP method :)

sdk/core/azure-core/test/ut/http_method.cpp
antkmsft: It can be done, but the failure logs aren't probably going to be that clear, and the code would not be identical, because it would not use `HttpMethod::<Method>`, but a copy of it instead, which is technically not the same thing, and likely do not reflect the most common usage in user's code.

antkmsft: So, I would rather leave it as is. Certainly would've been easier to remove the dup - it took me some time to have almost identical code repeated correctly. But that was kind of an intent - to use primitive methods and make minimum assumptions, and just write plain code that does verification. But I am not super strong on this one. If you feel strongly, I can refactor it into one method that gets invoked with different parameters.

sdk/core/azure-core/test/ut/http_method.cpp
antkmsft: My other candidate was `Eyjafjallajokull`, but in the end I thought it is slightly more likely to be picked for HTTP method name in the future (after 42th century).

ahsonkhan:
As part of the initial issue request, we also want to add a test for the http pipeline and send request using the HttpMethod created from "TRACE" and make sure that succeeds. That way, we ensure users can actually successfully pass in a request with a valid HTTP method, that happens to not be available as an enum value in our SDK today.

PR #2184, opened 2021-05-04
Return false when Context::TryGetValue has a type mismatch, rather than aborting.

ahsonkhan:
~Todo: Add unit tests~

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -174,13 +174,10 @@ namespace Azure { namespace Core {
     {
       for (auto ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
       {
-        if (ptr->Key == key)
+        // Only return true if we find a key-type pair that matches.
+        // Otherwise, keep looping through the context graph.
+        if (ptr->Key == key && typeid(T) != ptr->ValueType)
         {
-          if (typeid(T) != ptr->ValueType)
-          {
-            // type mismatch
-            return false;
-          }
           outputValue = *reinterpret_cast<const T*>(ptr->Value.get());
           return true;
         }
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -165,7 +165,8 @@ namespace Azure { namespace Core {
      * within the context tree.
      *
      * @return If found, returns `true`, with \p outputValue set to the value associated with the
-     * key found; otherwise returns `false`.
+     * key found; otherwise returns `false`, including when the key found is not of the expected
+     * type.
      *
      * @remark The \p outputValue is left unmodified it the \p key is not found.
      */
@@ -178,7 +179,7 @@ namespace Azure { namespace Core {
           if (typeid(T) != ptr->ValueType)
           {
             // type mismatch
-            std::abort();
+            return false;
           }
           outputValue = *reinterpret_cast<const T*>(ptr->Value.get());
           return true;
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -176,7 +176,7 @@ namespace Azure { namespace Core {
       {
         // Only return true if we find a key-type pair that matches.
         // Otherwise, keep looping through the context graph.
-        if (ptr->Key == key && typeid(T) != ptr->ValueType)
+        if (ptr->Key == key && typeid(T) == ptr->ValueType)
         {
           outputValue = *reinterpret_cast<const T*>(ptr->Value.get());
           return true;
--- a/sdk/core/azure-core/test/ut/context.cpp
+++ b/sdk/core/azure-core/test/ut/context.cpp
@@ -88,6 +88,36 @@ TEST(Context, BasicChar)
   EXPECT_TRUE(value == str);
 }
 
+TEST(Context, KeyTypePair)
+{
+  Context context;
+  Context::Key const key;
+  Context::Key const keyNotFound;
+
+  std::string s("Test String");
+
+  // New context from previous
+  auto c2 = context.WithValue(key, 123);
+  auto c3 = c2.WithValue(key, s);
+
+  int intValue = -1;
+  std::string strValue = "previous value";
+
+  EXPECT_FALSE(c2.TryGetValue<std::string>(keyNotFound, strValue));
+  EXPECT_FALSE(c2.TryGetValue<int>(keyNotFound, intValue));
+
+  EXPECT_FALSE(c2.TryGetValue<std::string>(key, strValue));
+  EXPECT_TRUE(strValue == "previous value");
+  EXPECT_TRUE(c2.TryGetValue<int>(key, intValue));
+  EXPECT_TRUE(intValue == 123);
+
+  EXPECT_TRUE(c3.TryGetValue<int>(key, intValue));
+  EXPECT_TRUE(intValue == 123);
+
+  EXPECT_TRUE(c3.TryGetValue<std::string>(key, strValue));
+  EXPECT_TRUE(strValue == s);
+}
+
 TEST(Context, IsCancelled)
 {
   auto duration = std::chrono::milliseconds(250);


JeffreyRichter (approved):

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -178,7 +179,7 @@ namespace Azure { namespace Core {
           if (typeid(T) != ptr->ValueType)
           {
             // type mismatch
-            std::abort();
+            return false;
```
vhvb1989, line 182: `false` has a different meaning. It is for when the key is not found.  In this case, the key is found but there is a logical/semantic error on the application that needs to be fixed (and can be fixed).  We must abort here to let people fix the bug in the code.

vhvb1989:
see: https://github.com/Azure/azure-sdk-for-cpp/pull/2170

JeffreyRichter:
I think returning false is ok. It means the key with the specified type is not found. It's not necessarily a bug because the same key could be stumbled upon.

- Jeffrey Richter (see Architecting Distributed Cloud Apps<https://na01.safelinks.protection.outlook.com/?url=http%3A%2F%2Faka.ms%2FRichterCloudApps&data=04%7C01%7C%7Cdc05b9290f9f445693ec08d8e1b30ed8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637507506469305857%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=TeBH6rge3q8VyFBwCrBXfSXND4oajfqayGLiZq1gmFo%3D&reserved=0> & Designing HTTP APIs<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D9Ng00IlBCtw%26list%3DPL9XzOCngAkqs4m0XdULJu_78nM3Ok3Q65%26index%3D1&data=04%7C01%7C%7Cdc05b9290f9f445693ec08d8e1b30ed8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637507506469305857%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=nzunz4x55lMRzHb6qJVXuRkrlAAI03wXjPdexaPavzs%3D&reserved=0>)

vhvb1989:
> I think returning false is ok. It means the key with the specified type is not found. It's not necessarily a bug because the same key could be stumbled upon.

Does it mean we support adding multiple key names just with different type?
if yes, then we should not return false or throw, we should then only check if the type if the same to return the value and if not, we just keep traversing/looking for a key with the right type.

And only after all the loop, we return false as `not found`.

But this is only if we want support same key-name with specific type.  If not, we should abort as there is a bug of someone using writing the wrong type to be gotten v/s what it was added to the context.

JeffreyRichter:
Good point. Yes, the loop should look for the first matching key-type pair and turn false only if no context matches.

- Jeffrey Richter (see Architecting Distributed Cloud Apps<https://na01.safelinks.protection.outlook.com/?url=http%3A%2F%2Faka.ms%2FRichterCloudApps&data=04%7C01%7C%7Cdc05b9290f9f445693ec08d8e1b30ed8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637507506469305857%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=TeBH6rge3q8VyFBwCrBXfSXND4oajfqayGLiZq1gmFo%3D&reserved=0> & Designing HTTP APIs<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D9Ng00IlBCtw%26list%3DPL9XzOCngAkqs4m0XdULJu_78nM3Ok3Q65%26index%3D1&data=04%7C01%7C%7Cdc05b9290f9f445693ec08d8e1b30ed8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637507506469305857%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=nzunz4x55lMRzHb6qJVXuRkrlAAI03wXjPdexaPavzs%3D&reserved=0>)

vhvb1989 (approved):
LGTM, the test cases you mentioned should be enough

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -165,21 +165,19 @@ namespace Azure { namespace Core {
...
+     * type.
      *
      * @remark The \p outputValue is left unmodified it the \p key is not found.
      */
     template <class T> bool TryGetValue(Key const& key, T& outputValue) const
     {
       for (auto ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
       {
-        if (ptr->Key == key)
+        // Only return true if we find a key-type pair that matches.
```
RickWinter, line 177: This is a contract violation as the code running after this point will NOT do what the customer expects.  I would expect this to be an abort().
Note: The keys are tied to a specific value instance.  They are not string keys.  Thus there should never be a key inserted that has an invalid type tied to it.

RickWinter (changes requested):
Can you elaborate on the scenario where we want to accept a Key/Value where the Value is not of the expected type.  Perhaps we could assert the type matches the Key, i.e. you can only insert the expected type.  Thus we would abort when the user creates the Value and the type doesn't match the expected type.

Then we could leave this abort as it should never be reachable.

sdk/core/azure-core/inc/azure/core/context.hpp
ahsonkhan: Contexts with the same key can be added in the graph to override a previous value, there is no reason to block changing the type of that value as an override as well. So, this isn't really considered a contract violation. It should continue to work as customers expect.

ahsonkhan:
Sharing some context from offline design discussion that motivates this change.

Unlike GoLang, the C++ Context type requires the user calling `TryGetValue` to know, up front, the type of the value associated with the particular key.

What happens if I don’t own the context (let’s say it is being passed-in), where I didn’t create the context calling WithValue, but some other component did, and we have a mechanism to share the lookup key (which is always a `Context::Key` type anyway).
That component is free to change the type of the value corresponding to that key, and now my existing code that might assume a particular value type, like int, breaks (with an abort).
There is no way for me to know what the type needs to be for me to successfully access the key.

Golang allows a mechanism to get the type and do a safe cast to that type so the user can write code that is resilient to the owner changing the type of the context value. Returning false from `TryGetValue` enables that type of tester-doer pattern in C++ where we don't have reflection style type assertion.

```C++
  Context context;
  Context::Key const key;

  // This works fine
  auto c2 = context.WithValue(key, true);
  bool value;
  EXPECT_TRUE(c2.TryGetValue<bool>(key, value)); // How do I know the value is of type bool?
  EXPECT_TRUE(value == true);

  // Tomorrow, the owner of context and c2 changes the value type to string
  // Now, existing code breaks
  c2 = context.WithValue(key, "true");
  bool value;
  EXPECT_TRUE(c2.TryGetValue<bool>(key, value)); // This aborts due to a type mismatch

```

Go developers can write code using context continues to work even when the owner of the context changes the value type:
https://play.golang.org/p/zMHnqR_g61V

sdk/core/azure-core/inc/azure/core/context.hpp
RickWinter: I agree you can add the same value to override.  I don't agree that changing the type of the object associated with that key is expected to work.  Their is a type coupling of the Key -> typeof(Value).  We need to help customers find their bugs when they try to put the wrong type in with a Key.

JeffreyRichter:
Why can't it be a coupling between key/type -> value?

- Jeffrey Richter (see Architecting Distributed Cloud Apps<https://na01.safelinks.protection.outlook.com/?url=http%3A%2F%2Faka.ms%2FRichterCloudApps&data=04%7C01%7C%7Cdc05b9290f9f445693ec08d8e1b30ed8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637507506469305857%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=TeBH6rge3q8VyFBwCrBXfSXND4oajfqayGLiZq1gmFo%3D&reserved=0> & Designing HTTP APIs<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D9Ng00IlBCtw%26list%3DPL9XzOCngAkqs4m0XdULJu_78nM3Ok3Q65%26index%3D1&data=04%7C01%7C%7Cdc05b9290f9f445693ec08d8e1b30ed8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637507506469305857%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=nzunz4x55lMRzHb6qJVXuRkrlAAI03wXjPdexaPavzs%3D&reserved=0>)

RickWinter:
@JeffreyRichter I believe what you are asking is what I'm suggesting.  They Key has an explicit type that is the value.  We should enforce that.

antkmsft (approved):

Jinming-Hu (approved):
I know it's still controversial whether or not we should abort in this case. I'm fine either way. I don't think we'll ever hit this edge case in the real world.

ahsonkhan:
We decided to defer adding this functionality until it is requested by customers in the future. Going from a precondition failure to returning false is a non-breaking change.

I extracted the test into https://github.com/Azure/azure-sdk-for-cpp/pull/2223

Hence, closing.

Issue #1853, opened 2021-03-11
Avoid, or minimize, exposing windows.h header from our public headers (either directly or indirectly)


ahsonkhan:
1) For places where we have private fields coming from that header, use `void*` and cast it to the appropriate type in the cpp source file. This is similar to:
https://github.com/Azure/azure-sdk-for-cpp/blob/a9c0637e462f80340ac7b912fe9c52df0b4d39ec/sdk/storage/azure-storage-common/inc/azure/storage/common/xml_wrapper.hpp#L40
https://github.com/Azure/azure-sdk-for-cpp/blob/a9c0637e462f80340ac7b912fe9c52df0b4d39ec/sdk/storage/azure-storage-common/src/xml_wrapper.cpp#L42

2) For places where windows.h is needed as an implementation, split that out of the header file and into cpp.
3) No public APIs should be accepting types from windows.h.

**Here's the list in order of priority:**
- [ ] body_stream.hpp https://github.com/Azure/azure-sdk-for-cpp/blob/3d6cd38658548747587adb84462a7a2d2910275b/sdk/core/azure-core/inc/azure/core/io/body_stream.hpp#L22
- [ ] hkeyholder.hpp (though it is an internal header, not exposed via any public headers) https://github.com/Azure/azure-sdk-for-cpp/blob/bc564dbddd0371e222105bd539bbbe8ba8e45fe1/sdk/core/azure-core/inc/azure/core/internal/hkeyholder.hpp#L21
- [ ] environment_log_level_listener_private.hpp (low priority, but is the include even needed here?) https://github.com/Azure/azure-sdk-for-cpp/blob/a9c0637e462f80340ac7b912fe9c52df0b4d39ec/sdk/core/azure-core/src/environment_log_level_listener_private.hpp#L16

Jinming-Hu:
I'll handle `file_io.hpp`

ahsonkhan:
@RickWinter let's consider adding this to the guidelines too.

RickWinter:
@ahsonkhan please split this work item into two tasks.  One for the work in Azure::Core and the other for work in Azure::Storage

ahsonkhan:
@RickWinter - done - https://github.com/Azure/azure-sdk-for-cpp/issues/2182

Issue #2182, opened 2021-05-03
Within Azure Storage, avoid, or minimize, exposing windows.h header from our public headers (either directly or indirectly)


ahsonkhan:
Related to https://github.com/Azure/azure-sdk-for-cpp/issues/1853, which is focusing on changes to Azure::Core.

1) For places where we have private fields coming from that header, use `void*` and cast it to the appropriate type in the cpp source file. This is similar to:
https://github.com/Azure/azure-sdk-for-cpp/blob/a9c0637e462f80340ac7b912fe9c52df0b4d39ec/sdk/storage/azure-storage-common/inc/azure/storage/common/xml_wrapper.hpp#L40
https://github.com/Azure/azure-sdk-for-cpp/blob/a9c0637e462f80340ac7b912fe9c52df0b4d39ec/sdk/storage/azure-storage-common/src/xml_wrapper.cpp#L42
2) For places where windows.h is needed as an implementation, split that out of the header file and into cpp.
3) No public APIs should be accepting types from windows.h.

**Here's the list in order of priority:**
- [ ] file_io.hpp (though it isn't currently exposed via other public headers) https://github.com/Azure/azure-sdk-for-cpp/blob/bc564dbddd0371e222105bd539bbbe8ba8e45fe1/sdk/storage/azure-storage-common/inc/azure/storage/common/file_io.hpp#L15

@Jinming-Hu, @katmsft  do you know of any other places that need fixing within storage?

PR #2188, opened 2021-05-04
Remove exposing windows.h header from our public headers and reduce risk of introducing it in non-public headers

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1853


sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -220,7 +220,7 @@ namespace Azure { namespace Storage { namespace Blobs {
 
     auto uploadBlockFunc = [&](int64_t offset, int64_t length, int64_t chunkId, int64_t numChunks) {
       Azure::Core::IO::_internal::RandomAccessFileBodyStream contentStream(
-          fileReader.GetHandle(), offset, length);
+          static_cast<void*>(fileReader.GetHandle()), offset, length);
```
ahsonkhan, line 223: This is left for storage to address: https://github.com/Azure/azure-sdk-for-cpp/issues/2182

sdk/core/azure-core/src/private/environment_log_level_listener.hpp
```diff
@@ -12,8 +12,6 @@
 #if !defined(NOMINMAX)
 #define NOMINMAX
 #endif
-
-#include <windows.h>
```
ahsonkhan, line 16: @antkmsft is there any reason for us to have this header included here? Same with the `WIN32_LEAN_AND_MEAN`/etc. above.

sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -17,7 +17,47 @@
 
 #include <windows.h>
 
-#include "azure/core/internal/hkey_holder.hpp"
+#if !defined(WINAPI_PARTITION_DESKTOP) \
+    || WINAPI_PARTITION_DESKTOP // See azure/core/platform.hpp for explanation.
+
+namespace Azure { namespace Core { namespace _internal {
```
ahsonkhan, line 23: @RickWinter what are your thoughts on the hkey_holder fix by moving the header into the only source that needs it? Specifically https://github.com/Azure/azure-sdk-for-cpp/pull/2188/commits/d713ea023bfb41d1a0b604afdfb5edadf023d079

sdk/core/azure-core/src/private/environment_log_level_listener.hpp
antkmsft: Yes - so it is only available on non-UWP Windows, and to detect that, we need windows.h

sdk/core/azure-core/inc/azure/core/io/body_stream.hpp
```diff
@@ -220,7 +219,7 @@ namespace Azure { namespace Core { namespace IO {
        * for the necessary duration. The caller is also responsible for closing it once they are
        * done.
        */
-      RandomAccessFileBodyStream(HANDLE fileHandle, int64_t offset, int64_t length)
+      RandomAccessFileBodyStream(void* fileHandle, int64_t offset, int64_t length)
```
antkmsft, line 222: Not a fan of `void*` here :(

antkmsft: maybe we don't avoid including windows.h here.
If you use `RandomAccessFileBodyStream`, you nave to have it included.
So maybe either be fine with including windows.h, or extract `RandomAccessFileBodyStream` to a separate header - so that if you don't use it, Win32 does not pop in your intellisense.

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: I think hkey_holder can be moved to sources.
And also I don't see you updating telemetry_policy that uses it.

antkmsft: should you delete it, make sure you do
```
try
{
   handle = OpenHandle() // pseudocode
   ...
   CloseHandle() // pseudocode
}
catch(...) // literally 3 dots
{
   CloseHandle() // pseudocode
   throw; // literally just "throw;" without an exception
}
```

ahsonkhan: > I think hkey_holder can be moved to sources.
And also I don't see you updating telemetry_policy that uses it.

I am confused. I did.

sdk/core/azure-core/inc/azure/core/io/body_stream.hpp
ahsonkhan: This type is internal, so `void*` doesn't bother me as much. Not removing `windows.h` here in `body_stream.hpp` would effectively mean all SDK users get exposed to it, which is precisely what we want to avoid, as part of this.

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: Got it! You moved it here. This is fine. Maybe better to move it to a private (src) header. But not critical, to me.
In a separate header or not, should we remove other windows.h header usages or not, this is a good change.

sdk/core/azure-core/inc/azure/core/io/body_stream.hpp
antkmsft: I see. If for internal - I don't care about `void*` too much :)
Maybe it is better to move internal stuff to a header in `internal/`.

antkmsft (approved):

sdk/core/azure-core/inc/azure/core/io/body_stream.hpp
```diff
@@ -19,7 +19,6 @@
 #if !defined(NOMINMAX)
 #define NOMINMAX
 #endif
-#include <windows.h>
```
Jinming-Hu, line 22:
```C++
#include <unistd.h>
#endif
```

Jinming-Hu: we should move `WIN32_LEAN_AND_MEAN` and `NOMINMAX` together with `windows.h`.

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -11,6 +11,12 @@
 #include <azure/storage/common/storage_common.hpp>
 #include <azure/storage/common/storage_switch_to_secondary_policy.hpp>
 
+#include <azure/core/platform.hpp>
```
Jinming-Hu, line 14: This header should go with the group above it.

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -11,6 +11,12 @@
...
+
+#if defined(AZ_PLATFORM_WINDOWS)
+#include <windows.h>
+#endif
```
Jinming-Hu, line 18: I prefer moving it after Line 5.

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -219,8 +225,13 @@ namespace Azure { namespace Storage { namespace Blobs {
     _internal::FileReader fileReader(fileName);
 
     auto uploadBlockFunc = [&](int64_t offset, int64_t length, int64_t chunkId, int64_t numChunks) {
+#if defined(AZ_PLATFORM_WINDOWS)
+      Azure::Core::IO::_internal::RandomAccessFileBodyStream contentStream(
+          reinterpret_cast<void*>(fileReader.GetHandle()), offset, length);
```
Jinming-Hu: I'll fix this in a separate PR.

`fileReader.GetHandle()` will return `void*` directly, so we don't need `if defined` here.

Jinming-Hu (approved):

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
ahsonkhan: I was having trouble making this work within the lambda, it was easier for me to just fix it in this PR. Already done, please review :)

ahsonkhan:
@Jinming-Hu can you help me figure out where the missing windows.h header errors in CI that are coming from in storage? If I just compile core, locally, things build fine.

sdk/storage/azure-storage-common/src/file_io.cpp
```diff
@@ -22,9 +26,11 @@ namespace Azure { namespace Storage { namespace _internal {
 #if defined(AZ_PLATFORM_WINDOWS)
   FileReader::FileReader(const std::string& filename)
   {
+    HANDLE fileHandle = NULL;
```
Jinming-Hu, line 29: I'm not sure if this is the right way to initialize handle

```suggestion
    HANDLE fileHandle = INVALID_HANDLE_VALUE;
```

or just

```suggestion
    HANDLE fileHandle;
```

sdk/storage/azure-storage-common/test/test_base.hpp
```diff
@@ -16,6 +16,12 @@
 #include <azure/storage/common/storage_common.hpp>
 #include <gtest/gtest.h>
 
+#include <azure/core/platform.hpp>
+
+#if defined(AZ_PLATFORM_WINDOWS)
+#include <windows.h>
+#endif
+
```
Jinming-Hu, line 24: Move to after Line 5

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
```diff
@@ -3,6 +3,12 @@
 
 #include "azure/storage/files/shares/share_file_client.hpp"
 
+#include <azure/core/platform.hpp>
+
+#if defined(AZ_PLATFORM_WINDOWS)
+#include <windows.h>
+#endif
```
Jinming-Hu, line 10: Why do we need `windows.h` in this file?

sdk/storage/azure-storage-common/test/test_base.hpp
Jinming-Hu: also need
```
#define WIN32_LEAN_AND_MEAN
#endif
#if !defined(NOMINMAX)
#define NOMINMAX
#endif
```

Jinming-Hu:
> @Jinming-Hu can you help me figure out where the missing windows.h header errors in CI that are coming from in storage? If I just compile core, locally, things build fine.
> 
> I don't know which file needs the header added explicitly.

`HINTERNET` in `win_http_transport.hpp`

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -803,7 +809,7 @@ int64_t CurlSession::OnRead(uint8_t* buffer, int64_t count, Context const& conte
 
   auto totalRead = int64_t();
   auto readRequestLength = this->m_isChunkedResponseType
-      ? std::min(this->m_chunkSize - this->m_sessionTotalRead, count)
+      ? (std::min)(this->m_chunkSize - this->m_sessionTotalRead, count)
```
Jinming-Hu, line 812: If we define `NOMINMAX` everytime before including `windows.h`. We don't need this extra `()`.

sdk/storage/azure-storage-common/src/file_io.cpp
```diff
@@ -12,6 +12,10 @@
 #include <unistd.h>
 #endif
 
+#if defined(AZ_PLATFORM_WINDOWS)
```
Jinming-Hu, line 15: #if !defined(WIN32_LEAN_AND_MEAN)
#define WIN32_LEAN_AND_MEAN
#endif
#if !defined(NOMINMAX)
#define NOMINMAX
#endif

sdk/core/azure-core/src/http/curl/curl.cpp
ahsonkhan: Yep, understood. I am thinking we should add the extra `()` across all uses, anyway, for good measure.

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/test/ut/macro_guard.cpp

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
```diff
@@ -3,6 +3,18 @@
...
+#if defined(AZ_PLATFORM_WINDOWS)
+#if !defined(WIN32_LEAN_AND_MEAN)
+#define WIN32_LEAN_AND_MEAN
+#endif
+#if !defined(NOMINMAX)
+#define NOMINMAX
+#endif
+#include <windows.h>
+#endif
+
```
Jinming-Hu, line 17:
```suggestion

```

sdk/core/azure-core/src/http/curl/curl.cpp
ahsonkhan: I agree. However, would that catch all of them? It doesn't cover the usage in the *.cpp files.

Jinming-Hu: It's just personal preference. I don't like the macros, so I always try my best to avoid including them.

ahsonkhan: In the case of this curl.cpp file, if I remove the `()` it gives me a compiler error for some reason.

sdk/core/azure-core/src/io/body_stream.cpp
```diff
@@ -99,29 +105,30 @@ FileBodyStream::FileBodyStream(const std::string& filename)
...
       throw std::runtime_error("Failed to get size of file. File name: '" + filename + "'");
     }
+    m_filehandle = static_cast<void*>(fileHandle);
     m_randomAccessFileBodyStream = std::make_unique<_internal::RandomAccessFileBodyStream>(
         _internal::RandomAccessFileBodyStream(m_filehandle, 0, fileSize.QuadPart));
   }
   catch (std::exception&)
   {
-    CloseHandle(m_filehandle);
+    CloseHandle(fileHandle);
```
RickWinter, line 131: The handle should be in a holder at the point of initialization.  That way we don't need to setup a try/catch in this method.

RickWinter (changes requested):
Although there is some niceness to do this change we should ensure we don't do it in a way that creates possible bugs.  The FileHandle should be a FileHandleHolder similar to the Hkeyholder.

sdk/core/azure-core/src/io/body_stream.cpp
ahsonkhan: That is unrelated to this PR and its intended change. The issue already exists in master.

I would have filed an issue, but this was discussed previously, and after extensive discussions with Jeff, we decided to avoid the "holder" pattern.
For example: https://github.com/Azure/azure-sdk-for-cpp/pull/1830#discussion_r591891880

RickWinter: Please file an issue, we should fix the code to use the holder pattern to avoid unnessecary handler creation and to ensure release.

RickWinter (approved):

sdk/core/azure-core/src/io/body_stream.cpp
ahsonkhan: https://github.com/Azure/azure-sdk-for-cpp/issues/2199

Issue #1120, opened 2020-12-10
Add UTF-8 and UTF-16 specific tests for hosts to verify correctness of transcoding in the transport layer and add code comments do document the behavior


ahsonkhan:
See https://github.com/Azure/azure-sdk-for-cpp/pull/897#discussion_r517808859

Verify that `WideCharToMultiByte` and `MultiByteToWideChar` are working as expected.

RickWinter:
This would be a data corruption bug if its wrong.  We need to test cases to cover this scenario before GA

PR #2198, opened 2021-05-05
Add transport layer unit tests with non-ascii host strings as input.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/1120

### DIFF
--- a/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
+++ b/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
@@ -35,9 +35,12 @@ inline std::wstring HttpMethodToWideString(HttpMethod method)
 // This assumes the input string is always null-terminated.
 std::wstring StringToWideString(const std::string& str)
 {
+  // Since the strings being converted to wstring can be provided by the end user, and can contain
+  // invalid characters, use the MB_ERR_INVALID_CHARS to validate and fail.
+
   // Passing in -1 so that the function processes the entire input string, including the terminating
   // null character.
-  int sizeNeeded = MultiByteToWideChar(CP_UTF8, 0, str.c_str(), -1, 0, 0);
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -35,9 +35,12 @@ inline std::wstring HttpMethodToWideString(HttpMethod method)
 // This assumes the input string is always null-terminated.
 std::wstring StringToWideString(const std::string& str)
 {
+  // Since the strings being converted to wstring can be provided by the end user, and can contain
+  // invalid characters, use the MB_ERR_INVALID_CHARS to validate and fail.
+
   // Passing in -1 so that the function processes the entire input string, including the terminating
   // null character.
-  int sizeNeeded = MultiByteToWideChar(CP_UTF8, 0, str.c_str(), -1, 0, 0);
+  int sizeNeeded = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.c_str(), -1, 0, 0);
```
ahsonkhan, line 43: Given this isn't really a valid scenario (we don't expect folks to have invalid encoded URL endpoints), and because we were still throwing a `TransportException`, changing from 0 to `MB_ERR_INVALID_CHARS` isn't really a significant bug fix, but rather a nice to have.

**Previously, we were throwing a TransportException, a bit later after sending a request to the bad URL, with the message:**
`"Error while getting a connection handle. Error Code: 12005."`

**Now, we fail earlier, and throw a TransportException right when transcoding, with the message:**
`"Unable to get the required transcoded size for the input string. Error Code: 1113."`

RickWinter (approved):

vhvb1989 (approved):

PR #2201, opened 2021-05-05
Remove exposing windows.h header from our public headers within storage.

ahsonkhan:
Follow up to https://github.com/Azure/azure-sdk-for-cpp/pull/2188, fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2182

### DIFF
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/file_io.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/file_io.hpp
@@ -5,16 +5,6 @@
 
 #include <azure/core/platform.hpp>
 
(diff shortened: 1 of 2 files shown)


antkmsft (approved):

ahsonkhan:
@Jinming-Hu I am waiting for your review before merging this. Should be fairly quick :)

Jinming-Hu (changes requested):
This is a .cpp file, why do we need to remove `windows.h` from .cpp file?

ahsonkhan:
> This is a .cpp file, why do we need to remove `windows.h` from .cpp file?

We don't, fixed the PR to update the header file instead. Good catch.

Jinming-Hu (approved):

PR #2210, opened 2021-05-06
Rename IsBufferedDownload to ShouldBufferResponse to avoid using a storage specific verb in general purpose Request object.

ahsonkhan:
Follow-up to https://github.com/Azure/azure-sdk-for-cpp/pull/2196#discussion_r627003594

"Download" is just one particular use case of buffering a raw response rather than streaming it, and it's specific to storage. The API name within core should be agnostic of how it just happens to be used today.

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Breaking Changes
 
 - Removed `Context::GetApplicationContext()` in favor of a new static data member `Context::ApplicationContext`.
-- Renamed `Request::IsDownloadViaStream()` to `IsBufferedDownload()`.
+- Renamed `Request::IsDownloadViaStream()` to `ShouldBufferResponse()`.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


ahsonkhan:
Here's the issue for adding the complementary method on RawResponse: https://github.com/Azure/azure-sdk-for-cpp/issues/2211

ahsonkhan:
@JeffreyRichter, @RickWinter, @vhvb1989 - waiting for approval on this.

ahsonkhan:
> This was not planned for the milestone but added after another change.

It's from https://github.com/Azure/azure-sdk-for-cpp/pull/2196 and https://github.com/Azure/azure-sdk-for-cpp/issues/2044, which are part of this milestone.

`Request::IsBufferedDownload` gives no indication that this bool only affects where and how the RawResponse is returned back to the user. This boolean is modifying the state of the returned response, whether the response should be buffered in a vector OR returned as a stream. It essentially tells whoever is processing the request to do something, hence the use of `Should`. `IsX` indicates a particular state of the object. The request isn't a buffered download.

RickWinter (approved):
Review with Victor before committing.

ahsonkhan:
Got a thumbs up from Victor, offline. Thanks.

JeffreyRichter (approved):

PR #2239, opened 2021-05-08
Correct options assignment for TokenCredentialOptions

antkmsft:
Currently, `TokenCredentialOptions` is empty, so assignment of `ClientOptions` part works just as good as if `static_cast<TokenCredentialOptions&>` was used, but the correct way is to use `TokenCredentialOptions&` that was the intent when the code was written.
Thanks to @ahsonkhan for finding this in the PR a long time ago, and I never go to correct this.


antkmsft:
@vhvb1989, currently, there is no way to test for it, because TokenCredentialOptions do not have any extra members, so technically the code without the fix so far works exactly as the code with it.

Issue #2252, opened 2021-05-11
BlobServiceClientTest.UndeleteBlobContainer live test failing because "400 The value of deleted-container-name header must match the destination container name."


ahsonkhan:
See https://github.com/Azure/azure-sdk-for-cpp/pull/2246

Linux
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=887515&view=logs&j=b28a1a35-334b-5681-3876-e5562b5e8062&t=8093ec4b-94b5-5b1c-251c-634eb19c00d0&l=408
```text
2021-05-11T00:00:16.9351667Z 294: [ RUN      ] BlobServiceClientTest.UndeleteBlobContainer
2021-05-11T00:00:16.9693168Z 294: unknown file: Failure
2021-05-11T00:00:16.9695340Z 294: C++ exception with description "400 The value of deleted-container-name header must match the destination container name.
2021-05-11T00:00:16.9696803Z 294: The value of deleted-container-name header must match the destination container name.
2021-05-11T00:00:16.9698383Z 294: RequestId:78f0dd6f-201e-0069-6cf8-459a55000000
2021-05-11T00:00:16.9699893Z 294: Time:2021-05-11T00:00:16.9702517Z
2021-05-11T00:00:16.9701374Z 294: Request ID: 78f0dd6f-201e-0069-6cf8-459a55000000" thrown in the test body.
2021-05-11T00:00:16.9702588Z 294: [  FAILED  ] BlobServiceClientTest.UndeleteBlobContainer (34 ms)
```

Mac
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=887515&view=logs&j=267db408-59fc-59d9-cfd0-fd1515cbe21c&t=da0f4332-ab51-5c8d-61ac-094f82163104&l=408
```text
2021-05-11T00:01:28.8280850Z 294: [ RUN      ] BlobServiceClientTest.UndeleteBlobContainer
2021-05-11T00:01:29.2030290Z 294: unknown file: Failure
2021-05-11T00:01:29.2033480Z 294: C++ exception with description "400 The value of deleted-container-name header must match the destination container name.
2021-05-11T00:01:29.2036660Z 294: The value of deleted-container-name header must match the destination container name.
2021-05-11T00:01:29.2038280Z 294: RequestId:805e6c25-901e-001d-26f8-45eb6c000000
[...]

PR #2263, opened 2021-05-13
`BlobServiceClient::UndeleteBlobContainer` doesn't support restoring a deleted container under a different name anymore.

Jinming-Hu:
fixed https://github.com/Azure/azure-sdk-for-cpp/issues/2252

Hi @JeffreyRichter , storage service team decided to take back the feature of restoring a deleted container under a different name. This introduces a breaking change in client side.

### DIFF
--- a/sdk/storage/azure-storage-blobs/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blobs/CHANGELOG.md
@@ -10,6 +10,7 @@
 - `BlobLeaseClient::Change()` updates internal lease id.
 - Removed `ContentType` from `GetBlockListResult`.
 - Moved `GetPageRangesResult` to detail namespace.
+- `BlobServiceClient::UndeleteBlobContainer` doesn't support restoring a deleted container under a different name anymore.
 
 ## 12.0.0-beta.10 (2021-04-16)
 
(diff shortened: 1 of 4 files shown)


vhvb1989 (approved):

ahsonkhan:
What impact does it have on the client/SDK and to the track2 end customers? What about the previous track1 users?

sdk/storage/azure-storage-blobs/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - `BlobLeaseClient::Change()` updates internal lease id.
 - Removed `ContentType` from `GetBlockListResult`.
 - Moved `GetPageRangesResult` to detail namespace.
+- `BlobServiceClient::UndeleteBlobContainer` doesn't support restoring a deleted container under a different name anymore.
```
ahsonkhan, line 13: This is driven primarily by the service-side behavioral change than the SDK, correct? Might be worth mentioning that.
```suggestion
- `BlobServiceClient::UndeleteBlobContainer` doesn't support restoring a deleted container under a different name because the Azure Blob Service doesn't support it.
```

Jinming-Hu:
> What impact does it have on the client/SDK and to the track2 end customers? What about the previous track1 users?

If Track2 customers were using this feature to restore a deleted container under a different name, it doesn't work anymore.

Track1 CPP sdk doesn't support soft delete or restore container at all. So no impact to them.

PR #2267, opened 2021-05-14
Update upper/lower/other casing in comments and docs

antkmsft:


sdk/storage/README.md
```diff
@@ -50,9 +50,9 @@ After Vcpkg is initialized and bootstrapped, you can install the dependencies:
 vcpkg.exe install libxml2:x64-windows-static curl:x64-windows-static
 ```
 
-#### Unix Platforms
+#### POSIX Platforms
```
Jinming-Hu, line 53: POSIX is not a platform.

Jinming-Hu: Unix platform is just fine.

Jinming-Hu (approved):

sdk/storage/README.md
antkmsft: Hmm, but then Linux is POSIX but not Unix(or UNIX)... POSIX seems the best short way to capture all of what we mean when we say "unix" casually while talking about macOS and Linux, i.e. non-Windows.

Jinming-Hu: In my opinion, Linux is a Unix-like operating system. It's compatible (or compliant) with POSIX.

antkmsft: You can meet "POSIX" platform in some occurrences: https://www.bing.com/search?q=%22POSIX+platform%22

Jinming-Hu: Google "unix platform", About 175,000 results
Google "POSIX platform", About 7,540 results

I think Unix Platform is better than POSIX platform. But it's just doc, not source code, and it won't affect customers because people understand what we're referring to. So whichever is fine by me. I'd like to hear other people's opinion.

antkmsft: I'm the same, if only I knew the 100% answer. And I get what you are saying. I think there are cases when we clearly can use "POSIX" in other places in our code, and it will be better there. Here - maybe not, but maybe it will be more consistent. 

And I totally get that even casually "Unix" is used more often then "POSIX", and everyone knows what that means. There are also language nerds arguing on "Linux" vs "GNU/Linux" - terminology disputes never interested me. But I am making a PR to somehow organize casing, and we probably need to settle on something. Although if you feel strong about this specific instance, I am not against removing it. I think you are right, let's collect more feedback across the team.

But I think it is easier to use "POSIX" everywhere, instead of "Unix" in one place (correctly), and "POSIX" in another (also correctly), but every time we'll have to think and distinguish between them. And I think, "POSIX platform" is about 99% correct and understandable term.

A bit related, I am also not completely sure if "ID" is right - maybe it is "Id", or even "id". But the Azure and MS docs spell them as "ID" in most of the cases. But I also won't be surprised if in 10 years from now writing it as "id" will become more popular just as today we spell "email" and not "E-mail" like in 90s.

vhvb1989 (approved):

ahsonkhan:
Part of https://github.com/Azure/azure-sdk-for-cpp/issues/2260
@antkmsft link to/mention the epic issue that the docs PRs are fixing so they are easy to find later from that issue.

Issue #2226, opened 2021-05-07
Consolidate use of size_t vs int32_t in public surface area


ahsonkhan:
Are our guidelines correct that we are allowed to use `size_t` in public surface area or should we always use `int32_t`?https://azure.github.io/azure-sdk/cpp_implementation.html#integer-sizes

We use `size_t` in some places (for example Hash), but most other places use `int32_t` (for example BodyStream).

https://github.com/Azure/azure-sdk-for-cpp/blob/1a71c46cbd5fd7f54365ab8cd74e812b6f63a391/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp#L61
https://github.com/Azure/azure-sdk-for-cpp/blob/1a71c46cbd5fd7f54365ab8cd74e812b6f63a391/sdk/core/azure-core/inc/azure/core/io/body_stream.hpp#L98-L101

ahsonkhan:
The guideline also recommends use of `int64_t` everywhere else, and not `int32_t`.

cc @BillyONeal as well since you helped us initially write these.

Jinming-Hu:
C++ Working Group Meeting Notes 6/10

```
Topic: Choice of integer types
Conclusion:
File length are int64_t  (signed)
Array indexes should be size_t or ptrdiff_t (unsigned)
Numbers passed into or returned by client that are transmitted/received over the wire should be signed (int64_t, or if it's completely safe int32_t)
If client library returns some calculated number that is not going through or coming from the wire then the client should use whatever is appropriate
```

ahsonkhan:
Fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/2285

Will open separate issues for the storage FileWriter change and private header/implementation fix-ups.
https://github.com/Azure/azure-sdk-for-cpp/issues/2315
https://github.com/Azure/azure-sdk-for-cpp/issues/2316

PR #2276, opened 2021-05-14
Explore removing use of `int` within header files to align with guidelines.

ahsonkhan:
Part of https://github.com/Azure/azure-sdk-for-cpp/issues/2226

Trying to focus mainly on headers, depending on the conclusion, we'd want to make cpp file changes as well, but that can be done later.

Assuming this is the guideline we want:
https://azure.github.io/azure-sdk/cpp_implementation.html#integer-sizes

image

Some places that we still use `int` is for things like unix file desriptor or where APIs like BIO_read from OpenSSL returns an int. Are those exceptions OK?
https://github.com/Azure/azure-sdk-for-cpp/blob/0b64ab3845d6850bc1666337f21640899e732ef5/sdk/core/azure-core/inc/azure/core/io/body_stream.hpp#L166
https://github.com/Azure/azure-sdk-for-cpp/blob/308e5363adaa00ace2f06ff0f1a31747a0a28062/sdk/core/azure-core/src/base64.cpp#L87

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/uuid.hpp
+++ b/sdk/core/azure-core/inc/azure/core/uuid.hpp
@@ -28,7 +28,7 @@ namespace Azure { namespace Core {
   class Uuid final {
 
   private:
-    static constexpr int UuidSize = 16;
+    static constexpr int32_t UuidSize = 16;
 
     uint8_t m_uuid[UuidSize];
     // The UUID reserved variants.
@@ -86,13 +86,13 @@ namespace Azure { namespace Core {
 #if defined(AZ_PLATFORM_WINDOWS)
       std::random_device rd;
 
-      for (int i = 0; i < UuidSize; i += 4)
+      for (size_t i = 0; i < UuidSize; i += 4)
       {
         const uint32_t x = rd();
         std::memcpy(uuid + i, &x, 4);
       }
 #elif defined(AZ_PLATFORM_POSIX)
-      int ret = RAND_bytes(uuid, UuidSize);
+      size_t ret = RAND_bytes(uuid, UuidSize);
       if (ret <= 0)
       {
         abort();
--- a/sdk/core/azure-core/inc/azure/core/uuid.hpp
+++ b/sdk/core/azure-core/inc/azure/core/uuid.hpp
@@ -92,7 +92,7 @@ namespace Azure { namespace Core {
         std::memcpy(uuid + i, &x, 4);
       }
 #elif defined(AZ_PLATFORM_POSIX)
-      size_t ret = RAND_bytes(uuid, UuidSize);
+      int ret = RAND_bytes(uuid, UuidSize);
       if (ret <= 0)
       {
         abort();
--- a/sdk/core/azure-core/inc/azure/core/uuid.hpp
+++ b/sdk/core/azure-core/inc/azure/core/uuid.hpp
@@ -28,7 +28,7 @@ namespace Azure { namespace Core {
   class Uuid final {
 
   private:
-    static constexpr int32_t UuidSize = 16;
+    static constexpr size_t UuidSize = 16;
 
     uint8_t m_uuid[UuidSize];
     // The UUID reserved variants.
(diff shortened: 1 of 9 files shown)


ahsonkhan:
cc @CaseyCarter and @BillyONeal

sdk/core/azure-core/src/private/package_version.hpp
```diff
@@ -24,13 +26,13 @@ namespace Azure { namespace Core { namespace _detail {
   class PackageVersion final {
   public:
     /// Major numeric identifier.
-    static constexpr int Major = AZURE_CORE_VERSION_MAJOR;
+    static constexpr int32_t Major = AZURE_CORE_VERSION_MAJOR;
```
ahsonkhan, line 29: Didn't modify storage, but if we do this, we'd want to make the same change there.

Jinming-Hu (approved):
Really don't like this. I believe we should use `int` for those cases we know the value isn't too big. This is suggested by Google C++ Style Guide.

The rule in our C++ guidelines doesn't make any sense to me. But I'll accept it as it's guideline.

ahsonkhan:
> This is suggested by Google C++ Style Guide.

For reference, I believe this is what @Jinming-Hu is referring to:
https://google.github.io/styleguide/cppguide.html#Integer_Types

image

CaseyCarter:
"Don't use `int` under any circumstances" is a silly guideline that I would ignore. Presumably this guideline was originally formulated as a prohibition against using `int` to index into or represent the size of memory buffers and containers - which _is_ sensible given that such sizes may be larger than an `int` can represent - and then "simplified" into its current state.

You can't even declare `main` without using `int`.

ahsonkhan:
> You can't even declare `main` without using `int`.

Haha, exactly :)

BillyONeal:
> "Don't use `int` under any circumstances" is a silly guideline that I would ignore. Presumably this guideline was originally formulated as a prohibition against using `int` to index into or represent the size of memory buffers and containers - which _is_ sensible given that such sizes may be larger than an `int` can represent - and then "simplified" into its current state.
> 
> You can't even declare `main` without using `int`.

The intent was that the couple of places it would be necessary would require review and we would put that in the guidelines. In particular, int shouldn't be in any public API surface, for indexing / loops, or in any data structure defined by the SDK, when used as an int. (e.g. POSIX fds are `int` but they aren't used that way by the application).

Notably, anything that might go over the wire needs to be int64_t in case the service(s) are changed to allow that, as happened when Storage adopted Jumbo Blobs. (That requirement came from Jeffrey Richter)

I think the current list of known exceptions is:

* POSIX fds, or other situations where the int is an opaque handle
* Interfaces that are forced to be defined in terms of int, like main
* SFIANE/enable_if as the =0 target in a template

and the list of "absolutely problematic" ones are:

* Any public API surface (should probably be int64_t)
* Memory buffer sizes (should be size_t or ptrdiff_t)
* File sizes (should be int64_t)

> Really don't like this. I believe we should use `int` for those cases we know the value isn't too big. This is suggested by Google C++ Style Guide.
> 
> The rule in our C++ guidelines doesn't make any sense to me. But I'll accept it as it's guideline.

Our experience is that people think the values aren't too big, until the values can be too big, and then it's a nightmare to fix later. For example, this is exactly what happened in cpprestsdk that caused it to fall over with inputs bigger than 2GiB.

> > This is suggested by Google C++ Style Guide.
> 
> For reference, I believe this is what @Jinming-Hu is referring to:
> https://google.github.io/styleguide/cppguide.html#Integer_Types

I disagree with Google's style guide here -- note that their style guide effectively assumes an LP64 platform (Linux). Their justification is "uniformity of declaration," that is, to be consistent with the large base of existing code already present at Google, and I don't believe it's particularly relevant here.

ahsonkhan:
> In particular, int shouldn't be in any public API surface, for indexing / loops, or in any data structure defined by the SDK, when used as an int. (e.g. POSIX fds are `int` but they aren't used that way by the application).

There are a couple of places we need to go fix. I will follow up and file issues.
https://github.com/Azure/azure-sdk-for-cpp/blob/e4a18b5eb3b556fae7bde96c43444f27f2fcbdb8/sdk/storage/azure-storage-files-shares/inc/azure/storage/files/shares/share_options.hpp#L638-L642

antkmsft (approved):

sdk/core/azure-core/inc/azure/core/uuid.hpp
```diff
@@ -28,7 +28,7 @@ namespace Azure { namespace Core {
   class Uuid final {
 
   private:
-    static constexpr int UuidSize = 16;
+    static constexpr int32_t UuidSize = 16;
```
CaseyCarter, line 31: Why not `size_t` here as well? Isn't this used as the size of memory buffers?

vhvb1989 (approved):

PR #2282, opened 2021-05-17
Remove Azure::Core::Diagnostics::Logger::Listener typedef

### LINKED ISSUE #2283: Remove Azure::Core::Diagnostics::Logger::Listener typedef
1. Typedef not that helpful in IntelliSense. `std::function<...>` is a bit better than `Logger::Listener`.
2. Typedef is not really that useful - it is not that you need to constantly need variables of `Logger:Listener` type referencing the type name.
3. Less entities to document an support
4. More API design freedom in terms of being non-breaking - prettier looking overloads for `std::function` with 3rd argument, `Logger` can be introduced as a class, so if we unreserve that name, naming that class can be easier in the future. If it is needed, it is possible to reintroduce the typedef in the future, but won't be possible to pull it out, and given `#`2, I don't think we need it right now.

antkmsft:
Closes #2283 (which is now approved for May).


sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -115,7 +115,6 @@ class DateTime final : public _detail::Clock::time_point {
 
   /**
    * @brief Construct an instance of #Azure::DateTime from base class.
-   *
```
ahsonkhan, line 118: Revert changes to these files. I believe @vhvb1989 added them intentionally to fix doxygen docs.

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
```diff
@@ -38,21 +38,13 @@ namespace Azure { namespace Core { namespace Diagnostics {
       Error = 4,
     };
 
-    /**
-     * @brief A callback function to receive Azure SDK log messages.
-     *
-     * @param level The log message level.
-     * @param message The log message text.
-     */
-    typedef std::function<void(Level level, std::string const& message)> Listener;
```
ahsonkhan, line 47: Any logger readme or docs that need updating?

ahsonkhan (changes requested):
Looks good, but revert the unrelated datetime header change.

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
antkmsft: Nope - but they need to be written :)

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Thanks - no idea where they came from. One would be ok, but all 3... I don't rebember touching them. Maybe was some bad merge.

ahsonkhan (approved):

PR #2285, opened 2021-05-17
Use size_t for buffers representing data in memory rather than int64_t.

ahsonkhan:
Part of https://github.com/Azure/azure-sdk-for-cpp/issues/2226

Mainly modifies BodyStream and derived types.

~Todo: Changelog~


antkmsft (approved):

CaseyCarter (approved):

Jinming-Hu (approved):
I always prefer `std::size_t` over `size_t`. What's the difference between them? Is one considered better than the other?

sdk/core/azure-core/src/io/body_stream.cpp
```diff
@@ -75,10 +75,11 @@ std::vector<uint8_t> BodyStream::ReadToEnd(Context const& context)
   }
 }
 
-int64_t MemoryBodyStream::OnRead(uint8_t* buffer, int64_t count, Context const& context)
+int64_t MemoryBodyStream::OnRead(uint8_t* buffer, size_t count, Context const& context)
```
RickWinter, line 78: On 32-bit systems this puts an upper limit of 4GB on the buffer size and the file size.  That means our API isn't compatible with 32-bit OS using LFS.

ahsonkhan:
> I always prefer `std::size_t` over `size_t`. What's the difference between them? Is one considered better than the other?

I don't know the difference. @CaseyCarter and @BillyONeal can you help provide some context here?

@Jinming-Hu why do you prefer one over the other if the difference isn't clear? :p

Looks like vector returns size_t:
https://en.cppreference.com/w/cpp/container/vector/size

Jinming-Hu:
> Looks like vector returns size_t:
https://en.cppreference.com/w/cpp/container/vector/size

Looks like it's `std::size_t`
https://en.cppreference.com/w/cpp/container/vector
image

ahsonkhan:
It looks like they are the same type, one is a typedef of the other.
image

I am fine with using `std:size_t` consistently.

Jinming-Hu:
> It looks like they are the same type, one is a typedef of the other.
> image
> 
> image
> 
> I am fine with using `std:size_t` consistently.

Do we need a guideline? For consistency?

BillyONeal:
> > I always prefer `std::size_t` over `size_t`. What's the difference between them? Is one considered better than the other?
> 
> I don't know the difference. @CaseyCarter and @BillyONeal can you help provide some context here?

They are defined to be identical. See http://eel.is/c++draft/cstddef.syn#1

I don't think adding the std:: adds value myself. In practice the plain version will be available more often (since someone needs to drag in `<cstddef>` to get the `std::` version).

CaseyCarter:
> I don't think adding the std:: adds value myself. In practice the plain version will be available more often (since someone needs to drag in `<cstddef>` to get the `std::` version).

`::size_t` (inherited from the C standard library) and `std::size_t` (from the C++ standard library) are specified to denote the same type. The correct name to use depends on which header you include to get access to the type. The header `<cstddef>` defines the name `std::size_t` (and may define `::size_t` as well); the header `<stddef.h>` defines the name `::size_t` (and may define `std::size_t` as well).

@BillyONeal and I differ on best practice with headers inherited from C. IIUC Billy generally prefers to include `<meow.h>` instead of `<cmeow>` and use the unqualified name `woof`, whereas I prefer to include `<cmeow>` and use `std::woof`. Billy's style acknowledges the reality that the major implementations put those names in the global namespace regardless of which header you include, and provides consistency of naming for C library things whether they appear in C or C++ source files. My style tries to reserve unqualified names for the local project and always qualifies names defined externally.

ahsonkhan:
@CaseyCarter, @BillyONeal - **Here's the higher order question that will motivate us being consistent within the SDK:**
If we use `std::size_t` in public surface area today, and change it to `size_t` in the future, or vice versa, would that be a source breaking change?

If yes, we need to be consistent across the board, I will just pick one and use it. If all things are generally equal, I prefer size_t because of less typing :)
If no, I think it would be OK for us to use whichever.

BillyONeal:
> @CaseyCarter, @BillyONeal - **Here's the higher order question that will motivate us being consistent within the SDK:**
> If we use `std::size_t` in public surface area today, and change it to `size_t` in the future, or vice versa, would that be a source breaking change?

No, they are identical.\*

(I can come up with pathological ways to break either like:

```
namespace myapp {
    constexpr int size_t = 42; // breaks plain "size_t" with ADL
    namespace std {
        constexpr int size_t = 42; // might break std::size_t depending on 'using namespace's
    }
}
```

that force the standard library to go full `::std::size_t` for things, but I wouldn't suggest that anyone actually care about such cases)

ahsonkhan:
In that case, this isn't something we need to worry all that much about, and no need to add noise to the guidelines (we can reserve the precious space in that doc for more customer impacting discussions). Thanks!

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2226

vhvb1989 (approved):

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
```diff
@@ -939,7 +941,9 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
     auto uploadPageFunc = [&](int64_t offset, int64_t length, int64_t chunkId, int64_t numChunks) {
       (void)chunkId;
       (void)numChunks;
-      Azure::Core::IO::MemoryBodyStream contentStream(buffer + offset, length);
+      // TODO: Investigate changing lambda parameters to be size_t, unless they need to be int64_t
+      // for some reason.
```
Jinming-Hu, line 945: Can you open an issue and assign it to me?

ahsonkhan: Sure.

ahsonkhan: https://github.com/Azure/azure-sdk-for-cpp/issues/2313

PR #2291, opened 2021-05-18
copy constructor for RequestFailedException

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/2281

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -92,5 +92,25 @@ namespace Azure { namespace Core {
     explicit RequestFailedException(
         const std::string& message,
         std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);
+
+    /**
+     * @brief Copy a RequestFailedException to construct a new one.
+     *
+     * @param other The one to be copied.
+     */
+    RequestFailedException(const RequestFailedException& other)
+        : std::runtime_error(other.Message), StatusCode(other.StatusCode),
+          ReasonPhrase(other.ReasonPhrase), ClientRequestId(other.ClientRequestId),
+          RequestId(other.RequestId), ErrorCode(other.ErrorCode), Message(other.Message),
+          RawResponse(std::make_unique<Azure::Core::Http::RawResponse>(*other.RawResponse))
+    {
+    }
+
+    /**
+     * @brief Move a RequestFailedException to construct a new one.
+     *
+     * @param other The one to be moved.
+     */
+    RequestFailedException(RequestFailedException&& other) = default;
   };
 }} // namespace Azure::Core
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -109,7 +109,8 @@ namespace Azure { namespace Core {
 
     /**
      * @brief Constructs a new `RequestFailedException`.
-     * @note Transfers ownership of the #Azure::Core::Http::RawResponse to the new `RequestFailedException`.
+     * @note Transfers ownership of the #Azure::Core::Http::RawResponse to the new
+     * `RequestFailedException`.
      *
      * @param other An rvalue reference for moving the `RequestFailedException`.
      */
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -94,9 +94,10 @@ namespace Azure { namespace Core {
         std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);
 
     /**
-     * @brief Copy a RequestFailedException to construct a new one.
+     * @brief Constructs a new `RequestFailedException` by copy.
+     * @note Copies the #Azure::Core::Http::RawResponse into the new `RequestFailedException`.
      *
-     * @param other The one to be copied.
+     * @param other The `RequestFailedException` to be copied.
      */
     RequestFailedException(const RequestFailedException& other)
         : std::runtime_error(other.Message), StatusCode(other.StatusCode),
@@ -107,9 +108,10 @@ namespace Azure { namespace Core {
     }
 
     /**
-     * @brief Move a RequestFailedException to construct a new one.
+     * @brief Constructs a new `RequestFailedException`.
+     * @note Transfers ownership of the #Azure::Core::Http::RawResponse to the new `RequestFailedException`.
      *
-     * @param other The one to be moved.
+     * @param other An rvalue reference for moving the `RequestFailedException`.
      */
     RequestFailedException(RequestFailedException&& other) = default;
   };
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/inc/azure/core/http/raw_response.hpp
```diff
@@ -76,10 +76,18 @@ namespace Azure { namespace Core { namespace Http {
...
+      AZURE_ASSERT(m_bodyStream == nullptr);
       // Copy body
       m_body = response.GetBody();
     }
 
+    /**
+     * @brief Move a raw response to construct a new one.
+     *
+     * @param response An rvalue reference for moving the raw response.
+     */
```
RickWinter, line 88:
```suggestion
    /**
     * @brief Constructs a new `RawResponse`.
     * @note Transfers ownership of the `RawResponse` to the new `RawResponse`.
     *
     * @param response An rvalue reference for moving the raw response.
     */
```

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -92,5 +92,25 @@ namespace Azure { namespace Core {
...
+          ReasonPhrase(other.ReasonPhrase), ClientRequestId(other.ClientRequestId),
+          RequestId(other.RequestId), ErrorCode(other.ErrorCode), Message(other.Message),
+          RawResponse(std::make_unique<Azure::Core::Http::RawResponse>(*other.RawResponse))
+    {
+    }
+
+    /**
+     * @brief Move a RequestFailedException to construct a new one.
+     *
+     * @param other The one to be moved.
```
RickWinter, line 112:
```suggestion
     * @brief Constructs a new `RequestFailedException`.
     * @note Transfers ownership of the #Azure::Core::Http::RawResponse to the new `RequestFailedException`.
     *
     * @param other An rvalue reference for moving the `RequestFailedException`.
```

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -92,5 +92,25 @@ namespace Azure { namespace Core {
     explicit RequestFailedException(
         const std::string& message,
         std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);
+
+    /**
+     * @brief Copy a RequestFailedException to construct a new one.
+     *
+     * @param other The one to be copied.
```
RickWinter, line 99:
```suggestion
    /**
     * @brief Constructs a new `RequestFailedException` by copy.
     * @note Copies the #Azure::Core::Http::RawResponse into the new `RequestFailedException`.
     *
     * @param other The `RequestFailedException` to be copied.
```

vhvb1989 (approved):

sdk/core/azure-core/inc/azure/core/http/raw_response.hpp
```diff
@@ -76,10 +76,19 @@ namespace Azure { namespace Core { namespace Http {
...
       m_body = response.GetBody();
     }
 
+    /**
+     * @brief Constructs a new `RawResponse`.
+     * @note Transfers ownership of the `RawResponse` to the new `RawResponse`.
+     *
+     * @param response An rvalue reference for moving the raw response.
+     */
+    RawResponse(RawResponse&& response) = default;
```
ahsonkhan, line 90: @CaseyCarter does the RawResponse type follow the 0-3-5 rule?

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -92,5 +92,28 @@ namespace Azure { namespace Core {
     explicit RequestFailedException(
         const std::string& message,
         std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);
+
+    /**
+     * @brief Constructs a new `RequestFailedException` by copy.
```
ahsonkhan, line 97:
```suggestion
     * @brief Constructs a new `RequestFailedException` by copying from an existing one.
```

sdk/core/azure-core/inc/azure/core/http/raw_response.hpp
```diff
@@ -76,10 +76,19 @@ namespace Azure { namespace Core { namespace Http {
...
       // Copy body
       m_body = response.GetBody();
     }
 
+    /**
+     * @brief Constructs a new `RawResponse`.
+     * @note Transfers ownership of the `RawResponse` to the new `RawResponse`.
+     *
+     * @param response An rvalue reference for moving the raw response.
+     */
```
ahsonkhan, line 89: This is a move ctor, so the comment needs to be updated to reflect that.

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -92,5 +92,28 @@ namespace Azure { namespace Core {
...
+    }
+
+    /**
+     * @brief Constructs a new `RequestFailedException`.
+     * @note Transfers ownership of the #Azure::Core::Http::RawResponse to the new
+     * `RequestFailedException`.
+     *
+     * @param other An rvalue reference for moving the `RequestFailedException`.
+     */
+    RequestFailedException(RequestFailedException&& other) = default;
```
ahsonkhan, line 117: This is a move ctor. Should the docs reflect that?

@CaseyCarter do we need to explicitly create this move ctor, or would the compiler create it for us?

sdk/core/azure-core/inc/azure/core/http/raw_response.hpp
```diff
@@ -76,10 +76,19 @@ namespace Azure { namespace Core { namespace Http {
             response.m_statusCode,
             response.m_reasonPhrase)
     {
+      AZURE_ASSERT(m_bodyStream == nullptr);
```
CaseyCarter, line 79: "This type is copyable except that sometimes copying them explodes and kills the program" doesn't feel like a great design to me. Could we use the type system to make raw responses that can be copied distinct from those that can't?

sdk/core/azure-core/inc/azure/core/http/raw_response.hpp
CaseyCarter: No, it doesn't. The class is correct (well, as much as something that's copyable but can't always be copied can be) but seeing that requires inspection and knowledge of the complicated rules for when and how special member functions are defined. I'd add:
```c++
~RawResponse() = default;
RawResponse& operator=(RawResponse const&) = delete;
RawResponse& operator=(RawResponse&&) = delete;
```
here somewhere to make it explicit that we've considered all five special member functions.

sdk/core/azure-core/inc/azure/core/exception.hpp
CaseyCarter: The compiler won't implicitly declare the move constructor since there's (now) an explicitly-declared copy constructor. If we want one, we need to explicitly default it like this.

sdk/core/azure-core/inc/azure/core/http/raw_response.hpp
CaseyCarter: A comment that this peculiar design is necessary to workaround #2281 would be a good idea. I hadn't noticed the issue before performing my review, I suspect other underinformed folks in the future will also wonder what's going on here.

Issue #2299, opened 2021-05-18
Docs: review and update Http (without policies) and IO namespaces


antkmsft:
Extracted from #2260:

Guidelines:
"Drop Represents, Defines and other "intro words" in the doc comment, and just state what it is."
"Use words, by adding (s) like "Gets", "Returns", "Destroys", etc. and not "Get", "Return", "Destroy""
"Add pre tags for preconditions"
"Add 2-5 liner code snippets for the most important or commonly used APIs, such as client ctors, that folks can use for getting started, maybe using \example"
"Avoid self-referential links on the same page"
"Add docs to Tparam (for example Operation)"
"Use @note to highlight important things to call out rather than @remark"

antkmsft:
Relevant catches:
operator #bool() and #GetValue() are not getting linked/referenced correctly. Make sure API name is correct.
Add documentation for HttpMethod extensible enum
HttpPolicy doc, add @details - WIP - Victor

antkmsft:
See also recent beta9 docs: https://azuresdkdocs.blob.core.windows.net/$web/cpp/azure-core/1.0.0-beta.9/annotated.html

PR #2332, opened 2021-05-20
Doc changes for HTTP and removing the "More" link

vhvb1989:


cmake-modules/AzureDoxygen.cmake
```diff
@@ -38,6 +38,10 @@ function(generate_documentation PROJECT_NAME PROJECT_VERSION)
         # Use MathJax instead of latex to render formulas
         set(DOXYGEN_USE_MATHJAX YES)
 
+        set(DOXYGEN_REPEAT_BRIEF NO)
+        #set(DOXYGEN_ALWAYS_DETAILED_SEC NO)
+        #set(DOXYGEN_BRIEF_MEMBER_DESC YES)
```
RickWinter, line 43: Add comments into this file for why we are setting certain values.  

Is there a reason to keep the two options that are commented out?

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+   * @brief The method to be performed on the resource identified by the Request.
    */
   class HttpMethod final {
   public:
     HttpMethod() = delete;
+
+    /**
+     * @brief Construct a new HTTP Method from a given string.
+     *
+     * @details The HttpMethod won't check if the given string is a known method defined as per any
```
RickWinter, line 103: @note
```suggestion
     * @note The HttpMethod won't check if the given string is a known method defined as per any
```

RickWinter (approved):

cmake-modules/AzureDoxygen.cmake
vhvb1989, line 43:
```suggestion
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -51,8 +51,8 @@ namespace Azure { namespace Core { namespace Http {
 
   /*********************  Exceptions  **********************/
   /**
-   * @brief HTTP transport layer error.
-   *
+   * @brief Provides consistent interface to handle errors occurred while performing an HTTP
```
antkmsft, line 54: Please rephrase to avoid "Provides..." in the class description.
(see #2299)

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -71,9 +71,9 @@ namespace Azure { namespace Core { namespace Http {
   };
 
   /**
-   * @brief Defines a range of bytes within an HTTP resource, starting at an `Offset` and ending at
-   * `Offset + Length - 1` inclusively.
+   * @brief Defines a range of bytes within an HTTP resource.
```
antkmsft, line 74: Rephrase to say what the struct is, not what it does. I.e. avoid "Defines"

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
   };
 
   /**
-   * HTTP request method.
-   *
+   * @brief The method to be performed on the resource identified by the Request.
```
antkmsft, line 94: I like the description here: no "defines" or "provides", we just go straight and say what the class **is**.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
    */
   class HttpMethod final {
   public:
     HttpMethod() = delete;
+
+    /**
+     * @brief Construct a new HTTP Method from a given string.
```
antkmsft, line 101: ...Constructs (see #2299)
```suggestion
     * @brief Constructs `HttpMethod` from string.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     *
+     * @note The HttpMethod won't check if the given string is a known method defined as per any
+     * RFC.
```
antkmsft, line 104:
```suggestion
     * @note Won't check if \p value is a known HTTP method defined as per any RFC.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     *
+     * @param value A given string to represent the HTTP Method.
+     */
     explicit HttpMethod(std::string value) : m_value(std::move(value)) {}
+
+    /**
+     * @brief Equal operator enables comparing two HttpMethod.
```
antkmsft, line 111:
```suggestion
     * @brief Compares two instances of `HttpMethod` for equality.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     *
+     * @param other Some HttpMethod instance to be compared.
```
antkmsft, line 113:
```suggestion
     * @param other Some `HttpMethod` instance to compare with.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     * @return `true` if instances are equal; otherwise, `false`.
+     */
     bool operator==(const HttpMethod& other) const { return m_value == other.m_value; }
+
+    /**
+     * @brief The non-equals operator allows comparing two HttpMethod.
```
antkmsft, line 119:
```suggestion
     * @brief Compares two instances of `HttpMethod` for equality.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     *
+     * @param other Some HttpMethod instance to be compared.
```
antkmsft, line 121:
```suggestion
     * @param other Some `HttpMethod` instance to compare with.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
     const std::string& ToString() const { return m_value; }
 
+    /**
+     * @brief The representation of a Get HttpMethod based on [RFC 2616]
+     * (https://datatracker.ietf.org/doc/html/rfc2616#section-9.3).
+     */
     AZ_CORE_DLLEXPORT const static HttpMethod Get;
+
+    /**
+     * @brief The representation of a Head HttpMethod based on [RFC 2616]
```
antkmsft, line 138:
```suggestion
     * @brief The representation of a `HEAD` HTTP method based on [RFC 2616]
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     * (https://datatracker.ietf.org/doc/html/rfc2616#section-9.4).
+     */
     AZ_CORE_DLLEXPORT const static HttpMethod Head;
+
+    /**
+     * @brief The representation of a Post HttpMethod based on [RFC 2616]
```
antkmsft, line 144:
```suggestion
     * @brief The representation of a `POST` HTTP method based on [RFC 2616]
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     * (https://datatracker.ietf.org/doc/html/rfc2616#section-9.5).
+     */
     AZ_CORE_DLLEXPORT const static HttpMethod Post;
+
+    /**
+     * @brief The representation of a Put HttpMethod based on [RFC 2616]
```
antkmsft, line 150:
```suggestion
     * @brief The representation of a `PUT` HTTP method based on [RFC 2616]
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     * (https://datatracker.ietf.org/doc/html/rfc2616#section-9.6).
+     */
     AZ_CORE_DLLEXPORT const static HttpMethod Put;
+
+    /**
+     * @brief The representation of a `Delete` HttpMethod based on [RFC 2616]
```
antkmsft, line 156:
```suggestion
     * @brief The representation of a `DELETE` HTTP method based on [RFC 2616]
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,77 @@ namespace Azure { namespace Core { namespace Http {
...
+     * (https://datatracker.ietf.org/doc/html/rfc2616#section-9.7).
+     */
     AZ_CORE_DLLEXPORT const static HttpMethod Delete;
+
+    /**
+     * @brief The representation of a `Patch` HttpMethod based on [RFC 5789]
```
antkmsft, line 162:
```suggestion
     * @brief The representation of a `PATCH` HTTP method based on [RFC 5789]
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -118,8 +173,10 @@ namespace Azure { namespace Core { namespace Http {
   }} // namespace Policies::_internal
 
   /**
-   * @brief HTTP request.
+   * @brief A request message from a client to a server.
    *
+   * @details Includes, within the first line of the message, the HttpMethod to be applied to the
```
antkmsft, line 178:
```suggestion
   * @details Includes, within the first line of the message, the HTTP method to be applied to the
```

antkmsft: TBH, I don't completely understand this being applied to a class description "Includes, within the first line of the message"?

antkmsft (approved):
Please follow up on my feedback before completing the PR.

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: "...identified by HTTP request"?

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: `HttpMethod` is resolved by doxygen as a type and adds a link to the type.

cmake-modules/AzureDoxygen.cmake
```diff
@@ -38,6 +38,8 @@ function(generate_documentation PROJECT_NAME PROJECT_VERSION)
         # Use MathJax instead of latex to render formulas
         set(DOXYGEN_USE_MATHJAX YES)
 
+        set(DOXYGEN_REPEAT_BRIEF NO)
```
ahsonkhan, line 41: I couldn't easily find what `DOXYGEN_REPEAT_BRIEF` does, I imagine you already know this Victor, can you share, and maybe add it as a comment?

ahsonkhan: Oh... is this that allows us to not have the extra empty new line after `brief`? Interesting.

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -135,7 +134,7 @@ class ETag final {
    */
   bool HasValue() const { return m_value.HasValue(); }
 
-  /*
+  /**
```
ahsonkhan, line 137: Nice! How did you find this issue?

sdk/core/azure-core/inc/azure/core/http/http.hpp
ahsonkhan:
```suggestion
   * @brief The method to be performed on the resource identified by the HTTP request.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,76 @@ namespace Azure { namespace Core { namespace Http {
...
-   * HTTP request method.
-   *
+   * @brief The method to be performed on the resource identified by the Request.
    */
   class HttpMethod final {
   public:
     HttpMethod() = delete;
+
+    /**
+     * @brief Constructs a new HttpMethod from a given string.
```
ahsonkhan, line 101: Would using `HttpMethod` in the description result in a circular link to the same page on the generated docs? I remember we wanted to avoid that.

So, it might be better to split it and say `HTTP method`.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,76 @@ namespace Azure { namespace Core { namespace Http {
...
+     *
+     * @note Won't check if \p value is a known HttpMethod defined as per any RFC.
```
ahsonkhan, line 103: nit:
```suggestion
     * @note This won't validate if \p value is a known HTTP method defined as per any RFC.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,76 @@ namespace Azure { namespace Core { namespace Http {
...
+     * @brief Compares two instances of `HttpMethod` for equality.
+     *
+     * @param other Some `HttpMethod` instance to compare with.
+     * @return `false` if instances are equal; otherwise, `true`.
+     */
     bool operator!=(const HttpMethod& other) const { return !(*this == other); }
+
+    /**
+     * @brief Returns the HttpMethod represented as a string.
+     */
```
ahsonkhan, line 127: Does this create a new string or a copy, every time? Should the caller store it in a local, or is it OK to keep calling `ToString()` in a loop?

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -91,22 +91,76 @@ namespace Azure { namespace Core { namespace Http {
...
     const std::string& ToString() const { return m_value; }
 
+    /**
+     * @brief The representation of a `GET` HttpMethod based on [RFC 2616]
+     * (https://datatracker.ietf.org/doc/html/rfc2616#section-9.3).
```
ahsonkhan, line 132: Good stuff, but this RFC has been obsoleted by others, so we may want to refer to those instead:
image

So, try this?
https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1

Also, will the link render as expected in doxygen?

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -160,7 +216,7 @@ namespace Azure { namespace Core { namespace Http {
     /**
      * @brief Construct an #Azure::Core::Http::Request.
      *
-     * @param httpMethod HTTP method.
+     * @param httpMethod HttpMethod.
```
ahsonkhan, line 219: Since you are fixing this already, let's do something a bit more meaningful, rather than being redundant.
We can copy/paste it across all such param docs.

How about this:

```suggestion
     * @param httpMethod The HTTP verb indicating the action this request is intended to take..
```

sdk/core/azure-core/inc/azure/core/url.hpp
```diff
@@ -40,8 +40,9 @@ namespace Azure { namespace Core {
 
   /**
    * @brief Represents the location where a request will be performed.
-   * It can be parsed and initialized from a string that contains all URL components (scheme, host,
-   * path, etc.). Authority is not currently supported.
+   *
+   * @details It can be parsed and initialized from a string that contains all URL components
+   * (scheme, host, path, etc.). Authority is not currently supported.
```
ahsonkhan, line 45: > Authority is not currently supported.

This is unrelated to the PR, but really? How come? What happens if someone creates a url (or provides a string) with authority? Do we have a test for it?

ahsonkhan (approved):
Some feedback.

Can you help me understand where/how this PR is addressing the "More" link issue. I don't see what we are changing that fixes that.

cmake-modules/AzureDoxygen.cmake
vhvb1989: > Oh... is this that allows us to not have the extra empty new line after brief? Interesting.

Not really. It is hard to explain by writing here. I will try, but, if it is not clear, we can have a call.

Whenever DOXYGEN_REPEAT_BRIEF is YES (default value), For a class like

```
/***
* @brief something
*/
class Something { ..  };
```

The generated documentation in the HTML page will be like

```
class Something  (Name of the class)

something brief (brief from docs)

...
...
...

Detail
something brief (brief from docs is mentioned here again)

The docs after the brief.
```

So, when turning DOXYGEN_REPEAT_BRIEF to NO, we won't see a detail section that repeats the brief when we only have a brief in the docs.  But in the case we have more docs next to the brief like:

```
/**
* @brief I am a brief
*  I am just the next thing after the brief. You can call me details.
* /
```

In this case, the generated docs are:

 ```
class Something  (Name of the class)

I am a brief

...
...
...

Detail
 I am just the next thing after the brief. You can call me details.

```

The `More` link depends directly on this.  
 -  If the docs contains only a brief, there will not be a More link because there will not be a details section.
 -  If the docs contains a brief and also more things after it (or even just an empty line after the brief), then there will be a `more` link that takes you to the details section to see the info after the brief (the brief is not repeated here because of DOXYGEN_REPEAT_BRIEF)

File: sdk/core/azure-core/inc/azure/core/etag.hpp (hunk 19 above)
Comment (vhvb1989) [reply sdk/core/azure-core/inc/azure/core/etag.hpp:137]: By luck while trying to see how we were documenting the `ToString` methods.  I was expecting to find this in the eTag docs and I didn't see it.

File: sdk/core/azure-core/inc/azure/core/http/http.hpp [hunk 26]
```diff
@@ -51,8 +51,8 @@ namespace Azure { namespace Core { namespace Http {
 
   /*********************  Exceptions  **********************/
   /**
-   * @brief HTTP transport layer error.
-   *
+   * @brief Consistent interface to handle errors occurred while performing an HTTP
+   * Request.
```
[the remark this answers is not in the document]
vhvb1989: Request matches the Request class and adds the link to it

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: Request is linked to the class:
image

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: Yes, but I think is fine, the link takes you the page with the info about an HTTPMethod

image

sdk/core/azure-core/inc/azure/core/etag.hpp
ahsonkhan: Maybe there is some regex we can run to detect this issue, if it is happening in more places.

sdk/core/azure-core/inc/azure/core/http/http.hpp
ahsonkhan: image

Where does clicking on `HttpMethod` in that public member function section go to? To the same page?

antkmsft: I think, in this case (and this is the only place I'm suggesting it, I believe), that we should say "Constructs `HttpMethod` from string", meaning, not the "HTTP method" as a broader concept, but specifically HttpMethod class instance. And we backtick it in doxygen documentation.
Plus,
* "a new" is not needed (`operator new` does the allocation, constructor does initialize. Factory method does both, but in this case it is a constructor and not a factory method). And, "a new", in the sense it was meant (i.e. casual and not C++-specific), is redundant, because there is not much other options.
* "from a given", sounds redundant. "Constructs X from Y" is a cleaner declarative-style description. And, "a given" is also redundant, because there is no much other options.

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: This is defined by the method signature and not by docs:

const std::string& ToString() ->  Returns a const ref to some state (not creating a new string). Caller can read the state but not mutate it

std::string& ToString() -> Caller can access the state and mutate it.

std::string ToString() -> Returns a new string on every call.


If the caller want to make a copy of the value read from the instance's state, they would do `auto copyName = instance.ToString();`  After the const ref is returned, the compiler makes a copy into `copyName`.  Otherwise, the caller would do `auto& readOnly = instance.ToString()` to access the memory and value without making copies

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: image

vhvb1989: Will update to 7231.  Thank you

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: I am just creating the link to the HttpMethod class:

image

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: I would not document it, and it is better for the customer code, if the performance here is important, to cache it once before the loop without making assumption how effective the SDK implementation is. Today we may return const reference, tomorrow we restructure internals and start returning a string by value. But if we made promises in documentation, we won't be able to do that.

sdk/core/azure-core/inc/azure/core/http/http.hpp
ahsonkhan: > and we backtick it in doxygen documentation.

Backtick to remove the link but keep it formatted as a "type", differentiated from the rest of the text, like `HttpMethod`?

I like that idea. It follows the spirit of the docs issue I shared above, to avoid unnecessary clickable links that aren't helpful.

sdk/core/azure-core/inc/azure/core/url.hpp
vhvb1989: The URL is currently only for Azure Services.  
I don't expect people using it to parse any URL in the world and to use it independently from the Azure SDK.

I would expect that if any Azure Service uses/requires URLs with the authority in it, then Azure Core will be updated to support it. Right now, it's not required.

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: Exactly - we've done it in other places already, Rick did it in his docs PR. So, when you need to refer to your own class name within the class documentation, we should use it to avoid self-referencing.

sdk/core/azure-core/inc/azure/core/http/http.hpp
ahsonkhan: That's fine, we can do a separate PR/fix for  the wording :)

sdk/core/azure-core/inc/azure/core/http/http.hpp
vhvb1989: backtick it in doxygen documentation does not remove the link

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
antkmsft: Then if we want to refer to class name, we should fully qualify it with namespace, I think we did it in other places... But in this specific case, I think the doc is talking more about "HTTP request" as a concept and not as a "Request class".

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: It is going to be self-reference which we decided to avoid

sdk/core/azure-core/inc/azure/core/http/http.hpp
ahsonkhan: Should we add some template remarks to all such const ref returning methods (or maybe specifically the string returning ones), to let novice users know that `auto copyName = instance.ToString();` will create a hidden copy, and you should generally start off as: `auto& readOnly = instance.ToString()`, unless there is a reason to create a copy?

ahsonkhan: > I would not document it, and I'd say, if the performance here is important to the customer, cache the result of the function once before the loop without making assumption how effective the SDK implementation is. Today we may return const reference, tomorrow we restructure internals and start returning a string by value. But if we made promises in documentation, we won't be able to do that.

Wouldn't that require a breaking change/change of signature, which we wouldn't do anyway?

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: Then, we should use `%`: https://www.doxygen.nl/manual/autolink.html#linkclass

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: It is fine, as user reads the docs, there will be plenty of references to the HttpMethod class in the function signatures, no need to turn everything into a hyperlink.

antkmsft: And I still don't know what the comment is trying to say, be it with the link to the class or not :)

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: I would not document the language itself, and I don't think that suggesting to use `auto&` is a good advice, unless it is in a perf bottleneck. Reference to something you don't control the lifetime has greater potential for bugs. I'd say, if you are a novice user, do a copy, it works better most of the time. Compiler will do good optimizations. If you are an expert, also use copy, unless you know you are in perf bottleneck. I suspect that `auto` type works this way in C++ for exactly this reason: it was added to the language later, in C++11, and the goal probably was that `string x = f()` and `auto x = f()` has the same meaning and type, because it is common pattern with less problems and assumptions. If an exact type is needed, other option is to use `decltype(f())` in place of auto, then your code is guaranteed to use reference when function returns reference, and may create a copy when it returns value. But I would leave that to compiler to do optimization in most cases.

PR #2362, opened 2021-05-27
Fix empty rethrows, update some catches to catch all

antkmsft:
`throw;` is for rethrowing an exception. It should be used within the `catch` blocks.

If used outside, the program does not know what to do - there is not an active exception to rethrow, so it calls `terminate()`. Which is not like throwing exception at all - it does not give user any chance to respond or finalize any stuff, it just most frequently core dumps.

https://en.cppreference.com/w/cpp/language/throw

Yes, I throw `std::exception()` - it is better than `throw;`. If you are thinking of something else specific to throw there, please suggest in the PR. [UPDATE: replaced with AZURE_UNREACHABLE_CODE() per PR suggestions]

Another thing in this PR, I updated `catch(const std::exception&)` to `catch(...)`, where we don;t care about the exception being derived from `std::exception`, and actually need to close the handle regardless.


antkmsft:
Inspired by #2361 and #2360.

sdk/keyvault/azure-security-keyvault-common/src/sha.cpp
```diff
@@ -88,7 +88,7 @@ class SHAWithOpenSSL final : public Azure::Core::Cryptography::Hash {
       }
       default:
         // imposible to get here
-        throw;
+        throw std::exception();
```
Jinming-Hu, line 91: We have an unreachable assertion defined in `azure_assert.hpp`

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -1034,7 +1034,7 @@ int64_t CurlSession::ResponseBufferParser::Parse(
         {
           // Should never happen that parser is not statusLIne or Headers and we still try
           // to parse more.
-          throw;
+          throw std::exception();
```
Jinming-Hu, line 1037:
```suggestion
          AZURE_UNREACHABLE_CODE();
```

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -996,7 +996,7 @@ int64_t CurlSession::ResponseBufferParser::Parse(
...
```
Jinming-Hu, line 999:
```suggestion
          AZURE_UNREACHABLE_CODE();
```

Jinming-Hu (changes requested):

sdk/keyvault/azure-security-keyvault-common/src/sha.cpp
antkmsft, line 91:
```suggestion
        AZURE_UNREACHABLE_CODE();
```

Jinming-Hu (approved):

sdk/core/azure-core/src/io/body_stream.cpp
```diff
@@ -130,7 +130,7 @@ FileBodyStream::FileBodyStream(const std::string& filename)
     m_randomAccessFileBodyStream = std::make_unique<_internal::RandomAccessFileBodyStream>(
         _internal::RandomAccessFileBodyStream(m_filehandle, 0, fileSize.QuadPart));
   }
-  catch (const std::exception&)
+  catch (...)
```
ahsonkhan, line 133: Good catch!

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -1034,7 +1034,7 @@ int64_t CurlSession::ResponseBufferParser::Parse(
         {
           // Should never happen that parser is not statusLIne or Headers and we still try
           // to parse more.
-          throw;
+          AZURE_UNREACHABLE_CODE();
```
ahsonkhan, line 1037: Nice. @vhvb1989 maybe take a quick look as well.

ahsonkhan (commented):
This is a great fix, thanks for discovering this Anton and auditing our try/catch and exception blocks.

@antkmsft Consider compiling some of the issues you discovered and coding guidelines you shared and put it in the C++ guidelines.

I recommend starting by filing an issue here: https://github.com/Azure/azure-sdk/issues?q=is%3Aopen+is%3Aissue+label%3AC%2B%2B+label%3A%22ArchBoard+Guidelines%22

PR #2364, opened 2021-05-28
fix blob/file read consistency issue

Jinming-Hu:
Hi @JeffreyRichter , this PR is to fix a data consistency issue where the blob/file is overwritten during a download process that requires more than one HTTP request. The behavior is now consistent across all language storage sdks after this PR is merged.

There's also another feature in other language sdk that allows customers to explicitly choose to risk this kind of inconsistency issue. It's like "Just download the blob/file and I don't care if it's modified during download". This feature can be useful when customers know for sure the data they care about won't be modified (like append only operations).
Do you think it's necessary for us to add such a feature in C++?

### DIFF
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -191,10 +191,7 @@ namespace Azure { namespace Storage { namespace Blobs {
         {
           newOptions.Range.Value().Length = options.Range.Value().Length.Value() - retryOffset;
         }
-        if (!newOptions.AccessConditions.IfMatch.HasValue())
-        {
-          newOptions.AccessConditions.IfMatch = eTag;
-        }
+        newOptions.AccessConditions.IfMatch = eTag;
         return std::move(Download(newOptions, context).Value.BodyStream);
       };
 
@@ -291,10 +288,7 @@ namespace Azure { namespace Storage { namespace Blobs {
             chunkOptions.Range = Core::Http::HttpRange();
             chunkOptions.Range.Value().Offset = offset;
             chunkOptions.Range.Value().Length = length;
-            if (!chunkOptions.AccessConditions.IfMatch.HasValue())
-            {
-              chunkOptions.AccessConditions.IfMatch = eTag;
-            }
+            chunkOptions.AccessConditions.IfMatch = eTag;
             auto chunk = Download(chunkOptions, context);
             int64_t bytesRead = chunk.Value.BodyStream->ReadToCount(
                 buffer + (offset - firstChunkOffset),
@@ -417,10 +411,7 @@ namespace Azure { namespace Storage { namespace Blobs {
             chunkOptions.Range = Core::Http::HttpRange();
             chunkOptions.Range.Value().Offset = offset;
             chunkOptions.Range.Value().Length = length;
-            if (!chunkOptions.AccessConditions.IfMatch.HasValue())
-            {
-              chunkOptions.AccessConditions.IfMatch = eTag;
-            }
+            chunkOptions.AccessConditions.IfMatch = eTag;
             auto chunk = Download(chunkOptions, context);
             bodyStreamToFile(
                 *(chunk.Value.BodyStream),
--- a/sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
+++ b/sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
@@ -272,8 +272,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
         auto newResponse = Download(newOptions, context);
         if (eTag != newResponse.Value.Details.ETag)
         {
-          throw Azure::Core::RequestFailedException(
-              "File was changed during the download process.");
+          throw Azure::Core::RequestFailedException("file was modified in the middle of download");
         }
         return std::move(newResponse.Value.BodyStream);
       };
@@ -669,6 +668,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
     }
 
     auto firstChunk = Download(firstChunkOptions, context);
+    const Azure::ETag etag = firstChunk.Value.Details.ETag;
 
     int64_t fileSize;
     int64_t fileRangeSize;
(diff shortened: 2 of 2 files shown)


vhvb1989 (approved):
If test cases for simulating this is hard, I would create an issue to make it at some point in the future

JeffreyRichter (approved):

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
```diff
@@ -272,8 +272,7 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
         auto newResponse = Download(newOptions, context);
         if (eTag != newResponse.Value.Details.ETag)
         {
-          throw Azure::Core::RequestFailedException(
-              "File was changed during the download process.");
+          throw Azure::Core::RequestFailedException("file was modified in the middle of download");
```
ahsonkhan, line 275:
```suggestion
          throw Azure::Core::RequestFailedException("File was modified in the middle of download.");
```

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
```diff
@@ -728,6 +728,11 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
             {
               throw Azure::Core::RequestFailedException("error when reading body stream");
             }
+            if (chunk.Value.Details.ETag != etag)
+            {
+              throw Azure::Core::RequestFailedException(
+                  "file was modified in the middle of download");
```
ahsonkhan, line 734:
```suggestion
              throw Azure::Core::RequestFailedException(
                  "File was modified in the middle of download.");
```

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
```diff
@@ -840,6 +846,11 @@ namespace Azure { namespace Storage { namespace Files { namespace Shares {
             chunkOptions.Range.Value().Offset = offset;
             chunkOptions.Range.Value().Length = length;
             auto chunk = Download(chunkOptions, context);
+            if (chunk.Value.Details.ETag != etag)
+            {
+              throw Azure::Core::RequestFailedException(
+                  "file was modified in the middle of download");
```
ahsonkhan, line 852:
```suggestion
                  "File was modified in the middle of download.");
```

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -191,10 +191,7 @@ namespace Azure { namespace Storage { namespace Blobs {
         {
           newOptions.Range.Value().Length = options.Range.Value().Length.Value() - retryOffset;
         }
-        if (!newOptions.AccessConditions.IfMatch.HasValue())
-        {
-          newOptions.AccessConditions.IfMatch = eTag;
-        }
+        newOptions.AccessConditions.IfMatch = eTag;
```
ahsonkhan, line 194: > this PR is to fix a data consistency issue where the blob/file is overwritten during a download process that requires more than one HTTP request. 

Help me understand how such change fixes this issue?
How did you test it?

ahsonkhan (approved):

ahsonkhan:
> Do you think it's necessary for us to add such a feature in C++?

@Jinming-Hu how do you think we would enable this feature?

@JeffreyRichter do we need it?

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: Before this change, customers can specify `*` for this `IfMatch` value. In that case, we do not throw exception even if the ETag from multiple HTTP requests doesn't match.

Jinming-Hu:
> @Jinming-Hu how do you think we would enable this feature?

Add another field in options bag, like `LockEtag`. If set to `false`, we don't memorize the first ETag or compare ETag on each response.

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: > How did you test it?

It's not easy to test. I feel confident about my code.

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
Jinming-Hu: We need to be consistent about exception messages (uppercase or lowercase for the first character, `.` at the end). What's the format of exception message in Core SDK?

JeffreyRichter:
The customer should be able to pass an explicit if-match etag and if the blob doesn't match the operation should fail.
If the customer does not pass an explicit if-match etag then the first downloaded chunk should get the etag and ensure that all other chunks match this etag - this prevents the possibility of data corruption.
It could be tested by using a large blob and inserting a delay between chunks and then changing the blob during this delay.

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/2366

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/2366

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/issues/2366

sdk/storage/azure-storage-files-shares/src/share_file_client.cpp
ahsonkhan: There might be mistakes in core about this (will check and see), but we should be using complete sentences that are customer-oriented so the developer can understand what went wrong and how they might need to handle those cases.

This is a good example (especially since we usability tested it, even though it may not be an exception):
https://github.com/Azure/azure-sdk-for-cpp/blob/dfd2933a66372bcc27e8563f5ef2329615f3a894/sdk/core/azure-core/inc/azure/core/io/body_stream.hpp#L61-L68

ahsonkhan:
> It could be tested by using a large blob and inserting a delay between chunks and then changing the blob during this delay.

@Jinming-Hu let us know the result of such a test, and if we can check it in.

Issue #2304, opened 2021-05-18
Docs: review and update docs for Azure::Core namespace


antkmsft:
In your initial PR, you mentioned it was only done for some classes?
Please close this item if you've done everything already.

Extracted from #2260:

Guidelines:
"Drop Represents, Defines and other "intro words" in the doc comment, and just state what it is."
"Use words, by adding (s) like "Gets", "Returns", "Destroys", etc. and not "Get", "Return", "Destroy""
"Add pre tags for preconditions"
"Add 2-5 liner code snippets for the most important or commonly used APIs, such as client ctors, that folks can use for getting started, maybe using \example"
"Avoid self-referential links on the same page" - use \`%ClassName\`
"Add docs to Tparam (for example Operation)"
"Use @note to highlight important things to call out rather than @remark"

antkmsft:
See also recent beta9 docs - please check that docs for the classes inside `Azure::Core` namespace look good: https://azuresdkdocs.blob.core.windows.net/$web/cpp/azure-core/1.0.0-beta.9/namespace_azure_1_1_core.html

Issue #2300, opened 2021-05-18
Docs: review and update docs for Azure:: (the root namespace, without Core), Cryptography


antkmsft:
Extracted from #2260:

Guidelines:
"Drop Represents, Defines and other "intro words" in the doc comment, and just state what it is."
"Use words, by adding (s) like "Gets", "Returns", "Destroys", etc. and not "Get", "Return", "Destroy""
"Add pre tags for preconditions"
"Add 2-5 liner code snippets for the most important or commonly used APIs, such as client ctors, that folks can use for getting started, maybe using \example"
"Avoid self-referential links on the same page" - use \`%ClassName\`
"Add docs to Tparam (for example Operation)"
"Use @note to highlight important things to call out rather than @remark"

antkmsft:
DateTime ctor remark reword, change to: This class can handle a DateTime...
 Add example format for Rfc format enum in the docs.
 Link the actual RFC doc in docs.
 ETag docs have references that used to be in Azure::Core, make sure to reference the right namespace.

`[Yesterday 1:01 PM] Anton Kolesnyk
    Your first question, I was talking about class docs, i.e.
/** * @brief Manages date and time in standardized string formats.
 * @note `DateTime` is convertible to `std::chrono::time_point` (see
 * https://en.cppreference.com/w/cpp/chrono/time_point).`

antkmsft:
See also recent beta9 docs: https://azuresdkdocs.blob.core.windows.net/$web/cpp/azure-core/1.0.0-beta.9/annotated.html

PR #2378, opened 2021-06-02
Core: Doxygen: Constructors, destructors, operators, and context parameter

### LINKED ISSUE #2298: Docs: Apply common wording
Extracted from #2260:

* Construtors, destructors, operators (including deleted and defaulted)
* Context parameter
* Boolean returns

"Add commonly used template description for all simple/defaulted virtual dtors that give the reason for why we have it."
"Have common phrasing as a starting point across all ctor/dtors for consistency. Constructs the <type name>.../Destructs the <type name> ..."
"Document delete functions like the default ctor, using a template statement."
"For bool returns, follow C, and do ; otherwise, false."

antkmsft:
Closes #2298.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/base64.hpp
+++ b/sdk/core/azure-core/inc/azure/core/base64.hpp
@@ -21,6 +21,10 @@ namespace Azure { namespace Core {
   class Convert final {
   private:
     // This type currently only contains static methods and hence disallowing instance creation.
+    /**
+     * @brief An instance of `%Convert` class cannot be created.
+     *
+     */
     Convert() = default;
 
   public:
--- a/sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
+++ b/sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
@@ -60,7 +60,7 @@ namespace Azure { namespace Core { namespace Credentials {
      *
      * @param tokenRequestContext #Azure::Core::Credentials::TokenRequestContext to get the token
      * in.
-     * @param context #Azure::Core::Context so that operation can be cancelled.
+     * @param context A context to control the request lifetime.
      *
      * @throw Azure::Core::Credentials::AuthenticationException Authentication error occurred.
      */
@@ -68,14 +68,30 @@ namespace Azure { namespace Core { namespace Credentials {
         TokenRequestContext const& tokenRequestContext,
         Context const& context) const = 0;
 
-    /// Destructor.
+    /**
+     * @brief Destructs `%TokenCredential`.
+     *
+     */
     virtual ~TokenCredential() = default;
 
   protected:
+    /**
+     * @brief Constructs a default instance of `%TokenCredential`.
+     *
+     */
     TokenCredential() {}
 
   private:
+    /**
+     * @brief `%TokenCredential` does not allow copy construction.
+     *
+     */
     TokenCredential(TokenCredential const&) = delete;
+
+    /**
+     * @brief `%TokenCredential` does not allow assignment.
+     *
+     */
     void operator=(TokenCredential const&) = delete;
   };
 
@@ -87,7 +103,7 @@ namespace Azure { namespace Core { namespace Credentials {
 
   public:
     /**
-     * @brief Constructs with message string.
+     * @brief Constructs `%AuthenticationException` with a message string.
      *
      * @param message Message string.
      */
--- a/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
+++ b/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
@@ -45,12 +45,14 @@ namespace Azure { namespace Core { namespace Cryptography {
      */
     virtual std::vector<uint8_t> OnFinal(const uint8_t* data, std::size_t length) = 0;
 
-  public:
+  protected:
     /**
-     * @brief Construct a default instance of #Azure::Core::Cryptography::Hash.
+     * @brief Constructs a default instance of `%Hash`.
      *
      */
     Hash() = default;
+
+  public:
 
     /**
      * @brief Used to append partial binary input data to compute the hash in a streaming fashion.
@@ -95,7 +97,7 @@ namespace Azure { namespace Core { namespace Cryptography {
     std::vector<uint8_t> Final() { return Final(nullptr, 0); }
 
     /**
-     * @brief Cleanup any state when destroying the instance of #Azure::Core::Cryptography::Hash.
+     * @brief Destructs `%Hash`.
      *
      */
     virtual ~Hash() = default;
@@ -103,8 +105,16 @@ namespace Azure { namespace Core { namespace Cryptography {
   private:
     bool m_isDone = false;
 
-    // Delete the copy constructor, along with the assignment operator.
+    /**
+     * @brief `%Hash` does not allow copy construction.
+     *
+     */
     Hash(Hash const&) = delete;
+
+    /**
+     * @brief `%Hash` does not allow assignment.
+     *
+     */
     void operator=(Hash const&) = delete;
   };
 
@@ -122,7 +132,7 @@ namespace Azure { namespace Core { namespace Cryptography {
     Md5Hash();
 
     /**
-     * @brief Cleanup any state when destroying the instance of #Azure::Core::Cryptography::Md5Hash.
+     * @brief Destructs `%Md5Hash`.
      *
      */
     ~Md5Hash() override;
--- a/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
+++ b/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
@@ -53,7 +53,6 @@ namespace Azure { namespace Core { namespace Cryptography {
     Hash() = default;
 
   public:
-
     /**
      * @brief Used to append partial binary input data to compute the hash in a streaming fashion.
      * @remark Once all the data has been added, call #Azure::Core::Cryptography::Hash::Final() to
--- a/sdk/core/azure-core/inc/azure/core/datetime.hpp
+++ b/sdk/core/azure-core/inc/azure/core/datetime.hpp
@@ -85,13 +85,13 @@ private:
 
 public:
   /**
-   * @brief Construct a default instance of #Azure::DateTime (00:00:00.0000000 on January
-   * 1st, 0001).
+   * @brief Constructs a default instance of `%DateTime` (`00:00:00.0000000 on January 1st, 0001`).
+   *
    */
   constexpr DateTime() : time_point() {}
 
   /**
-   * @brief Construct an instance of #Azure::DateTime.
+   * @brief Constructs an instance of `%DateTime`.
    *
    * @param year Year.
    * @param month Month.
@@ -114,15 +114,15 @@ public:
   }
 
   /**
-   * @brief Construct an instance of #Azure::DateTime from base class.
+   * @brief Constructs an instance of `DateTime` from a `time_point`.
    *
    */
   constexpr DateTime(time_point const& timePoint) : time_point(timePoint) {}
 
   /**
-   * @brief Construct an instance of #Azure::DateTime from
-   * `std::chrono::system_clock::time_point`.
+   * @brief Construct an instance of `%DateTime` from `std::chrono::system_clock::time_point`.
    * @param systemTime A value of `std::chrono::system_clock::time_point`.
+   *
    */
   DateTime(std::chrono::system_clock::time_point const& systemTime)
       : DateTime(
--- a/sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
+++ b/sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
@@ -54,7 +54,17 @@ namespace Azure { namespace Core { namespace Diagnostics {
     static void SetLevel(Level level);
 
   private:
+    /**
+     * @brief An instance of `%Logger` class cannot be created.
+     *
+     */
     Logger() = delete;
+
+    /**
+     * @brief An instance of `%Logger` class cannot be destructed, because no instance can be
+     * created.
+     *
+     */
     ~Logger() = delete;
   };
 }}} // namespace Azure::Core::Diagnostics
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -65,7 +65,7 @@ namespace Azure { namespace Core {
     std::unique_ptr<Azure::Core::Http::RawResponse> RawResponse;
 
     /**
-     * @brief Constructs a new `RequestFailedException` object.
+     * @brief Constructs a new `%RequestFailedException` with a message string.
      *
      * @note An Exception without an HTTP raw response represents an exception that happened
      * before sending the request to the server.
@@ -78,7 +78,7 @@ namespace Azure { namespace Core {
     }
 
     /**
-     * @brief Constructs a new `RequestFailedException` object with an HTTP raw response.
+     * @brief Constructs a new `%RequestFailedException` object with an HTTP raw response.
      *
      * @note The HTTP raw response is parsed to populate information expected from all Azure
      * Services like the status code, reason phrase and some headers like the request ID. A concrete
@@ -93,10 +93,10 @@ namespace Azure { namespace Core {
         std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);
 
     /**
-     * @brief Constructs a new `RequestFailedException` by copying from an existing one.
+     * @brief Constructs a new `%RequestFailedException` by copying from an existing one.
      * @note Copies the #Azure::Core::Http::RawResponse into the new `RequestFailedException`.
      *
-     * @param other The `RequestFailedException` to be copied.
+     * @param other The `%RequestFailedException` to be copied.
      */
     RequestFailedException(const RequestFailedException& other)
         : std::runtime_error(other.Message), StatusCode(other.StatusCode),
@@ -109,9 +109,29 @@ namespace Azure { namespace Core {
     {
     }
 
+    /**
+     * @brief Constructs a new `%RequestFailedException` by moving in an existing one.
+     * @param other The `%RequestFailedException` to move in.
+     */
     RequestFailedException(RequestFailedException&& other) = default;
+
+    /**
+     * @brief An instance of `%RequestFailedException` class cannot be assigned.
+     *
+     */
     RequestFailedException& operator=(const RequestFailedException&) = delete;
+
+    /**
+     * @brief An instance of `%RequestFailedException` class cannot be moved into another instance
+     * after creation.
+     *
+     */
     RequestFailedException& operator=(RequestFailedException&&) = delete;
+
+    /**
+     * @brief Destructs `%RequestFailedException`.
+     *
+     */
     ~RequestFailedException() = default;
   };
 }} // namespace Azure::Core
--- a/sdk/core/azure-core/inc/azure/core/http/http.hpp
+++ b/sdk/core/azure-core/inc/azure/core/http/http.hpp
@@ -51,13 +51,12 @@ namespace Azure { namespace Core { namespace Http {
 
   /*********************  Exceptions  **********************/
   /**
-   * @brief Consistent interface to handle errors occurred while performing an HTTP
-   * Request.
+   * @brief An error while sending the HTTP request with the transport adapter.
    */
   class TransportException final : public Azure::Core::RequestFailedException {
   public:
     /**
-     * @brief An error while sending the HTTP request with the transport adapter.
+     * @brief Constructs `%TransportException` with \p message string.
      *
      * @remark The transport policy will throw this error whenever the transport adapter fail to
      * perform a request.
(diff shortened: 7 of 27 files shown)


sdk/core/azure-core/inc/azure/core/base64.hpp
```diff
@@ -21,6 +21,10 @@ namespace Azure { namespace Core {
   class Convert final {
   private:
     // This type currently only contains static methods and hence disallowing instance creation.
+    /**
+     * @brief An instance of `%Convert` class cannot be created.
```
ahsonkhan, line 25: What does the `%` symbol do and can you share how this renders in doxygen?

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
```diff
@@ -60,22 +60,38 @@ namespace Azure { namespace Core { namespace Credentials {
      *
      * @param tokenRequestContext #Azure::Core::Credentials::TokenRequestContext to get the token
      * in.
-     * @param context #Azure::Core::Context so that operation can be cancelled.
+     * @param context A context to control the request lifetime.
```
ahsonkhan, line 63: In this context (no pun intended), for getting an access token, is there a request being sent whose lifetime is actually being controlled by the context, or does it server a different purpose?

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
```diff
@@ -60,22 +60,38 @@ namespace Azure { namespace Core { namespace Credentials {
...
 
   private:
+    /**
+     * @brief `%TokenCredential` does not allow copy construction.
+     *
+     */
     TokenCredential(TokenCredential const&) = delete;
+
+    /**
+     * @brief `%TokenCredential` does not allow assignment.
```
ahsonkhan, line 92: 👍

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -45,13 +45,14 @@ namespace Azure { namespace Core { namespace Cryptography {
      */
     virtual std::vector<uint8_t> OnFinal(const uint8_t* data, std::size_t length) = 0;
 
-  public:
+  protected:
     /**
-     * @brief Construct a default instance of #Azure::Core::Cryptography::Hash.
+     * @brief Constructs a default instance of `%Hash`.
```
ahsonkhan, line 50: Just to confirm, isn't this a breaking change, it needs to be called out in the CL, and please try to avoid such subtle changes in large doc updates.

Why do we need to make this change? Please revert changing from public to protected in this PR and open a separate PR for this, and let's discuss the reasons/merits there.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -114,15 +114,15 @@ class DateTime final : public _detail::Clock::time_point {
   }
 
   /**
-   * @brief Construct an instance of #Azure::DateTime from base class.
+   * @brief Constructs an instance of `DateTime` from a `time_point`.
```
ahsonkhan, line 117: Why no `%` here? Can you please help me understand the difference.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -114,15 +114,15 @@ class DateTime final : public _detail::Clock::time_point {
...
    *
    */
   constexpr DateTime(time_point const& timePoint) : time_point(timePoint) {}
 
   /**
-   * @brief Construct an instance of #Azure::DateTime from
-   * `std::chrono::system_clock::time_point`.
+   * @brief Construct an instance of `%DateTime` from `std::chrono::system_clock::time_point`.
    * @param systemTime A value of `std::chrono::system_clock::time_point`.
+   *
```
ahsonkhan, line 125: @vhvb1989 is this extra new line required still?

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
```diff
@@ -54,7 +54,17 @@ namespace Azure { namespace Core { namespace Diagnostics {
     static void SetLevel(Level level);
 
   private:
+    /**
+     * @brief An instance of `%Logger` class cannot be created.
+     *
```
ahsonkhan, line 59: Same here, @vhvb1989 is this extra new line required still?
I thought we didn't need it anymore. @antkmsft  please confirm and check if we are consistent elsewhere (it was modified recently). I think removing this extra line is necessary to get rid of the "more..." link.

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
```diff
@@ -54,7 +54,17 @@ namespace Azure { namespace Core { namespace Diagnostics {
...
+     */
     Logger() = delete;
+
+    /**
+     * @brief An instance of `%Logger` class cannot be destructed, because no instance can be
+     * created.
```
ahsonkhan, line 65: Nice!

ahsonkhan (changes requested):
I was about to block this PR because of the subtle breaking change :(

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -65,7 +65,7 @@ namespace Azure { namespace Core {
     std::unique_ptr<Azure::Core::Http::RawResponse> RawResponse;
 
     /**
-     * @brief Constructs a new `RequestFailedException` object.
+     * @brief Constructs a new `%RequestFailedException` with a message string.
```
ahsonkhan, line 68: Consistency nit (consider fixing elsewhere too):
```suggestion
     * @brief Constructs a new `%RequestFailedException` with a \p message string.
```

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -95,70 +94,68 @@ namespace Azure { namespace Core { namespace Http {
    */
   class HttpMethod final {
   public:
-    HttpMethod() = delete;
```
ahsonkhan, line 98: Why did we get rid of this?

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
antkmsft: @vhvb1989, should we ever have them, or only sometimes, or never?

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
```diff
@@ -169,24 +169,52 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
...
+     * @param other Other `%HttpPolicy` to move.
+     *
+     */
+    HttpPolicy(HttpPolicy&& other) = default;
   };
 
   /**
    * @brief The next HTTP policy in the stack sequence of policies.
+   * @note `%NextHttpPolicy` is an abstraction representing a next line in the stack sequence of
+   * policies, from the caller's perspective.
```
ahsonkhan, line 217: This is a bit confusing to parse.

Maybe modify as something like this:
`...representing the next policy to execute in the sequence of policies.`

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
```diff
@@ -201,9 +229,9 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
      * @brief Constructs an abstraction representing a next line in the stack sequence of policies,
      * from the caller's perspective.
      *
-     * @param index An sequential index of this policy in the stack sequence of policies.
-     * @param policies A vector of unique pointers next in the line to be invoked after the
-     * current policy.
+     * @param index A sequential index of this policy in the stack sequence of policies.
```
ahsonkhan, line 232: Good catch!

sdk/core/azure-core/inc/azure/core/nullable.hpp
```diff
@@ -74,7 +82,8 @@ template <class T> class Nullable final {
   }
 
   /**
-   * @brief Destructs the Nullable, destructs the contained value if there is one.
+   * @brief Destructs the `%Nullable`, calling the destructor for the contained value if there is
+   * one.
```
ahsonkhan, line 86: Great comment, containing an important detail.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Without `%`, it inserts a link. They render just as you'd expect - without the `%` and not as hyperlink. You can look in the build artifacts - https://dev.azure.com/azure-sdk/public/_build/results?buildId=922992&view=artifacts&pathAsName=false&type=publishedArtifacts

sdk/core/azure-core/inc/azure/core/response.hpp
```diff
@@ -31,8 +31,7 @@ template <class T> class Response final {
   std::unique_ptr<Azure::Core::Http::RawResponse> RawResponse;
 
   /**
-   * @brief Initialize an #Azure::Response<T> with the value and raw response returned by the
-   * service.
+   * @brief Constructs a `%Response` with the value and raw response returned by the service.
```
ahsonkhan, line 34: Does this show the templated type correctly in doxygen, like `Response<T>`?

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: Since the instances of the class can't be created because it is abstract, only the classes deriving from it will have access to this constructor, for which `protected` is sufficient.

ahsonkhan:
Other than the code breaking changes which I need to understand the reasoning behind, mostly nits/small suggestions to fix up (and the extra new line).

sdk/core/azure-core/inc/azure/core/base64.hpp
antkmsft: https://www.doxygen.nl/manual/autolink.html#linkclass

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
antkmsft: Yes, the request is being sent. Credential is client. Other clients give you files (storage) or secrets (keyvault). Credential classes retreive auth tokens from the server. This is an oversimplification, to an extent.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: It should be there, I missed it!

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: If custom constructor is present, default constructor does not get generated either way.

antkmsft: But I think later we should explicitly list and document all the default operations for all the code. We have a work item for that. We currently don't do that, and it was not worth it to document the deletion of the constructor that happens by the C++ rules anyway.

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
antkmsft: I think to combine would be better "`is an abstraction representing the next policy in the stack sequence of policies, from the caller's perspective.`".

sdk/core/azure-core/inc/azure/core/response.hpp
antkmsft: It would print the name in monospace font without the link, i.e without the self-reference.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Will do a follow up PR.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Maybe will be in a follow up PR, depending on the answer

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
antkmsft: Maybe will be in a follow up PR, depending on the answer

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: I think we should leave it, it is not technically a breaking change, since you can't call a public constructor of an abstract class, other than deriving from it, for which `protected` constructor is accessible.

sdk/core/azure-core/inc/azure/core/http/http.hpp
```diff
@@ -51,13 +51,12 @@ namespace Azure { namespace Core { namespace Http {
...
   /**
-   * @brief Consistent interface to handle errors occurred while performing an HTTP
-   * Request.
+   * @brief An error while sending the HTTP request with the transport adapter.
    */
   class TransportException final : public Azure::Core::RequestFailedException {
   public:
     /**
-     * @brief An error while sending the HTTP request with the transport adapter.
+     * @brief Constructs `%TransportException` with \p message string.
```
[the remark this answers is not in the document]
antkmsft: Will take to the follow up PR.

sdk/core/azure-core/inc/azure/core/exception.hpp
antkmsft: Will take to the follow up PR.

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
antkmsft: Will bring in the follow up PR.

sdk/core/azure-core/inc/azure/core/http/http.hpp
antkmsft: I am planning to not take this to the follow up PR.

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: I am planning to not take this to the follow up PR.

sdk/core/azure-core/inc/azure/core/response.hpp
antkmsft: Please LMK if you think we should be monospacing it as `Response<T>` (then it should apply to `Nullable<T>` as well). I don't think we should.

antkmsft: So far, not taking this to a follow up PR.

sdk/core/azure-core/inc/azure/core/diagnostics/logger.hpp
vhvb1989: We want it for docs inside a class, like properties and methods.

We don't want it for documentation out of classes, like struct/class/namespace/global stuff

antkmsft: Victor: "We want it for docs inside a class, like properties and methods.

We don't want it for documentation out of classes, like struct/class/namespace/global stuff"

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Victor "We want it for docs inside a class, like properties and methods.

We don't want it for documentation out of classes, like struct/class/namespace/global stuff"

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
ahsonkhan: What do you mean by "stack sequence of policies" and what does "the caller's perspective" mean? Those two phrases confuse me a bit, so I question their usefulness. Your call of course :)

sdk/core/azure-core/inc/azure/core/response.hpp
ahsonkhan: > monospacing it as `Response<T>`

What do you mean by that?

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: #2390

sdk/core/azure-core/inc/azure/core/exception.hpp
antkmsft: #2390

sdk/core/azure-core/inc/azure/core/http/http.hpp
ahsonkhan: @CaseyCarter for the extensible enum pattern, should we explicitly delete the default ctor or leave it to the compiler to generate?
The guideline doesn't have it:

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
antkmsft: Maybe I am biased, because I think I know exactly what this means, and it makes perfect sense to me.
"next" policy is in relation to who? To the caller. I.e. whenever you see this and yo can call it, it is next relative to you, not in the absolute sense.
And as of the stack. Policies are a stack.

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
ahsonkhan: Do we need to define a ctor at all in this case? For example, BodyStream is an abstract class, that doesn't have a ctor.

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
antkmsft: I like the wording that I put in ##2378. If you give it to me, I'll leave it as I have it in that PR. But I don't feel super strongly on this, if you think of a better wording, please propose it in that PR.

sdk/core/azure-core/inc/azure/core/http/http.hpp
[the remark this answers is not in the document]
antkmsft: #2390

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: You are right - we don't need to (we would've need if if we needed a custom one, but since we are using the default - we don't need to define it at all

sdk/core/azure-core/inc/azure/core/datetime.hpp
ahsonkhan: I think a few other instances were missed. For example:
https://github.com/antkmsft/azure-sdk-for-cpp/blob/6185cba9e6701a1943f48f52963c963637cd8e52/sdk/core/azure-core/inc/azure/core/datetime.hpp#L52

https://github.com/antkmsft/azure-sdk-for-cpp/blob/6185cba9e6701a1943f48f52963c963637cd8e52/sdk/core/azure-core/inc/azure/core/datetime.hpp#L134-L136

https://github.com/antkmsft/azure-sdk-for-cpp/blob/6185cba9e6701a1943f48f52963c963637cd8e52/sdk/core/azure-core/inc/azure/core/datetime.hpp#L172

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: Or, if there were custom constructors with parameters, but we also wanted to have a parameterless one.

antkmsft: I added that to #2390.

sdk/core/azure-core/inc/azure/core/response.hpp
ahsonkhan: Nevermind, I get what you mean by `monospacing`, the "code" like font. I was more curious about the template `<T>`, maybe we should explicitly add that whenever referring to the type in the docs. So, not `Response` but rather `Response<T>`

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Not fixing all of them so far, only the places I was adding. There are methods next to constructors, and they do still do self-reference. There are work items to go and update classes in each namespace, divided by the team member, it can be covered there. I'll go and update comments there.

antkmsft: I updated #2299, #2300, and #2304

sdk/core/azure-core/inc/azure/core/response.hpp
antkmsft: See how it renders in the published artifacts - https://dev.azure.com/azure-sdk/public/_build/results?buildId=922992&view=artifacts&pathAsName=false&type=publishedArtifacts

antkmsft: I.e. 
image

antkmsft: > > I don't think we should.
> I am curious, how come?

Doxygen refers to it as simply `Response`, not `Response<>` or `Response<T>`. `T` is also a type parameter name, although a very common one. What if we rename `T` to `ResponseType`, should we start referring to it as `Response<ResponseType>`? I think of `<T>` as an implementation detail. It could've been used to distinguish between different `Response` types, should've had a templated and not templated versions, but this is not the case.

antkmsft: And if we'd have to emphasize it is a template, I think more pedantic way to call it would be `Response<>`, not `Response<T>`.

ahsonkhan: Makes sense, I think we can leave it as you have it for now.

sdk/core/azure-core/inc/azure/core/http/http.hpp
CaseyCarter: I'm perfectly happy for the default constructor to be implicitly deleted without calling it out explicitly. This isn't as likely to cause confusion as are the arcane rules around when move special members are generated or suppressed which motivates the rule of 0/3/5 guideline.

PR #2379, opened 2021-06-02
Fix potential overflow of casting

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -385,12 +385,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       }
     };
 
-    bodyStreamToFile(
-        *(firstChunk.Value.BodyStream),
-        fileWriter,
-        0,
-        firstChunkLength,
-        context);
+    bodyStreamToFile(*(firstChunk.Value.BodyStream), fileWriter, 0, firstChunkLength, context);
     firstChunk.Value.BodyStream.reset();
 
     auto returnTypeConverter = [](Azure::Response<Models::DownloadBlobResult>& response) {
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -255,7 +255,7 @@ namespace Azure { namespace Storage { namespace Blobs {
     }
     firstChunkLength = std::min(firstChunkLength, blobRangeSize);
 
-    if (blobRangeSize > std::numeric_limits<size_t>::max()
+    if (static_cast<uint64_t>(blobRangeSize) > std::numeric_limits<size_t>::max()
         || static_cast<size_t>(blobRangeSize) > bufferSize)
     {
       throw Azure::Core::RequestFailedException(
@@ -373,7 +373,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       std::vector<uint8_t> buffer(bufferSize);
       while (length > 0)
       {
-        size_t readSize = std::min(bufferSize, static_cast<size_t>(length));
+        size_t readSize = static_cast<size_t>(std::min<int64_t>(bufferSize, length));
         size_t bytesRead = stream.ReadToCount(buffer.data(), readSize, context);
         if (bytesRead != readSize)
         {
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -255,7 +255,8 @@ namespace Azure { namespace Storage { namespace Blobs {
     }
     firstChunkLength = std::min(firstChunkLength, blobRangeSize);
 
-    if (static_cast<size_t>(blobRangeSize) > bufferSize)
+    if (blobRangeSize > std::numeric_limits<size_t>::max()
+        || static_cast<size_t>(blobRangeSize) > bufferSize)
     {
       throw Azure::Core::RequestFailedException(
           "buffer is not big enough, blob range size is " + std::to_string(blobRangeSize));
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -367,13 +367,13 @@ namespace Azure { namespace Storage { namespace Blobs {
     auto bodyStreamToFile = [](Azure::Core::IO::BodyStream& stream,
                                _internal::FileWriter& fileWriter,
                                int64_t offset,
-                               size_t length,
+                               int64_t length,
                                const Azure::Core::Context& context) {
       constexpr size_t bufferSize = 4 * 1024 * 1024;
       std::vector<uint8_t> buffer(bufferSize);
       while (length > 0)
       {
-        size_t readSize = std::min(bufferSize, length);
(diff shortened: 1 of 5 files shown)


sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -366,13 +367,13 @@ namespace Azure { namespace Storage { namespace Blobs {
...
                                int64_t offset,
-                               size_t length,
+                               int64_t length,
                                const Azure::Core::Context& context) {
       constexpr size_t bufferSize = 4 * 1024 * 1024;
       std::vector<uint8_t> buffer(bufferSize);
       while (length > 0)
       {
-        size_t readSize = std::min(bufferSize, length);
+        size_t readSize = std::min(bufferSize, static_cast<size_t>(length));
```
Jinming-Hu, line 376: This is still problematic. What if `length` is exact `0x100000000`? It will become `0` after casting to `size_`, then it will lead to a infinite loop.

Jinming-Hu:
Hi @ahsonkhan , I want to use `uint64_t` for all sizes, and `int64_t` for all offsets (within storage sdk, so core sdk will not be affected). The API interfaces that are public to customers will remain `int64_t` so our customers are not affected.

The advantage is that I no longer need to consider the comparison between `int64_t` and `size_t`. (which is nightmare because on 32bit platform, `size_t` will be promoted to signed 64 bit integer, while on 64bit platform, `int64_t` will be converted to unsigned)

This can also save us a lot of `static_cast`s. What do you think?

sdk/storage/azure-storage-blobs/src/blob_client.cpp
antkmsft: `readSize = static_cast<size_t>(std::min<int64_t>(bufferSize, length))`?

antkmsft (approved):
Please fix the problematic line that you've identified, other than that, LGTM.

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: Yeah, that will work. But I just feel bad about the castings, casting `bufferSize` to `int64_t` first, then casting min of the two back to `size_t`.

What do you think of my [proposal](https://github.com/Azure/azure-sdk-for-cpp/pull/2379#issuecomment-852687282)?

antkmsft: Sounds reasonable to me, but I am not deep into the size_t vs int64_t discussion.
I am fine with int64_t. But either way, I here would care more about the tactical stuff which is to make this `std::min` non-problematic, whichever way you chose.

vhvb1989 (approved):

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -255,7 +255,8 @@ namespace Azure { namespace Storage { namespace Blobs {
     }
     firstChunkLength = std::min(firstChunkLength, blobRangeSize);
 
-    if (static_cast<size_t>(blobRangeSize) > bufferSize)
+    if (static_cast<uint64_t>(blobRangeSize) > std::numeric_limits<size_t>::max()
```
ahsonkhan, line 258: So, I am guessing we are casting `blobRangeSize` to uint64_t to also validate that the int64_t wasn't negative. Is that needed? Is it possible for it to be < 0, and if so, consider adding a comment here to explain that. since it isn't obvious (though I do like this way of validating two conditions at the same time).

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
```diff
@@ -127,6 +127,11 @@ namespace Azure { namespace Storage { namespace Blobs {
     constexpr int64_t MaxBlockNumber = 50000;
     constexpr int64_t BlockGrainSize = 1 * 1024 * 1024;
 
+    if (static_cast<uint64_t>(options.TransferOptions.SingleUploadThreshold)
+        > std::numeric_limits<size_t>::max())
+    {
+      throw Azure::Core::RequestFailedException("Single upload threshold is too big");
```
ahsonkhan, line 133: What happens if it is set to a negative number? The exception message won't be accurate.

sdk/storage/azure-storage-common/src/crypt.cpp
```diff
@@ -151,6 +151,10 @@ namespace Azure { namespace Storage {
 
     std::vector<uint8_t> Sha256(const std::vector<uint8_t>& data)
     {
+      if (data.size() > std::numeric_limits<ULONG>::max())
```
ahsonkhan, line 154: How do we decide what's too big? Should this be a precondition instead of throwing, since it is easy for the caller to validate the input argument?

cc @JeffreyRichter

sdk/storage/azure-storage-common/src/file_io.cpp
```diff
@@ -31,7 +31,12 @@ namespace Azure { namespace Storage { namespace _internal {
   FileReader::FileReader(const std::string& filename)
   {
     int sizeNeeded = MultiByteToWideChar(
-        CP_UTF8, MB_ERR_INVALID_CHARS, filename.data(), int(filename.length()), nullptr, 0);
+        CP_UTF8,
+        MB_ERR_INVALID_CHARS,
+        filename.data(),
+        static_cast<int>(filename.length()),
```
ahsonkhan, line 37: Why is this cast safe to do? Isn't the filename provided by the end user? There should be a static assert here or somewhere up the call stack ensuring the filename is small enough.

sdk/storage/azure-storage-common/src/crypt.cpp
JeffreyRichter: Yes, a precondition seems right instead of throwing.

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: > I am guessing we are casting blobRangeSize to uint64_t to also validate that the int64_t wasn't negative. Is that needed?

No, that wasn't my intention. This is to check casting `blobRangeSize` to `size_t` won't lead to overflow.

sdk/storage/azure-storage-blobs/src/block_blob_client.cpp
Jinming-Hu: I don't think negative value makes any sense. So it'll be undefined behavior. If the customer doesn't provide a valid value, we don't guarantee what would happen.

sdk/storage/azure-storage-common/src/file_io.cpp
Jinming-Hu: Filename larger than `INT_MAX` doesn't make any sense on any OS or filesystem. So I think this case is safe.

sdk/storage/azure-storage-common/src/crypt.cpp
Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/pull/2421

PR #2384, opened 2021-06-02
Add sample for transactional checksum

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/CMakeLists.txt
+++ b/sdk/storage/azure-storage-blobs/CMakeLists.txt
@@ -115,6 +115,7 @@ if(BUILD_STORAGE_SAMPLES)
         sample/blob_getting_started.cpp
         sample/blob_list_operation.cpp
         sample/blob_sas.cpp
+        sample/transactional_checksum.cpp
   )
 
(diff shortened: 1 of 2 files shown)


antkmsft (approved):
I a not a fan how our samples are done, I think if they were in a separate repo or a folder, fully standalone-capable, i.e. each with their own `main()` and CMakeFile, without the `SAMPLE()`macro, they would be more useful for the people how are not familiar with the SDK, they would illustrate the process from beginning to end, and they could be modified for customer to start their own app by modifying our samples. These samples don't give that.
But, it is not the goal of this PR to change the way we do samples. Without that, this sample, in alignment with other existing ones, looks ok to me.

Maybe, it is worth to put more empty lines between groups of statements.
Also, `buffer = ...ReadToEnd()` does not read too well, because it is declared 20 or so lines above. But it I won't block this PR due to that.

PR #2390, opened 2021-06-03
Doxygen PR follow-up

antkmsft:
Follow up on #2378 feedback that was published after the PR got completed.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
+++ b/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
@@ -45,6 +45,13 @@ namespace Azure { namespace Core { namespace Cryptography {
      */
     virtual std::vector<uint8_t> OnFinal(const uint8_t* data, std::size_t length) = 0;
 
+  protected:
+    /**
+     * @brief Constructs a default instance of `%Hash`.
+     *
+     */
+    Hash() = default;
+
   public:
     /**
      * @brief Used to append partial binary input data to compute the hash in a streaming fashion.
--- a/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
+++ b/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
@@ -45,13 +45,6 @@ namespace Azure { namespace Core { namespace Cryptography {
      */
     virtual std::vector<uint8_t> OnFinal(const uint8_t* data, std::size_t length) = 0;
 
-  protected:
-    /**
-     * @brief Constructs a default instance of `%Hash`.
-     *
-     */
-    Hash() = default;
-
   public:
     /**
      * @brief Used to append partial binary input data to compute the hash in a streaming fashion.
--- a/sdk/core/azure-core/inc/azure/core/datetime.hpp
+++ b/sdk/core/azure-core/inc/azure/core/datetime.hpp
@@ -114,7 +114,7 @@ public:
   }
 
   /**
-   * @brief Constructs an instance of `DateTime` from a `time_point`.
+   * @brief Constructs an instance of `%DateTime` from a `time_point`.
    *
    */
   constexpr DateTime(time_point const& timePoint) : time_point(timePoint) {}
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -65,7 +65,7 @@ namespace Azure { namespace Core {
     std::unique_ptr<Azure::Core::Http::RawResponse> RawResponse;
 
     /**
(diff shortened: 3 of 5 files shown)


vhvb1989 (approved):

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -45,13 +45,6 @@ namespace Azure { namespace Core { namespace Cryptography {
      */
     virtual std::vector<uint8_t> OnFinal(const uint8_t* data, std::size_t length) = 0;
 
-  protected:
-    /**
-     * @brief Constructs a default instance of `%Hash`.
-     *
-     */
-    Hash() = default;
```
ahsonkhan, line 53: So, removing the public ctor all together - breaking change?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -114,7 +114,7 @@ class DateTime final : public _detail::Clock::time_point {
   }
 
   /**
-   * @brief Constructs an instance of `DateTime` from a `time_point`.
+   * @brief Constructs an instance of `%DateTime` from a `time_point`.
```
ahsonkhan, line 117: Fyi, there are some other instances of self-referential DateTime links, if you want to fix those.

ahsonkhan (approved):

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: I think there should be a separate effort, I updated instructions for #2299, #2300, and #2304

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: No customer could ever call it. other than from a derived class. And for the derived class, nothing changes.
But, we can't delete it, because apparently there is custom constructor declaration.

antkmsft: I reverted this change from the PR. Left it as `protected` (and previous PR moved it out from `public` to `protected`, but as I explained above it does not impact customer code because you can't call a public constructor of an abstract class other than from a derived class).

ahsonkhan: > But, we can't delete it, because apparently there is custom constructor declaration.

The change you currently have is fine, but I think I am a bit confused now. Why do we need the explicit ctor declaration?

antkmsft: Apparently, `Hash(Hash const&) = delete` is in the category of custom-defined constructors: https://onlinegdb.com/oygrmmNa1

antkmsft: I think we really should explicitly list all the default constructors and operators at all times, for ultimate clarity and minimum surprises.

antkmsft: Even if they align with the (currently) default compiler behavior.

antkmsft: Because you `=delete` one, and list of default operations changes.

ahsonkhan: Rather than the parameterless ctor being protected and defaulted, should it be private and deleted similar to `Hash(Hash const&)`?

antkmsft: We can't do that - then the class will be non-inheritable, because it does not have any other constructor, and derived classes won't be able to construct the base class.

Jinming-Hu: > I think we really should explicitly list all the default constructors and operators at all times, for ultimate clarity and minimum surprises.

I always want to do the opposite, that is, defining constructors/destructors/assignment operators only when necessary.

In the case of `Hash` base class. I can see deleting copy constructor and copy assignment operator is necessary, because the inherited class usually has some kind of handle and thus is not copiable.
Since we defined copy constructor, we also need to define default (parameterless) constructor. Customers are not supposed to call this constructor directly, so it should be `protected`.
Virtual destructor is also necessary because it's perfectly reasonable that we use a pointer to `Hash` base class to hold  an inherited class object which can be `Md5Hash` or `Crc64Hash` or maybe `Sha256Hash` if we support it in the future. The actual concrete type doesn't have to be determined at compile-time, it can be determined at run-time.

So to sum up, the current definition of `Hash` base class looks perfect to me, and we don't need to make any changes to it,

Issue #2319, opened 2021-05-19
unify the usage of `std::size_t` or `size_t`


Jinming-Hu:
also `ptrdiff_t`

also functions like `memcpy` or `std::memcpy`

antkmsft:
`int64_t` is also an `std::` - https://en.cppreference.com/w/cpp/header/cstdint.

ahsonkhan:
Fixed most of the concerns here outside of storage, here: https://github.com/Azure/azure-sdk-for-cpp/pull/2415

Therefore, this issue can now be used to track storage fix-ups only.

PR #2415, opened 2021-06-07
Consolidate the use of std::size_t and always use size_t instead, since it is a "primitive" type.

ahsonkhan:
Fixes the azure core and keyvault parts of https://github.com/Azure/azure-sdk-for-cpp/issues/2319

We don't have an instances of types like `std::[u]int`, and functions like `std::memcpy` can remain as is. Left other things, within storage, like `ptrdiff_t` to be fixed separately.

### DIFF
--- a/samples/integration/vcpkg-storage/src/main.cpp
+++ b/samples/integration/vcpkg-storage/src/main.cpp
@@ -41,7 +41,7 @@ int main()
   {
     std::cout << metadata.first << ":" << metadata.second << std::endl;
   }
-  blobContent.resize(static_cast<std::size_t>(properties.BlobSize));
+  blobContent.resize(static_cast<size_t>(properties.BlobSize));
 
   blobClient.DownloadTo(reinterpret_cast<uint8_t*>(&blobContent[0]), blobContent.size());
 
--- a/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
+++ b/sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
@@ -33,7 +33,7 @@ namespace Azure { namespace Core { namespace Cryptography {
      * calculation.
      * @param length The size of the data provided.
      */
-    virtual void OnAppend(const uint8_t* data, std::size_t length) = 0;
+    virtual void OnAppend(const uint8_t* data, size_t length) = 0;
 
     /**
      * @brief Computes the hash value of the specified binary input data, including any previously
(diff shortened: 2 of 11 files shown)


antkmsft:
I think we should go the other way, and to use C++ types and not C types. We should replace c-style headers like `#include <stdint.h>` with `#include <cstdint>`, refer to size_t as `std::size_t`, and (u)intN_t as `std::_(u)intN_t`.

antkmsft:
BTW, for the includes, we should do that foar all stdlib includes - i.e. `#include <cstdio>` instead of `stdio.h`, and if there are more, do the same for them as well.

ahsonkhan:
> I think we should go the other way, and to use C++ types and not C types. We should replace c-style headers like `#include <stdint.h>` with `#include <cstdint>`, refer to size_t as `std::size_t`, and (u)intN_t as `std::_(u)intN_t`.

We had this discussion in other threads previously, we have converged moreso towards size_t instead, so this is just completing the few leftover places. I prefer not to revert the direction at this point, doesn't seem worth the effort for a customer impact perspective.
Feel free to read the previous discussion starting from https://github.com/Azure/azure-sdk-for-cpp/pull/2285#pullrequestreview-661558933

> BTW, for the includes, we should do that foar all stdlib includes - i.e. `#include <cstdio>` instead of `stdio.h`, and if there are more, do the same for them as well.

That's a separate issue, I have no strong opinion either way, up to you. From previous discussions with C++ folks, there is no clear consensus, so to me, this issue/fix doesn't meet the bar to spend a lot of time on, we can work on more impactful features and bug fixes :)

Jinming-Hu (approved):

Jinming-Hu:
> > I think we should go the other way, and to use C++ types and not C types. We should replace c-style headers like `#include <stdint.h>` with `#include <cstdint>`, refer to size_t as `std::size_t`, and (u)intN_t as `std::_(u)intN_t`.
> 
> We had this discussion in other threads previously, we have converged moreso towards size_t instead, so this is just completing the few leftover places. I prefer not to revert the direction at this point, doesn't seem worth the effort for a customer impact perspective.
> Feel free to read the previous discussion starting from [#2285 (review)](https://github.com/Azure/azure-sdk-for-cpp/pull/2285#pullrequestreview-661558933)
> 
> > BTW, for the includes, we should do that foar all stdlib includes - i.e. `#include <cstdio>` instead of `stdio.h`, and if there are more, do the same for them as well.
> 
> That's a separate issue, I have no strong opinion either way, up to you. From previous discussions with C++ folks, there is no clear consensus, so to me, this issue/fix doesn't meet the bar to spend a lot of time on, we can work on more impactful features and bug fixes :)

Yes, this is a separate issue. And I'd vote for `<cstdio>`.

vhvb1989 (approved):

PR #2516, opened 2021-06-30
Rename KeyOperation to KeyOperationType to avoid confusion with LRO

### LINKED ISSUE #2507: Rename KeyOperation to KeyOperationType to avoid confusion with LRO `Operation<T>` types
From https://apiview.dev/Assemblies/Review/2b96c2a3509844fcaf0106f1af1d43f1#002e5a4dd371402b875d3e7124a79b9a

cc @vhvb1989, @RickWinter

gearama:
closes #2507
# Pull Request Checklist

- [ ] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [ ] Comments in source
- [X] No typos
- [ ] Update changelog
- [X] Not work-in-progress
- [X] Self review of PR done
- [ ] Any breaking changes?


antkmsft (approved):

RickWinter (changes requested):
@ahsonkhan I believe this came out of the archboard review.  However I'm questioning if the right decision was reached in that group.  I've looked at 4 other SDKs and they are calling this KeyOperation(s).
I know its nice to have Operation suffix always being related to Operation.  In this case the name is what it is b/c it maps to the json that is expected.  Changing the name on this value could make the LLC->HLC grow up story confusing.

heaths (approved):

ahsonkhan:
> In this case the name is what it is b/c it maps to the json that is expected.

The JSON calls it key_ops:
https://docs.microsoft.com/en-us/rest/api/keyvault/update-key/update-key#request-body

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_operation_type.hpp
```diff
@@ -19,30 +19,30 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Keys {
    * @brief An operation that can be performed with the key.
    *
    */
-  class KeyOperation final {
+  class KeyOperationType final {
```
ahsonkhan, line 22: @gearama This type rename needs a change log entry since it is a breaking change.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_operation_type.hpp
```diff
@@ -19,30 +19,30 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Keys {
...
-    explicit KeyOperation(std::string operation) : m_operation(std::move(operation)) {}
+    explicit KeyOperationType(std::string operation) : m_operation(std::move(operation)) {}
 
     /**
-     * @brief Construct a default Key operation.
+     * @brief Construct a default Key operation type.
      *
      */
-    KeyOperation() = default;
+    KeyOperationType() = default;
```
ahsonkhan, line 38: I don't feel strongly about this given what other languages have already done with the type name, but also want to account for the archboard feedback, so we have three good options (in order of my preference, but I would be happy with any):
1) Change to `KeyOperations`.
2) Merge this PR as is, using `KeyOperationType`.
2) Close this PR.

One thing we should make sure though, is the subsequent language SDKs (like GoLang/etc.) follow the "new" naming pattern, using this as precedence, rather than having a "third" naming approach.

@tg-msft, @heaths, @jhendrixMSFT - what's the plan on naming for future SDKs? Joel, what would GoLang do for the name of such a type. Does it also use "Operation" suffix for LROs?

jhendrixMSFT: Can you link to the corresponding swagger?

gearama: switching to option 1 is problematic as KeyOperations with this exact capitalization appears in several other places in the code. 
We would need to do more refactoring than a simple name change / string replace to achieve our goal. 
Thus we can push this to the next release, or take the change as it is.

gearama: @ahsonkhan , @jhendrixMSFT what is the swagger doc and where it is located ?

heaths: See https://github.com/Azure/azure-rest-api-specs/blob/c234651f2885c07ff4a77e50c883145949fda5fa/specification/keyvault/data-plane/Microsoft.KeyVault/stable/7.2/keys.json#L1378-L1397.

I also don't like `KeyOperations` because it's not a flags-style enum. `key_ops` is an array of individual `KeyOperation` types. Adding the "Type" suffix is not uncommon. You see it in .NET quite often when not having the "Type" suffix would be vague or ambiguous.

jhendrixMSFT: Go honors the swagger, so the name would be `JsonWebKeyOperation`.

gearama:
Closing this PR as we decided to leave things as is as it is not actually causing any issues.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_operation_type.hpp
ahsonkhan: > switching to option 1 is problematic as KeyOperations with this exact capitalization appears in several other places in the code.

Ah yes, good point @gearama, in C++ we can't duplicate type and field/method name (at least that is what I remember from a previous discussion):
https://github.com/Azure/azure-sdk-for-cpp/blob/fda88bf7b8389575b9d973821efc604442e10493/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_create_options.hpp#L42
https://github.com/Azure/azure-sdk-for-cpp/blob/fda88bf7b8389575b9d973821efc604442e10493/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_vault_key.hpp#L85

https://godbolt.org/z/8fjxn487j
```C++
#include <vector>

struct Foo
{

};

struct Bar
{
 std::vector<Foo> Foo;
 //bool Foo();
};
```

@CaseyCarter can correct me if I am mistaken.

PR #2524, opened 2021-06-30
Move POSIX time converter from KeyVault to Core

### LINKED ISSUE #2503: UnixTimeConverter is a common datetime utility that isn't specific to KeyVault
UnixTimeConverter to/from DateTime is a very common transformation. It should be in Azure Core since we know future services will need it.

It is currently internal within KeyVault Common. If we are happy with the design, we should consider making it public and moving it to core. Otherwise, even if it remains internal, the right place for it is Azure Core (in .NET, it is provided by the layer below, i.e. the underlying language framework).

https://github.com/Azure/azure-sdk-for-cpp/blob/0874e1c1c9047b9ccf6caac5057f67c9b92144a7/sdk/keyvault/azure-security-keyvault-common/inc/azure/keyvault/common/internal/unix_time_helper.hpp#L15-L21

https://github.com/Azure/azure-sdk-for-net/search?p=1&q=FromUnixTimeSeconds
Attestation, ContainerRegistry, ACS, FormRecognizer, etc.

cc @RickWinter, @JeffreyRichter, @vhvb1989

antkmsft:
Closes #2503


sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -267,4 +267,50 @@ inline bool operator>=(std::chrono::system_clock::time_point const& tp, DateTime
...
+    PosixTimeConverter() = delete;
+
+    /**
+     * @brief An instance of `%PosixTimeConverter` class cannot be destructed, because no instance
+     * can be created.
+     *
+     */
+    ~PosixTimeConverter() = delete;
+  };
+} // namespace _internal
```
ahsonkhan, line 314: nit: I think this is the first time we are adding `::_internal` to the root Azure namespace.

Is it possible to make this type internal but nested within `DateTime` or some other namespace to reduce noise?

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -267,4 +267,50 @@ inline bool operator>=(std::chrono::system_clock::time_point const& tp, DateTime
...
+     *
+     */
+    PosixTimeConverter() = delete;
+
+    /**
+     * @brief An instance of `%PosixTimeConverter` class cannot be destructed, because no instance
+     * can be created.
+     *
+     */
+    ~PosixTimeConverter() = delete;
```
ahsonkhan, line 312: Other than the rename from Unix -> Posix and adding these private ctor/dtor deletions, was anything else changed that is worth a review?

ahsonkhan (approved):

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: No.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: Unfortunately, we can't do a nested internal class.
We can move it to `Azure::Core::_internal`. I have no strong preference. If you do - let me know, otherwise we have to decide on something, and current code will go in.

On the other hand, `_internal` should not clutter things. I think it is ok.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: * I rewritten `DateTimeToPosixTime()` to not use temporary variable.
* Doxy comment updates (nothing big, just touchups)

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -267,4 +267,50 @@ inline bool operator>=(std::chrono::system_clock::time_point const& tp, DateTime
   return (dt <= tp);
 }
 
+namespace _internal {
```
RickWinter, line 270: I was expecting this be Azure::Core::_internal.

antkmsft: Sounds good - I'll moved it to `Core`, and once CI succeeds, I am checking it in.

PR #2534, opened 2021-07-01
Update readme headers retroactively

antkmsft:
I don't see why we can't update the document "retroactively".


ahsonkhan:
The changelogs have already been published as part of the release notes, tags, and published documents.
I recommend reducing the change surface area and only modifying the headings **going forward**. There might be links out in the wild to existing changelog headings which will get broken now if we change headings like this.

There is no real value to do this retroactively, not worth the risk, imo.

antkmsft:
Good point about the link, although not likely.

PR #2537, opened 2021-07-01
Consider making the RFE ctor as an internal factory method.

ahsonkhan:
Currently, the public RFE ctors look weird and it is hard to discern which parameter is optional:
```C++
explicit RequestFailedException(std::string const& message);
explicit RequestFailedException( const std::string& message, std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);

// newly added
explicit RequestFailedException(std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse);
```

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -147,7 +147,7 @@ namespace Azure { namespace Core {
     ~RequestFailedException() = default;
 
   private:
-    std::string getRawResponseField(
+    std::string GetRawResponseField(
         std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse,
         std::string fieldName);
   };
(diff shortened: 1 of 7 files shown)


sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -145,10 +133,23 @@ namespace Azure { namespace Core {
...
     ~RequestFailedException() = default;
-
-  private:
-    std::string GetRawResponseField(
-        std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse,
-        std::string fieldName);
   };
+
+  namespace _internal {
+    class ExceptionFactory final {
```
ahsonkhan, line 139: @JeffreyRichter what do you think about this name and approach rather than a public ctor?

ahsonkhan:
Superseded by https://github.com/Azure/azure-sdk-for-cpp/pull/2539

PR #2539, opened 2021-07-01
Make sure RFE Message is only set to the response body payload.

ahsonkhan:
Follow up from offline discussion on the right RFE semantics.

### DIFF
--- a/sdk/core/azure-core/src/exception.cpp
+++ b/sdk/core/azure-core/src/exception.cpp
@@ -32,17 +32,21 @@ namespace Azure { namespace Core {
 
   RequestFailedException::RequestFailedException(
       std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse)
-      : std::runtime_error(GetRawResponseField(rawResponse, "message"))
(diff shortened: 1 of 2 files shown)


JeffreyRichter (approved):

sdk/core/azure-core/src/exception.cpp
```diff
@@ -32,17 +32,21 @@ namespace Azure { namespace Core {
...
-    RequestId = HttpShared::GetHeaderOrEmptyString(headers, HttpShared::MsRequestId);
-    ClientRequestId = HttpShared::GetHeaderOrEmptyString(headers, HttpShared::MsClientRequestId);
-    Message = this->what();
     RawResponse = std::move(rawResponse);
+
+    // The response body may or may not have these fields
+    ErrorCode = GetRawResponseField(RawResponse, "code");
+    Message = GetRawResponseField(RawResponse, "message");
+
+    ClientRequestId = HttpShared::GetHeaderOrEmptyString(headers, HttpShared::MsClientRequestId);
```
ahsonkhan, line 48: Re-ordered them as part of code hygiene (not correctness)

Issue #2513, opened 2021-06-29
Remove Sha256 from storage code


Jinming-Hu:
https://github.com/Azure/azure-sdk-for-cpp/blob/fda88bf7b8389575b9d973821efc604442e10493/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp#L40

Jinming-Hu:
blocked by https://github.com/Azure/azure-sdk-for-cpp/issues/2500

ahsonkhan:
> blocked by #2500

@Jinming-Hu what do you mean by this? Is removing Sha256 from storage not feasible yet?

Jinming-Hu:
> > blocked by #2500
> 
> @Jinming-Hu what do you mean by this? Is removing Sha256 from storage not feasible yet?

It's being used in storage test code.

I'd have to move the implementation of `Sha256` from storage common package to test code. Then after you guys move `Sha256` from keyvault to core (as we talked this morning in the working group meeting), I'd remove `Sha256()` from storage test code and use the one in core package. That sounds like a lot of hassle.

Why not wait until you've moved SHA-256 to core package, then I can use the it in storage test code and remove the one in storage common package.

ahsonkhan:
Before we do this, let's make sure we validate that a test dependency change from storage to core doesn't require storage vcpkg package to take a dependency on a higher azure-core version (and cause a vcpkg CI failure).

It might be worth moving the storage impl to the tests for now, or wait until after the next release, so we have enough time to validate this.

Jinming-Hu:
I'd prefer waiting.

ahsonkhan:
> Before we do this, let's make sure we validate that a test dependency change from storage to core doesn't require storage vcpkg package to take a dependency on a higher azure-core version (and cause a vcpkg CI failure).

@Jinming-Hu were you able to verify this?

Jinming-Hu:
This sha256 function is only used in test code, and `vcpkg install` won't build test code at all. So I think it should be fine.

PR #2585, opened 2021-07-09
Remove sha256 hash from storage internal code

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/2513

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
@@ -37,7 +37,6 @@ namespace Azure { namespace Storage {
   };
 
   namespace _internal {
-    std::vector<uint8_t> Sha256(const std::vector<uint8_t>& data);
     std::vector<uint8_t> HmacSha256(
         const std::vector<uint8_t>& data,
         const std::vector<uint8_t>& key);
--- a/sdk/storage/azure-storage-common/src/crypt.cpp
+++ b/sdk/storage/azure-storage-common/src/crypt.cpp
@@ -85,7 +85,6 @@ namespace Azure { namespace Storage {
     enum class AlgorithmType
     {
       HmacSha256,
-      Sha256,
     };
 
     struct AlgorithmProviderInstance final
@@ -97,7 +96,7 @@ namespace Azure { namespace Storage {
       AlgorithmProviderInstance(AlgorithmType type)
       {
         const wchar_t* algorithmId = nullptr;
-        if (type == AlgorithmType::HmacSha256 || type == AlgorithmType::Sha256)
+        if (type == AlgorithmType::HmacSha256)
         {
           algorithmId = BCRYPT_SHA256_ALGORITHM;
         }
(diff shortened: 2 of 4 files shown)


katmsft (approved):

sdk/storage/azure-storage-common/test/crypt_functions_test.cpp
```diff
@@ -15,16 +15,6 @@ namespace Azure { namespace Storage { namespace Test {
...
 
-  TEST(CryptFunctionsTest, Sha256)
-  {
-    EXPECT_EQ(
-        Azure::Core::Convert::Base64Encode(_internal::Sha256(ToBinaryVector(""))),
-        "47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=");
-    EXPECT_EQ(
-        Azure::Core::Convert::Base64Encode(_internal::Sha256(ToBinaryVector("Hello Azure!"))),
-        "Mjzwx2mqGHb9FSgjm33ShNmXYndkgvwA6tQmEiskOHg=");
-  }
```
ahsonkhan, line 26: Do we have tests and are we confident that the `_internal::Sha256` that is being removed and the new `Azure::Core::Cryptography::_internal::Sha256Hash` being used instead, have the same semantics and behavior?

I recommend keeping this test case, and just porting it to use the new one.

sdk/storage/azure-storage-common/inc/azure/storage/common/crypt.hpp
```diff
@@ -37,7 +37,6 @@ namespace Azure { namespace Storage {
   };
 
   namespace _internal {
-    std::vector<uint8_t> Sha256(const std::vector<uint8_t>& data);
```
ahsonkhan, line 40: In general, making breaking changes to `_internal` is not a good idea as it will inevitably break versioning between dependency updates, even if the user only calls and uses `public` surface area. We should therefore carefully review such changes from a customer impact perspective.

This is a rare example of a removal/breaking change to _internal that happens to be OK, only because:
- When the API **originally** shipped it had no callers within the SDK product code, and continues not to have any (across any version), and
- The API and type is more or less in a leaf node within the dependency graph, which has the added benefit of making the audit of usages easier.

It serves as a good example for how we can go about it :)

FYI @JeffreyRichter, @RickWinter

sdk/storage/azure-storage-common/src/crypt.cpp
```diff
@@ -97,7 +96,7 @@ namespace Azure { namespace Storage {
       AlgorithmProviderInstance(AlgorithmType type)
       {
         const wchar_t* algorithmId = nullptr;
-        if (type == AlgorithmType::HmacSha256 || type == AlgorithmType::Sha256)
```
ahsonkhan, line 100: nit: With just one such type of hashing algorithm now, we could simplify the implementation a bit and remove `AlgorithmType`, but it is fine to keep that for now, if we expect new algorithms to get added and supported later.

sdk/storage/azure-storage-common/test/crypt_functions_test.cpp
Jinming-Hu: I don't think that'd be necessary. SHA-256 in very standardized. It'd be horrible if we compute hash with two implementations and get different results.

PR #2596, opened 2021-07-09
adding backupKey as return type for backup key

vhvb1989:


sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/backup_key.hpp
```diff
@@ -0,0 +1,44 @@
...
+     * @param backupKey The backup key array data.
+     */
+    explicit BackupKey(std::vector<uint8_t> backupKey) : m_value(std::move(backupKey)) {}
+
+    /**
+     * @brief Return the backup key.
+     *
+     */
+    std::vector<uint8_t> const& GetValue() const { return m_value; }
+  };
```
ahsonkhan, line 42: @heaths and @tg-msft we decided to create a "model type" wrapping the `std::vector<uint8_t>` return type from the `BackupKey` operation to future proof that API in case the service adds more things to it.

https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/keyvault/Azure.Security.KeyVault.Keys/src/KeyClient.cs#L883

We believe this is a good change following the C++ and general SDK guidelines around return types, since we can't overload based on that, in the future, and we value language specific consistency over consistency across languages. If you have any thoughts, on the contrary, please let me know :)

Similar to:
https://github.com/Azure/azure-sdk-for-net/blob/7e0315a4c606a504088476d00e335490e72c6fcc/sdk/keyvault/Microsoft.Azure.KeyVault/src/Generated/Models/BackupKeyResult.cs#L21

Essentially, we are leveraging the benefit of hindsight that we happen to have in C++.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_client.hpp
```diff
@@ -341,7 +342,7 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Keys {
      * @param name The name of the key.
      * @param context A #Azure::Core::Context controlling the request lifetime.
      */
-    Azure::Response<std::vector<uint8_t>> BackupKey(
+    Azure::Response<Azure::Security::KeyVault::Keys::BackupKey> BackupKey(
```
ahsonkhan, line 345: @heaths and @tg-msft I think `BackupKey` is probably the right name here, how come .NET decided to call it `BackupKeyResult`?

Trying to page-back and rationalize the model type naming table again:
https://azure.github.io/azure-sdk/general_design.html#model-types

image

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/backup_key.hpp
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param backupKey The backup key array data.
+     */
+    explicit BackupKey(std::vector<uint8_t> backupKey) : m_value(std::move(backupKey)) {}
+
+    /**
+     * @brief Return the backup key.
+     *
+     */
+    std::vector<uint8_t> const& GetValue() const { return m_value; }
```
ahsonkhan, line 41: This should follow a general "model struct" pattern, with public fields, similar to options bag, rather then the private field with a `GetValue()` method.

Just like to `struct Azure::Storage::Blobs::Models::BlobProperties`:
https://github.com/Azure/azure-sdk-for-cpp/blob/07861a9687449ec360299a90a0b824c0eeab0903/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/protocol/blob_rest_client.hpp#L1321-L1335

But then, because we'd have a naming conflict between BackupKey the byte array field, and the type, we may want to call the type BackupKeyResult (or name the field differently, based on what the rest API/JSON calls it).
```suggestion
  /**
   * The backup key array data.
   */
  struct BackupKeyResult final
  {
    std::vector<uint8_t> BackupKey;
```

cc @JeffreyRichter

ahsonkhan (changes requested):
The returned model types should be simple struct data bags, similar to struct options, which makes extending them easy, and simpler for the customer to use.

```C++
  struct BackupKeyResult final
  {
  /**
   * The backup key array data.
   */
    std::vector<uint8_t> BackupKey;
  }
```

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/backup_key.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <cstdint>
+#include <vector>
+
+namespace Azure { namespace Security { namespace KeyVault { namespace Keys {
+
+  /**
+   * @brief The JsonWebKey types.
+   *
+   */
+  class BackupKey final {
```
ahsonkhan, line 23: The brief doc string for this type doesn't match.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/backup_key.hpp
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ * @brief Defines the backupKey type.
```
ahsonkhan, line 6:
```suggestion
 * @brief Defines the BackupKey model type.
```

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_client.hpp
heaths: We don't - we return a `byte[]` array. Why are you returning something else that the other SDKs don't do?

* https://github.com/Azure/azure-sdk-for-net/blob/c216595c466f15518f2ad8b6733841d3c8009370/sdk/keyvault/Azure.Security.KeyVault.Keys/src/KeyClient.cs#L883
* https://github.com/Azure/azure-sdk-for-java/blob/0747a4129d3ae48fe2a8f54070a79fb3faf16913/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/KeyClient.java#L656
* https://github.com/Azure/azure-sdk-for-python/blob/e75c11b16477185b344cb8cd672253c88a587a98/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py#L511
* https://github.com/Azure/azure-sdk-for-js/blob/9656690691ac5aa3fd17eb3e334d9bceba51efc3/sdk/keyvault/keyvault-keys/src/index.ts#L623

heaths (changes requested):
Should just return a byte array as you already were doing. All the other languages do.

ahsonkhan:
> Should just return a byte array as you already were doing. All the other languages do.

Not returning a struct wrapping model type doesn't future proof the API for if (or when) the service adds more fields to the response JSON, which isn't considered a breaking change from the service perspective. We can't overload on return types, so what would we do in such a case in other languages? This allows us to version without having to workaround things or make breaking changes, at least that is the primary motivation here.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_client.hpp
ahsonkhan: What about https://github.com/Azure/azure-sdk-for-net/search?q=BackupKeyResult?

heaths:
Java nor C# (.NET) can overload on return types. If that happens, and we're told that's incredibly unlikely, we'd add a new method. There's no idiomatic reason to differ here, so I think remaining consistent with all the other tier 1 languages makes sense.

ahsonkhan:
> There's no idiomatic reason to differ here, so I think remaining consistent with all the other tier 1 languages makes sense.

I imagined local consistency with a language SDK (like C++) would be better.

Do we continue to add APIs that return a single "primitive" type in newer SDKs in .NET rather than wrapping them in a model type? Wouldn't auto-generated SDKs always generate a model type?

Thought the guidelines allow "unstructured" payload return types, I thought this was recommended (i.e. if we had hindsight, we'd re-do it):
✅ DO return a model type if the content has a schema and can be deserialized.
https://azure.github.io/azure-sdk/cpp_introduction.html#cpp-design-client-return-model-type

heaths:
Key Vault predates many of those guidelines (and helped establish a few, like extensible enums). Given the service team's doubt they'd change the return type, we decided to expose the byte array directly rather than make the caller go through extra steps.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/key_client.hpp
ahsonkhan: From offline discussion, that is from the old track 1 SDK, so ignore it.

heaths (approved):
Per offline conversation, approving.

ahsonkhan:
Regarding naming of types, ideally we'd want something like this, but C++ doesn't allow it:

```C++
Azure::Response<BackupKey> BackupKey(...);

struct BackupKey
{
  std::vector<uint8_t> BackupKey;
}
```

Which of these two do folks prefer (I am leaning towards 1)? @heaths?

**Options 1:**

```C++
Azure::Response<BackupKeyResult> BackupKey(...);

struct BackupKeyResult
{
  std::vector<uint8_t> BackupKey;
}
```

**OR Option 2:**

```C++
Azure::Response<BackupKey> BackupKey(...);

struct BackupKey
{
  std::vector<uint8_t> Key;
}
```

ahsonkhan:
Superseded by https://github.com/Azure/azure-sdk-for-cpp/pull/2619

Issue #2599, opened 2021-07-12
Remove the MaxPageResults options on KeyVault GetPropertiesOfKeys and others for now, unless a user scenario requries it


ahsonkhan:
Remove this:
https://github.com/Azure/azure-sdk-for-cpp/blob/46e333d32d71966d51b3deef23dcbda2945c417b/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/list_keys_result.hpp#L128

And leave out the query parameter within the request:
https://github.com/Azure/azure-sdk-for-cpp/blob/07861a9687449ec360299a90a0b824c0eeab0903/sdk/keyvault/azure-security-keyvault-keys/src/key_client.cpp#L54

cc @heaths, @vhvb1989

PR #2620, opened 2021-07-14
Removed the `MaxPageResults` field from `GetPropertiesOfKeysOptions`, `GetPropertiesOfKeyVersionsOptions`, and `GetDeletedKeysOptions`.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2599

### DIFF
--- a/sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
+++ b/sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Removed access to `Iv` field member from `EncryptParameters` and `DecryptParameters`.
 - Removed `Encrypt(EncryptionAlgorithm, std::vector, context)`.
 - Removed `Decrypt(DecryptAlgorithm, std::vector, context)`.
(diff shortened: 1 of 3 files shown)


RickWinter:
What is the motivation for not including this?  Its valid and would be there if codegen was creating the package

heaths:
@RickWinter we decided not to include because:

1. The service max is already so low: 25 max.
2. The actual page size can vary from 0 (with subsequent pages still) to maxresults and we though it would leave to customer confusion: "I asked for 5 results per page, but you gave me back only 2. What gives?"

As such, we felt it provided little to no value given the payload size of each element is pretty small. JWKs (keys) are the largest and even they aren't that big.

heaths (approved):

ahsonkhan:
@RickWinter, we discussed this offline. @heaths captured the rationale from that meeting well.

Reference REST API:
https://docs.microsoft.com/en-us/rest/api/keyvault/get-secrets/get-secrets
image

We can always add it back if needed by a customer scenario.

Issue #2624, opened 2021-07-14
Fix the namespace and header path mismatch for _detail vs /internal/


ahsonkhan:
The file is publicly accessible within `/internal/` but the namespace is `_detail`. If this header doesn't need to be public, move it to the `/src/` folder under `/private/`. If it is truly `/internal/` (which it shouldn't given the KeyVault keys package is a leaf node), then change the namespace to reflect that (`_internal`).

https://github.com/Azure/azure-sdk-for-cpp/blob/5ddebe53bf0ad2220ed09d5ef889d4b95fc7e733/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/internal/cryptography/local_cryptography_provider.hpp#L18-L25

cc @vhvb1989, @RickWinter

PR #2625, opened 2021-07-14
Removed the /internal/ directory in Keyvault Keys and move headers to private that don't need to be public.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2624


sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/cryptography/cryptography_client.hpp
```diff
@@ -52,18 +52,6 @@ namespace Azure {
...
-     * @brief Provides a #CryptographyProvider that performs operations in the Key Vault Keys
-     * Server.
-     *
-     * @return A cryptographic client to perform operations on the server.
-     */
-    std::shared_ptr<Azure::Security::KeyVault::Keys::Cryptography::_detail::CryptographyProvider>
-    RemoteClient() const
-    {
-      return m_remoteProvider;
-    }
```
ahsonkhan, line 65: This wasn't used by anything.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/cryptography/cryptography_client.hpp
```diff
@@ -14,12 +14,12 @@
 #include <azure/core/io/body_stream.hpp>
 
 #include "azure/keyvault/keys/cryptography/cryptography_client_options.hpp"
+#include "azure/keyvault/keys/cryptography/cryptography_provider.hpp"
```
ahsonkhan, line 17: I wish I could make these headers private as well, by forward declaration, similar to how @antkmsft did it in identity, but due to the inheritance relationship between `CryptographyProvider` and `RemoteCryptographyClient`, it was trickier :(

https://github.com/Azure/azure-sdk-for-cpp/blob/5ddebe53bf0ad2220ed09d5ef889d4b95fc7e733/sdk/identity/azure-identity/inc/azure/identity/managed_identity_credential.hpp#L18-L20

The following public header has these private fields that forces me to keep the headers public, that I wanted to hide:
https://github.com/Azure/azure-sdk-for-cpp/blob/5ddebe53bf0ad2220ed09d5ef889d4b95fc7e733/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/cryptography/cryptography_client.hpp#L37-L45

I want to do the following (and remove the header includes), but then I get builds errors:
  namespace _detail {
    struct CryptographyProvider;
    struct RemoteCryptographyClient;
  } // namespace _detail

image

@antkmsft or @CaseyCarter, any suggestions?

RickWinter (approved):

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/cryptography/cryptography_client.hpp
antkmsft: Ahson, I think you can make work if you remove the headers, but you'll need to move destructor from the header (right now, it is "in the header" - it is default-generated, but it is in header). You need to move the destructor from header to source. See what I did in the identity.

```cpp
// this is internal header. Move it to src
namespace Azure { namespace _internal {
  class InternalClass
  {
  public:
    void f();
  };
}}
```

```cpp
// this is public header
namespace Azure
{
  namespace _internal
  {
    class InternalClass; // this is just a forward declaration
  }

  class PublicClass
  {
  private:
    std::unique_ptr<_internal::InternalClass> m_internalClass;
  public:
    ~PublicClass(); // you need this
    // note that we only declare the destructor, but do not define it in the header
  }
}
```

```cpp
// this is PublicClass.cpp

#inlcude <azure/internal_class.hpp> // <-- we did not include it in the header, but we DO include it here

PublicClass::~PublicClass() = default;
```

What is happening:
1. If you don't declare destructor, it still gets generated automatically, and gets inserted into the code that works with the class.
2. In order to generate the destructor, it needs to know, who to destruct InternalClass. And for that, it needs to have full definition of a class, only the forward declaration isn't enough.
3. What we do: we declare destructor, but we define it in the cpp file. In cpp file, we make sure that the destructor has all the knowledge it needs about the InternalClass.
4. The client code does no longer needs to generate the destructor. It has enough to invoke it - the destructor declaration. So, the client code invokes it as any other function that sits in the cpp file.

I.e. it is the same principle as if you move an inline function from the header to the source: the header no longer needs to include all the headers in order to make that inline function work.

antkmsft: @ahsonkhan, I managed to move these headers to private, and all compiles: https://github.com/ahsonkhan/azure-sdk-for-cpp/pull/1

It wasn't only the things that I've told - `cryptography_client.hpp` misses a few headers that is should be including, i.e. it includes `wrap_result`, but does not include `unwrap_result` for some reason. But it should. That was a second source of compilation errors.

As you look into the PR I created in your repo, most of the changes are formatting changes due to changed indentation due to forward declarations, the rest I did there is just as I told. I clearly did not have enough time to make as many non-formatting changes as the  diff shows at first glance :)

You can make sure for yourself, and if you approve the PR in your repo, it will make it as a commit to this PR.

ahsonkhan: @antkmsft I imagine your approach is better, but in the interest of time, let me ask two pertinent questions:
1) Is there anything functionally wrong or broken with the current approach in the PR?
2) Is there anything preventing us to move from the current approach to yours in the future, without making any breaking changes?

If the answer to both is no, I prefer merging this PR as is, and make the improvements in the future. Let me know!

ahsonkhan: OK, I understand what you did. This works really well! Thanks.

ahsonkhan: Remind me to buy you lunch when we meet :)

antkmsft: No problem! I am glad we went with the change.

As of your question earlier, I do not think it is critical to get it now, just a very nice to have. But what I'd look on is the fact that we never will be able to remove a public header, and also if including `X.hpp` was bringing declarations of public classes `X`, `Y`, and `Z` in version 1.0 GA, (because `X.hpp` was including `Y.hpp` and `Z.hpp`), we'll have to keep that behavior forever, even if `X.hpp` does not require `Z`: including `X.hpp` will have to bring `Z` declaration forever. As long as we are fine in that aspect, moving headers to internal can be done in future version.

But I am glad we've went with the change - it is pretty small, easy provable (in short, either builds or not, with a few nuances), and sometimes I think it is easier to make something than to ensure it is possible to safely do the planned work in the future.

ahsonkhan: Great points. I agree!

PR #2626, opened 2021-07-14
Use forward declaration to avoid exposing private/src headers from public headers.

ahsonkhan:
Follow up from https://github.com/Azure/azure-sdk-for-cpp/pull/2618

**The relevant commit to review is:** https://github.com/Azure/azure-sdk-for-cpp/commit/46977110a1c954dec07aa7a7ca0b0edfb2f6e11d

I had to forward declare both, to avoid the cycle and the circular dependency error: can't delete an incomplete type.
https://stackoverflow.com/questions/44633726/circular-dependency-cant-delete-an-incomplete-type

@CaseyCarter, is this safe (and "correct") to do?


sdk/keyvault/azure-security-keyvault-keys/src/private/keyvault_protocol.hpp
```diff
@@ -21,6 +21,10 @@
 #include <string>
 #include <vector>
 
+namespace Azure { namespace Security { namespace KeyVault { namespace Keys {
+  class KeyClient;
+}}}} // namespace Azure::Security::KeyVault::Keys
```
ahsonkhan, line 26: @antkmsft maybe you have better insights on how to manage this cyclical dependency, similar to https://github.com/Azure/azure-sdk-for-cpp/pull/2625#discussion_r669405309

gearama (approved):

sdk/keyvault/azure-security-keyvault-keys/src/private/keyvault_protocol.hpp
ahsonkhan: Discussed with @CaseyCarter offline, both patterns are OK.

PR #2636, opened 2021-07-15
Use std::vector instead of std::list for KeyOperation arrays.

ahsonkhan:
Addressing the API review feedback:
https://apiview.dev/Assemblies/Review/da8d0ecdf87c49238b75fee069413029#3508f13ee9d3483ea2e4223592ed08df


sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Updated `BackupKey()` API return type to `BackupKeyResult` model type.
 - Renamed `KeyPropertiesPageResult` to `KeyPropertiesPagedResponse`.
 - Renamed `DeletedKeyPageResult` to `DeletedKeyPagedResponse`.
+- Changed the container for `KeyOperations` from `std::list` to `std::vector` within `CreateKeyOptions` and `UpdateKeyProperties()`.
```
heaths, line 24: Hard to tell from the limited PR view, but do you support ImportKeyOptions as well? I'd also expect to see this change for the `JsonWebKey` and anywhere else key_ops is used.

heaths (changes requested):
Unless you don't have the full surface area for Keys (except Crypto) that .NET does, this seems to be missing a few changes.

sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
ahsonkhan: We have `ImportKeyOptions` and `JsonWebKey`, yes, but I didn't see any usage of list there.
https://github.com/Azure/azure-sdk-for-cpp/blob/90bdfd840832d04cb6572803dcf47543ab8137e2/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/import_key_options.hpp#L28

`JsonWebKey` already uses `std::vector`, so that's why we didn't need to fix it.
https://github.com/Azure/azure-sdk-for-cpp/blob/90bdfd840832d04cb6572803dcf47543ab8137e2/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/json_web_key.hpp#L63

Any others come to mind?

I did a `std::list` regex search across the source code, and I didn't find any other instances of it during the first pass. I can try doing that again, to double check.

ahsonkhan:
> Unless you don't have the full surface area for Keys (except Crypto) that .NET does, this seems to be missing a few changes.

Looks like I caught them all, or they were already `std::vector`.

image

sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
ahsonkhan: Where's the list in `ImportKeyOptions`?
https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.keys.importkeyoptions?view=azure-dotnet

ahsonkhan: The rest are already correct, so no need to mention in Changelog:
https://github.com/Azure/azure-sdk-for-cpp/blob/90bdfd840832d04cb6572803dcf47543ab8137e2/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/json_web_key.hpp#L53-L63

vhvb1989 (approved):

sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
heaths: I didn't mean to imply there *were* other instances. Wanted to make sure you checked to catch them all i.e., I'll leave that exercise for the code owner. I can approve now.

heaths (approved):

sdk/keyvault/azure-security-keyvault-keys/CHANGELOG.md
ahsonkhan: @vhvb1989 if you find any other place (that I somehow missed), let me know :)

Issue #2694, opened 2021-08-02
Documentation warnings break Werror build with recent clang


davidchisnall:
**Describe the bug**
Compiling with clang 11 (possibly other versions) fails with a number of warnings triggered from Wdocumentation.

This is particularly problematic because the vcpkg build enables Werror by default and so it is not possible to install from vcpkg on platforms using clang 11 as the compiler.

**To Reproduce**

Compile with clang 11.

**Expected behavior**

I believe Werror is valuable to have enabled by default for anyone building from source manually because it catches a lot of errors but installing via vcpkg should not fail as a result of developers using a different compiler to the CI systems.  

**Screenshots**

**Setup (please complete the following information):**
 - OS: FreeBSD
 - IDE : No
 - Version of the Library used: git head and whatever vcpkg currently uses.

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ X] Bug Description Added
- [ X] Repro Steps Added
- [ X] Setup information Added

PR #2697, opened 2021-08-02
clang doc fixes

vhvb1989:
fixes: https://github.com/Azure/azure-sdk-for-cpp/issues/2694


antkmsft (approved):
Thank you Victor!
Hmm, strange, I think at a time there used to be a Clang CI leg that would fail for documentation, and it would've failed for this kind of `@param` mismatch. Or was it in C repo?

Issue #2716, opened 2021-08-08
Fix cmake build with custom OpenSSL library


kshcherbatov:
**Describe the bug**
To compile dynamic Azure SDK libraries I link them against certain OpenSSL distributed with our product:

```
$ /snap/cmake/912/bin/cmake .. -DOPENSSL_ROOT_DIR=$TREASURE_ROOT/OpenSSL/1.1.1  -DOPENSSL_INCLUDE_DIR=$TREASURE_ROOT/OpenSSL/1.1.1/include -DOPENSSL_CRYPTO_LIBRARY=$TREASURE_ROOT/OpenSSL/1.1.1/a64_lnx/lib/libcrypto.so.1.1 -DOPENSSL_SSL_LIBRARY=$TREASURE_ROOT/OpenSSL/1.1.1/a64_lnx/lib/libssl.so.1.1 -DCMAKE_INSTALL_PREFIX=$TREASURE_ROOT/Azure/main/ -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DAZ_ALL_LIBRARIES=OFF

$ make clean
$ make -j5
```

```
In file included from /home/hadoop/Desktop/azure-sdk-for-cpp/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp:17:0,
                 from /home/hadoop/Desktop/azure-sdk-for-cpp/sdk/core/azure-core/inc/azure/core/internal/client_options.hpp:13,
                 from /home/hadoop/Desktop/azure-sdk-for-cpp/sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keys/cryptography/cryptography_client_options.hpp:13,          
                 from /home/hadoop/Desktop/azure-sdk-for-cpp/sdk/keyvault/azure-security-keyvault-keys/src/cryptography/cryptography_client_options.cpp:4:
/home/hadoop/Desktop/azure-sdk-for-cpp/sdk/core/azure-core/inc/azure/core/uuid.hpp:20:10: fatal error: openssl/rand.h: No such file or directory
 #include <openssl/rand.h> //for RAND_bytes
```

Here is a git format-patch output for the one-liner-fix proposed:
[0001-Fix-cmake-build-with-custom-OpenSSL-library.patch.txt](https://github.com/Azure/azure-sdk-for-cpp/files/6950464/0001-Fix-cmake-build-with-custom-OpenSSL-library.patch.txt)

[...]

Issue #2560, opened 2021-07-06
Core SDK should link OpenSSL as PUBLIC since it uses openssl functions in header files


Jinming-Hu:
https://github.com/Azure/azure-sdk-for-cpp/blob/cc1a7d22ce49543a9fd57d3b19b080952a23e787/sdk/core/azure-core/CMakeLists.txt#L145

This should be `target_link_libraries(azure-core PUBLIC OpenSSL::SSL)` because openSSL headers is used here in a header file.

https://github.com/Azure/azure-sdk-for-cpp/blob/cc1a7d22ce49543a9fd57d3b19b080952a23e787/sdk/core/azure-core/inc/azure/core/uuid.hpp#L20

This will lead into a build error if OpenSSL is not in the system include path. Like

```
In file included from ~/Desktop/azure-sdk-for-cpp/sdk/core/azure-core/test/ut/bearer_token_authentication_policy_test.cpp:4:
In file included from ~/Desktop/azure-sdk-for-cpp/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp:17:
~/Desktop/azure-sdk-for-cpp/sdk/core/azure-core/inc/azure/core/uuid.hpp:20:10: fatal error: 'openssl/rand.h' file not found
#include <openssl/rand.h> //for RAND_bytes
         ^~~~~~~~~~~~~~~~
4 warnings and 1 error generated.
make[2]: *** [sdk/core/azure-core/test/ut/CMakeFiles/azure-core-test.dir/bearer_token_authentication_policy_test.cpp.o] Error 1
make[1]: *** [sdk/core/azure-core/test/ut/CMakeFiles/azure-core-test.dir/all] Error 2
```

PR #2734, opened 2021-08-11
Fix build error on some platforms

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/2716
closes https://github.com/Azure/azure-sdk-for-cpp/issues/2560

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/uuid.hpp
+++ b/sdk/core/azure-core/inc/azure/core/uuid.hpp
@@ -11,14 +11,7 @@
 #include "azure/core/platform.hpp"
 
 #include <cstring>
-#include <new> // for placement new
-#include <random>
 #include <string>
-#include <utility> // for swap and move
-
-#if defined(AZ_PLATFORM_POSIX)
-#include <openssl/rand.h> //for RAND_bytes
-#endif
 
 namespace Azure { namespace Core {
   /**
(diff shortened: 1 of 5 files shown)


Jinming-Hu:
@vhvb1989 @antkmsft @RickWinter Can you please help review this PR?

antkmsft:
Do you have a repro? Can we check if this is because `uuid.hpp` includes an openssl header? The rest of the includes are in src files, and we've missed this one. The proper fix then, I think, is to move that part of a function into a non-inlineable function, which is implemented in cpp file and uses `RAND_bytes`.

Jinming-Hu:
> @Jinming-Hu Do you have a repro? Can we check if this is because `uuid.hpp` includes an openssl header? The rest of the includes are in src files, and we've missed this one. The proper fix then, I think, is to move that part of a function into a non-inlineable function, which is implemented in cpp file and uses `RAND_bytes`.

Done, please review it again.

sdk/core/azure-core/inc/azure/core/uuid.hpp
```diff
@@ -44,71 +37,12 @@ namespace Azure { namespace Core {
...
-      std::random_device rd;
-
-      for (size_t i = 0; i < UuidSize; i += 4)
-      {
-        const uint32_t x = rd();
-        std::memcpy(uuid + i, &x, 4);
-      }
-#elif defined(AZ_PLATFORM_POSIX)
-      // This static cast is safe since we know Uuid size, which is a const, will always fit an int.
-      int ret = RAND_bytes(uuid, static_cast<int>(UuidSize));
```
antkmsft, line 95: I was thinking of encapsulating this specific function call to a non-inline private function, but making the entire functions non-inlineable probably works too. Try to get @RickWinter to review, he wrote this.

sdk/core/azure-core/inc/azure/core/uuid.hpp
```diff
@@ -11,14 +11,7 @@
 #include "azure/core/platform.hpp"
 
 #include <cstring>
```
antkmsft, line 13: With this header as is, with all the function bodies moved to cpp, I don't think `#niclude <cstring>` (not `<string>`) is needed anymore, so should be moved to cpp file. `<cstring>` is for `std::snprintf()`, I think.

antkmsft (approved):
Ideally, if you move the entirety of functions to cpp (inline => non-inline), I'd recommend to get @RickWinter 's approval as he is the author. But if Rick does not get to see these changes, I think the changes are safe. But I think that if you encapsulate just the `RAND_bytes()` call, it is less of a change (not that it is better/worse)

sdk/core/azure-core/inc/azure/core/uuid.hpp
Jinming-Hu: `<cstdio>` is for `std::snprintf()`. `<cstring>` is for `std::memcpy`.

sdk/core/azure-core/inc/azure/core/uuid.hpp
Jinming-Hu: I don't think there's any difference. After moving the entire function body to `.cpp` file, the header file now looks more concise.

sdk/core/azure-core/inc/azure/core/uuid.hpp
antkmsft: Still, if we keep it as is, with just the function signatures, I think, memcpy is not used in the header, just in the cpp.

Jinming-Hu:
@RickWinter I moved the function body of uuid impl to `.cpp` file, so that we don't have to include openssl in header file, so as to fix #2716, #2560 these two issues.

Can you please review this PR?

sdk/core/azure-core/inc/azure/core/uuid.hpp
antkmsft: I can't disagree, it is only the perf that could be a bit different, although I don't believe it makes a difference, in this place.

antkmsft: If we only keep the class declaration here, I think we should move `#include <cstring>` from the header to the source.

Jinming-Hu: We still need `<cstring>` because we're using `std::memcpy` in the private constructor.

antkmsft: Ah, I see it now, thanks for clarification! GitHub hid that line, and I didn't see it. Sure then!

antkmsft:
I think we need a changelog entry to the "Other Changes" section, "Fixed  compilation error on POSIX platforms where openssl was not available.", something like that.

RickWinter (approved):

vhvb1989 (approved):
cool!  thank you!

PR #2791, opened 2021-08-24
Update keyvault codeowners to reduce noise to signal ratio

ahsonkhan:
Trying to reduce the noise to signal ratio and manage GitHub notifications to when I am explicitly tagged for relevant reviews and conversations, rather than during day-to-day keyvault code churn and features.


vhvb1989 (approved):
You should keep reviewing all team's PRs.  The number of PRs is usually under the 10 per week anyway. Please don't let parts of the SDK without your reviews

PR #2817, opened 2021-09-01
Optional RTTI idea demo

antkmsft:
This is a demo of the idea of how I think it could be done, as possible idea for Victor.

I did not check it extensively.

`__cpp_rtti` is C++20 (https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros), if I understand correctly, so we need a better mechanism, probably something compiler specific. Or, we can define a Cmake configuration option.

And this is more to illustrate the non-criticalness of the RTTI in Context implementation.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -79,7 +79,9 @@ namespace Azure { namespace Core {
       std::atomic<DateTime::rep> Deadline;
       Context::Key Key;
       std::shared_ptr<void> Value;
+#if __cpp_rtti
       const std::type_info& ValueType;
+#endif
 
       static constexpr DateTime::rep ToDateTimeRepresentation(DateTime const& dateTime)
       {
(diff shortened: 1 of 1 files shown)


RickWinter:
We could do something like this.  We'd need to add some safe-guards on the cast since the reinterpret_cast isn't safe if don't know explicitly what type the underlying type is.

antkmsft:
@RickWinter, current behavior is not that much different, it only checks in assert, which is only available in debug builds. We don't throw, or anything like it. I think it is ok to say that if you want to debug the correctness of context, compile it as Debug with RTTI, otherwise, it is undefined behavior, i.e. basically, an reinterpret_cast, which is not necessarily the end of the world (reason to quit immediately) for the customer app.

antkmsft:
@vhvb1989 It will keep running the same, without type-checking. You can see the code, and the only place where the type info is used is in assert, which only make it work in debug builds.

Jinming-Hu:
What's the motivation of removing RTTI info from Context in the first place? @antkmsft @vhvb1989

antkmsft:
@Jinming-Hu, RTTI takes up resulting binary space, and that is something that is important for Cognitive. They compile without RTTI, but if Core does, then they should too, and they can't afford that.

Jinming-Hu:
> @Jinming-Hu, RTTI takes up resulting binary space, and that is something that is important for Cognitive. They compile without RTTI, but if Core does, then they should too, and they can't afford that.

But size of `std::type_info` is only 16 bytes. Does it really matter? Or are you referring to other kinds of binary space.

PR #3018, opened 2021-11-01
Lro

### LINKED ISSUE #2896: Archboard review

gearama:
fixes #2896 
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?


sdk/keyvault/azure-security-keyvault-certificates/inc/azure/keyvault/certificates/certificate_client_operations.hpp
```diff
@@ -23,16 +23,17 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Certificat
...
-      : public Azure::Core::Operation<CertificateOperationProperties> {
+      : public Azure::Core::Operation<KeyVaultCertificateWithPolicy> {
 
     friend class CertificateClient;
 
   private:
     std::shared_ptr<CertificateClient> m_certificateClient;
-    CertificateOperationProperties m_value;
+    KeyVaultCertificateWithPolicy m_value;
+    CertificateOperationProperties m_properties;
```
vhvb1989, line 33: I think this can be public?  Or what's the reason to keep it private?

sdk/keyvault/azure-security-keyvault-certificates/inc/azure/keyvault/certificates/certificate_client_operations.hpp
```diff
@@ -69,8 +70,9 @@ namespace Azure { namespace Security { namespace KeyVault { namespace Certificat
      *
      * @return A CertificateOperationProperties object.
      */
-    CertificateOperationProperties Value() const override { return m_value; }
+    KeyVaultCertificateWithPolicy Value() const override { return m_value; }
 
+    CertificateOperationProperties Properties() const { return m_properties; }
```
vhvb1989, line 75: We can avoid this getter method by making Properties a public field

sdk/keyvault/azure-security-keyvault-certificates/src/certificate_client_operations.cpp
```diff
@@ -9,7 +9,7 @@
 
 using namespace Azure::Security::KeyVault::Certificates;
 
-Azure::Response<CertificateOperationProperties> CreateCertificateOperation::PollUntilDoneInternal(
+Azure::Response<KeyVaultCertificateWithPolicy> CreateCertificateOperation::PollUntilDoneInternal(
     std::chrono::milliseconds period,
     Azure::Core::Context& context)
```
vhvb1989, line 14: We need to check for context cancellation.  Either inside the while(true) loop from this function or within the Poll( context ) function.

sdk/keyvault/azure-security-keyvault-certificates/src/certificate_client_operations.cpp
```diff
@@ -23,7 +23,14 @@ Azure::Response<CertificateOperationProperties> CreateCertificateOperation::Poll
...
 
-  return Azure::Response<CertificateOperationProperties>(
+  if (!m_properties.Error)
+  {
+    auto response = m_certificateClient->GetCertificate(m_properties.Name);
+    m_value = response.Value;
+    m_rawResponse = std::move(response.RawResponse);
+  }
+
+  return Azure::Response<KeyVaultCertificateWithPolicy>(
```
vhvb1989, line 33: so, we are returning this even if `m_properties.Error == true` ??  But only assigning m_value when m_propertie.Error is false?  
What is it going to happen on error?  do we return an empty certificate?  

Shouldn't we throw if Error == true?  and make the error part of the Exception?

sdk/keyvault/azure-security-keyvault-certificates/src/certificate_client_operations.cpp
```diff
@@ -59,22 +66,22 @@ std::unique_ptr<Azure::Core::Http::RawResponse> CreateCertificateOperation::Poll
 
   if (m_status == Azure::Core::OperationStatus::Succeeded)
```
vhvb1989, line 67: I think there's a bug about how we are setting the m_status:

image


We are using the `HttpStatusCode` from the operation response to se the m_status.  We should be parsing the OperationProperties to know about the operation current status and use it to set m_status.

Right now, as soon as we get a valid Operation response, we are setting m_status == Succeded, but the operation might have come with a payload that contains status: 'pending'.

sdk/keyvault/azure-security-keyvault-certificates/inc/azure/keyvault/certificates/certificate_client_operations.hpp
gearama: so it's read only , i don't want poeole to modify it, especially while polling

sdk/keyvault/azure-security-keyvault-certificates/src/certificate_client_operations.cpp
gearama: we never do that anywhere else , so i'm not sure why we would suddenly do it here ? ism't that looked at during the SendRequest method , in the http pipeline ?

sdk/keyvault/azure-security-keyvault-certificates/inc/azure/keyvault/certificates/certificate_client_operations.hpp
gearama: i want it private, see previous comment

sdk/keyvault/azure-security-keyvault-certificates/src/certificate_client_operations.cpp
gearama: we discussed this in a call

vhvb1989 (approved):

PR #3059, opened 2021-11-05
Proposal: Breaking Change: Make AZURE_ASSERT and related macros internal

### LINKED ISSUE #2834: Make Azure Assert internal
1. Move `AzureNoReturnPath()` to `Azure::_internal` namespace. Currently, it is in global namespace.
2. Move `core/azure_assert.hpp` header to `core/internal/azure_assert.hpp`.

If we ever to take breaking changes, this would be among the ones I'd take.

antkmsft:
Closes #2834.

It will be a breaking change.
Should we release it?
Should we wait, and group all the breaking changes?
Should we never fix it?

I do not have a strong opinion, but I think we should discuss the possibility. And this PR is a way to start that discussion.

If we are not taking the entire change, then we should take the change that moves `AzureNoReturnPath()` from a global namespace into an Azure namespace, that one is much less of a breaking change. It is theoretically possible that a customer was invoking `AzureNoReturnPath()` in their code, but I do not think it is likely. Any objections about just moving the function? - See PR #3060

Should there be a parallel set of macros and a new location and names that we should start using from now on? Then, when we have to release a breaking change, we only break in the currently known places, but not the new places that will start using the new header.

We also can create a compatibility flag, when you build core with it, it puts back the assert header, but otherwise it is not available.


sdk/storage/azure-storage-blobs/src/blob_lease_client.cpp
```diff
@@ -3,7 +3,7 @@
 
 #include "azure/storage/blobs/blob_lease_client.hpp"
 
-#include <azure/core/azure_assert.hpp>
+#include "azure/core/internal/assert.hpp"
```
Jinming-Hu, line 6:
```suggestion
#include <azure/core/internal/assert.hpp>
```

Jinming-Hu: We use `""` for headers within the package, and `<>` for headers out of the package.

Jinming-Hu (approved):
Changes to storage code LGTM.

JeffreyRichter (approved):

ahsonkhan:
@antkmsft Why do we want to make this change? Can you clarify what the benefit is exactly (to customers, SDK authors, maintainers like us)?

Some questions that are typically useful to answer when thinking about breaking changes:
1) What is the customer benefit or impact of making this change? Why is now the right time to make this change, rather than later
2) How many customers would be negatively impacted by this, and if significant, what is the workaround for them, if they wanted the previous behavior? What could we do in the future, if customers report a pain point?
3) Did we consider alternatives which solve this problem without a breaking change?

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ set(
     inc/azure/core/http/raw_response.hpp
     inc/azure/core/http/policies/policy.hpp
     inc/azure/core/http/transport.hpp
+    inc/azure/core/internal/assert.hpp
```
ahsonkhan, line 68: nit: Can we keep the name as azure_assert or az_assert, just to avoid confusion with others that are similar like `cassert` or `assert.h`?

sdk/core/azure-core/inc/azure/core/internal/assert.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <string>
+
+#if defined(NDEBUG)
+
+/*
+ * NDEBUG = defined = Build is on Release
+ * Define _azure_ASSERT to call abort directly on exp == false
+ */
+
+#define _azure_ASSERT(exp) \
```
ahsonkhan, line 27: Other than the under-score renaming, putting the method in a namespace, and the file move/rename, any other changes to this file?

Doesn't look like it, but just verifying.

image

sdk/core/azure-core/inc/azure/core.hpp
```diff
@@ -9,14 +9,14 @@
...
 // azure/core
-#include "azure/core/azure_assert.hpp"
 #include "azure/core/base64.hpp"
 #include "azure/core/case_insensitive_containers.hpp"
 #include "azure/core/context.hpp"
 #include "azure/core/datetime.hpp"
 #include "azure/core/dll_import_export.hpp"
 #include "azure/core/etag.hpp"
 #include "azure/core/exception.hpp"
+#include "azure/core/internal/assert.hpp"
```
ahsonkhan, line 19: Don't include internal headers in the simplified header that the end customer uses.
```suggestion
```

ahsonkhan (commented):
I think the change is OK, but want to understand the rationale and discuss some of the questions before approving.

Also, made some suggestions.

vhvb1989 (approved):

sdk/core/azure-core/inc/azure/core/internal/assert.hpp
antkmsft: No, only these two changes - prefix change and moving the function to a namespace.

sdk/core/azure-core/CMakeLists.txt
antkmsft: I don't think there will be confusion, and it does not align with the other naming patterns, but you feel more strongly about this, so I'll make the change - after all, it is internal header.

sdk/core/azure-core/inc/azure/core.hpp
antkmsft: Thank you good catch!

antkmsft:
@ahsonkhan

> Why do we want to make this change? Can you clarify what the benefit is exactly (to customers, SDK authors, maintainers like us)?

The benefit is that it will be clear for customers that they should not be using `AZURE_ASSERT()` in their code, and we won't have to maintain it as public API.
The benefit of a namespace change is that when it is a global function, we are outside of the scope that we took for ourselves, so should customer want to have a global function with this name, there will be name collision in linker.
Both events are unlikely to happen, but still, a breaking change, is a breaking change.

> Some questions that are typically useful to answer when thinking about breaking changes:
> Reason for Change: What is the customer benefit or impact of making this change? Why is now the right time to make this change, rather than later?

To correct the wrong impression we may have created that this macro is intended for public use.
If we make this sooner, there's less change that customers will use it.

> Recommended Action: How would the customer observe this change? How many customers would be negatively impacted by this, and if significant, what is the workaround for them, if they wanted the previous behavior? What could we do in the future, if customers report a pain point? And how can we minimize the impact of the breaking change?

They could use the standard `assert()`.

> Did we consider alternatives which solve this problem without a breaking change?
> Versioning Info: In what version was this API introduced that had the "previous" behavior? And in what version will we get the new behavior?

1.0 GA.

PR #3068, opened 2021-11-05
Disable unit tests for staticLibCurl since they are failing on Mac

ahsonkhan:
Tracking the actual issue here:
https://github.com/Azure/azure-sdk-for-cpp/issues/3065

The static lib curl is a private implementation only, the test issue is not as significant to resolve, since no one is using that transport adapter. This is a stop gap to unblock the release and get main to a  green state.


antkmsft (approved):
Please create a work item to bring them back (I assume, just more fixes are needed to make them work properly / not run on Mac etc).

vhvb1989 (approved):

ahsonkhan:
> Please create a work item to bring them back (I assume, just more fixes are needed to make them work properly / not run on Mac etc).

Already did :)
https://github.com/Azure/azure-sdk-for-cpp/issues/3065

PR #3085, opened 2021-11-09
Remove `LICENSE` file duplicates

antkmsft:
Inspired by https://github.com/Azure/azure-sdk-for-c/pull/1987

We have one at https://github.com/Azure/azure-sdk-for-cpp/blob/main/LICENSE.txt.
Plus, it is not even that every service has them - some do, some don't (Identity, Storage).
License text is the same.


sdk/core/azure-core/LICENSE
```diff
@@ -1,21 +0,0 @@
-    MIT License
```
ahsonkhan, line 1: Unlike Embedded C, we do ship "individual packages" in C++ when publishing to vcpkg. I wonder if the better fix is to keep the duplicate LICENSE file for all packages in their individual directories.

For context: https://github.com/Azure/azure-sdk-for-c/pull/1960#discussion_r742431357

@weshaggard and @RickWinter what do you think?

antkmsft: For vcpkg, we do take the one from the root of the repo and install in with each package with the name that vcpkg expects: https://github.com/Azure/azure-sdk-for-cpp/blob/main/cmake-modules/AzureVcpkg.cmake#L71-L72

weshaggard: I would say we keep only the one at the root given we can copy it into the vcpkg output when we produce the package.

Jinming-Hu (approved):

sdk/core/azure-core/LICENSE
RickWinter: This is different than all the other SDKs.  Why do we want to break from standard SDKs guidlines for C++.  C++ has packages

RickWinter (changes requested):
Why are we getting rid of the LICENSE in each package.  This is a change from standard SDK designs.

RickWinter:
The single LICENSE in the Embedded C repo makes sense because that SDK is a monolith.  The C++ repo is individual packages.  I'd prefer we follow standard practices with the LICENSE file in each package.

sdk/core/azure-core/LICENSE
weshaggard: The standard is to have one license file in the root of the repo. We only duplicate the file in cases where the source layout is interesting. For example in .NET repo we only have the one license file in the root of the repo.

RickWinter: Oh interesting, and good to know.  Would make sense to add EngSys process to copy the file before removing it from everywhere.  I know we don't have process to copy the file into the package.  Granted the package is bigger than it should be right now.

antkmsft: We do copy the license from the root of the repo with each vcpackage. See the comments above for details. LICENSE files in each folder are not utilized. Some services/packages don't even have them (Storage).

antkmsft: @RickWinter would you approve this PR given the information above?

RickWinter (approved):

Issue #3086, opened 2021-11-09
Missing API in blob service: SyncCopyFromUri


Jinming-Hu:

ahsonkhan:
I am curious, how did you find out about this being missing?

Jinming-Hu:
> I am curious, how did you find out about this being missing?

reported by other team member

PR #3098, opened 2021-11-11
New API: CopyFromUri, which copies a blob synchronously

Jinming-Hu:
closes https://github.com/Azure/azure-sdk-for-cpp/issues/3086

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-blobs/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blobs/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features Added
 
-- New API: `BlobClient::CopyFromUri`.
+- New API: `BlobClient::CopyFromUri()`.
 
 ### Breaking Changes
 
--- a/sdk/storage/azure-storage-blobs/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blobs/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Release History
 
+## 12.3.0 (Unreleased)
+
+### Features Added
+
+- New API: `BlobClient::CopyFromUri`.
+
+### Breaking Changes
+
+### Bugs Fixed
+
+### Other Changes
+
 ## 12.2.1 (2021-11-08)
 
 ### Other Changes
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
@@ -203,7 +203,7 @@ namespace Azure { namespace Storage { namespace Blobs {
         const Azure::Core::Context& context = Azure::Core::Context()) const;
 
     /**
-     * @brief Copies data at from the source to this blob synchronously.
+     * @brief Copies data from the source to this blob, synchronously.
      *
      * @param sourceUri Specifies the URL of the source blob. The value may be a URL of up to 2 KB
      * in length that specifies a blob. The value should be URL-encoded as it would appear in a
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
@@ -208,7 +208,7 @@ namespace Azure { namespace Storage { namespace Blobs {
      * @param sourceUri Specifies the URL of the source blob. The value may be a URL of up to 2 KB
      * in length that specifies a blob. The value should be URL-encoded as it would appear in a
      * request URI. The source blob must either be public or must be authorized via a shared access
-     * signature. If the size of thesource blob is greater than 256 MB, the request will fail with
+     * signature. If the size of the source blob is greater than 256 MB, the request will fail with
      * 409 (Conflict). The blob type of the source blob has to be block blob.
      * @param options Optional parameters to execute this function.
      * @param context Context for cancelling long running operations.
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
@@ -203,6 +203,23 @@ namespace Azure { namespace Storage { namespace Blobs {
         const Azure::Core::Context& context = Azure::Core::Context()) const;
 
     /**
+     * @brief Copies data at from the source to this blob synchronously.
+     *
+     * @param sourceUri Specifies the URL of the source blob. The value may be a URL of up to 2 KB
+     * in length that specifies a blob. The value should be URL-encoded as it would appear in a
+     * request URI. The source blob must either be public or must be authorized via a shared access
+     * signature. If the size of thesource blob is greater than 256 MB, the request will fail with
+     * 409 (Conflict). The blob type of the source blob has to be block blob.
+     * @param options Optional parameters to execute this function.
+     * @param context Context for cancelling long running operations.
+     * @return A CopyBlobFromUriResult describing the copy result.
+     */
+    Azure::Response<Models::CopyBlobFromUriResult> CopyFromUri(
+        const std::string& sourceUri,
+        const CopyBlobFromUriOptions& options = CopyBlobFromUriOptions(),
+        const Azure::Core::Context& context = Azure::Core::Context()) const;
+
+    /**
      * @brief Copies data at from the source to this blob.
      *
      * @param sourceUri
--- a/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
+++ b/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
@@ -498,6 +498,52 @@ namespace Azure { namespace Storage { namespace Blobs {
   };
 
   /**
+   * @brief Optional parameters for #Azure::Storage::Blobs::BlobClient::CopyFromUri.
+   */
+  struct CopyBlobFromUriOptions
+  {
+    /**
+     * @brief Specifies user-defined name-value pairs associated with the blob. If no
+     * name-value pairs are specified, the operation will copy the metadata from the source blob or
+     * file to the destination blob. If one or more name-value pairs are specified, the destination
+     * blob is created with the specified metadata, and metadata is not copied from the source blob
+     * or file.
+     */
+    Storage::Metadata Metadata;
+
+    /**
+     * @brief The tags to set for this blob.
+     */
+    std::map<std::string, std::string> Tags;
+
+    /**
+     * @brief Optional conditions that must be met to perform this operation.
+     */
+    BlobAccessConditions AccessConditions;
+
+    /**
+     * @brief Optional conditions that the source must meet to perform this operation.
+     *
+     * @note Lease access condition only works for API versions before 2012-02-12.
+     */
+    struct : public Azure::ModifiedConditions, public Azure::MatchConditions
+    {
+    } SourceAccessConditions;
+
+    /**
+     * @brief Specifies the tier to be set on the target blob.
+     */
+    Azure::Nullable<Models::AccessTier> AccessTier;
+
+    /**
+     * @brief Hash of the blob content. This hash is used to verify the integrity of
+     * the blob during transport. When this header is specified, the storage service checks the hash
+     * that has arrived with the one that was sent.
+     */
+    Azure::Nullable<ContentHash> TransactionalContentHash;
+  };
+
+  /**
    * @brief Optional parameters for #Azure::Storage::Blobs::BlobClient::AbortCopyFromUri.
    */
   struct AbortBlobCopyFromUriOptions final
--- a/sdk/storage/azure-storage-blobs/src/blob_client.cpp
+++ b/sdk/storage/azure-storage-blobs/src/blob_client.cpp
@@ -3,6 +3,7 @@
 
 #include "azure/storage/blobs/blob_client.hpp"
 
+#include <azure/core/azure_assert.hpp>
 #include <azure/core/http/policies/policy.hpp>
 #include <azure/storage/common/internal/concurrent_transfer.hpp>
 #include <azure/storage/common/internal/constants.hpp>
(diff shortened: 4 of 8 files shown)


JeffreyRichter (approved):

sdk/storage/azure-storage-blobs/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Release History
 
+## 12.3.0 (Unreleased)
+
+### Features Added
+
+- New API: `BlobClient::CopyFromUri`.
```
ahsonkhan, line 7: nit:
```suggestion
- New API: `BlobClient::CopyFromUri()`.
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
```diff
@@ -202,6 +202,23 @@ namespace Azure { namespace Storage { namespace Blobs {
         const SetBlobAccessTierOptions& options = SetBlobAccessTierOptions(),
         const Azure::Core::Context& context = Azure::Core::Context()) const;
 
+    /**
+     * @brief Copies data at from the source to this blob synchronously.
```
ahsonkhan, line 206:
```suggestion
     * @brief Copies data from the source to this blob, synchronously.
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
```diff
@@ -202,6 +202,23 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+     *
+     * @param sourceUri Specifies the URL of the source blob. The value may be a URL of up to 2 KB
+     * in length that specifies a blob. The value should be URL-encoded as it would appear in a
+     * request URI. The source blob must either be public or must be authorized via a shared access
+     * signature. If the size of thesource blob is greater than 256 MB, the request will fail with
```
ahsonkhan, line 211:
```suggestion
     * signature. If the size of the source blob is greater than 256 MB, the request will fail with
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
```diff
@@ -497,6 +497,52 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    BlobAccessConditions AccessConditions;
+
+    /**
+     * @brief Optional conditions that the source must meet to perform this operation.
+     *
+     * @note Lease access condition only works for API versions before 2012-02-12.
+     */
+    struct : public Azure::ModifiedConditions, public Azure::MatchConditions
+    {
+    } SourceAccessConditions;
```
ahsonkhan, line 531: This is an interesting code style. Do we do this elsewhere? And why do we have this empty struct?

1) Unnamed struct, with name after?
2) Struct that inherits from multiple others (rather than class)?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
```diff
@@ -202,6 +202,23 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+     * 409 (Conflict). The blob type of the source blob has to be block blob.
+     * @param options Optional parameters to execute this function.
+     * @param context Context for cancelling long running operations.
+     * @return A CopyBlobFromUriResult describing the copy result.
+     */
+    Azure::Response<Models::CopyBlobFromUriResult> CopyFromUri(
```
ahsonkhan, line 217: Is the reason why this isn't a long-running oepration

The LRO version, `StartCopyFromUri()`, returns `BlobProperties`, why is this returning `CopyBlobFromUriResult`?

https://github.com/Azure/azure-sdk-for-cpp/blob/fda88bf7b8389575b9d973821efc604442e10493/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp#L171

Is that not also synchronous? When does the user call one vs the other?

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -531,6 +532,38 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    protocolLayerOptions.SourceIfModifiedSince = options.SourceAccessConditions.IfModifiedSince;
+    protocolLayerOptions.SourceIfUnmodifiedSince = options.SourceAccessConditions.IfUnmodifiedSince;
+    protocolLayerOptions.SourceIfMatch = options.SourceAccessConditions.IfMatch;
+    protocolLayerOptions.SourceIfNoneMatch = options.SourceAccessConditions.IfNoneMatch;
+    if (options.TransactionalContentHash.HasValue())
+    {
+      AZURE_ASSERT_MSG(
+          options.TransactionalContentHash.Value().Algorithm == HashAlgorithm::Md5,
+          "This operation only supports MD5 transactional content hash.");
+      protocolLayerOptions.TransactionalContentHash = options.TransactionalContentHash;
```
ahsonkhan, line 560: This should be documented in the API signature and in the options class.

sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
```diff
@@ -210,53 +210,6 @@ namespace Azure { namespace Storage { namespace Test {
     EXPECT_NO_THROW(appendBlobClient.Delete(options));
   }
 
-  TEST_F(AppendBlobClientTest, SourceBlobAccessConditions)
```
ahsonkhan, line 213: Why are we removing this (especially in this PR)? Looks like we might be losing some coverage from `AppendBlobClient`, no?

ahsonkhan (commented):
Some feedback and questions

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
Jinming-Hu: > Do we do this elsewhere?

Yes, mainly for various combinations of access conditions.

>  And why do we have this empty struct?

It's not empty. It inherits from two other structs, which makes it effectively a union of the two structs.

> Unnamed struct, with name after?

It has the same effect as

```C++
struct StructWillOnlyBeUsedOnce { // bla bla };
StructWillOnlyBeUsedOnce SourceAccessConditions;
```

except that `StructWillOnlyBeUsedOnce` is not accessible elsewhere.

> Struct that inherits from multiple others (rather than class)?

All its members are public and it works like a struct. So I think `struct` is better than `class` here.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
Jinming-Hu: `StartCopyFromUri` starts the copy operation and then returns before the operation is completed. Storage service will continue copying the blob in the background. Client side can check if the copy is completed by getting the blobs' properties, which includes a field of copy status. That's why this is an LRO operation and why `Poll()` returns `BlobProperties`.

In constrast, `CopyFromUri` copies the blob synchronously, which means by the time the REST API returns, the copy operation is either finished or failed. That's why it's not an LRO. The return type `CopyBlobFromUriResult` is just to follow our naming pattern, `[Operation]Result`.

Jinming-Hu: `CopyFromUri` can only be used to copy small blobs (<256MB, IIRC). `StartCopyFromUri` can be used on any size of blobs.

> When does the user call one vs the other?

If a customer knows for sure a blob is relatively small and wants to write simple code, he can just write one line of code
`blob_client.CopyFromUri(url);`, which makes exactly one REST API call.
If he knows the blob can be large and the copy operation can take quite some time to finish, he'll call `StartCopyFromUri`.

sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
Jinming-Hu: It's moved to `block_blob_client_test.cpp`.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
Jinming-Hu: This is a convenient way to define structs that will only be used once, and it can help us avoid the trouble of naming things.

For example, how would you want to name this struct with four access conditions? Another operation supports only 3 access conditions (`IfMatch`, `IfNoneMatch`, `IfModifiedSince`), how do you want to name that? With anonymous struct, we don't need to worry about that at all.

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -531,6 +532,38 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
+    protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
+    protocolLayerOptions.IfMatch = options.AccessConditions.IfMatch;
+    protocolLayerOptions.IfNoneMatch = options.AccessConditions.IfNoneMatch;
+    protocolLayerOptions.IfTags = options.AccessConditions.TagConditions;
+    protocolLayerOptions.SourceIfModifiedSince = options.SourceAccessConditions.IfModifiedSince;
+    protocolLayerOptions.SourceIfUnmodifiedSince = options.SourceAccessConditions.IfUnmodifiedSince;
+    protocolLayerOptions.SourceIfMatch = options.SourceAccessConditions.IfMatch;
+    protocolLayerOptions.SourceIfNoneMatch = options.SourceAccessConditions.IfNoneMatch;
+    if (options.TransactionalContentHash.HasValue())
```
katmsft, line 555: Does the API even support transactional content hash? The request body should be empty right?

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -531,6 +532,38 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+  Azure::Response<Models::CopyBlobFromUriResult> BlobClient::CopyFromUri(
+      const std::string& sourceUri,
+      const CopyBlobFromUriOptions& options,
+      const Azure::Core::Context& context) const
+  {
+    _detail::BlobRestClient::Blob::CopyBlobFromUriOptions protocolLayerOptions;
+    protocolLayerOptions.Metadata = options.Metadata;
+    protocolLayerOptions.Tags = options.Tags;
+    protocolLayerOptions.SourceUri = sourceUri;
+    protocolLayerOptions.AccessTier = options.AccessTier;
```
katmsft, line 544: Official doc does not allow access tier. Could you please double-check/target test whether it is supported?
https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url

sdk/storage/azure-storage-blobs/src/blob_client.cpp
```diff
@@ -531,6 +532,38 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+    protocolLayerOptions.LeaseId = options.AccessConditions.LeaseId;
+    protocolLayerOptions.IfModifiedSince = options.AccessConditions.IfModifiedSince;
+    protocolLayerOptions.IfUnmodifiedSince = options.AccessConditions.IfUnmodifiedSince;
+    protocolLayerOptions.IfMatch = options.AccessConditions.IfMatch;
+    protocolLayerOptions.IfNoneMatch = options.AccessConditions.IfNoneMatch;
+    protocolLayerOptions.IfTags = options.AccessConditions.TagConditions;
```
katmsft, line 550: Tag condition seems to be not supported for this API.

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: content of the blob to be copied

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: It's supported according to swagger https://github.com/Azure/azure-rest-api-specs/blob/2fbdd238b9fc19884e2d0491a338f7f6fc2aa1d7/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-02-10/blob.json#L6150

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: It's supported according to swagger https://github.com/Azure/azure-rest-api-specs/blob/2fbdd238b9fc19884e2d0491a338f7f6fc2aa1d7/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-02-10/blob.json#L6123

Jinming-Hu: and it's covered by unit test

sdk/storage/azure-storage-blobs/src/blob_client.cpp
Jinming-Hu: and it's covered by unit test

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
ahsonkhan: Hmm, interesting. I was thinking about the rules/logic on when and how such a struct definition would be code-gen'd.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_client.hpp
ahsonkhan: Ok, gotcha, that makes sense now. Thanks for clarifying.

sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
ahsonkhan: Where exactly? I don't see any use of an `AppendBlobClient` or calls to `appendBlobClient.StartCopyFromUri()` within `BlockBlobClientTest`.

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/blob/eab04fc80040164203859eeea8840331e53e0c9b/sdk/storage/azure-storage-blobs/test/ut/block_blob_client_test.cpp#L349

Jinming-Hu: https://github.com/Azure/azure-sdk-for-cpp/blob/eab04fc80040164203859eeea8840331e53e0c9b/sdk/storage/azure-storage-blobs/test/ut/block_blob_client_test.cpp#L1230

ahsonkhan: Aren't those block blob clients, and not append blob clients?

Jinming-Hu: Why does it matter? Asyn copy works for all kinds of blobs.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp
Jinming-Hu: Generate like this

```C++
struct CopyBlobFromUriOptions {
  Storage::Metadata Metadata;
  std::map<std::string, std::string> Tags;
  BlobAccessConditions AccessConditions;
  (anonymous123) SourceAccessConditions;
  ...
};
```

and `(anonymous123)` has a link to the definition of the type, so if you click on it, it jumps to

```C++
struct (anonymous123) : public Azure::ModifiedConditions, public Azure::MatchConditions
{
};
```

PR #3111, opened 2021-11-16
Do not initialize statics in inline functions

antkmsft:
This came up during improving code coverage.

But there is a problem, if you initialize static variable in an inline method, because each translation unit gets its own copy of an inline function, so there will be multiple statics. It would be just an ineffectiveness (duplicates) if the variable is immutable, which are the cases here. But should've these been mutable variables, then it will be out of sync between different translation units.

### DIFF
--- a/sdk/core/azure-core/src/etag.cpp
+++ b/sdk/core/azure-core/src/etag.cpp
@@ -7,6 +7,6 @@ using Azure::ETag;
 
 const ETag& ETag::Any()
 {
-  static ETag any = ETag("*"); // LCOV_EXCL_LINE
+  static ETag any = ETag("*");
   return any;
 }
--- a/sdk/core/azure-core/src/etag.cpp
+++ b/sdk/core/azure-core/src/etag.cpp
@@ -1,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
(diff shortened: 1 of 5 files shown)


sdk/core/azure-core/src/etag.cpp
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include "azure/core/etag.hpp"
+
+using Azure::ETag;
+
+const ETag& ETag::Any()
+{
+  static ETag any = ETag("*"); // LCOV_EXCL_LINE
```
ahsonkhan, line 10: > LCOV_EXCL_LINE

What's this and why do we need it?

ahsonkhan (approved):

sdk/core/azure-core/src/etag.cpp
antkmsft: Thanks! This may be added later, as part of a different PR that addresses code coverage. I missed the removal of this comment for this PR.

antkmsft:
```suggestion
  static ETag any = ETag("*");
```

Jinming-Hu (approved):

vhvb1989 (approved):

PR #3120, opened 2021-11-18
Exclude test, samples, and 3rd party code from code coverage

antkmsft:
Much more clarity, less noise, more accurate, reliable, and actionable coverage numbers.

We should only measure SDK code as our first priority, and we should do it well.
In the future, when we figure out how to do it, we can measure the amount of code that is covered by samples. But the way we do it currently is not the way how we should do it anyways.

It does not make sense to measure the coverage numbers of tests. Little sense/actionability for samples and 3rd party.

### DIFF
--- a/sdk/core/azure-core/CMakeLists.txt
+++ b/sdk/core/azure-core/CMakeLists.txt
@@ -143,7 +143,7 @@ target_include_directories(
 add_library(Azure::azure-core ALIAS azure-core)
 
 # coverage. Has no effect if BUILD_CODE_COVERAGE is OFF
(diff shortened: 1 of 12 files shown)


antkmsft:
Before: https://dev.azure.com/azure-sdk/public/_build/results?buildId=1201286&view=codecoverage-tab
After: https://dev.azure.com/azure-sdk/public/_build/results?buildId=1202238&view=codecoverage-tab

vhvb1989 (approved):
Awesome!
I was forgotten about the exclude arg support in the macro!

Thank you!

PR #3143, opened 2021-11-29
Fix unit tests on Mac

antkmsft:


sdk/identity/azure-identity/test/ut/client_secret_credential_test.cpp
```diff
@@ -60,8 +60,8 @@ TEST(ClientSecretCredential, Regular)
   EXPECT_EQ(response.AccessToken.Token, "ACCESSTOKEN1");
 
   using namespace std::chrono_literals;
-  EXPECT_GT(response.AccessToken.ExpiresOn, response.EarliestExpiration + 3600s);
-  EXPECT_LT(response.AccessToken.ExpiresOn, response.LatestExpiration + 3600s);
+  EXPECT_GE(response.AccessToken.ExpiresOn, response.EarliestExpiration + 3600s);
```
ahsonkhan, line 63: Do we understand the underlying reason why this test was unreliable, only on a Mac?

Is it OK for the access token expire time to be equal to the earliest expiration + 1 hr, or should it always be greater than?

ahsonkhan (approved):

sdk/identity/azure-identity/test/ut/client_secret_credential_test.cpp
antkmsft: It is probably some clever compiler optimization. In the ideal world, where some instructions are instant or take less than 1 ms, the timestamp when the token was issued by calling `chrono::...::now()` and then when the tests would take the timestamp by using the same `chrono::...::now()`, that would be the same timestamp. I never thought it would be realistically possible, but on Mac they somehow did it. It is not at all a logical error if the timestamps are equal - in fact it is perfection I did not think was possible.

ahsonkhan: Hmm... Any idea why it only started failing recently?

PR #3152, opened 2021-12-02
Use code coverage exclude annotations

antkmsft:
Some are false positives. Lcov doesn't do very well with static initializations, constructor initialization lists, constructing from `const char*`.
Others are not coverable by design. Asserts (abort in case of failure), `_azure_UNREACHABLE_CODE()`, `_azure_ASSERT()` never gets to the `[noreturn]` part where it terminates the program.
I prefer clean and actionable coverage reports. If something is not green, it means it is not covered. If it is non-coverable - by design, or due to lcov bug, we review it and annotate so it gets excluded. Results are more trustworthy, rather than to know that some unknown amount of code is non-coverable there's never will be done anything about it (it is the fact whether we like it or not). No, I like accuracy, it removes this mystery from the system.


eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -77,6 +77,10 @@ jobs:
         condition: and(succeeded(), ne(variables['AptDependencies'], ''))
         displayName: Install dependencies from apt
 
+      - pwsh: sudo pip install $(PipDependencies)
```
benbp, line 80: Why is `sudo` necessary here?

cmake-modules/AzureCodeCoverage.cmake
```diff
@@ -55,6 +55,8 @@ function(add_gcovr_html)
             -r ${BASEDIR} ${EXCLUDE_ARGS}
             --object-directory=${PROJECT_BINARY_DIR}
             -o ${args_TARGET_NAME}/index.html
+            --exclude-unreachable-branches
+            --exclude-throw-branches
```
ahsonkhan, line 59: Are we sure we want to exclude throw branches? We have `EXPECT_THROW` for such testing, no?

cmake-modules/AzureCodeCoverage.cmake
```diff
@@ -55,6 +55,8 @@ function(add_gcovr_html)
             -r ${BASEDIR} ${EXCLUDE_ARGS}
             --object-directory=${PROJECT_BINARY_DIR}
             -o ${args_TARGET_NAME}/index.html
+            --exclude-unreachable-branches
```
ahsonkhan, line 58: Wouldn't we want to get rid of unreachable branches in the code, rather than skip it for coverage?

Can you share a couple of concrete examples of these?

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -48,7 +48,8 @@ jobs:
         VcpkgInstall: 'curl[ssl] libxml2 openssl'
         VCPKG_DEFAULT_TRIPLET: 'x64-linux'
         CmakeArgs: ' -DBUILD_TESTING=ON -DBUILD_PERFORMANCE_TESTS=ON -DRUN_LONG_UNIT_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -DBUILD_CODE_COVERAGE=ON'
-        AptDependencies: 'gcovr lcov'
+        AptDependencies: 'lcov'
+        PipDependencies: 'gcovr'
```
ahsonkhan, line 52: Why is `pip` needed now? Was `apt` not working previously?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -205,6 +206,7 @@ namespace Azure { namespace Core {
      */
     template <class T> bool TryGetValue(Key const& key, T& outputValue) const
     {
+      // LCOV_EXCL_START
       for (auto ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
```
ahsonkhan, line 210: This we should be testing, imo. It is part of correct usage of `TryGetValue`, no? Why are we skipping it?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -246,7 +250,7 @@ namespace Azure { namespace Core {
     {
       if (IsCancelled())
       {
-        throw OperationCancelledException("Request was cancelled by context.");
+        throw OperationCancelledException("Request was cancelled by context."); // LCOV_EXCL_LINE
```
ahsonkhan, line 253: Same here. Why are we unable to test this?

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
```diff
@@ -23,7 +23,7 @@ namespace Azure { namespace Core { namespace Credentials {
   /**
    * @brief An access token is used to authenticate requests.
    */
-  struct AccessToken final
+  struct AccessToken final // LCOV_EXCL_LINE
```
ahsonkhan, line 26: Seems odd to exclude this line. Is it excluding an entire struct? Why just this type? This seems buggy.

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
```diff
@@ -64,8 +64,11 @@ namespace Azure { namespace Core { namespace Cryptography {
      */
     void Append(const uint8_t* data, size_t length)
     {
+      // LCOV_EXCL_START
       _azure_ASSERT(data || length == 0);
```
ahsonkhan, line 68: In the Embedded C SDK, we had a mechanism to test preconditions. Can we do something similar to these azure asserts? They are meaningful for correctness for the user, and hence having them covered by tests has value.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -214,7 +215,12 @@ inline bool operator==(DateTime const& dt, std::chrono::system_clock::time_point
 
 inline bool operator<(DateTime const& dt, std::chrono::system_clock::time_point const& tp)
 {
-  return dt < DateTime(tp);
+  if (dt < DateTime(tp)) // LCOV_EXCL_LINE
+  {
+    return true;
+  }
+
+  return false;
```
ahsonkhan, line 223: It isn't great that we have to contort the code-style to be a bit more confusing in order to satisfy code-coverage.

The immediate question for code snippet like this would be, why is it not written the way it was previously.

Why is `dt < DateTime(tp);` not being able to get covered?

sdk/core/azure-core/inc/azure/core/etag.hpp
```diff
@@ -87,7 +87,7 @@ class ETag final {
       return !left.m_value && !right.m_value;
     }
 
-    switch (comparisonKind)
+    switch (comparisonKind) // LCOV_EXCL_LINE
```
ahsonkhan, line 90: If we had `default: _azure_UNREACHABLE_CODE()`, could we get rid of this comment?

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
```diff
@@ -402,8 +404,11 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
           std::string const& componentName,
           std::string const& componentVersion,
           TelemetryOptions options = TelemetryOptions())
-          : m_telemetryId(BuildTelemetryId(componentName, componentVersion, options.ApplicationId))
-      {
+          : m_telemetryId(BuildTelemetryId(
+              componentName,
+              componentVersion,
+              options.ApplicationId)) // LCOV_EXCL_START
+      { // LCOV_EXCL_STOP
```
ahsonkhan, line 411: It's hard to see when we'd use such a pattern. Starting/stopping with just one line, which happens to be a brace, is weird :(

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
```diff
@@ -367,10 +367,12 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
           NextHttpPolicy nextPolicy,
           Context const& context) const override
       {
+        // LCOV_EXCL_START
         auto uuid = Uuid::CreateUuid().ToString();
 
         request.SetHeader(RequestIdHeader, uuid);
         return nextPolicy.Send(request, context);
+        // LCOV_EXCL_STOP
       }
```
ahsonkhan, line 376: Why can't this get covered?

sdk/core/azure-core/inc/azure/core/internal/cryptography/sha_hash.hpp
```diff
@@ -33,7 +33,7 @@ namespace Azure { namespace Core { namespace Cryptography { namespace _internal
      * @brief Cleanup any state when destroying the instance of #Sha256Hash.
      *
      */
-    ~Sha256Hash() {}
+    ~Sha256Hash() override = default;
```
ahsonkhan, line 36: nit: seems like an unrelated change

Semantically, is this identical to just using `{}`?

sdk/core/azure-core/inc/azure/core/internal/http/pipeline.hpp
```diff
@@ -47,13 +47,13 @@ namespace Azure { namespace Core { namespace Http { namespace _internal {
...
+        throw std::invalid_argument("policies cannot be empty"); // LCOV_EXCL_LINE
       }
 
-      m_policies.reserve(policies.size());
+      m_policies.reserve(policies.size()); // LCOV_EXCL_LINE
       for (auto& policy : policies)
       {
-        m_policies.emplace_back(policy->Clone());
+        m_policies.emplace_back(policy->Clone()); // LCOV_EXCL_LINE
       }
```
ahsonkhan, line 57: This just looks like a test is missing to cover the ctor with multiple policies. Is that not the case?

sdk/core/azure-core/inc/azure/core/url.hpp
```diff
@@ -28,7 +28,7 @@ namespace Azure { namespace Core {
           auto separator = '?';
           for (const auto& q : encodedQueryParameters)
           {
-            queryStr += separator + q.first + '=' + q.second;
+            queryStr += separator + q.first + '=' + q.second; // LCOV_EXCL_LINE
             separator = '&';
```
ahsonkhan, line 32: Why is line 31 not covered, but line 32 is? 🤔

sdk/core/azure-core/src/operation_status.cpp
```diff
@@ -11,4 +11,6 @@ namespace Azure { namespace Core {
   const OperationStatus OperationStatus::Failed{"Failed"};
   const OperationStatus OperationStatus::Cancelled{"Cancelled"};
 
+  // LCOV_EXCL_START
 }} // namespace Azure::Core
+// LCOV_EXCL_STOP
```
ahsonkhan, line 16: How would we know when to do these type of exclusion? I am not sure if this is a good idea. Given anything over 95% starts to have diminishing returns in terms of effort, could we leave out these single-line "lack of coverage" as part of that 5%, and not have such annotations?

sdk/core/ci.yml
```diff
@@ -43,7 +43,7 @@ stages:
       LiveTestCtestRegex: '"azure-core.|json-test"'
       LiveTestTimeoutInMinutes: 90 # default is 60 min. We need a little longer on worst case for Win+jsonTests
       LineCoverageTarget: 81
-      BranchCoverageTarget: 43
+      BranchCoverageTarget: 50
```
ahsonkhan, line 46: 🥳

sdk/identity/ci.yml
```diff
@@ -28,8 +28,8 @@ stages:
       ServiceDirectory: identity
       CtestRegex: azure-identity-unittest.
       LiveTestCtestRegex: azure-identity-livetest.
-      LineCoverageTarget: 99
-      BranchCoverageTarget: 62
+      LineCoverageTarget: 100
+      BranchCoverageTarget: 100
```
ahsonkhan, line 32: 🎉 

Are we sure we want to keep this as the bar? Seems really high to maintain without putting a bunch of (seemingly) arbitrary annotations. ~95 allows us that flexibility.

I am curious to know what you and others think.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
antkmsft: It is a way to get the latest `gcovr`, it is not easily available via apt.

ahsonkhan: That is useful info, might be worth putting it in as a comment.

cmake-modules/AzureCodeCoverage.cmake
antkmsft: It is supposed to be for the cases when you pass a string literal to a function that expects `std::string`: theoretically, that might be a moment when your system runs out a last piece of RAM, and `std::string` fails to create, so it might throw.

cmake-modules/AzureCodeCoverage.cmake
antkmsft: This was an attempt to combat false positive noise.

sdk/core/azure-core/inc/azure/core/context.hpp
antkmsft: It is not that we don't test the exception being thrown, it is construction of a string from a string literal.

sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp
antkmsft: It is gcovr's false positives.

sdk/core/azure-core/inc/azure/core/cryptography/hash.hpp
antkmsft: I agree, it can be done.

sdk/core/azure-core/inc/azure/core/datetime.hpp
antkmsft: It is covered, but gcovr gives false positive. Doesn't do well for inline functions, I thik.

That extra dance with return true/false is to show when you look at the lines, that both returns are green lines, i.e. covered.

sdk/core/azure-core/inc/azure/core/etag.hpp
antkmsft: Gcovr on this line won't be satisfied that we don't pass a value that brings us to `default` label, even though the label itself is excluded.

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
antkmsft: Gcovr false positives, or the compiler-generated code that we don't see/have no ability to change :(

sdk/core/azure-core/inc/azure/core/internal/cryptography/sha_hash.hpp
antkmsft: Yes, it is identical, maybe the default didn't make the gcovr think that it was not covered. I'll have to check.

sdk/core/azure-core/inc/azure/core/url.hpp
antkmsft: Because theoretically every string concatenation might throw, if system runs out of memory.

sdk/core/azure-core/src/operation_status.cpp
antkmsft: There is no way to know in advance. The idea was that you look at each file, and if you see a noise, you mark it with annotations, indicating that it has been reviewed and nothing can be done.

sdk/identity/ci.yml
antkmsft: I don't disagree, especially 100 won't make it possible to simply submit a draft PR without dropping these numbers temporarily. But also to be fair, this might be the problem with any number - 95, 90, if you add a lot o code.

Under the approach in this PR, where you mark everything unactionable (noise) with annotations, 100% is actually a good result, meaning that everything has been addressed. The moment something drops to 99%, you know there is higher chance of a real coverage problem. Every line you'd see in red would mean it is not covered, without the inherent mysterious 5% coverage gap in the system that we admit we can't do anything about.

antkmsft:
I no longer think it is such a good idea, or is worth it. I don't think we should do it now. Maybe in the future we might return to it, using the ideas/finding from this PR. But now, I am closing it. The "evolution" of this PR is #3179. Thanks @ahsonkhan for spending time to review and your thoughts, I am sorry this will not go anywhere now, but your feedback was valuable.

PR #3154, opened 2021-12-02
Update vcpkg and dependencies to the newer commit sha from 11-10-2021, after last SDK release

ahsonkhan:
Testing this to fix https://github.com/Azure/azure-sdk-for-cpp/issues/3100

Some things to consider:
- Report the version in CI
  -  We log/report the vcpkg SHA during the `Set Vcpkg Variables` step:
  - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1232672&view=logs&j=117dcd10-cb90-5257-bbaf-0e8ca4e59cfe&t=c93086b3-22be-5e77-35d0-48311420342d
  - https://github.com/Azure/azure-sdk-for-cpp/blob/c534a1e16004346b574242b58e54728a522b9456/eng/pipelines/templates/steps/vcpkg.yml#L15
- Have some mechanism to always stay up-to-date, or at least notified when we are out-of-date (maybe compare to the list of fetched tags from the vcpkg repo)


vhvb1989 (approved):

ahsonkhan:
Tracking issue of intermittent failure if it occurs again: https://github.com/Azure/azure-sdk-for-cpp/issues/3160

eng/vcpkg-commit.txt
```diff
@@ -1 +1 @@
-tags/2020.11
\ No newline at end of file
+fcd3a3bb53356cc954f7fa07cee2c99b253d1a49
```
weshaggard, line 1: Do we end to use a sha here? Is there no tag to use instead?

ahsonkhan: Yep, talked to vcpkg folks, and the recommendation for getting package dependencies from the registry is to use recent commit shas (that are updated at some feasible cadence we decide) and have two separate test pipelines (one for making sure the SDK builds on newer versions, and also one pinned to the oldest set of dependencies we want to support). They don't publish the tags, given the vcpkg tool and the registry are decoupled.

> I think the "maximally mature" approach would be the following:
a. You should have a build using the "oldest" set of dependencies you ostensibly support
b. You should have a build using a semi-current version that you update reasonably often (1/mo, for example)
c. You should have a smoke-test CI or automated PR that tries to grab latest. When that passes, it should merge into (b)

weshaggard: OK. I just wanted to make sure we had a plan. 

One of the issues I see with using raw sha's is you don't easily have information about how old it is without going and digging into the repo itself. Perhaps we should add a comment in the file to help better document which how old the version it, or at least have that metadata in the commit message for this repo.

antkmsft (approved):

Jinming-Hu (approved):

eng/vcpkg-commit.txt
ahsonkhan: https://github.com/Azure/azure-sdk-for-cpp/issues/3165

sdk/storage/azure-storage-common/CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ if(WIN32)
     target_compile_options(azure-storage-common PUBLIC /wd28204 /wd28020)
 else()
     find_package(LibXml2 REQUIRED)
-    target_include_directories(azure-storage-common PRIVATE ${LIBXML2_INCLUDE_DIRS})
+    target_include_directories(azure-storage-common SYSTEM PRIVATE ${LIBXML2_INCLUDE_DIRS})
```
ahsonkhan, line 88: https://github.com/Azure/azure-sdk-for-cpp/issues/3167

In this case, we didn't want to apply the "doxygen doc correctness" warnings as errors to libxml2, but we do want it for our own SDK (i.e. `-Wdocumentation-unknown-command`):
https://github.com/Azure/azure-sdk-for-cpp/blob/fda88bf7b8389575b9d973821efc604442e10493/cmake-modules/AzureGlobalCompileOptions.cmake#L27

PR #3163, opened 2021-12-07
Add a few simple unit tests to boost coverage

antkmsft:


sdk/core/azure-core/src/http/curl/curl_connection_private.hpp
```diff
@@ -57,7 +57,7 @@ namespace Azure { namespace Core { namespace Http {
    *
    */
   class CurlNetworkConnection {
-  protected:
+  private:
     bool m_isShutDown = false;
```
gearama, line 61: do we still need this then ?

antkmsft: I am not touching the design too much, but as I understand the intent was that the base class has `Shutdown()`/`IsShutdown()` functions, which modify/read the `m_isShutdown`. Now, the derived class' `Shutdown()` was overloading the `Shutdown()`, did some additional work, in in the end, instead of calling base class' `Shutdown()`, it did the same job as base class' `Shutdown()` by modifying the base class boolean data member directly, so it was made protected. Now I made the derived class to call the base class' `Shutdown()`, and now there is no need in exposing the data member via `protected`. Plus, in the code coverage, base class `Shutdown()` does show up.
Should we make them pure virtual in the base class, if that's what you ask, I don't know, but it will be a bigger change, while her I am just moving things around a bit without changing the behavior or the design too much.

PR #3177, opened 2021-12-10
Fix DateTime::Parse validation for dates past year 9999 due to adjustments

antkmsft:


gearama (approved):

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -236,7 +236,7 @@ void ValidateDate(
           + (hour * OneHourIn100ns) + (minute * OneMinuteIn100ns) + (second * OneSecondIn100ns)
           + fracSec + (roundFracSecUp ? 1 : 0);
 
-      constexpr auto FracSecTo9999_12_31_23_59_59_9999999 = (31 * OneDayIn100ns)
+      constexpr auto FracSecTo9999_12_31_23_59_59_9999999 = (30 * OneDayIn100ns)
```
RickWinter, line 239: From your change you are setting the max value to be 1 day less than before.  Thus this name is no longer ...12_31..., it would now be ...12_30...

RickWinter (changes requested):

sdk/core/azure-core/src/datetime.cpp
antkmsft: No, it is correct. That's why I also made this mistake initially:
December 1st + 1 day = December 2nd.
December 1st + 2 days = December 3rd.
...
December 1st + 30 days = December 31st.

... + 23 hours + 59 minutes + 59 seconds + 9999999 fractional seconds = Dec 31st EOD, the very last fractional second of the day.

antkmsft: @RickWinter, please see the explanation above. I also must note that an alternative form to write the fix would be `FracSecTo9999_12_31_23_59_59_9999999 = (31 * OneDayIn100ns) - 1;` (i.e. without adding 23 hours and so on), it will have the same value.

PR #3179, opened 2021-12-13
Exclude uncoverable lines from code coverage

antkmsft:
There is a PR that does this more extensively: https://github.com/Azure/azure-sdk-for-cpp/pull/3152, and you might disagree with it.

In that PR, we are working around gcovr false positives, and excluding `_azure_ASSERT` and `_azure_UNREACHABLE_CODE`. But you might say that it is not worth it to go after gcovr false positives, or that it is possible to cover assert macros by writing a debug version of `AzureNoReturnPath` with test hooks.

But these three instances marked in this PR aren't any of that, and code comments explain why it is not possible to cover them.

### DIFF
--- a/sdk/core/azure-core/src/datetime.cpp
+++ b/sdk/core/azure-core/src/datetime.cpp
@@ -342,6 +342,9 @@ T ParseNumber(
   }
 
   ThrowParseError(description);
+
+  // ThrowParseError() will always throw, but there's no way to tell that to compiler, so a return
+  // statement is required. It is not possible to cover the return line with tests.
   return T(); // LCOV_EXCL_LINE
 }
 
--- a/sdk/core/azure-core/src/datetime.cpp
+++ b/sdk/core/azure-core/src/datetime.cpp
@@ -342,7 +342,7 @@ T ParseNumber(
   }
 
   ThrowParseError(description);
-  return T();
+  return T(); // LCOV_EXCL_LINE
 }
 
 template <typename T>
--- a/sdk/identity/azure-identity/src/managed_identity_credential.cpp
+++ b/sdk/identity/azure-identity/src/managed_identity_credential.cpp
@@ -20,7 +20,9 @@ std::unique_ptr<_detail::ManagedIdentitySource> CreateManagedIdentitySource(
          AzureArcManagedIdentitySource::Create,
          ImdsManagedIdentitySource::Create};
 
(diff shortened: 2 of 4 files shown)


sdk/identity/azure-identity/src/managed_identity_credential.cpp
```diff
@@ -20,16 +20,20 @@ std::unique_ptr<_detail::ManagedIdentitySource> CreateManagedIdentitySource(
...
   {
     if (auto source = create(clientId, options))
     {
       return source;
     }
   }
 
+  // LCOV_EXCL_START
   throw AuthenticationException(
       "ManagedIdentityCredential authentication unavailable. No Managed Identity endpoint found.");
```
ahsonkhan, line 35: If this will never happen, should we throw here or terminate with `AzureNoReturnPath`?

sdk/core/azure-core/src/datetime.cpp
```diff
@@ -342,7 +342,10 @@ T ParseNumber(
   }
 
   ThrowParseError(description);
-  return T();
+
+  // ThrowParseError() will always throw, but there's no way to tell that to compiler, so a return
+  // statement is required. It is not possible to cover the return line with tests.
+  return T(); // LCOV_EXCL_LINE
```
ahsonkhan, line 348: 👍

Is this the type of place we'd use `AzureNoReturnPath`?

gearama (approved):

ahsonkhan (approved):

sdk/core/azure-core/src/datetime.cpp
antkmsft: Maybe, the `_azure_UNREACHABLE_CODE()`. But I'd do it in a separate PR.

sdk/identity/azure-identity/src/managed_identity_credential.cpp
antkmsft: Maybe, I tend to like exceptions more, but maybe we should think about using the `_azure_UNREACHABLE_CODE()`. But I'd do it in a separate PR.

antkmsft: #3185

sdk/core/azure-core/src/datetime.cpp
antkmsft: #3185

PR #3186, opened 2021-12-15
Macro rename compatibility layer

antkmsft:
We made a breaking change, but if we still want to keep compatibility at least for some time, complete this PR before the Core release.
If not, we'll have to release all the updated packages in one vcpkg PR.
All the packages should depend on azure_core >= 1.4.0

We have 3 options, really:
1. Complete this PR, release core, release anything else as needed, remove this header in future core releases.
2. Don't go with this PR, release all packages (including all keyvault) in one vcpkg PR.
3. Revert the breaking change before anyone releases.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
+++ b/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
@@ -1,6 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
+#pragma once
+
+/**
+ * @file
+ * @brief Deprecated assert macros.
+ *
+ */
+
 #include "azure/core/internal/azure_assert.hpp"
 
 #define AZURE_ASSERT(exp) _azure_ASSERT(exp)
--- a/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
+++ b/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
@@ -5,7 +5,7 @@
 
 /**
  * @file
- * @brief Deprecated assert macros.
+ * @brief Deprecated assert macros. It is only temporarily made available for compatibility and should NOT be used by any callers outside of the SDK.
  *
  */
 
--- a/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
+++ b/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
@@ -5,7 +5,8 @@
 
 /**
  * @file
- * @brief Deprecated assert macros. It is only temporarily made available for compatibility and should NOT be used by any callers outside of the SDK.
+ * @brief Deprecated assert macros. It is only temporarily made available for compatibility and
+ * should NOT be used by any callers outside of the SDK.
  *
  */
 
--- a/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
+++ b/sdk/core/azure-core/inc/azure/core/azure_assert.hpp
@@ -1,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include "azure/core/internal/azure_assert.hpp"
+
+#define AZURE_ASSERT(exp) _azure_ASSERT(exp)
+#define AZURE_ASSERT_MSG(exp, msg) _azure_ASSERT_MSG(exp, msg)
+#define AZURE_ASSERT_FALSE(exp) _azure_ASSERT_FALSE(exp)
+#define AZURE_UNREACHABLE_CODE() _azure_UNREACHABLE_CODE()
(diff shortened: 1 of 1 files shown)


vhvb1989 (approved):

ahsonkhan:
Related to: https://github.com/Azure/azure-sdk-for-cpp/pull/3059

sdk/core/azure-core/inc/azure/core/azure_assert.hpp
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ahsonkhan, line 1: Shouldn't this be under `/internal/`?

ahsonkhan (changes requested):
Please hold off on merging until we can compare with what the state of the header and macro was before https://github.com/Azure/azure-sdk-for-cpp/pull/3059 and evaluate that this actually mitigates the break without introducing a need for another break in the future.

I thought it was an `/internal` header?

antkmsft:
@ahsonkhan No, this wasn't an internal header. It was exactly at this location: https://github.com/Azure/azure-sdk-for-cpp/blob/5e0f1c9ac888c8e55f677671a43ab18625d5974f/sdk/core/azure-core/inc/azure/core/azure_assert.hpp

sdk/core/azure-core/inc/azure/core/azure_assert.hpp
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+/**
+ * @file
+ * @brief Deprecated assert macros.
```
ahsonkhan, line 8: Maybe we should be more explicit in the comment.

```suggestion
 * @brief Deprecated assert macros. It is only temporarily made available for compatibility and should NOT be used by any callers outside of the SDK.
```

sdk/core/azure-core/inc/azure/core/azure_assert.hpp
```diff
@@ -0,0 +1,18 @@
...
+ *
+ */
+
+#include "azure/core/internal/azure_assert.hpp"
+
+#define AZURE_ASSERT(exp) _azure_ASSERT(exp)
+#define AZURE_ASSERT_MSG(exp, msg) _azure_ASSERT_MSG(exp, msg)
+#define AZURE_ASSERT_FALSE(exp) _azure_ASSERT_FALSE(exp)
+#define AZURE_UNREACHABLE_CODE() _azure_UNREACHABLE_CODE()
+#define AZURE_NOT_IMPLEMENTED() _azure_NOT_IMPLEMENTED()
```
ahsonkhan, line 18: Are all of these used within upstream consumers like storage? If some aren't, we could remove those.

sdk/core/azure-core/inc/azure/core/azure_assert.hpp
ahsonkhan: Oh, right, gotcha. I misread the other PR. Thanks!

sdk/core/azure-core/inc/azure/core/azure_assert.hpp
antkmsft: I checked prior public non-beta releases, all of them were used.

ahsonkhan:
We know everything in main works together well. We might want to test to increase the confidence that this change will work as-is to enable upstream libraries to ship on vcpkg, if they keep their dependency versions pinned to the older version. Particularly things like this:
https://github.com/Azure/azure-sdk-for-cpp/blob/5f97ac1b3d8b3ec356116796474350f228910aff/sdk/storage/azure-storage-common/vcpkg/vcpkg.json#L14-L18

antkmsft:
@ahsonkhan, I just summarized all 3 options we really have in the PR title. I don't think we have other options.
If we go with this PR, other packages will have to do `version>=` upgrades not this time, but the next time we go with the change that removes this header (after all the libraries shipped at least once, using the new `_azure_ASSERT` macro).

antkmsft:
If storage needs to go with the GA release, we will also need to release core `1.3.2`, or revert the breaking change.

ahsonkhan (approved):
We'll revert calling the internal macros in client libraries for now, until they are ready to remove the calls to the public ones, to help stage the changes, and decouple releases. This is something they might decide to never do:
https://github.com/Azure/azure-sdk-for-cpp/pull/3217

sdk/core/azure-core/inc/azure/core/azure_assert.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#pragma once
+
+/**
+ * @file
+ * @brief Deprecated assert macros. It is only temporarily made available for compatibility and
+ * should NOT be used by any callers outside of the SDK.
+ *
+ */
+
```
ahsonkhan, line 12: Since we brought these back, updated the changelog entry:
https://github.com/Azure/azure-sdk-for-cpp/pull/3255

Issue #3188, opened 2021-12-15
Broken CI gate ( gcc5 ) 


vhvb1989:

ahsonkhan:
`cmake build` step failing on `ValidateLive Linux_x64_gcc5_with_unit_test`


https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1251047&view=logs&j=0e119b68-c922-5ad7-2413-fa5b213ae9b8&t=502cd50d-53e8-51de-5d28-96b76afcaf7e&l=194
```text
2021-12-15T19:03:58.7394915Z [ 47%] Building CXX object sdk/core/azure-core/test/ut/CMakeFiles/azure-core-test.dir/sha_test.cpp.o
2021-12-15T19:03:58.8246737Z /mnt/vss/_work/1/s/sdk/core/azure-core/test/ut/retry_policy_test.cpp: In instantiation of ‘{anonymous}::RetryPolicyTest::RetryPolicyTest(const Azure::Core::Http::Policies::RetryOptions&, std::function<bool(const Azure::Core::Http::Policies::RetryOptions&, int, std::chrono::duration<long int, std::ratio<1l, 1000l> >&, double)>, std::function<bool(const Azure::Core::Http::RawResponse&, const Azure::Core::Http::Policies::RetryOptions&, int, std::chrono::duration<long int, std::ratio<1l, 1000l> >&, double)>)::<lambda(auto:1, auto:2, auto:3, auto:4)> [with auto:1 = Azure::Core::Http::Policies::RetryOptions; auto:2 = int; auto:3 = std::chrono::duration<long int, std::ratio<1l, 1000l> >; auto:4 = double]’:
[...]

PR #3190, opened 2021-12-15
Fix nightly tests compilation on GCC5

antkmsft:
Closes https://github.com/Azure/azure-sdk-for-cpp/issues/3188

### DIFF
--- a/sdk/core/azure-core/test/ut/retry_policy_test.cpp
+++ b/sdk/core/azure-core/test/ut/retry_policy_test.cpp
@@ -72,16 +72,17 @@ public:
       decltype(m_shouldRetryOnResponse) shouldRetryOnResponse)
       : RetryPolicy(retryOptions),
         m_shouldRetryOnTransportFailure(
-            shouldRetryOnTransportFailure != nullptr
+            shouldRetryOnTransportFailure != nullptr //
                 ? shouldRetryOnTransportFailure
                 : static_cast<decltype(m_shouldRetryOnTransportFailure)>( //
                     [this](auto options, auto attempt, auto retryAfter, auto jitter) {
                       retryAfter = std::chrono::milliseconds(0);
                       auto ignore = decltype(retryAfter)();
-                      return BaseShouldRetryOnTransportFailure(options, attempt, ignore, jitter);
+                      return this->BaseShouldRetryOnTransportFailure(
+                          options, attempt, ignore, jitter);
                     })),
         m_shouldRetryOnResponse(
-            shouldRetryOnResponse != nullptr
+            shouldRetryOnResponse != nullptr //
                 ? shouldRetryOnResponse
                 : static_cast<decltype(m_shouldRetryOnResponse)>( //
                     [this](
@@ -92,7 +93,8 @@ public:
                         auto jitter) {
                       retryAfter = std::chrono::milliseconds(0);
                       auto ignore = decltype(retryAfter)();
-                      return BaseShouldRetryOnResponse(response, options, attempt, ignore, jitter);
+                      return this->BaseShouldRetryOnResponse(
+                          response, options, attempt, ignore, jitter);
                     }))
   {
   }
--- a/sdk/core/azure-core/test/ut/retry_policy_test.cpp
+++ b/sdk/core/azure-core/test/ut/retry_policy_test.cpp
@@ -58,7 +58,8 @@ public:
           [this](auto options, auto attempt, auto retryAfter, auto jitter) {
             retryAfter = std::chrono::milliseconds(0);
             auto ignore = decltype(retryAfter)();
-            return RetryPolicy::ShouldRetryOnTransportFailure(options, attempt, ignore, jitter);
+            return this->RetryPolicy::ShouldRetryOnTransportFailure(
+                options, attempt, ignore, jitter);
           });
     }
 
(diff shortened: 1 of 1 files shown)


vhvb1989 (approved):

vhvb1989 (approved):

sdk/core/azure-core/test/ut/retry_policy_test.cpp
```diff
@@ -56,23 +56,29 @@ class RetryPolicyTest final : public RetryPolicy {
...
-            : [&](auto options, auto attempt, auto retryAfter, auto jitter) {
-              retryAfter = std::chrono::milliseconds(0);
-              auto ignore = decltype(retryAfter)();
-              return RetryPolicy::ShouldRetryOnTransportFailure(options, attempt, ignore, jitter);
-        }),
+        m_shouldRetryOnTransportFailure(
+            shouldRetryOnTransportFailure != nullptr
+                ? shouldRetryOnTransportFailure
+                : static_cast<decltype(shouldRetryOnTransportFailure)>(
+                    [&](auto options, auto attempt, auto retryAfter, auto jitter) {
```
ahsonkhan, line 63: Can you help me understand what the issue was and how this cast helps fix it?

antkmsft: GCC5 does not do implicit conversion from a lambda to `std::function`, so it thinks two sides of a ternary operator return different types, which is not allowed. Other thing, on GCC5, lambda `[&]` written within the member initializer, is not smart enough to grab `this` pointer and use it when resolving the function name.

antkmsft: So, in the end, the clearest form seems to be to re-initialize in the constructor body. It appears in GCC5, lambdas used within the initializer context are treated as if they are written outside of class context, so they think can't call protected functions.
I could've tried to write a static function that returns the std::function, and use it in the initializer, but I think the current way it is written, should be the most clear to understand.

sdk/core/azure-core/test/ut/retry_policy_test.cpp
```diff
@@ -55,25 +55,31 @@ class RetryPolicyTest final : public RetryPolicy {
...
-              retryAfter = std::chrono::milliseconds(0);
-              auto ignore = decltype(retryAfter)();
-              return RetryPolicy::ShouldRetryOnResponse(
-                  response, options, attempt, ignore, jitter);
-        })
-  {
+            };
+    }
+
+    if (m_shouldRetryOnResponse == nullptr)
```
ahsonkhan, line 71: Just a bit question, could this be written as follows?
```C++
if (!m_shouldRetryOnResponse)
```

sdk/core/azure-core/test/ut/retry_policy_test.cpp
```diff
@@ -55,25 +55,31 @@ class RetryPolicyTest final : public RetryPolicy {
...
           double)> shouldRetryOnResponse)
-      : RetryPolicy(retryOptions),
-      m_shouldRetryOnTransportFailure(
-          shouldRetryOnTransportFailure != nullptr
-            ? shouldRetryOnTransportFailure
-            : [&](auto options, auto attempt, auto retryAfter, auto jitter) {
+      : RetryPolicy(retryOptions), m_shouldRetryOnTransportFailure(shouldRetryOnTransportFailure),
+        m_shouldRetryOnResponse(shouldRetryOnResponse)
+  {
+    if (m_shouldRetryOnTransportFailure == nullptr)
```
ahsonkhan, line 61: Agreed, it's a bit cleaner this way.

sdk/core/azure-core/test/ut/retry_policy_test.cpp
antkmsft: For sure, it is a matter of taste. I prefer `==/!= nullptr`, for non-booleans, and `!isSomething` for booleans.

ahsonkhan (approved):
As long as it passes CI + nightly test run, and is readable, looks good to me :)

sdk/core/azure-core/test/ut/retry_policy_test.cpp
ahsonkhan: Yeah, seems reasonable. I think I have seen both in the codebase.

antkmsft: I think, especially here, when the callback is totally named, how a boolean could've been named: `shouldSomething`

sdk/core/azure-core/test/ut/retry_policy_test.cpp
```diff
@@ -45,34 +45,57 @@ class RetryPolicyTest final : public RetryPolicy {
...
         m_shouldRetryOnResponse(
-            shouldRetryOnResponse != nullptr
-            ? shouldRetryOnResponse
-            : [&](RawResponse const& response, auto options, auto attempt, auto retryAfter, auto jitter) {
-              retryAfter = std::chrono::milliseconds(0);
-              auto ignore = decltype(retryAfter)();
-              return RetryPolicy::ShouldRetryOnResponse(
-                  response, options, attempt, ignore, jitter);
-        })
+            shouldRetryOnResponse != nullptr //
```
ahsonkhan, line 85: empty comments?

sdk/core/azure-core/test/ut/retry_policy_test.cpp
```diff
@@ -45,34 +45,57 @@ class RetryPolicyTest final : public RetryPolicy {
...
+                ? shouldRetryOnResponse
+                : static_cast<decltype(m_shouldRetryOnResponse)>( //
+                    [this](
```
ahsonkhan, line 88: It's a bit difficult for me to parse and understand what this is doing without really spending some time on it, including the use of decltype, and where this std::function get's used. Some comments would be helpful, especially to indicate why it is written this way, so that future modifiers know not to hit the gcc5 issue.

Issue #2543, opened 2021-07-01
Consider obsoleting or removing the two param RFE ctor, there should be no callers


ahsonkhan:
The semantics of this API are confusing, and hence no one within the SDK calls it. End customers aren't expected to call it either.
https://github.com/Azure/azure-sdk-for-cpp/blob/d0ca42643f4c81261ecaa8ba0d75ceb7d3127a48/sdk/core/azure-core/inc/azure/core/exception.hpp#L91-L93

PR #3193, opened 2021-12-16
StorageException: Reuse code from RequestFailedException

janbernloehr:
This PR modifies the creation of `StorageException` to reuse code of its baseclass `RequestFailedException` to initialize the shared fields. In particular, this fixes the issue that `ErrorCode` was not set.

# Pull Request Checklist

- [x] Doxygen docs
- [ ] Unit tests
  - [x] PR is single purpose
  - [ ] Related issue listed
- [x] Comments in source
- [x] No typos
- [ ] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -145,10 +145,9 @@ namespace Azure { namespace Core {
      *
      */
     ~RequestFailedException() = default;
-
-  private:
-    std::string GetRawResponseField(
-        std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse,
-        std::string fieldName);
   };
+
+  std::string GetRawResponseField(
+      std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse,
+      std::string fieldName);
 }} // namespace Azure::Core
--- a/sdk/storage/azure-storage-common/inc/azure/storage/common/storage_exception.hpp
+++ b/sdk/storage/azure-storage-common/inc/azure/storage/common/storage_exception.hpp
@@ -38,5 +38,13 @@ namespace Azure { namespace Storage {
      */
     static StorageException CreateFromResponse(
         std::unique_ptr<Azure::Core::Http::RawResponse> response);
+
+  private:
+    StorageException(
+        const std::string& whatArg,
+        std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse)
+        : RequestFailedException(whatArg, std::move(rawResponse))
+    {
+    }
   };
 }} // namespace Azure::Storage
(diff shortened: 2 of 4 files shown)


sdk/storage/azure-storage-common/inc/azure/storage/common/storage_exception.hpp
```diff
@@ -38,5 +38,13 @@ namespace Azure { namespace Storage {
      */
     static StorageException CreateFromResponse(
         std::unique_ptr<Azure::Core::Http::RawResponse> response);
+
+  private:
+    StorageException(
+        const std::string& whatArg,
+        std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse)
+        : RequestFailedException(whatArg, std::move(rawResponse))
```
ahsonkhan, line 46: We were planning on removing this 2-param ctor from RFE. Is calling it necessary for this case?

https://github.com/Azure/azure-sdk-for-cpp/issues/2543

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -145,10 +145,9 @@ namespace Azure { namespace Core {
...
      */
     ~RequestFailedException() = default;
-
-  private:
-    std::string GetRawResponseField(
-        std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse,
-        std::string fieldName);
   };
+
+  std::string GetRawResponseField(
```
ahsonkhan, line 150: If we really want to make this public, which I am not sure we do atm, I don't think exposing it from outside of the scope of a class like RFE (or probably RawResponse), makes sense.

ahsonkhan (commented):
Thanks for the contribution @janbernloehr!

Can you please clarify what the issue is that you found with Error Code not being set, when it should be, and your expectations around that?

ahsonkhan:
@JeffreyRichter, @Jinming-Hu - what are your thoughts here?

JeffreyRichter:
The latest thinking (and what we're doing in Go) is that Core defines a single RequestFailedException and this is thrown whenever a HTTP response has a non-2xx status code. Inside this, there is the raw HTTP response, a field with the HTTP status code and a string field that returns an operation-specific error code. Azure has a standard error response see [here](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#handling-errors). The 2 fields I mention are the ONLY fields that are part of Azure error response contract and these are the ONLY 2 fields customers could reference in their code to make an error recovery decision. It is also useful to have the exception type's ToString() method return the same string that would be passed to the logger. We expect customers to call this method to record/log the failed exception in order to fix their code. 

How close can we get to this without breaking customers? Perhaps we make Core's RequestFailedException do what I mention above and then StorageException can simply derive off of it and inherit the behavior in its entirety? (And add other stuff if necessary to avoid breaks).

janbernloehr:
I sure would be interested to bring an improvement to master but my feeling is that we need more input from the team owning the library on how the change should be laid out.

PR #3215, opened 2022-01-06
Removing the two param RFE ctor, there should be no callers

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/2543


antkmsft (commented):
👍 drop it! The only thing that is using this is the test that I just recently wrote, and it is failing the PR.

antkmsft (changes requested):
Drop the entire `TEST(RequestFailedException, Message)` unit test - it was written to test this constructor.

sdk/core/azure-core/src/exception.cpp
```diff
@@ -16,9 +16,8 @@ using namespace Azure::Core::Http::_internal;
 namespace Azure { namespace Core {
 
   RequestFailedException::RequestFailedException(
-      const std::string& what,
       std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse)
-      : std::runtime_error(what)
+      : std::runtime_error("Received an HTTP unsuccessful status code.")
```
antkmsft, line 20: Delegate to `: RequestFailedException(std::string const&)`, but move its body from header to here (or there will be linking problems when building DLL).

ahsonkhan: I prefer to keep this change isolated, and avoid making too much shuffle to the header/src file.
Is this suggested change necessary for correctness to this PR? If not, maybe we can do this as a separate clean-up.

ahsonkhan: https://github.com/Azure/azure-sdk-for-cpp/pull/3315/commits/42a99cda0f3923a7fbec337117ba9592fbfba778

antkmsft: We are about to cut a beta release, there is probably no better time for that sort of a cleanup. That would be a small change.
With constructor delegation, the code becomes more maintainable, and we make all constructors non-inline which is better to prevent potential linking errors when maintaining.

If you decide to not add it here, please create a work item to clean it up.

antkmsft (approved):
Regarding delegating constructor, we are about to cut a beta release, there is probably no better time for that sort of a cleanup. That would be a small change.
With constructor delegation, the code becomes more maintainable, and we make all constructors non-inline which is better to prevent potential linking errors when maintaining.

If you decide to not add it here, please create a work item to clean it up.

sdk/core/azure-core/src/exception.cpp
antkmsft: I commented before I saw, sepaarate PR instead of a work item works too!

PR #3217, opened 2022-01-06
Update client libraries to use the original public assert macros

ahsonkhan:
Follow-up to https://github.com/Azure/azure-sdk-for-cpp/pull/3186, and reverts most of the upstream changes (outside of core) from https://github.com/Azure/azure-sdk-for-cpp/pull/3059.

This way client libraries can, on their own pace, update these whenever they are ready to release, and not tied to the release of Azure Core.

Depends on https://github.com/Azure/azure-sdk-for-cpp/pull/3311


gearama (approved):

Jinming-Hu (approved):

antkmsft (changes requested):
I thought we are going with the compatibility layer, until we maybe can remove it some time in the future. If so, libraries should be using the new macro. Rick was going to make a decision, have you heard from him?

**And if we're removing as you propose, we should remove the new macro header and the new macros, but keep the namespace change for the NoReturnPath.** There is no sense to keen the new header around then, it will only get more confusing with new thing in place and nothing using it, or if both macros are allowed to be used in the codebase at the same time.

If we're not removing but keep the compat layer, it is ok for the client libs to request the newest core version. We do it all the time for other internal features. Old libs will compile too.

ahsonkhan:
> Rick was going to make a decision, have you heard from him?

Yes, we discussed this in stand-up last week, and agreed to keep the compat layer and not break the macros.

> If we're not removing but keep the compat layer, it is ok for the client libs to request the newest core version. We do it all the time for other internal features. Old libs will compile too.

Each library SDK can do that independently whenever they wish to upgrade to the new macro, at their own pace, after azure core has GA'd and the new macros are available in that vcpkg registry. We don't have to do so eagerly, yet.

antkmsft:
If they are releasing next month, and core does not, then we can revert their code to old macros.

There are 2 ways at this point: revert the new header completely + update client libraries to use old, or keep everything as-is: all code base uses new macro, and for the old code, there is compat macro. Having the new macro, but the current code base using the old macro is the worst of both worlds. And even for what? Storage is not even releasing this month, and so far we're not expecting to not release core next month.

ahsonkhan:
Moving out to March since it only impacts client libraries (like storage) which we aren't shipping this milestone.

Jinming-Hu (approved):

PR #3228, opened 2022-01-10
Create Attestation SDK

LarryOsterman:
# Initial checkpoint for Microsoft Azure Attestation service SDK.

This pull request implements the skeleton of an SDK for the Microsoft Azure Attestation service. 

It implements the bare minimum functionality but defines patterns and practices which will be used throughout the SDK.

The current codebase only implements two APIs: `AttestSgxEnclave` and `AttestOpenEnclave`, other APIs will be implemented in other pull requests.

The current codebase implements model serialization functionality and implements Attestation Token validation functionality. It also implements the cryptographic functionality required for the SDK.

Note that the current implementation takes a dependency on OpenSSL. There is also the skeleton of a bcrypt based crypto layer, if it turns out that a general OpenSSL dependency cannot be applied.


CMakeSettings.json
```diff
@@ -15,6 +15,16 @@
...
+        },
+        {
+          "name": "LIB_EAY_RELEASE",
+          "value": "G:/Az/LarryO/vcpkg/packages/openssl_x64-windows-static/lib/libcrypto.lib",
+          "type": "FILEPATH"
+        },
+        {
+          "name": "LIB_EAY_DEBUG",
+          "value": "G:/Az/LarryO/vcpkg/packages/openssl_x64-windows-static/lib/libcrypto.lib",
+          "type": "FILEPATH"
```
vhvb1989, line 27: Do not bring local paths to the repo :P

Keep this change to be only on your side

sdk/attestation/azure-attestation.sln
```diff
@@ -0,0 +1,67 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
vhvb1989, line 2: Do not bring sln files

sdk/attestation/azure-security-attestation/test/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+include(GoogleTest)
+
+add_executable (
+     azure-security-attestation-test
+     ut/macro_guard.cpp
+     ut/crypto_test.cpp
+     ut/attestation_test.cpp
+)
+
```
vhvb1989, line 23:
```suggestion
create_per_service_target_build(azure-security-attestation-test attestation)
```

Add this so your test can be built and run on CI

CMakeSettings.json
LarryOsterman: @#$@#$@#$@#$ - that wasn't supposed to be in the pull request. removing.

sdk/attestation/azure-security-attestation/test/CMakeLists.txt
LarryOsterman: Added. Thanks Victor.

sdk/attestation/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+      - Name: ATTESTATION_AAD_URL
+        Value: https://NotRealAttestationInstanceaad.wus.attest.azure.net
+      - Name: AZURE_TENANT_ID
+        Value: "33333333-3333-3333-3333-333333333333"
+      - Name: AZURE_CLIENT_ID
+        Value: "non-real-client"
+      - Name: AZURE_CLIENT_SECRET
+        Value: "non-real-secret"
+      - Name: locationShortName
+        Value: imaginaryLocation
```
vhvb1989, line 46: Try with: 

```suggestion
        Value: "imaginaryLocation"
```

sdk/attestation/azure-security-attestation/test/ut/attestation_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+        m_endpoint = "https://shared" + shortLocation + "." + shortLocation + ".attest.azure.net";
+      }
+      else if (GetParam() == "Aad")
+      {
+        m_endpoint = GetEnv("ATTESTATION_AAD_URL");
+      }
+      else if (GetParam() == "Isolated")
+      {
+        m_endpoint = GetEnv("ATTESTATION_ISOLATED_URL");
+      }
```
vhvb1989, line 38: nit 

```suggestion
      std::string mode(GetParam());
      if (mode == "Shared")
      {
        std::string shortLocation(GetEnv("locationShortName"));
        m_endpoint = "https://shared" + shortLocation + "." + shortLocation + ".attest.azure.net";
      }
      else if (mode == "Aad")
      {
        m_endpoint = GetEnv("ATTESTATION_AAD_URL");
      }
      else if (mode == "Isolated")
      {
        m_endpoint = GetEnv("ATTESTATION_ISOLATED_URL");
      }
```

sdk/attestation/ci.yml
LarryOsterman: I tried that, it didn't help :(.

.devcontainer/devcontainer.json
```diff
@@ -18,14 +18,14 @@
   // Do not run as root. See https://aka.ms/vscode-remote/containers/non-root.
   "remoteUser": "azure-sdk-for-cpp",
   "containerEnv": {
-    "AZURE_CLIENT_ID": "",
-    "AZURE_CLIENT_SECRET": "",
-    "AZURE_KEYVAULT_HSM_URL": "",
-    "AZURE_KEYVAULT_URL": "",
+    "AZURE_CLIENT_ID": "272b264a-2ef7-4491-b833-344e673e3bfd",
+    "AZURE_CLIENT_SECRET": "45d480eb-6b2e-46f9-9708-b39de1977b0e",
```
vhvb1989: hope this is not real :P.  We don't want real values pushed to main.

If it is real, you might want to rotate the secret now, as it is already in git story

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <azure/core/internal/json/json.hpp>
+#include <shared_mutex>
+#include <string>
+
+#include "attestation_client_options.hpp"
+#include <azure/attestation/attestation_client_models.hpp>
```
vhvb1989, line 11: we use `#include < ... >`  for external includes and `#include " ... "`  for same-lib includes.

So, these both includes should use quotes.

The first one, it doesn't matter if the header is in the same path, we would use `#include "azure/attestation/attestation_client_options.hpp"`

eng/pipelines/templates/stages/platform-matrix.json
```diff
@@ -69,6 +69,40 @@
...
+            "TargetArchitecture": {
+                "x86": {
+                    "CMAKE_GENERATOR_PLATFORM": "Win32"
+                },
+                "x64": {
+                    "CMAKE_GENERATOR_PLATFORM": "x64",
+                    "VCPKG_DEFAULT_TRIPLET": "x64-uwp"
+                }
+            }
+        },
```
vhvb1989, line 105: this should be gone

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <azure/core/internal/http/pipeline.hpp>
```
vhvb1989, line 12: do not include `internal` in public headers

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace Azure { namespace Security { namespace Attestation {
+
+  using namespace Azure::Core;
```
vhvb1989, line 16: We don't use `using namespace ...` from headers, as it gets applied into customer code.

We only allow `using namespace ..` from cpp files

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+     * @param endpoint The URL address where the client will send the requests to.
+     * @param credential OPTIONAL The authentication method to use (required for TPM attestation).
+     * @param options The options to customize the client behavior.
+     */
+    explicit AttestationClient(
+        std::string const& endpoint,
+        std::shared_ptr<Core::Credentials::TokenCredential const> credential,
+        AttestationClientOptions options = AttestationClientOptions());
+
+    explicit AttestationClient(
```
vhvb1989, line 47: missing docs

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+        std::string const& endpoint,
+        AttestationClientOptions options = AttestationClientOptions())
+        : AttestationClient(endpoint, nullptr, options)
+    {
+    }
+
+    std::string ClientVersion() const;
```
vhvb1989, line 54: missing docs

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
...
+        : m_endpoint(attestationClient.m_endpoint), m_apiVersion(attestationClient.m_apiVersion),
+          m_pipeline(attestationClient.m_pipeline){};
+
+    /**
+     * @brief Retrieve the metadata signing certificates for this attestation instance.
+     *
+     * @returns Attestation Metadata.
+     */
+    Response<AttestationOpenIdMetadata> GetOpenIdMetadata(
+        Azure::Core::Context const& context = Azure::Core::Context::ApplicationContext);
```
vhvb1989, line 70: If the API won't change the client, it should be const as:
```suggestion
    Response<AttestationOpenIdMetadata> GetOpenIdMetadata(
        Azure::Core::Context const& context = Azure::Core::Context::ApplicationContext) const;
```

Add `const` to all API methods which won't mutate the client

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,122 @@
...
+     */
+    ::std::string Issuer;
+    ::std::string JsonWebKeySetUrl;
+    ::std::vector<::std::string> SupportedResponseTypes;
+    ::std::vector<::std::string> SupportedTokenSigningAlgorithms;
+    ::std::vector<::std::string> SupportedClaims;
+
+    AttestationOpenIdMetadata() {}
+    ~AttestationOpenIdMetadata() {}
+    AttestationOpenIdMetadata(AttestationOpenIdMetadata&& that) noexcept
```
vhvb1989, line 43: missing docs

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include <unordered_map>
+#include <vector>
+
+namespace Azure { namespace Security { namespace Attestation {
+  class AttestationClient;
+  /**
+   * @brief Contains identity and other basic properties of a Certificate.
+   *
+   */
+  struct AttestationOpenIdMetadata final
```
vhvb1989, line 31: nit. when adding a constructor, we use `class`.  It's a convention I've never liked within the team, but you'll hear about it

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,122 @@
...
+  {
+    /**
+     */
+    ::std::string Issuer;
```
vhvb1989, line 35: replace `::std` to `std`.  Or make all ocurrences of the file to follow `::std`  .  We don't use it like that, anyway

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,122 @@
...
+  struct AttestationSigningKey
+  {
+    std::string SigningPrivateKey; /// A PEM encoded RSA or ECDSA private key which will be used to
+                                   /// sign an attestation token.
+    std::string SigningCertificate; /// A PEM encoded X.509 certificate which will be sent to the
+                                    /// attestation service to validate an attestation token. The
+                                    /// public key embedded in the certificate MUST be the public
+                                    /// key of the SigningPrivateKey.
+  };
+}}} // namespace Azure::Security::Attestation
```
vhvb1989, line 122:
```suggestion
}}} // namespace Azure::Security::Attestation

```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * @file
+ * @brief Defines the supported options to create a Attestation clients.
+ *
+ */
+
+#pragma once
+
+#include <azure/core/internal/client_options.hpp>
```
vhvb1989, line 12: This is the only internal header we are currently allowed to leak :)

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -0,0 +1,123 @@
...
+    AttestationDataType DataType;
+  };
+
+  struct AttestOptions final
+  {
+    AttestationData RuntimeData;
+    AttestationData InittimeData = {};
+    std::string DraftPolicyForAttestation = {};
+  };
+}}} // namespace Azure::Security::Attestation
```
vhvb1989, line 123:
```suggestion
}}} // namespace Azure::Security::Attestation

```

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+  std::vector<std::unique_ptr<HttpPolicy>> perRetrypolicies;
+  if (credential)
+  {
+    //      Azure::Core::Credentials::TokenRequestContext const tokenContext
+    //          = {{_internal::UrlScope::GetScopeFromUrl(m_vaultUrl)}};
+
+    //      perRetrypolicies.emplace_back(
+    //          std::make_unique<BearerTokenAuthenticationPolicy>(credential,
+    //          std::move(tokenContext)));
```
vhvb1989, line 40: ?

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+Azure::Response<std::vector<AttestationSigner>>
+AttestationClient::GetAttestationSigningCertificates(Azure::Core::Context const& context)
+{
+  auto request
+      = AttestationCommonRequest::CreateRequest(m_endpoint, HttpMethod::Get, {"certs"}, nullptr);
+
+  auto response = AttestationCommonRequest::SendRequest(*m_pipeline, request, context);
+  auto jsonWebKeySet(JsonWebKeySetSerializer::Deserialize(response));
```
vhvb1989, line 71: It is good for readability to know that is going to be the type.  `auto` is amazing, but in cases like this, hides the type because we don't know what Deserialize returns.

it's like `auto something(GetSomething())` v/s `std::String something(GetSomething());`

vhvb1989 (commented):
partial review

sdk/attestation/azure-security-attestation/CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+
+include(AzureVcpkg)
+include(AzureVersion)
+include(AzureCodeCoverage)
+include(AzureTransportAdapters)
+include(AzureDoxygen)
+include(AzureGlobalCompileOptions)
+include(AzureConfigRTTI)
+include(AzureBuildTargetForCI)
+
```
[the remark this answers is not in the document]
LarryOsterman: What's a "map"? And how do I know if a component creates one?

sdk/attestation/azure-security-attestation/CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+)
+
+#if(WIN32)
+#set(
+#    AZURE_ATTESTATION_SOURCE
+#    ${AZURE_ATTESTATION_SOURCE}
+#    src/private/crypto/bcrypt/bcryptcrypto.cpp
+#    src/private/crypto/bcrypt/bcryptkeys.cpp
+#    src/private/crypto/bcrypt/bcryptcert.cpp)
+#else()
```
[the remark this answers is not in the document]
LarryOsterman: Until I'm done with the attestation SDK, I don't want to throw away the work I did for the bcrypt crypto layer. Once I'm done and I decide I don't need them but until then, I'd like to keep them so I don't lose two days of work.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <azure/core/internal/json/json.hpp>
```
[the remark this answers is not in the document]
LarryOsterman: That's going to be ... interesting for internal headers.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: The rule I've used is "if it should come from `-I` or `INCLUDES`, use <>, otherwise use "". That's the semantics the compiler uses, I believe (it's close)

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: Docs pass is a future work item

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: Most of my API methods will mutate the client,. But adding const anyway.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
LarryOsterman: IIRC I added the std because of a compilation error earlier. Not seing it now though.

eng/pipelines/templates/stages/platform-matrix.json
vhvb1989, line 105: remove this

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
LarryOsterman: Curious as to what the suggestion here is - I don't see a difference (or did you add a crlf at the end of the line?)

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: THis was cut and pasted from KV, it's template logic for when I add the ability to have authenticated clients (currently all clients are unauthenticated). Since I don't know what to do, I left the KV boiler plate and I'll fix it later.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: FooSerializer::Serialize take a foo and returns a string - always. FooSerializer::Deserialize take a response or json and returns a Foo - always.

This is a convention applied to all my model serializer types.

OTOH, the SendRequest response type should be called out explicitly.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
vhvb1989: I used to do it in another way as well.  Then @RickWinter made it a guideline https://vhvb1989.github.io/azure-sdk/cpp_implementation.html#cpp-design-physical-include-quotes

And we all follow and honor the guidelines because its something beyond the good, the bad and the ugly  =D

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
vhvb1989: > I feel fairly strongly that struct == class with all members public.

<3 <3 <3  Where have you been all this time Larry!!! hahahaha   Now we just need to convince @RickWinter @ahsonkhan and @antkmsft about it.  ( I don't remember where's @gearama 's stand about this ... )

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
vhvb1989: Yes, just the end of file new line

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
vhvb1989: I won't make a big deal about it 😄

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
ahsonkhan: I think we are on the same page of what's the difference between struct vs class in C++, Victor :)

There's also the following, that Larry said, which largely holds true as well.
> More importantly, model types should always be structs.

The convention was, for model and option types, which almost always just contain public members ("properties"), or option bags, and are otherwise stateless, they should be structs. Everything else, including things with functions, and state ("private fields"), should be classes.

I guess the question here is about "always" vs "usually". I don't see a ctor here, as expected, so struct makes sense, like all other typical model types.

sdk/attestation/azure-security-attestation/CMakeLists.txt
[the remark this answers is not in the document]
danieljurek: I have not added docs. Here's what to do: 

* Map files can only be produced when there's an executable unit of code (e.g. a test or sample program) not a static library.
* In relevant CMakeLists.txt files add the line (the `<target-name>` should be globally unique): 
```cmake
create_map_file(<target-name> <target-name>.map)
```

This will produce a map file with the name of your target. Example: https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core-test/CMakeLists.txt#L49

* The engineering system automatically scoops up all `**/*.map` files in the repo and uploads them as an artifact for the build. The folder structure is flattened so `.map` files end up in the same artifact folder. Make sure the target name is globally unique or you'll get undefined behavior.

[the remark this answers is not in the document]
LarryOsterman: Oh - MAP files are the classic text dump of symbols? not a cmake construct? That makes so much more sense.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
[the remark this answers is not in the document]
LarryOsterman: All parsing is done in headers, not cpp files. That allows the compiler to do a better job of inlining (LTCG and comdat folding helps but putting the code in headers is generally better, IMHO).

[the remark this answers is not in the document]
vhvb1989: We've had issues in the past by having implementation in headers.

Something related to dynamic linking.  There were some limits about what to do implement and what to avoid implementing on headers.

There was also something related to making compilation time longer.

Anyway, you could create a private header (within the src/ folder) and add inline implementation to it.  Then consume the file from the client.cpp implementation.  But avoid bringing the json header to public client, since the json header is huge (it's all the lib in a header).

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
LarryOsterman: I nuked the ctor in the most recent revision because it wasn't needed (I wasn't seeing the default move ctor being called when I debugged the code so I was trying to ensure it was present).

However my general principle holds: Model types (outputs of the API calls) should be structs. They also should be immutable (all fields are const). That last one drives to an interesting problem: Initializing const fields can only be done through a ctor. The default move and copy ctors will likely work well, but that isn't the case for initializing (this is especially true for complicated model types (ten-twenty fields).

I totally agree that model types SHOULD NOT (RFC 2119) have methods. But I think a ctor makes sense if we have immutable model types. Let's take this offline to email rather than GH comments - it's not germane to this particular PR right now, but may be in the future (one of my model types has 17 fields)

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: FWIW, I'll make this more explicit in the documentation for the serializer methods.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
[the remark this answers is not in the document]
LarryOsterman: Raw functions in headers breaks when you have headers in multiple files. Putting the functions in a class resolves that issue. 

And my parsing functions are in a private header - they're COMPLETELY not a part of the public contract.

sdk/attestation/azure-security-attestation/cgmanifest.json
```diff
@@ -0,0 +1,36 @@
...
+            },
+            "DevelopmentDependency": true
+        },
+        {
+            "Component": {
+                "Type": "other",
+                "Other": {
+                    "Name": "doxygen",
+                    "Version": "1.8.20",
+                    "DownloadUrl": "http://doxygen.nl/files/doxygen-1.8.20-setup.exe"
```
danieljurek, line 30: This is the same as in other cgmanfiest.json files around the repo... However, in order to prevent MITM attacks we've been downloading doxygen from our own storage account via HTTPS (this package was initially downloaded from SourceForge via HTTPS) 

https://github.com/Azure/azure-sdk-for-cpp/blob/c10de791cceeb393694e86c912267d5faa89bb94/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L84 

To that end, I think `DownloadUrl` should be `https://azuresdktooling.blob.core.windows.net/dependencies/doxygen-1.8.18-setup.exe` if we intend to accurately reflect where we fetch the file. 

I think @vhvb1989 worked on cgmanifest files... Is there a reason we use this http URL instead of the URL from which we're hosting the file?

eng/pipelines/templates/steps/vcpkg.yml
```diff
@@ -33,6 +33,7 @@ steps:
         $(Agent.JobName)
         | "$(VcpkgCommit)"
         | $(Agent.Os)
+        | variables['${{ parameters.DependenciesVariableName }}']
```
danieljurek, line 36: This is the right way to go. Our cache will be more useful this way if the dependencies vary based on the job executing. 

However, I think this only outputs the literal string: 

`| variables['vcpkg.deps']` 

To get this to output the list of dependencies we should expand the variable using [macro syntax](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#macro-syntax-variables) as it is an input to a task. 

```suggestion 
        | $(${{ parameters.DependenciesVariableName }})
```

This syntax is annoying to read but it basically says: 

`${{ parameters.DependenciesVariableName }}` -- Before expanding the YAML into steps, rewrite the YAML with this string replaced with the parameter value. Result: `vcpkg.deps` (or whatever is set for that parameter) 

`$(x)` -- At runtime, substitute the variable whose name is `x` into the string that is being built for this task. 

Since the `${{ parameters.___ }}` syntax mutates the YAML that will be processed we'll end up writing the variable name into the macro syntax which will then insert the variable value in the string.

danieljurek (commented):
A couple of comments. Thanks for updating (AND SORTING!!!) the word list :)

eng/pipelines/templates/steps/vcpkg.yml
LarryOsterman: That makes sense to me, let me try it and see what it does.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,135 @@
...
+  struct AttestationOpenIdMetadata final
+  {
+    /// The issuer which will be used for tokens generated by this instance.
+    std::string Issuer;
+
+    /// A URI which can be used to retrieve the AttestationSigner
+    /// objects returned by the attestation service.
+    std::string JsonWebKeySetUrl;
+
+    /// The response types that are supported by the service.
```
gearama, line 41: comments not in proper format 
/**
     * @brief Define policies to be called one time for every HTTP request from an SDK client.
     *
     */

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -0,0 +1,224 @@
...
+  /// attestation SDK.
+  using TokenValidationCallbackFn = std::function<
+      void(std::string const& rawToken, Models::AttestationSigner const& tokenSigner)>;
+
+  /** @brief The AttestationTokenValidationOptions represents a set of options which control how
+   * attestation tokens are validated. */
+  struct AttestationTokenValidationOptions final
+  {
+    /// Controls whether attestation tokens are validated at all.
+    ///
```
gearama, line 62: /**
     * @brief Define policies to be called one time for every HTTP request from an SDK client.
     *
     */

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+    /// Nonce which is sent to the attestation service to allow a caller to prevent replay attacks.
+    Azure::Nullable<std::string> Nonce{};
+
+    /// A test hook which allows developers to test attestation policies before they commit them to
+    /// the service.
+    Azure::Nullable<std::string> DraftPolicyForAttestation{};
+
+    /// Specifies the options which should be used to validate the attestation token returned
+    /// by the attestation service. If not provided by the caller, the token validation options
```
gearama, line 200: /**
     * @brief Define policies to be called one time for every HTTP request from an SDK client.
     *
     */

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,158 @@
...
+AttestationClient::AttestationClient(
+    std::string const& endpoint,
+    std::shared_ptr<Core::Credentials::TokenCredential const> credential,
+    AttestationClientOptions options)
+    : m_endpoint(endpoint), m_tokenValidationOptions(options.TokenValidationOptions)
+{
+  std::vector<std::unique_ptr<HttpPolicy>> perRetrypolicies;
+  if (credential)
+  {
+    //      Azure::Core::Credentials::TokenRequestContext const tokenContext
```
gearama, line 37: remove commented out code

sdk/attestation/azure-security-attestation/src/private/attestation_client_models_private.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+    static std::string PemFromX5c(std::string const& x5c);
+    operator AttestationSigner&&() { return std::move(m_signer); }
+    static std::string SerializeToJson(AttestationSigner const& signer);
+  };
+
+  struct AttestSgxEnclaveRequest
+  {
+    std::vector<uint8_t> Quote;
+    Azure::Nullable<AttestationData> InitTimeData;
```
gearama, line 79: probably need comments here

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
```diff
@@ -0,0 +1,390 @@
...
+              Models::AttestationSigner{Azure::Nullable<std::string>(), pemEncodedChain});
+        }
+      }
+      return returnValue;
+    }
+
+    /// @brief Given a set of possible signers, verify that the token was signed by one of those
+    /// signers.
+    /// @param possibleSigners - a list of possible AttestationSigner objects that could have
+    /// signed the token.
```
gearama, line 171: comments format here

gearama (commented):
I'm not familiar with this SDK so don't hold the following comments against me. Feel free to ignore them if they don't apply. 

Could we break it up in a few PRs ? we've done it before for other big PRs.
first one that lays down the file/folder structure and the basic cmake files, another on that adds the encrytion part etc etc.
83 files to review is a bit of a stretch to actually understand what it's doing. 
I suspect you would want more than basic comment check and code layout out of this PR. :)

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: At some point in the near future, this code will become live - it's currently a placeholder for what I hope to build in the next set of work items.

LarryOsterman:
> I'm not familiar with this SDK so don't hold the following comments against me. Feel free to ignore them if they don't apply.
> 
> Could we break it up in a few PRs ? we've done it before for other big PRs. first one that lays down the file/folder structure and the basic cmake files, another on that adds the encrytion part etc etc. 83 files to review is a bit of a stretch to actually understand what it's doing. I suspect you would want more than basic comment check and code layout out of this PR. :)

It doesn't help much, but 20 of the 83 files are recordings and thus somewhat irrelevant to the review. But there are still 60 or so files to be reviewed.

This PR is what I consider MVP for the attestation SDK - it implements the simplest API in the SDK but builds the infrastructure which will be used for all subsequent APIs.

What I'm primarily looking for here is twofold:

1.  Detailed review of the files *outside* the attestation SDK. These are core files that affect other SDKs and thus it's critical that it be correct.
2. Coding style and overall feedback on the conventions being used. This is a completely new area to me and I don't know all the conventions (things like "doxygen comments need to use C style multi-line comments and not single line comments").

Other areas that may bear attention are the Crypto APIs (src/private/crypto - 4 files) and the serializers (src/private/attestation_deserializers_private (2 files)).

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <shared_mutex>
+#include <string>
+
+#include "attestation_client.hpp"
+#include "attestation_client_models.hpp"
+#include "attestation_client_options.hpp"
```
vhvb1989, line 11: use the form #include "azure/attestation/attestation_client_models.hpp"

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <shared_mutex>
+#include <string>
+
+#include "attestation_client.hpp"
```
vhvb1989, line 9: this is confusing..  is it the same file including itself?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+
+namespace Azure { namespace Core { namespace Http { namespace _internal {
+  class HttpPipeline;
+}}}} // namespace Azure::Core::Http::_internal
+
+namespace Azure { namespace Security { namespace Attestation {
+  /** @brief An AttestationTokenHeader represents common properties in an the RFC 7515 JSON Web
+   * Token.
+   */
+  struct AttestationTokenHeader
```
vhvb1989, line 21: move to models?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+     *
+     */
+    Azure::Nullable<std::vector<std::string>> X509CertificateChain;
+  };
+
+  /** An AttestationToken represents an RFC 7519 JSON Web Token returned from the attestation
+   * service with the specialized body type.
+   * <typeparam name="T"></typeparam> The type which represents the body of the attestation token.
+   */
+  template <typename T> class AttestationToken final {
```
vhvb1989, line 121: move to models?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+   *
+   * The Attestation Signing Keys describe the keys which will be used to sign tokens generated by
+   * the attestation service. All tokens emitted by the attestation service will be signed by one
+   * of the certificates listed in the attestation signing keys.
+   *
+   */
+
+  class AttestationClient final {
+  private:
+    Azure::Core::Url m_endpoint;
```
vhvb1989, line 306: Include <azure/core/url.hpp>

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+        : AttestationClient(endpoint, nullptr, options)
+    {
+    }
+
+    /**
+     * @brief Returns the version the client was configured with.
+     *
+     * @returns The API version used when communicating with the attestation service.
+     */
+    std::string ClientVersion() const;
```
vhvb1989, line 347: nit. return a `std::string const&` instead of `std::string` to let caller decide if they want to make a copy or just read the value.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+    /**
+     * @brief Construct a new Key Client object from another key client.
+     *
+     * @param attestationClient An existing attestation client.
+     */
+    explicit AttestationClient(AttestationClient const& attestationClient)
+        : m_endpoint(attestationClient.m_endpoint), m_apiVersion(attestationClient.m_apiVersion),
+          m_pipeline(attestationClient.m_pipeline){};
```
vhvb1989, line 355: what about the clientOptions?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+    /**
+     * Retrieves metadata about the attestation signing keys in use by the attestation service.
+     *
+     * Retrieve the OpenID metadata for this attestation service instance..
+     *
+     * @return an {@link AttestationOpenIdMetadata} containing metadata about the specified service
+     * instance.
+     */
+    Response<Models::AttestationOpenIdMetadata> GetOpenIdMetadata(
+        Azure::Core::Context const& context = Azure::Core::Context::ApplicationContext) const;
```
vhvb1989, line 366: include <azure/core/context.hpp>

sdk/attestation/azure-security-attestation/CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+    AZURE_ATTESTATION_SOURCE
+    ${AZURE_ATTESTATION_SOURCE}
+    src/private/crypto/openssl/opensslcrypto.cpp
+    src/private/crypto/openssl/opensslkeys.cpp
+    src/private/crypto/openssl/opensslcert.cpp)
+#endif()
+
+create_per_service_target_build(azure-security-attestation attestation)
+
+add_library(azure-security-attestation ${AZURE_ATTESTATION_HEADER} ${AZURE_ATTESTATION_SOURCE} "src/private/attestation_client_private.hpp")
```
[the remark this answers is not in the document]
LarryOsterman: I have literally no idea. But if I try to use PUBLIC, cmake complains that it cannot find source file PUBLIC.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * @brief Retrieve the attestation signing certificates for this attestation instance.
+     *
+     * @returns Attestation Metadata.
+     */
+    Response<std::vector<Models::AttestationSigner>> GetAttestationSigningCertificates(
```
vhvb1989, line 373: might want to wrap the vector inside a `AttestationSigningCertificageResult` struct as a way to support returning more than just the vector in the future.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,410 @@
...
+     * service.
+     * @param options - Options to the attestation request (runtime data, inittime data, etc).
+     * @param context - Context for the operation.
+     *
+     * @returns Response<AttestationToken<AttestationResult>> - The result of the attestation
+     * operation
+     */
+    Response<AttestationToken<Models::AttestationResult>> AttestOpenEnclave(
+        std::vector<uint8_t> const& openEnclaveReportToAttest,
+        AttestOptions options = AttestOptions(),
```
vhvb1989, line 406: might want to consider having operation-options unique per operation.  Right now, `OpenEnclave` and `SgxEnclave` are getting marry forever :)

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ * @brief Defines the Azure Attestatoin API types.
```
vhvb1989, line 6: typo

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,135 @@
...
+    /// to sign attestation tokens.
+    std::vector<std::string> SupportedTokenSigningAlgorithms;
+
+    /// A list of claims which may be returned by the attestation service.
+    std::vector<std::string> SupportedClaims;
+  };
+
+  /** @brief An AttestationSigner represents an X .509 certificate and KeyID pair.
+   *
+   * There are two use scenarios for an AttestationSigner:
```
vhvb1989, line 54: might want to use @note or @detail to highlight this info

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+    /// The security version number of the SGX enclave.
+    Azure::Nullable<int> Svn;
+
+    /// A JSON encoded string representing the collateral which was used
+    /// to perform the attestation operation.
+    std::string SgxCollateral;
+  };
+
+}}}} // namespace Azure::Security::Attestation::Models
```
vhvb1989, line 135: add new line

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto response = AttestationCommonRequest::SendRequest(*m_pipeline, request, context);
+  std::string responseToken = AttestationServiceTokenResponseSerializer::Deserialize(response);
+  auto token
+      = AttestationTokenInternal<AttestationResult, AttestationResultSerializer>(responseToken);
+  return Response<AttestationToken<AttestationResult>>(token, std::move(response));
+}
+
+void AttestationClient::CacheAttestationSigners(Azure::Core::Context const& context) const
+{
+  std::unique_lock<std::shared_timed_mutex> stateLock(m_sharedStateLock);
```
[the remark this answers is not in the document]
LarryOsterman: I'm confused. This impl *is* in a cpp file. Where should I be moving it to?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
vhvb1989, line 135: add new line

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include "azure/attestation/attestation_client.hpp"
+#include <shared_mutex>
+
+#include "azure/core/base64.hpp"
+#include "azure/core/http/policies/policy.hpp"
+#include "azure/core/internal/http/pipeline.hpp"
```
vhvb1989, line 9: change quotes for diamond braces <>

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    //      perRetrypolicies.emplace_back(
+    //          std::make_unique<BearerTokenAuthenticationPolicy>(credential,
+    //          std::move(tokenContext)));
+  }
+  m_apiVersion = options.Version.ToString();
+  std::vector<std::unique_ptr<HttpPolicy>> perCallpolicies;
+
+  m_pipeline = std::make_shared<Azure::Core::Http::_internal::HttpPipeline>(
+      options,
```
vhvb1989, line 48: the `AttestationClientOptions` contains a `TokenOptions` right?  Where are you persisting those options insde the client?

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto token
+      = AttestationTokenInternal<AttestationResult, AttestationResultSerializer>(responseToken);
+
+  // Validate the token returned by the service. Use the cached attestation signers in the
+  // validation.
+  CacheAttestationSigners(context);
+  token.ValidateToken(
+      options.TokenValidationOptions.HasValue() ? options.TokenValidationOptions.Value()
+                                                : this->m_tokenValidationOptions,
+      m_attestationSigners);
```
vhvb1989, line 118: Why do we need to validate here? Is that something that the Service ask customers to do as per documentation?  or is this just the SDK trying to exceed expectations?

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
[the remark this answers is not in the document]
vhvb1989: my bad.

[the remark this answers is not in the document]
vhvb1989: However, you can create the mutex within this cpp file ( use unamed namespace ) and then remove the mutex field from the the class in the hpp

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: It's not a model type by the definition of model types, so why should it be in models?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: It's not strictly a model type, should it still go in models?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: There are no client options, but I did forget tokenvalidationoptions.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: You told me in an email that the convention was that all files in the same repo used "" and not <>.

Azure/Core is in the same repo as Azure/Attestation, so shouldn't it use the "" convention?

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: See line 32 of attestation_client.cpp.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: We should probably discuss this offline. Essentially the answer is: The client needs to be able to ensure that the token returned from the attestation service was actually generated by an SGX enclave contained within the attestation service. The token validation options defines a set of checks which are used to validate the token. There's one step that cannot be performed by the SDK (for <reasons>), that's the purpose of the client callback which allows customers to perform additional validations beyond the basic validations performed by the service.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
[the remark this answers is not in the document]
LarryOsterman: How do I scope that unnamed namespace mutex to the specific instance of the attestation client? Otherwise we serialize all clients on the mutex, which can lead to boxcar issues (maybe not a huge deal in practice, but I wouldn't want to use that as an examplar of how to design multi-threaded access).

sdk/attestation/azure-security-attestation/src/private/jsonhelpers.cpp
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  void JsonHelpers::SetField(
+      Azure::Core::Json::_internal::json& object,
+      Azure::Core::Json::_internal::json& fieldValue,
+      std::string const& fieldName)
+  {
+    object[fieldName] = fieldValue;
+  }
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 299: new line. But, if you just use JsonOptional, I don't think you need this file at all.

sdk/attestation/azure-security-attestation/src/private/jsonhelpers_private.hpp
```diff
@@ -0,0 +1,128 @@
...
+    static void SetField(
+        Azure::Core::Json::_internal::json& object,
+        Azure::DateTime const& fieldValue,
+        std::string const& fieldName);
+    static void SetField(
+        Azure::Core::Json::_internal::json& object,
+        Azure::Core::Json::_internal::json& fieldValue,
+        std::string const& fieldName);
+  };
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 128: JsonOptional would cover all this scenarios.

sdk/attestation/cmake_install.cmake
```diff
@@ -0,0 +1,40 @@
+# Install script for directory: G:/Az/LarryO/azure-sdk-for-cpp/sdk/attestation
```
vhvb1989, line 1: remove this file

sdk/attestation/azure-security-attestation/src/attestation_common_request.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+Azure::Core::Http::Request AttestationCommonRequest::CreateRequest(
+    Azure::Core::Url url,
+    Azure::Core::Http::HttpMethod method,
+    std::vector<std::string> const& path,
+    Azure::Core::IO::BodyStream* content)
+{
+  using namespace Azure::Core::Http;
+  return CreateRequest(url, "", method, path, content);
```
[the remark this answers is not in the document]
LarryOsterman: Some APIs we send require an API version, others do not (because the standard for that API does not include an API version). Rather than have a "CreateRequest" and "CreateUnversionedRequest", I put the API version in each API when it's needed.

sdk/attestation/azure-security-attestation/CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+
+#if(WIN32)
+#    find_package(WIL CONFIG REQUIRED)
+#    target_link_libraries(WIL::WIL)
+#    target_link_libraries(azure-core PRIVATE bcrypt crypt32)
+#else()
+    # Required for Hashing ( md5 and sha ).
+    find_package(OpenSSL REQUIRED)
+    target_link_libraries(azure-core PUBLIC OpenSSL::Crypto OpenSSL::SSL)
```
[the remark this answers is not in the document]
LarryOsterman: What should I link to other than azure-core?

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,280 @@
...
+   * The Deserialize method takes an instance of a json object and returns an instance of the Xxxx
+   * type.
+   *
+   */
+
+  AttestationOpenIdMetadata OpenIdMetadataSerializer::Deserialize(
+      std::unique_ptr<Azure::Core::Http::RawResponse>& response)
+  {
+    Models::AttestationOpenIdMetadata returnValue;
+    auto parsedBody = Azure::Core::Json::_internal::json::parse(response->GetBody());
```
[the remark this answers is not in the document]
LarryOsterman: GetBody returns a reference to the underlying storage. Is there still a copy made?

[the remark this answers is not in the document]
LarryOsterman: FWIW, I single stepped through the code - there are no temporaries created between the call to GetBody and the invocation of the parser in the JSON library.

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,280 @@
...
+      Azure::Core::Json::_internal::json const& parsedBody)
+  {
+    if (!parsedBody.contains("token"))
+    {
+      throw std::runtime_error("Field 'token' not found in Attestation Service Response");
+    }
+    if (!parsedBody["token"].is_string())
+    {
+      throw std::runtime_error("Field 'token' is not a string");
+    }
```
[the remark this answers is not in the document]
LarryOsterman: The code is going to throw if this value is not present in the response from the service. It's not optional.

The only question is if the exception thrown makes sense or if it is completely meaningless to the customer.

I could change the deserializer to just call parsedBody["token"].get<std::string>() but that's not going to generate a comfortable experience for the developer if somethign goes wrong.

sdk/attestation/azure-security-attestation/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
+      $<INSTALL_INTERFACE:include>
+      # Disable dependency on azure-core until azure-core has released to vcpkg
+      # ${azure-core-cpp_INCLUDE_DIRS}
+    PRIVATE
+      "src/private"
+)
+
+# Disable dependency on azure-core until azure-core has released to vcpkg
+target_link_libraries(azure-security-attestation PUBLIC Azure::azure-core)
```
vhvb1989, line 101: Here, Link crypto only to your library, 

```suggestion
target_link_libraries(azure-security-attestation PUBLIC Azure::azure-core OpenSSL::Crypto OpenSSL::SSL)
```

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
```diff
@@ -0,0 +1,410 @@
...
+      // Now check the expiration time
+      ValidateTokenTimeElements(validationOptions);
+
+      // And finally check the issuer.
+      ValidateTokenIssuer(validationOptions);
+    }
+
+    operator AttestationToken<T>&&() { return std::move(m_token); }
+  };
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 410:
```suggestion
}}}} // namespace Azure::Security::Attestation::_detail

```

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,297 @@
...
+   * The Deserialize method takes an instance of a json object and returns an instance of the Xxxx
+   * type.
+   *
+   */
+
+  AttestationOpenIdMetadata OpenIdMetadataSerializer::Deserialize(
+      std::unique_ptr<Azure::Core::Http::RawResponse>& response)
+  {
+    Models::AttestationOpenIdMetadata returnValue;
+    auto parsedBody = Azure::Core::Json::_internal::json::parse(response->GetBody());
```
vhvb1989, line 56: nit
```suggestion
    auto const parsedBody = Azure::Core::Json::_internal::json::parse(response->GetBody());
```

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,297 @@
...
+          return {
+              {"data", Azure::Core::_internal::Base64Url::Base64UrlEncode(value.Data)},
+              {"dataType", value.DataType.ToString()}};
+        });
+
+    JsonOptional::SetFromNullable(request.Nonce, serializedRequest, "nonce");
+    JsonOptional::SetFromNullable(
+        request.DraftPolicyForAttestation, serializedRequest, "draftPolicyForAttestation");
+    return serializedRequest.dump();
+  }
```
vhvb1989, line 99: would be worth it to define the function and re-use instead of defining it twice:


```suggestion
  std::string AttestSgxEnclaveRequestSerializer::Serialize(AttestSgxEnclaveRequest const& request)
  {
    Azure::Core::Json::_internal::json serializedRequest;
    serializedRequest["quote"] = Azure::Core::_internal::Base64Url::Base64UrlEncode(request.Quote);

    auto decorator = [](AttestationData const& value) -> std::map<std::string, std::string> {
          return {
              {"data", Azure::Core::_internal::Base64Url::Base64UrlEncode(value.Data)},
              {"dataType", value.DataType.ToString()}};
        };

    JsonOptional::SetFromNullable<AttestationData, std::map<std::string, std::string>>(
        request.RunTimeData,
        serializedRequest,
        "runtimeData",
        decorator
        );

    JsonOptional::SetFromNullable<AttestationData, std::map<std::string, std::string>>(
        request.InitTimeData,
        serializedRequest,
        "inittimeData",
        decorator
        );

    JsonOptional::SetFromNullable(request.Nonce, serializedRequest, "nonce");
    JsonOptional::SetFromNullable(
        request.DraftPolicyForAttestation, serializedRequest, "draftPolicyForAttestation");
    return serializedRequest.dump();
  }
```

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    JsonOptional::SetFromNullable<AttestationData, std::map<std::string, std::string>>(
+        request.RunTimeData,
+        serializedRequest,
+        "runtimeData",
+        [](AttestationData const& value) -> std::map<std::string, std::string> {
+          return {
+              {"data", Azure::Core::_internal::Base64Url::Base64UrlEncode(value.Data)},
+              {"dataType", value.DataType.ToString()}};
+        });
```
vhvb1989, line 115: here is the same decorator fn than before. re-use if possible

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,297 @@
...
+    Azure::Core::Json::_internal::json serializedHeader;
+    JsonOptional::SetFromNullable(tokenHeader.Algorithm, serializedHeader, "alg");
+    JsonOptional::SetFromNullable(tokenHeader.KeyURL, serializedHeader, "jku");
+    JsonOptional::SetFromNullable<AttestationSigner, std::string>(
+        tokenHeader.Key, serializedHeader, "jwk", [](AttestationSigner const& signer) {
+          return AttestationSignerInternal::SerializeToJson(signer);
+        });
+    JsonOptional::SetFromNullable(tokenHeader.ContentType, serializedHeader, "cty");
+    JsonOptional::SetFromNullable(tokenHeader.Critical, serializedHeader, "crit");
+    JsonOptional::SetFromNullable(tokenHeader.KeyId, serializedHeader, "kid");
```
vhvb1989, line 225: nit.
We might want to move all those hardcoded strings to a constants header.  We could re-use many.

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
```diff
@@ -0,0 +1,297 @@
...
+    Azure::Nullable<std::string> kty;
+    JsonOptional::SetIfExists(kty, jwk, "kty");
+    if (kty.HasValue())
+    {
+      returnValue.kty = kty.Value();
+    }
+    else
+    {
+      throw std::runtime_error("Could not find required field 'kty' in JSON Web Key");
+    }
```
vhvb1989, line 249: We can avoid creating the temporal value here:

```suggestion

    JsonOptional::SetIfExists(returnValue.kty, jwk, "kty");
    
    // returnValue.kty gets initialized to default (no value) and won't be changed by SetIfExists if kty is not found
    if (!returnValue.kty)
    {
      throw std::runtime_error("Could not find required field 'kty' in JSON Web Key");
    }
```

sdk/attestation/azure-security-attestation/src/private/attestation_deserializers_private.cpp
LarryOsterman: @ahsonkhan What will autorest be doing for the same constants? Will it be using manifest constants or creating a separate manifest constant header? Since this is a functional replacement for autorest generated code (and will be removed once the autorest serializers are created), I want to follow the pattern autorest is going to use.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,426 @@
...
+   * which can be used to retrieve the OpenId Metadata document and signing keys from the service.
+   *
+   *
+   * The OpenId Metadata document contains properties which describe the attestation service.
+   *
+   * The Attestation Signing Keys describe the keys which will be used to sign tokens generated by
+   * the attestation service. All tokens emitted by the attestation service will be signed by one
+   * of the certificates listed in the attestation signing keys.
+   *
+   */
```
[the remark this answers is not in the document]
LarryOsterman: This will appear in the documentation for the AttestationClient object, and this documents the behavior of the AttestationClient object.

From discussions with the language architects, It's too heavyweight to be in the readme. I've not yet ported the readme over, but you can look at the [Attestation Client readme for Java](https://docs.microsoft.com/en-us/java/api/overview/azure/security-attestation-readme?view=azure-java-stable) to see where the readme will be coming from.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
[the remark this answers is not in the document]
LarryOsterman: The `mutable` storage class indicates that a member may be changed when called from a const function. And `CacheAttestationSigners` is marked as const because it's called from `const` functions.

In general, I strongly prefer that locks be scoped as tightly as possible to their use case. Is there a reason you want this lock to be global? Is it solely to avoid having the `#include <mutex>`? If so, what are the rules w.r.t. standard headers? Which ones should be avoided? Encouraged?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,426 @@
...
+
+  class AttestationClient final {
+  private:
+    Azure::Core::Url m_endpoint;
+    std::string m_apiVersion;
+    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
+    AttestationTokenValidationOptions m_tokenValidationOptions;
+
+    mutable std::shared_timed_mutex m_sharedStateLock;
```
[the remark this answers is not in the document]
LarryOsterman: Is there a problem with referencing `std::shared_time_mutex` from the client implementation?

sdk/attestation/azure-security-attestation/src/private/crypto/openssl/opensslcert.cpp
```diff
@@ -0,0 +1,451 @@
...
+  std::unique_ptr<Cryptography::X509Certificate> OpenSSLX509Certificate::CreateFromPrivateKey(
+      std::unique_ptr<Cryptography::AsymmetricKey> const& key,
+      std::string const& subjectName)
+  {
+    auto utcTime(time(nullptr));
+    time_t expirationTime;
+    {
+      // This certificate expires in 8 hours.
+      struct std::tm currentTime;
+#if _WINDOWS
```
vhvb1989, line 412: nit.

our general guideline is to use `#if defined(MACRO)` instead of `#if MACRO`.   

Also, how do we know `_WINDOWS` is always there on Windows?  
We use `#if defined(_WIN32)` in our platform header from Azure-core (sdk/core/azure-core/inc/azure/core/platform.hpp),  you might want to follow that.

You might want to include platform.hpp and use `#if defined(AZ_PLATFORM_WINDOWS)`

sdk/attestation/azure-security-attestation/src/private/crypto/openssl/opensslcert.hpp
```diff
@@ -0,0 +1,231 @@
...
+      ss << "Unknown Certificate Key Type: " << std::to_string(nid) << " for certificate "
+         << GetSubjectName() << " " << GetIssuerName() << GetThumbprint();
+      throw std::invalid_argument(ss.str());
+      // LCOV_EXCL_STOP
+    }
+
+    static std::unique_ptr<X509Certificate> Import(std::string const& pemEncodedKey);
+  };
+
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 231:
```suggestion
}}}} // namespace Azure::Security::Attestation::_detail

```

sdk/attestation/azure-security-attestation/src/private/crypto/openssl/opensslkeys.hpp
```diff
@@ -0,0 +1,73 @@
...
+    RsaOpenSSLAsymmetricKey(openssl_evp_pkey&& key) : OpenSSLAsymmetricKey(std::move(key)) {}
+  };
+
+  class EcdsaOpenSSLAsymmetricKey : public OpenSSLAsymmetricKey {
+  public:
+    EcdsaOpenSSLAsymmetricKey();
+    EcdsaOpenSSLAsymmetricKey(openssl_evp_pkey&& key) : OpenSSLAsymmetricKey(std::move(key)) {}
+  };
+
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 73:
```suggestion
}}}} // namespace Azure::Security::Attestation::_detail

```

sdk/attestation/azure-security-attestation/src/private/jsonhelpers.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  std::map<std::string, std::string> JsonHelpers::DecorateAttestationData(
+      AttestationData const& value)
+  {
+    return {
+        {"data", Azure::Core::_internal::Base64Url::Base64UrlEncode(value.Data)},
+        {"dataType", value.DataType.ToString()}};
+  }
+
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 105:
```suggestion
}}}} // namespace Azure::Security::Attestation::_detail

```

sdk/attestation/azure-security-attestation/src/private/jsonhelpers_private.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+    static std::map<std::string, std::string> DecorateAttestationData(
+        Azure::Security::Attestation::AttestationData const& data);
+
+    static void SetIfExistsJson(
+        Azure::Nullable<std::string>& rv,
+        const Azure::Core::Json::_internal::json& field,
+        const std::string& fieldName);
+  };
+}}}} // namespace Azure::Security::Attestation::_detail
```
vhvb1989, line 42:
```suggestion
}}}} // namespace Azure::Security::Attestation::_detail

```

vhvb1989 (approved):
Hopefully my last comments.

Apply the end of file lines and lets move everything else to a next iteration in the interest of browser handling this PR :P

sdk/attestation/azure-security-attestation/NOTICE.txt
```diff
@@ -0,0 +1,32 @@
...
+Redmond, WA 98052
+USA
+
+Notwithstanding any other terms, you may reverse engineer this software to the
+extent required to debug changes to any libraries licensed under the GNU Lesser
+General Public License.
+
+------------------------------------------------------------------------------
+
+Azure SDK for C++ uses third-party libraries or other resources that may be
```
ahsonkhan, line 23: Do we have any third party library dependencies we are including to add to the notice?

For example, other services have a few other things:
https://github.com/Azure/azure-sdk-for-cpp/blob/771b7382c2382c72567af959794c9999479f760d/sdk/storage/azure-storage-blobs/NOTICE.txt#L34

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,324 @@
...
+   */
+  template <typename T> class AttestationToken final {
+  public:
+    /**
+     * @brief The full RFC 7515 JWS/JWT token returned by the attestation service.
+     */
+    std::string RawToken;
+
+    /**
+     * @brief      The elements of the raw token which will be signed by the Signature.
```
ahsonkhan, line 30: nit: spacing inconsistency (here/elsewhere), something to clean up across the board

```suggestion
     * @brief The elements of the raw token which will be signed by the Signature.
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+    /**
+     * @brief Returns the API version the client was configured with.
+     *
+     * @returns The API version used when communicating with the attestation service.
+     */
+    std::string const& ClientVersion() const { return m_apiVersion; }
+
+    /**
+     * @brief Construct a new Key Client object from another key client.
```
ahsonkhan, line 262: Comment mismatch, maybe copy/paste error

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * @brief Attest an SGX enclave, returning an attestation token representing the result
+     * of the attestation operation.
+     *
+     * @param sgxQuoteToAttest - SGX Quote to be validated by the attestation service.
+     * @param options - Options to the attestation request (runtime data, inittime data, etc).
+     * @param context - Context for the operation.
+     *
+     * @returns Response<{@link AttestationToken}<{@link AttestationResult}>> - The result of the
+     * attestation operation
```
ahsonkhan, line 299: nit:
```suggestion
     * attestation operation.
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -0,0 +1,324 @@
...
+     */
+    Response<AttestationToken<Models::AttestationResult>> AttestSgxEnclave(
```
ahsonkhan, line 301: This is an interesting approach and design (returning a `AttestationToken<T>` templated type which isn't a "model"). I am wondering how other language SDKs, like .NET expose the return type, and if it is similar to what we are doing here.

sdk/attestation/azure-security-attestation/CMakeLists.txt
```diff
@@ -0,0 +1,106 @@
...
+    ${AZURE_ATTESTATION_HEADER}
+    ${AZURE_ATTESTATION_SOURCE})
+
+target_include_directories(
+  azure-security-attestation
+    PUBLIC
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>
+      $<INSTALL_INTERFACE:include>
+      # Disable dependency on azure-core until azure-core has released to vcpkg
+      # ${azure-core-cpp_INCLUDE_DIRS}
```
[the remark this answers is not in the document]
LarryOsterman: I have no idea. This was in the template and I happily copied it because I thought it was something that would be picked up later.

sdk/attestation/azure-security-attestation/NOTICE.txt
LarryOsterman: Not to my knowledge. OpenSSL should already be included in the 3rd party library dependencies because it's used on Linux.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: I would actually prefer to return an `AttestationResponse` which derives from `Response`, however `Response` is final :(. That's what I did in the other languages.

PR #3237, opened 2022-01-12
kv-common vcpkg fix: Option A: revert to original constructor

antkmsft:
Fix for the vcpkg failing to build `azure-security-keyvault-common-cpp` that we deletedfrom our repo a long ago, but it is still in vcpkg.

The build error is the following:
```
sdk\keyvault\azure-security-keyvault-common\src\keyvault_exception.cpp(35):
error C2664: 'Azure::Core::RequestFailedException::RequestFailedException(
  const std::string &,std::unique_ptr<Azure::Core::Http::RawResponse,
  std::default_delete<Azure::Core::Http::RawResponse>> &)':
    cannot convert argument 2
      from 'std::unique_ptr<Azure::Core::Http::RawResponse,std::default_delete<Azure::Core::Http::RawResponse>>'
      to 'std::unique_ptr<Azure::Core::Http::RawResponse,std::default_delete<Azure::Core::Http::RawResponse>> &'
```

Which was caused by us updating the constructor just recently. This PR reverts that.

 But with this PR, there is not clear path to drop the constructor in the future, as we wanted (#3215).

Other options are #3238 and #3239.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/exception.hpp
+++ b/sdk/core/azure-core/inc/azure/core/exception.hpp
@@ -90,7 +90,7 @@ namespace Azure { namespace Core {
      */
     explicit RequestFailedException(
         const std::string& what,
-        std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse);
+        std::unique_ptr<Azure::Core::Http::RawResponse> rawResponse);
 
     /**
      * @brief Constructs a new `%RequestFailedException` object with an HTTP raw response.
--- a/sdk/core/azure-core/src/exception.cpp
+++ b/sdk/core/azure-core/src/exception.cpp
@@ -17,7 +17,7 @@ namespace Azure { namespace Core {
 
   RequestFailedException::RequestFailedException(
       const std::string& what,
-      std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse)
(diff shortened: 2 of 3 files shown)


ahsonkhan:
Because of the time in which KeyVault common beta for released, this fix alone won't  be enough.

We also had the macro rename from `AZURE_UNREACHABLE_CODE` to `AZURE_UNREACHABLE_CODE()` prior to Azure Core, GA.

sdk/core/azure-core/inc/azure/core/exception.hpp
```diff
@@ -90,7 +90,7 @@ namespace Azure { namespace Core {
      */
     explicit RequestFailedException(
         const std::string& what,
-        std::unique_ptr<Azure::Core::Http::RawResponse>& rawResponse);
```
ahsonkhan, line 93: You mentioned this was a recent change. Was that a change, one that is technically breaking?

antkmsft: Yes, it is technically breaking, if the customer was creating instances of RequestFailedException with the other constructor.

antkmsft:
@ahsonkhan, are you sure, that `AZURE_UNREACHABLE_CODE` update has happened in prior releases, why is it only now that it is broken?

PR #3239, opened 2022-01-12
kv-common vcpkg fix: Option C: release a no-op version of kv-common, then re-delete it from our SDK

antkmsft:
This is the best option, in my opinion.

We will have to re-create the kv-common, release it together with core, then we can delete it from our SDK again, as we did initially.


ahsonkhan:
If we are able to remove the port from the registry, that would be ideal. Let's aim for that first.

If we can't, for some reason, then I agree, this might be a good way to go as a workaround, if it has to come to that.

PR #3298, opened 2022-02-01
Fix unit test compilation on UWP

antkmsft:
This is NOT the fix for the missing `VCRUNTIME140D_APP.dll`.
This is to successfully compile all of them before it would get to that point.

Gtest does not support `ASSERT_DEATH()` on UWP: https://github.com/google/googletest/blob/100f6fbf5f81a82d163c1e29735e8a2936eacd4f/googletest/include/gtest/internal/gtest-port.h#L605-L613

### DIFF
--- a/sdk/core/azure-core/test/ut/bodystream_test.cpp
+++ b/sdk/core/azure-core/test/ut/bodystream_test.cpp
@@ -34,6 +34,7 @@ TEST(BodyStream, Rewind)
 {
   TestBodyStream tb;
 
+#if GTEST_HAS_DEATH_TEST
 #if defined(NDEBUG)
   // Release build won't provide assert msg
   ASSERT_DEATH(tb.Rewind(), "");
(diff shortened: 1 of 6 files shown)


vhvb1989 (approved):
Wonder how UWP gates were passing without this fix?

Anyway, the UWP gates are gone for now :) , but I am approving, since we will bring back UWP eventually

antkmsft:
@vhvb1989, mostlikely, CI was not building tests. Plus even today, it wouldn't be able to run them due to "`app.dll`" issue.
PR that builds them is https://github.com/Azure/azure-sdk-for-cpp/pull/3298, and that one has this compilation fix from this PR, and is stuck on the DLL issue now.
I agree, eventually we'll have to fix UWP. This is a step in that direction, it clearly would do no harm at least. If Gtest does not have that macro defined, `ASSERT_DEATH()` can't compile, be it UWP or any other platform where `GTEST_HAS_DEATH_TEST` might be undefined.

PR #3312, opened 2022-02-04
Try deprecating the public Azure Assert macros using the deprecated attribute.

ahsonkhan:


sdk/core/azure-core/inc/azure/core/azure_assert.hpp
```diff
@@ -12,8 +12,22 @@
 
 #include "azure/core/internal/azure_assert.hpp"
 
+[[deprecated("The Azure Assert macros are meant for internal use within the SDK only, use the STL "
+             "assert instead.")]]
 #define AZURE_ASSERT(exp) _azure_ASSERT(exp)
```
ahsonkhan, line 17: @LarryOsterman looks like I can't really use these attributes on macros. Do you know of an easy way to do this in C++?

If not, then maybe this isn't worth it and we leave them as is.

I get a build error in VS:
> Severity	Code	Description	Project	File	Line	Suppression State
Error (active)	E1847	attributes are not allowed here	azure-core.lib (sdk\core\azure-core\azure-core.lib) - x64-DebugWithTests	E:\GitHub\Fork\azure-sdk-for-cpp\sdk\core\azure-core\inc\azure\core\azure_assert.hpp	15	

FYI @antkmsft

antkmsft: If that would work, that would be great, but I don't think it is going to be possible to add attributes to macros.
If I understand correctly, the model is that in theory you can have a separate program that does preprocessing, but knows nothing about C++. It would do macro substitutions and feed the input to the C++ compiler which knows nothing about macros and does not see them at all, because macros are already substituted.

What I think can work is that we add an inline function `[[deprecated(...)]] void DeprecatedAsureAssert(...) { AzureNoReturnPath(...); }`, so the old macros would call that, and the new macros would call AzureNoReturnPath() directly. In that case, in substituted code, it will be using the function that is marked as deprecated, while if code uses the new macro, it would not.

antkmsft: #3320

LarryOsterman: What Anton said: You can add the attribute to functions, enumerations, structures, etc (not enumerators in C++ 14). But as Anton said, the C preprocessor is logically a separate executable (/user/bin/cpp) - in oldentimes, when clang/gcc didn't exist, the C compiler (cc) ran cpp, cc, then ld as a part of the process of turning a .c file to a.out. It's only in recent history that the preprocessor was burned into the compiler.

PR #3330, opened 2022-02-08
Improve the quality of the WinHTTP transport error message.

LarryOsterman:
# Improve the quality of the WinHTTP exception error message.

The WInHTTP error message used to simply report the Win32 error code returned by WinHTTP. This change updates the error generation logic in WinHTTP to add the error text from WIndows using the `FormatMessage` API.

### DIFF
--- a/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
+++ b/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
@@ -201,8 +201,26 @@ std::string GetHeadersAsString(Azure::Core::Http::Request const& request)
 void GetErrorAndThrow(const std::string& exceptionMessage)
 {
   DWORD error = GetLastError();
-  throw Azure::Core::Http::TransportException(
-      exceptionMessage + " Error Code: " + std::to_string(error) + ".");
+  char* errorString = nullptr;
+  std::string errorMessage = exceptionMessage + " Error Code: " + std::to_string(error);
+
+  if (FormatMessage(
+          FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_FROM_HMODULE | FORMAT_MESSAGE_ALLOCATE_BUFFER,
+          GetModuleHandle("winhttp.dll"),
+          error,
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -201,8 +201,26 @@ std::string GetHeadersAsString(Azure::Core::Http::Request const& request)
...
+          error,
+          MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+          reinterpret_cast<LPSTR>(&errorString),
+          0,
+          nullptr)
+      != 0)
+  {
+    errorMessage += ": ";
+    errorMessage += errorString;
+    LocalFree(errorString);
```
antkmsft, line 219: This is probably super defensive, but we can additionally guard from a memory leak in a rare case when nothing is right:
```suggestion
    try
    {
      errorMessage += ": ";
      errorMessage += errorString;
    }
    catch (...)
    {
      LocalFree(errorString);
      throw;
    }

    LocalFree(errorString);
```

antkmsft (approved):

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: You're 100% right (and it's been annoying me), but I think I have a slightly better solution... Let me make the change and see if you like it.

vhvb1989 (commented):
Is there a way to test this?

I wonder how one message looks against the other one.

Since we are not adding unit test, we will be dropping some coverage right?

LarryOsterman:
> Is there a way to test this?
> 
> I wonder how one message looks against the other one.
> 
> Since we are not adding unit test, we will be dropping some coverage right?

When I ran the tests with the WinHTTP pipeline configured, this was hit - I've stepped through this code manually.

Old error:
`"Error while sending a request. Error Code: 12044.`

New error:
`"Error while sending a request. Error Code: 12044: A certificate is required to complete client authentication
.`

Yes, the trailing crlf is annoying. If it bothers me enough, I'll fix it in a later pull request.

PR #3355, opened 2022-02-17
Created Attestation Administration client

LarryOsterman:
Create the attestation service administration client and implement the GetAttestationPolicy API in the administration client.

Also created a skeleton of the readme.md (based on the Java SDK, so there are likely Java artifacts left behind).

Added support for nightly live tests in the WestUS region.


LarryOsterman:
Ping :)

LarryOsterman:
Ping :).

LarryOsterman:
Ping :)

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+      perRetrypolicies.emplace_back(
+          std::make_unique<BearerTokenAuthenticationPolicy>(credential, tokenContext));
+    }
+    m_apiVersion = options.Version.ToString();
+    std::vector<std::unique_ptr<HttpPolicy>> perCallpolicies;
+
+    m_pipeline = std::make_shared<Azure::Core::Http::_internal::HttpPipeline>(
+        options,
+        "Attestation",
```
vhvb1989, line 59: mmm, since we are using the same service package for admin and non-admin, telemetry won't make a difference between them right?
At least you change this to `Attestation-admin`.  But I am not sure.

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
```diff
@@ -0,0 +1,150 @@
...
+        PackageVersion::ToString(),
+        std::move(perRetrypolicies),
+        std::move(perCallpolicies));
+  }
+
+  namespace {
+    std::shared_timed_mutex SharedStateLock;
+  }
```
vhvb1989, line 67: nit.
This namespace should go at the root namespace level instead of nested.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -21,23 +23,25 @@ using namespace Azure::Security::Attestation::_detail;
...
-    //          = {{_internal::UrlScope::GetScopeFromUrl(m_vaultUrl)}};
+    m_credentials = credential;
+    Azure::Core::Credentials::TokenRequestContext const tokenContext
+        = {{"https://attest.azure.net/.default"}};
 
-    //      perRetrypolicies.emplace_back(
-    //          std::make_unique<BearerTokenAuthenticationPolicy>(credential,
-    //          std::move(tokenContext)));
+    perRetrypolicies.emplace_back(
+        std::make_unique<BearerTokenAuthenticationPolicy>(credential, tokenContext));
```
vhvb1989, line 44: As you mentioned before, there is some duplicated code between the Admin client and non-admin client.

If you want to avoid the duplicated code, create a private header and define methods with the duplicated code. Then include the private header to both adminClient.cpp and client.cpp and call the methods from there.  This way you don't need to create class hierarchy.

So, both client constructors could just call `m_pipeline = PrivateClassHelper.CreatePipeline(... , ... , ); `

sdk/attestation/azure-security-attestation/test/ut/administration_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential> m_credential;
+    std::string m_endpoint;
+
+    // Create
+    virtual void SetUp() override
+    {
+      Azure::Core::Test::TestBase::SetUpTestBase(AZURE_TEST_RECORDING_DIR);
+
+      std::string mode(std::get<0>(GetParam()));
+      //      std::string mode(GetParam());
```
vhvb1989, line 31: remove
```suggestion

```

sdk/attestation/azure-security-attestation/test/ut/administration_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+  protected:
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential> m_credential;
+    std::string m_endpoint;
+
+    // Create
+    virtual void SetUp() override
+    {
+      Azure::Core::Test::TestBase::SetUpTestBase(AZURE_TEST_RECORDING_DIR);
+
+      std::string mode(std::get<0>(GetParam()));
```
vhvb1989, line 30: nit
```suggestion
      std::string const mode(std::get<0>(GetParam()));
```

sdk/attestation/azure-security-attestation/test/ut/administration_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+      //      std::string mode(GetParam());
+      if (mode == "Shared")
+      {
+        std::string shortLocation(GetEnv("LOCATION_SHORT_NAME"));
```
vhvb1989, line 34: nit
```suggestion
        std::string const shortLocation(GetEnv("LOCATION_SHORT_NAME"));
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -0,0 +1,121 @@
...
+        : m_endpoint(attestationClient.m_endpoint), m_apiVersion(attestationClient.m_apiVersion),
+          m_pipeline(attestationClient.m_pipeline),
+          m_tokenValidationOptions(attestationClient.m_tokenValidationOptions){};
+
+    /** @brief Creates a new {@link AttestationClient} with the same URL and HTTP
+     * pipeline as this AttestationAdministrationClient.
+     *
+     * @return A new {@link AttestationClient} instance..
+     */
+    AttestationClient AsAttestationClient() const;
```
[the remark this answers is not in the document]
LarryOsterman: Correct - it is a peer client. And I could see making it `GetAttestationClient`. I'm still not 100% sure I want to copy this from storage though. It turns out that the implementation is a bit tricky.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -0,0 +1,121 @@
...
+        GetPolicyOptions const& options = GetPolicyOptions(),
+        Azure::Core::Context const& context = Azure::Core::Context::ApplicationContext) const;
+
+  protected:
+    Azure::Core::Url m_endpoint;
+    std::string m_apiVersion;
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential const> m_credentials;
+    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
+    AttestationTokenValidationOptions m_tokenValidationOptions;
+
```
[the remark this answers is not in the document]
LarryOsterman: It was marked as protected because I wanted to share it with "friend" classes.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -315,10 +210,24 @@ namespace Azure { namespace Security { namespace Attestation {
...
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential const> m_credentials;
+    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
+    AttestationTokenValidationOptions m_tokenValidationOptions;
+
+    mutable std::vector<Models::AttestationSigner> m_attestationSigners;
+
+    std::vector<Models::AttestationSigner> const& GetAttestationSigners(
+        Azure::Core::Context const& context) const;
+
+    friend class AttestationAdministrationClient;
```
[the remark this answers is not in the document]
LarryOsterman: They're friends of each other to allow me to create one from the other. That requires pulling out internal state from one and passing it to the other.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -249,4 +272,16 @@ namespace Azure { namespace Security { namespace Attestation {
...
+  /** @brief Parameters sent to the attestation service when retrieving an attestation policy.
+   */
+  struct GetPolicyOptions final
+  {
+    /** @brief Specifies the options which should be used to validate the attestation token returned
+     * by the attestation service.
+     * @details If not provided by the caller, the token validation options
+     * specified when the @{link AttestationAdministrationClient} was created will be used.
+     */
+    Azure::Nullable<AttestationTokenValidationOptions> TokenValidationOptions{};
```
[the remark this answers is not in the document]
LarryOsterman: Yes, {} adds a default value - the default value for Azure::Nullable<AttestationTokenValidationOptions>, which is an empty Azure::Nullable.

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
LarryOsterman: Telemetry shouldn't really distinguish between them. Certainly not on the service side it isn't distinguished.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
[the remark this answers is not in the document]
vhvb1989: Friend classes can access private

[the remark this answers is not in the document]
LarryOsterman: Interesting. I had always assumed private was truly private.

sdk/attestation/azure-security-attestation/test/ut/attestation_collateral.cpp
```diff
@@ -0,0 +1,355 @@
...
+    return openEnclaveReport;
+    //    return std::vector<uint8_t>(openEnclaveReport.begin() + 0x10,
+    //    openEnclaveReport.end());
+  }
+  std::vector<uint8_t> AttestationCollateral::RuntimeData()
+  {
+    return Azure::Core::_internal::Base64Url::Base64UrlDecode(SgxEnclaveHeldData);
+    //    return std::vector<uint8_t>(RuntimeJWKClaim.begin(), RuntimeJWKClaim.end());
+  }
+}}}} // namespace Azure::Security::Attestation::Test
```
vhvb1989, line 355:
```suggestion
}}}} // namespace Azure::Security::Attestation::Test

```

sdk/attestation/azure-security-attestation/test/ut/attestation_test.cpp
```diff
@@ -55,8 +55,18 @@ namespace Azure { namespace Security { namespace Attestation { namespace Test {
...
+      std::shared_ptr<Azure::Identity::ClientSecretCredential const> credential;
+      if (m_testContext.IsPlaybackMode())
+      {
+        // Skip validating time stamps if using recordings.
+        options.TokenValidationOptions.ValidateNotBeforeTime = false;
+        options.TokenValidationOptions.ValidateExpirationTime = false;
+      }
+      else
+      {
+        credential = std::make_shared<Azure::Identity::ClientSecretCredential const>(
```
vhvb1989, line 67: same suggestion as before to handle client with a credential on testing

vhvb1989 (approved):
Added minor comments and suggestions.

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
vhvb1989: see line 30: `namespace Azure { namespace Security { namespace Attestation {`

and line 150: `}}} // namespace Azure::Security::Attestation`

sdk/attestation/azure-security-attestation/test/ut/administration_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+      // `InitTestClient` takes care of setting up Record&Playback.
+      auto options = InitClientOptions<
+          Azure::Security::Attestation::AttestationAdministrationClientOptions>();
+      std::shared_ptr<Azure::Identity::ClientSecretCredential const> credential;
+      if (m_testContext.IsPlaybackMode())
+      {
+        // Skip validating time stamps if using recordings.
+        options.TokenValidationOptions.ValidateNotBeforeTime = false;
+        options.TokenValidationOptions.ValidateExpirationTime = false;
```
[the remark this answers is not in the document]
LarryOsterman: Taking this discussion offline to email.

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
LarryOsterman: @#$@#$ - I was looking at attestation_client, not admin_client.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
 # Azure Attestation Package client library for C++
 
-Azure Template Package client library for C++ (`azure-security-attestation`)  matches necessary patterns that the development team has established to create a unified SDK written in the C++ programming language. These libraries follow the Azure SDK Design Guidelines for C++.
+Microsoft Azure Attestation (preview) is a unified solution for remotely verifying the trustworthiness of a platform and integrity 
+of the binaries running inside it. The service supports attestation of the platforms backed by Trusted Platform Modules (TPMs) 
+alongside the ability to attest to the state of Trusted Execution Environments (TEEs) such as Intel� Software Guard Extensions
```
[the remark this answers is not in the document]
LarryOsterman: @#$@#$. IntelliJ loves to put in non-ascii characters. Fixed.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
+### Install Dependencies
+
+#### Windows
+
+On Windows, dependencies are managed by [vcpkg](https://github.com/microsoft/vcpkg). You can reference the [Quick Start](https://github.com/microsoft/vcpkg#quick-start-windows) to quickly set yourself up.
+After Vcpkg is initialized and bootstrapped, you can install the dependencies:
+```BatchFile
+vcpkg.exe install curl:x64-windows-static
+vcpkg.exe install openssl:x64-windows-static
+```
```
ahsonkhan, line 44: The end user shouldn't need to install these dependencies independently.

The acquisition for any C++ client library and Azure SDK package should be:
`vcpkg install <package_name>:<optional triplet>`

And that should bring in the dependencies needed.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
+cmake --build .
+```
+Then you can consume the built library with the header files.
+make/ninja install is work in progress.
+
+
+### Via NuGet
+WIP
+TODO when ready.
+
```
ahsonkhan, line 95: I think we can remove this section for now, unless we have a strong reason to use NuGet as an acquisition point, today.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
+```
+
+## Dependencies
+
+  - [Azure Core SDK](https://github.com/Azure/azure-sdk-for-cpp/blob/main/README.md)
+  - OpenSSL
+
+
+
```
ahsonkhan, line 119: nit: remove extra new lines

```suggestion
```

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
+- [Attestation policy management certificate management](#policy-management-certificate-management) (yes, policy management management).
+
+Each attestation instance operates in one of three separate modes of operation:
+
+- Isolated mode.
+- AAD mode.
+- Shared mode.
+
+## Isolated Mode Attestation Instances.
+In "Isolated" mode, the customer indicates that they want to ensure that Microsoft administrators cannot influence the inputs or outputs of an attestation service instance. When the attestation service instance is running in Isolated mode, the customer is expected to provide additional proof that they are authorized to make changes to the attestation service instance.
```
ahsonkhan, line 148: nit: be consistent, add new line after heading

There's a markdown linter extension in VS Code which helps point out such formatting or other linting issues, if you want to use that while updating the doc :)

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
 # Azure Attestation Package client library for C++
 
-Azure Template Package client library for C++ (`azure-security-attestation`)  matches necessary patterns that the development team has established to create a unified SDK written in the C++ programming language. These libraries follow the Azure SDK Design Guidelines for C++.
+Microsoft Azure Attestation (preview) is a unified solution for remotely verifying the trustworthiness of a platform and integrity 
+of the binaries running inside it. The service supports attestation of the platforms backed by Trusted Platform Modules (TPMs) 
+alongside the ability to attest to the state of Trusted Execution Environments (TEEs) such as Intel� Software Guard Extensions 
+(SGX) enclaves and Virtualization-based Security (VBS) enclaves.
```
ahsonkhan, line 6: I really like the first use of abbreviations spelt out for naïve readers :)

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
+// Note that the "policy" calls require authentication.
+AttestationAdministrationClient client = attestationBuilder
+    .endpoint(endpoint)
+    .credential(new DefaultAzureCredentialBuilder().build())
+    .buildClient();
+```
+
+### Retrieve current attestation policy for OpenEnclave
+
+Use the `getAttestationPolicy` API to retrieve the current attestation policy for a given TEE.
```
ahsonkhan, line 249: 🍵

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
+When an attestation instance is in AAD mode, the caller can use a convenience method to set an unsigned attestation
+policy on the instance.
+
+```java readme-sample-set-unsigned-policy
+// Set the listed policy on an attestation instance. Please note that this particular policy will deny all
+// attestation requests and should not be used in production.
+PolicyResult policyResult = client.setAttestationPolicy(AttestationType.OPEN_ENCLAVE,
+    "version=1.0; authorizationrules{=> deny();}; issuancerules{};");
+System.out.printf("Policy set for OpenEnclave result: %s\n", policyResult.getPolicyResolution());
+```
```
ahsonkhan, line 267: Java samples in C++?! Blasphemy :p

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -0,0 +1,109 @@
...
+}}}} // namespace Azure::Core::Http::_internal
+
+namespace Azure { namespace Security { namespace Attestation {
+
+  /**
+   *
+   * The AttestationAdministrationClient implements the functionality required by the
+   * "Administration" family of attestation service APIs.
+   *
+   * @note: Attestation administration APIs cannot be used on shared attestation service instances.
```
ahsonkhan, line 23: Where do we imagine this doc comment to be visible for the customer? Would it render in the generated doxygen as we expected on the `AttestationAdministrationClient` class? Should we split it up in a `@brief` and other annotations?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+  public:
+    /**
+     * @brief Destructor.
+     *
+     */
+    virtual ~AttestationAdministrationClient() = default;
+
+    /**
+     * @brief Construct a new Attestation Administration Client object
```
ahsonkhan, line 57: nit: complete doc sentence with period
```suggestion
     * @brief Construct a new Attestation Administration Client object.
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -0,0 +1,109 @@
...
+     */
+    explicit AttestationAdministrationClient(
+        AttestationAdministrationClient const& attestationClient)
+        : m_endpoint(attestationClient.m_endpoint), m_apiVersion(attestationClient.m_apiVersion),
+          m_pipeline(attestationClient.m_pipeline),
+          m_tokenValidationOptions(attestationClient.m_tokenValidationOptions){};
+
+    /** @brief Retrieves the attestation policy for the specified attestation type.
+     */
+
```
ahsonkhan, line 90: consistency nit: remove space
```suggestion
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -0,0 +1,109 @@
...
+        Azure::Core::Context const& context = Azure::Core::Context::ApplicationContext) const;
+
+  private:
+    Azure::Core::Url m_endpoint;
+    std::string m_apiVersion;
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential const> m_credentials;
+    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
+    AttestationTokenValidationOptions m_tokenValidationOptions;
+
+    mutable std::vector<Models::AttestationSigner> m_attestationSigners;
```
ahsonkhan, line 103: I haven't seen this `mutable` keyword being used before in our SDK. For my edification, what's the purpose of it, generally, and why do we need to use it in this instance?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -187,6 +239,102 @@ namespace Azure { namespace Security { namespace Attestation { namespace Models
...
+
+    // RFC 7519 properties.
+
+    /**
+     *  The Expiration time for this attestation token.
+     *
+     * After this time, the token cannot be considered valid.
+     *
+     * See <a href='https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4'>RFC 7519
+     * Section 4.1.4</a> for more information.
```
ahsonkhan, line 276: Just curious, does it render as we expect in the generated doxygen docs?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
```diff
@@ -274,4 +422,73 @@ namespace Azure { namespace Security { namespace Attestation { namespace Models
...
+    /**
+     * @brief Construct a new PolicyResolution object
+     *
+     * @param modification The string resolution used for the result of an attestation policy
+     * operation.
+     */
+    PolicyModification(std::string modification) : m_policyModification(std::move(modification)) {}
+
+    /**
+     * @brief Enable comparing the ext enum.
```
ahsonkhan, line 444: nit: spell it out
```suggestion
     * @brief Enable comparing the extensible enum.
```

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // Construct a token whose body is the policy, but whose token is the response from the
+    // service.
+    auto returnedToken = AttestationTokenInternal<std::string>(responseToken, returnPolicy);
+    return Response<AttestationToken<std::string>>(returnedToken, std::move(response));
+  }
+
+  std::vector<AttestationSigner> const& AttestationAdministrationClient::GetAttestationSigners(
+      Azure::Core::Context const& context) const
+  {
+    std::unique_lock<std::shared_timed_mutex> stateLock(SharedStateLock);
```
ahsonkhan: I am not familiar enough with the implementation detail or requirements from the attestation SDK, so take this question with a grain of salt, and largely out of curiosity.

We have typically avoided taking locks or mutexes in the client libraries. Why is this necessary?

sdk/attestation/azure-security-attestation/test/ut/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ add_executable (
      macro_guard.cpp
      crypto_test.cpp
      attestation_test.cpp
-     token_test.cpp)
+     token_test.cpp administration_test.cpp "attestation_collateral.cpp")
```
ahsonkhan, line 22: nit: fix inconsistency of format for file includes

ahsonkhan (commented):
Partial review, will try to find some time to review the rest of the implementation and API surface later :)

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
 # Azure Attestation Package client library for C++
 
-Azure Template Package client library for C++ (`azure-security-attestation`)  matches necessary patterns that the development team has established to create a unified SDK written in the C++ programming language. These libraries follow the Azure SDK Design Guidelines for C++.
+Microsoft Azure Attestation (preview) is a unified solution for remotely verifying the trustworthiness of a platform and integrity 
+of the binaries running inside it. The service supports attestation of the platforms backed by Trusted Platform Modules (TPMs)
```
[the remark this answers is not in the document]
LarryOsterman: Editors which don't implement line wrap reasonably (like Visual Studio) turn a single line paragraph into an unreadable mess.

Manually inserting line breaks makes the text readable.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
 # Azure Attestation Package client library for C++
 
-Azure Template Package client library for C++ (`azure-security-attestation`)  matches necessary patterns that the development team has established to create a unified SDK written in the C++ programming language. These libraries follow the Azure SDK Design Guidelines for C++.
+Microsoft Azure Attestation (preview) is a unified solution for remotely verifying the trustworthiness of a platform and integrity
```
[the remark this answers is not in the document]
LarryOsterman: Azure uses "preview" for services, not "beta/stable". At the time this text was written, the attestation service was in preview, not GA, so text referring to the service says "preview", text referring to the SDK says "beta"

sdk/attestation/azure-security-attestation/README.md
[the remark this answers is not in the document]
ahsonkhan: That's true, but what line length do we maintain here? It could make edits more difficult, to maintain the line wrapping manually. What about other editors like VS Code? If I add a few words, how do I fixup the manual line breaks? Does it become a chain of edits?

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,39 +1,362 @@
...
 
-Azure Template Package client library for C++ (`azure-security-attestation`)  matches necessary patterns that the development team has established to create a unified SDK written in the C++ programming language. These libraries follow the Azure SDK Design Guidelines for C++.
+Microsoft Azure Attestation (preview) is a unified solution for remotely verifying the trustworthiness of a platform and integrity 
+of the binaries running inside it. The service supports attestation of the platforms backed by Trusted Platform Modules (TPMs) 
+alongside the ability to attest to the state of Trusted Execution Environments (TEEs) such as Intel� Software Guard Extensions 
+(SGX) enclaves and Virtualization-based Security (VBS) enclaves.
 
-The library allows client libraries to expose common functionality in a consistent fashion.  Once you learn how to use these APIs in one client library, you will know how to use them in other client libraries.
+Attestation is a process for demonstrating that software binaries were properly instantiated on a trusted platform. Remote 
+relying parties can then gain confidence that only such intended software is running on trusted hardware.
```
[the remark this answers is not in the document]
LarryOsterman: A relying party is someone who relies on the output of the attestation service and makes a trust decision based on that. It is a well-known term of art in the attestation space.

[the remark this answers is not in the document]
LarryOsterman: If there was a "glossary of terms" section in a readme file, I might consider adding this to it.

sdk/attestation/azure-security-attestation/README.md
LarryOsterman: This text was taken from keyvault's readme. If that is the case, we need to update KV (and likely other readme.md files).

sdk/attestation/azure-security-attestation/README.md
LarryOsterman: Again, stolen from KV.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_models.hpp
LarryOsterman: What about this wouldn't render properly?

sdk/attestation/azure-security-attestation/src/attestation_administration_client.cpp
LarryOsterman: It takes several hundred ms to retrieve the attestation signing certificates. If we don't cache them on the client, an attestation client which makes more than one API call to the service is going to burn hundreds of milliseconds of wall time retrieving the exact same data again and again, once for every single call into the service.

This avoids that cost.

ahsonkhan: Thanks! That would be great as a comment in source, to explain the rationale :)

sdk/attestation/azure-security-attestation/test/ut/CMakeLists.txt
```diff
@@ -19,7 +19,8 @@ add_executable (
      macro_guard.cpp
      crypto_test.cpp
      attestation_test.cpp
-     token_test.cpp)
+     token_test.cpp administration_test.cpp
```
ahsonkhan, line 22: consistency nit:
```suggestion
     token_test.cpp
     administration_test.cpp
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
LarryOsterman: We claim that SDK clients are immutable. That means that service methods should all be declared as "const". 

By using `mutable` we limit the potential damage of breaking const-ness to just the variables which are known to be mutable.

ahsonkhan: Again, up to you, but it feels wrong for us to claim one thing, convey intent using `const` and then break it.

If we want to limit the potential damage of breaking const-ness, rather than marking the variable as mutable, why not mark the method that is mutating it as non-const? And now the intent is clear? Or is that less desirable in this case compared to using `mutable`?

Without that, this variable can be mutated in all `const` methods, rather than isolating it to just the few that need to mutate it. Unless, all of them do mutate it, then that's a different question :)
Is that the case here?

Maybe @JeffreyRichter has some thoughts on immutability of SDK clients and when we should use `mutable` vs `const` on the particular method itself.

LarryOsterman: The compiler will not allow you to call a non-const method from a const method. The only way I know of working around that is the `const_cast<Foo*>(this)` trick, and that 100% *is* code smell.

LarryOsterman: The signing certificates are per-attestation instance (per-url), not per-client. But if you have 5 clients talking to 5 separate attestation service instances you will have five separate sets of certificates.

So implementing an attestation service specific pipeline policy to handle caching these values might be possible. It would be tricky, because we only want to have it enabled on certain API calls (it would be bad to invoke this policy on the `GetAttestationSigningCertificates` API for example). Another slight complication: If the customer says they don't want token validation, we don't ever want to retrieve the signing certificates. And that happens on a per-API basis (note: Not currently implemented - but it's on the list).

But it is worth considering, since we would have complete control over the policy.

OTOH, All this does is to move the mutable item from the client into a policy object. It doesn't change the architecture.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
LarryOsterman: I hope that documentation comments appear in the documentation. Or at least in intellisense.

PR #3366, opened 2022-02-19
Fixed Live Tests for Attestation SDK.

LarryOsterman:
# Fixed live tests.


RickWinter (approved):

Jinming-Hu (approved):

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -244,10 +246,26 @@ namespace Azure { namespace Core { namespace Test {
...
+      // Check to ensure that the filename is upper case
+      {
+        std::string ucName = name;
+        std::transform(ucName.begin(), ucName.end(), ucName.begin(), [](char const& ch) {
+          return static_cast<char>(std::toupper(ch));
+        });
+        if (ucName != name)
+        {
+          throw std::runtime_error("All Azure SDK environment variables must be all upper case.");
+        }
```
vhvb1989, line 265: use `StringExtensions::ToLower` and `StringExtensions::LocaleInvariantCaseInsensitiveEqual` from Azure core

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -262,8 +280,11 @@ namespace Azure { namespace Core { namespace Test {
      * @brief Run before each test.
      *
      */
-    void SetUpTestBase(std::string const& baseRecordingPath)
+    void SetUpTestBase(std::string const& serviceName, std::string const& baseRecordingPath)
```
vhvb1989, line 283: Since Attestation is the only one who needs this and won't use the test-resourses.json to set its env vars, can you move this new arg to the end as optional with default value empty. Then, only Attestation tests would use it.

We can also rename the input arg to something like `envVarPrefix` and add a note about what it does for you when it is  passed to SetUpTestBase.

This way we don't need to add the changes to Key Vault and Storage tests. Since they don't need it

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: Makes sense. I'll make that change.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: Attestation *does* use test-resources.json to set its environment variables, just not these three. That's because it's *important* that test-resources.json files be able to be shared across azure SDKs (similar to the test-resources-pre.ps1 scripts).

I will be confirming that KeyVault and STorage's live tests continue to work - this change should *not* require any change to them except for adding the words "KEYVAULT" and "STORAGE" to the ctor.

All the other SDKs call that variable "directory" - I thought "serviceName" was more descriptive. But I could call it servicePrefix or something like that.

antkmsft (approved):

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -276,6 +297,24 @@ namespace Azure { namespace Core { namespace Test {
           Sanitize(testNameInfo->test_suite_name()), Sanitize(testNameInfo->name()));
       m_testContext.RecordingPath = recordingPath;
       m_interceptor = std::make_unique<Azure::Core::Test::InterceptorManager>(m_testContext);
+
+      if (!m_testContext.IsPlaybackMode())
+      {
+        auto SetBuiltinEnvironment
+            = [](std::string const& serviceName, std::string const& targetVariable) {
+                std::string targetValue = Azure::Core::_internal::Environment::GetVariable(
+                    (serviceName + targetVariable).c_str());
```
vhvb1989, line 306: I kept thinking about this `serviceName` inpur argument.

What if we get rid of it and we just introduce a new environment variable called `AZURE_ENV_PREFIX` or something like that and we use it instead of this input args.  

That would make test-implementation agnostic of it and only the user/developer and the test-fw would use it for what you are doing here.

Then, we can set that env var on the internal ci pipeline for live test. The internal pipeline knows the service names as it is given in the ci.yaml file.

This would also allow developers running tests for more than one service to keep using the same ENV VARS for all services instead of setting more variables.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: I profoundly disagree with this assertion. Somehow I (and others) have managed to develop SDKs for the 4 primary azure languages without modifying the test-resources.json. But for some reason, C++ is different from all the other languages.

I'm still trying to understand why the same solution that every other language has implemented is not idiomatic C++.  Could you please help me understand which C++ idioms this breaks?

As far as I can tell, my change simply updates the C++ SDK to use the same patterns that every other Azure SDK does - it updates the SetUpTestBase method to include the serviceDirectory, exactly as is done for all the other languages (each language has an idiomaticly appropriate version for the language - for Python, it's done in the `EnvironmentVariableLoader`, for C#, it's handled in the `TestEnvironment` class. Sometimes it's not obvious where it is, but a similar mapping exists for all the Azure languages I've worked with.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: Doesn't this solution (adding new environment variables) require modifying new-testresources.ps1 in eng\common to emit them in the tests? That seems highly intrusive and unlikely to pass muster with the engineering system team.

Note that my change does *not* modify the AZURE_XXX variables if the service specific variable isn't set, which means that people working in multiple services will continue to be able to use the AZURE_XXX variables.

Because of your concerns, I'll update the SeUpTestBase function such that if AZURE_XXX is already set, the code does nothing. That allows you to locally set AZURE_XXX in your environment and have it work.

sdk/core/azure-core/src/strings.cpp
```diff
@@ -65,22 +65,53 @@ const unsigned char LocaleInvariantLowercaseTable[256] = {
...
-    }
+    std::transform(result.begin(), result.end(), result.begin(), [](char const ch) {
+      return StringExtensions::ToLower(ch);
+    });
+    return result;
+  }
+
+  unsigned char StringExtensions::ToUpper(unsigned char const symbol) noexcept
+  {
+    return LocaleInvariantUppercaseTable[symbol];
```
vhvb1989, line 106: Can we re-use the same table from Lower?   If the symbol is > 64 && < 90 -> `minus 20`.  Basically, check the range of `symbol` and add or remove `20` so it becomes `toUpper`.

It would keep constant time complexity w/o adding the extra bytes to the memory for the entire table.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
vhvb1989: > Doesn't this solution (adding new environment variables) require modifying new-testresources.ps1 in eng\common

No.  The new env var is added to `sdk/attestation/ci.yml` in the `TestEnv` section.  And it replaces the input arg for the SetUpTestBase().  It would check for the env var instead of checking the inputArg.

And the Servive name is already in the ci.yml as the `ServiceDirectory`

sdk/core/azure-core/src/strings.cpp
LarryOsterman: The entire purpose of the table is to avoid branches in the ToUpper/ToLower function. 

If I check the range, and add/subtract values, I introduce branches, thus defeating the point of the table.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
vhvb1989: Well, we are introducing a business-rule to our code based for `keeping a test cofing file` portable across languages. In C++, the idiomatic rules for the code (the things why people would choose C++ and not other languages) are things like performance, memory, or something related.  It's not expected to see a pattern where API needs an input value so it can make test-configuration (and only the test-configuration, not the test-cases or test-data) portable to other languages.

If we would really like to make the test-configuration the same for all languages, it would need to be in the eng/common folder. Or pulled from some repository.  Whatever that is not a duplicated copy for each language.

If in the future, how do other language's developer know about this contract?  How do we avoid breaking it?  What if there's a features that go live only in one language in the future and requires changes to the test-configuration. There are multiple things why I think keeping that file portable is not easy to maintain and does not bring too much value against the effort to keep it portable.

LarryOsterman: This issue *only* applies to service implementers. It does not apply to the developers who use the SDK. Developers who use the SDK always use AZURE_CLIENT_ID, AZURE_TENANT_ID, and AZURE_CLIENT_SECRET. 

Service implementers follow the patterns established by the SDK - so for Python developers, they pass the serviceDirectory parameter from new-testresources.ps1 to the PowerShellPreparer (now called EnvironmentVariableLoader) which is used for their tests. For C#, the sdk figures out the serviceDirectory value by looking at the running assembly (it also guesses the recordings directory from the assembly), other languages have similar checks - some are explicit, some are implicit.

What I'm proposing in this pull request is that the pattern for C++ service implementers should include them specifying the serviceDirectory parameter that they specified in their call to new-testresources.ps1.

Also, the testresources.json defines the service specific configuration parameters, and the test-resources-pre.ps1 and test-resources-post.ps1 scripts perform actions before and after the testresources ARM template for the service is run. They do *not* belong in the eng\common directory, because they're not common to all services. Instead they belong in a per-service directory. And the should be able to be shared between different languages because fundamentally each language is interacting with the same service.

If a service brings up new functionality that requires that the arm template for the service be modified (infrequent, but it happens), the arm template should be able to be shared across different languages without modification - that way as each service's SDK team implements the new functionality, they operate across a common service baseline.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: Interesting idea. It's not nearly as discoverable as adding a parameter to SetUpTestBase(), but I can make that solution work.

sdk/core/azure-core/src/strings.cpp
LarryOsterman: If you have concerns about adding the additional 256 bytes to test images, I could probably rework the table to reflect what transform should be applied to the character - 0 means don't change, 1 means upper case character, 2 means lower case character. That changes `ToLower` to be:

``` C++
std::string const StringExtensions::ToLower(const std::string& src) noexcept
{
  unsigned char val = LocaleInvariantCaseTable[symbol];
  if (val == 0 || val == 2)
  {
    return symbol;
  }
  return symbol + 'a' - 'A';
}
```
and `ToUpper` becomes:
``` C++
std::string const StringExtensions::ToUpper(const std::string& src) noexcept
{
  unsigned char val = LocaleInvariantCaseTable[symbol];
  if (val == 0 || val == 1)
  {
    return symbol;
  }
  return symbol - 'a' - 'A';
}
```

That adds 2 branches to each call to ToUpper or ToLower.

Compare that with the somewhat more traditional implementation:
``` C++
std::string const StringExtensions::ToLower(const std::string& src) noexcept
{
  if (symbol >= 'A' && symbol <= 'Z')
  {
    return symbol + 'a' - 'A';
  }
  return symbol;
}
```

OTOH, that is functionally equivalent to `std::tolower(symbol)`, so I'm not 100% sure there's additional value provided by this.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: There's a problem. The TestEnv section from ci.yml is not provided to live tests. There doesn't appear to be a mechanism to specify environment variables to the ctest template :(.

So while this is an incredibly attractive option, it doesn't appear to be possible.

LarryOsterman: Hmm. I might have a solution

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
vhvb1989: > This issue only applies to service implementers. It does not apply to the developers 

Everyone from the community can be a service implementer.

> Service implementers follow the patterns established by the SDK

We don't have one for C++.  Are we trying to follow what other language's decided instead of creating the pattern for C++ within the team?  

> If a service brings up new functionality that requires that the arm template for the service be modified (infrequent, but it happens), the arm template should be able to be shared across different languages without modification 

Can you update the attestation arm template and add the env vars mapping in the outputs like the other services in C++?  And then share the updated arm template across languages?  It should still work for all other languages.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
vhvb1989: ci.yml pass the parameter `ServiceDirectory` to `eng/pipelines/templates/stages/archetype-sdk-client.yml` and this one pass the parameter later to `/eng/pipelines/templates/jobs/archetype-sdk-tests.yml` for liveTests.

All you need is to add: `AZURE_ENV_VAR_PREFIX: ${{ parameters.ServiceDirectory }}` to line 150:
image


This will set the service name to that env var.  Then you read that env var from the test-fw

vhvb1989: Note:  `ServiceDirectory` is lowercase. We need to uppercase it.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: > Can you update the attestation arm template and add the env vars mapping in the outputs like the other services in C++? And then share the updated arm template across languages? It should still work for all other languages.

This is the crux of my issue.  Every language other than C++ solves the problem of mapping <serviceDirectory>_XXX to AZURE_XXX inside the test framework.

C++ is special, requiring modifications to the test-resources.json for live tests to pass, and those modifications are inappropriate for the patterns used in all the other Azure languages I'm familiar with.

I'm pushing HARD against that pattern, because IMHO test-resources.json files should not be language specific.

FWIW, my current version of the solution solves this problem entirely inside the test framework, without requiring any modification to tests or json files.

sdk/core/azure-core/src/strings.cpp
vhvb1989: std::tolower works on the default `C` locale:
https://en.cppreference.com/w/cpp/string/byte/tolower

I can live having both tables.  It's not such a big deal.  

@CaseyCarter  would you have any suggestions on how to avoid having 2 tables and keeping the LocaleInvariant?

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: That is essentially my "I might have a solution". However from what I've seen, adding it to variables isn't sufficient - you need to add it to the `env:` section of the `ctest` task to ensure that it is set when ctest runs.

And IMHO the environment variable should reflect the name of the parameter - It's refered to as "ServiceDirectory" in every other part of the ES (and in every other language), so it becomes AZURE_SERVICE_DIRECTORY. That way people who are working on the ES will understand the construct.

LarryOsterman: PS: When you reply to a resolved conversation, please unresolve it - it took me almost 10 minutes to find this comment (I *hate* the github pull request UX).

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
vhvb1989: @weshaggard  @benbp @danieljurek  Is C++ violating any rule by using test-resources.json to do `env var mapping` as prt of the `outputs`? like:

https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/storage/test-resources.json#L323

If yes, we might want to update and fix this for all C++ services, so we keep the pattern consistent within C++

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
vhvb1989: > so it becomes AZURE_SERVICE_DIRECTORY

I am OK with any name :) .  I was suggesting something to self-describe what it will do when it is set.  But we can add documentation notes somewhere (maybe in the test fw) to describe this env var meaning

sdk/core/azure-core/src/strings.cpp
vhvb1989: ah!.  It might be that it is only the `StringComparison` method the one that is locale invariant because for string A and B we would use the same table.  I am guessing...

LarryOsterman: FWIW, having the table is a good thing, IMHO - as I mentioned, it should be dramatically faster than using `std::toupper`/`std::tolower`. And we have the ability of modifying it to be different from the `c` locale in the future if we discover it's necessary.

CaseyCarter: The point of the table is that it's not associated with any locale. `std::to{lower,upper}` respect the current locale, these functions using this table behave as they do in the locale named "The 'C' locale", which is the default if the program has specified no other locale. We don't want our behavior to depend on what - if any - locale a user may have set, so we have these locale-invariant functions.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: Do you mean something like this existing comment:
```C++
     * @note If AZURE_TENANT_ID, AZURE_CLIENT_ID, or AZURE_CLIENT_SECRET are not available in the
     * environment, the AZURE_SERVICE_DIRECTORY environment variable is used to set those values
     * with the values emitted by the New-TestResources.ps1 script.
```

LarryOsterman: Btw, for completeness: I confirmed that `variables` are added to the environment for the test, so I've added it to the variables section in `archetype-sdk-tests.yml`.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -276,6 +301,38 @@ namespace Azure { namespace Core { namespace Test {
...
+            if (!targetValue.empty())
+            {
+              Azure::Core::_internal::Environment::SetVariable(
+                  azureVariable.c_str(), targetValue.c_str());
+            }
+          }
+        };
+        SetBuiltinEnvironment("_TENANT_ID");
+        SetBuiltinEnvironment("_CLIENT_ID");
+        SetBuiltinEnvironment("_CLIENT_SECRET");
```
vhvb1989, line 334: There's one little details still making me wonder...  

- We are setting env vars with the tests, which is never something we should :(
- We are setting tenant, client and secret on SetUpBaseTest.  What if the test doesn't even need this values?  It seems risky to me to set this values whenever it is not playback mode (record & live).
- We are setting env vars and we are not unsetting them, I don't think that's good.

Question ->  Instead of adding this logic to `SetUpBaseTest`, can we add it to `GetEnv()` ?  So it is when the test try to get `AZURE_TENANT_ID` when we run this logic and fallback to `ATTESTATION_TENANT_ID` ?  And we don't set the env var, but only return the value of

LarryOsterman: Good points. I thought about adding it to `GetEnv` (actually that was my first implementation - I decided to follow the python pattern because it resulted in a simpler experience), But moving the logic to GetEnv makes sense. It will be slightly more complicated, but reasonable.

FWIW, as far as unsetting environment variables - environment variables are per-process, there is no durable effect outside the current process.

benbp:
yml/env var changes LGTM.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -149,6 +149,8 @@ jobs:
     CmakeArgs: ""
     AZURE_TEST_MODE: "LIVE"
     AZURE_LOG_LEVEL: "verbose"
+    # Surface the ServiceDirectory parameter as an environment variable so tests can take advantage of it.
+    AZURE_SERVICE_DIRECTORY: ${{ parameters.ServiceDirectory }}
```
benbp, line 153: @LarryOsterman I created https://github.com/Azure/azure-sdk-tools/pull/2810 to set this in the core deployment script, which should also make it slightly easier to run locally if copy/pasting the env var outputs from that script.

LarryOsterman: Once your PR is approved and committed, I'll remove this since it makes this change redundant.

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -243,13 +243,61 @@ namespace Azure { namespace Core { namespace Test {
...
-
+#if !defined(NDEBUG)
+      // The azure CI pipeline uppercases all EnvVar values from ci.yml files.
+      // That means that any mixed case strings will not be found when run from the CI
+      // pipeline. Check to make sure that the developer only passed in an upper case environment
+      // variable.
+      {
+        if (name != Azure::Core::_internal::StringExtensions::ToUpper(name))
+        {
+          throw std::runtime_error("All Azure SDK environment variables must be all upper case.");
```
vhvb1989, line 272: Here's an idea. It can be done after this PR anyway.

Instead of throw if `name` is not all UPPERCASE, let's always call `ToUpper(name)` and document that the c++ test-fw do the same as the ADO and works with upper case env vars only.

We are "kind" of doing something like this already with `AZURE_TEST_MODE`, where the value is checked regardless of the case: https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core-test/src/interceptor_manager.cpp#L29

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -243,13 +243,61 @@ namespace Azure { namespace Core { namespace Test {
...
+        }
+      }
+#endif
+      auto ret = Azure::Core::_internal::Environment::GetVariable(name.c_str());
       if (ret.empty())
       {
+        if (!m_testContext.IsPlaybackMode() && name.find("AZURE_") == 0)
```
vhvb1989, line 279: nit.
Move "AZURE_" to a const variable and re-use it below on line 294 for calling substr()

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -303,6 +303,9 @@ void WinHttpTransport::CreateRequestHandle(std::unique_ptr<_detail::HandleManage
 {
   const std::string& path = handleManager->m_request.GetUrl().GetRelativeUrl();
   HttpMethod requestMethod = handleManager->m_request.GetMethod();
+  bool requestSecureHttp(
```
vhvb1989, line 306: nit.

```suggestion
  bool const requestSecureHttp(
```

vhvb1989 (approved):
Beautiful!  Thank you!

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
LarryOsterman: That's an interesting suggestion - my biggest concern with it is that penalizes developers who provide their names in uppercase always. My change was intentionally only on debug builds to reduce the performance consequences (otoh, we're about to spend many milliseconds talking to a remote service, so....)

LarryOsterman: Created Issue #3375 to track this suggestion.

vhvb1989: It would be only for test-fw.

benbp (approved):

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -330,6 +330,23 @@ void WinHttpTransport::CreateRequestHandle(std::unique_ptr<_detail::HandleManage
...
+    //
+    // Note: If/When TLS client certificate support is added to the pipeline, this line may need to
+    // be revisited.
+    if (!WinHttpSetOption(
+            handleManager->m_requestHandle,
+            WINHTTP_OPTION_CLIENT_CERT_CONTEXT,
+            WINHTTP_NO_CLIENT_CERT_CONTEXT,
+            0))
+    {
+      GetErrorAndThrow("Error while setting client cert context to ignore..");
```
ahsonkhan, line 347: nit: Extra period
```suggestion
      GetErrorAndThrow("Error while setting client cert context to ignore.");
```

sdk/core/azure-core/inc/azure/core/internal/strings.hpp
```diff
@@ -33,8 +33,10 @@ namespace Azure { namespace Core { namespace _internal {
     static bool LocaleInvariantCaseInsensitiveEqual(
         const std::string& lhs,
         const std::string& rhs) noexcept;
-    static std::string const ToLower(const std::string& src) noexcept;
-    static unsigned char ToLower(const unsigned char src) noexcept;
+    static std::string const ToLower(std::string const& src) noexcept;
+    static unsigned char ToLower(unsigned char const src) noexcept;
+    static std::string const ToUpper(std::string const& src) noexcept;
+    static unsigned char ToUpper(unsigned char const src) noexcept;
```
ahsonkhan, line 39: This API addition, as a complement to `ToLower` makes sense to add.

However, unlike private APIs in the `_detail` namespace, we cannot make breaking changes to things in `_internal`. Therefore, we typically only add public surface area to such headers within the SDK/Azure Core when an actual upstream service SDK package within the repo needs it. In this case, this looks to only be used by unit tests.

In general, we'd avoid adding APIs to the SDK that are only used by unit tests :)

PR #3543, opened 2022-04-11
API Review Feedback for Attestation SDK

LarryOsterman:
# Pull Request Checklist

- [ ] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [X] Comments in source
- [X] No typos
- [x] Update changelog
- [ ] Not work-in-progress
- [X] Self review of PR done
- [X] Any breaking changes?


gearama (approved):
Assuming it's just feedback from the archboard

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
```diff
@@ -204,12 +204,11 @@ namespace Azure { namespace Security { namespace Attestation { namespace _detail
...
 
+    // Set the token body based on the bodyToSet parameter provided.
     template <typename Ty>
-    void SetTokenBody(
-        Azure::Core::Json::_internal::json const& jsonBody,
-        Azure::Nullable<Ty> bodyToSet)
+    void SetTokenBody(Azure::Core::Json::_internal::json const& jsonBody, Ty const* const bodyToSet)
     {
-      if (bodyToSet)
+      if (bodyToSet != nullptr)
```
vhvb1989, line 211: why do we need the `!= nullptr` ??   I feel to me like you can remove it

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
```diff
@@ -231,7 +231,7 @@ namespace Azure { namespace Security { namespace Attestation { namespace _detail
      *
      * @param jwt - the JSON Web Token/JSON Web Signature to be parsed.
      */
-    AttestationTokenInternal(std::string const& jwt, Azure::Nullable<T> preferredBody = {})
+    AttestationTokenInternal(std::string const& jwt, T const* const preferredBody = nullptr)
```
vhvb1989, line 234: the preferredBody is not mentioned in the doxygent comments. It's not required here for non-public.
But I would still like to know what it is used for.

vhvb1989 (changes requested):
Change the public APIs from returning Azure::Core::Url

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -66,11 +66,11 @@ namespace Azure { namespace Security { namespace Attestation {
...
 
     /**
-     * @brief Returns the API version the client was configured with.
+     * @brief Returns the Endpoint which the client is communicating with.
      *
-     * @returns The API version used when communicating with the attestation service.
+     * @returns The remote endpoint used when communicating with the attestation service.
      */
-    std::string const& ClientVersion() const { return m_apiVersion; }
+    Azure::Core::Url const& Endpoint() const { return m_endpoint; }
```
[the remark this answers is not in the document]
LarryOsterman: I'll make the change, but why? It's a public type, why not return it?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -126,7 +127,7 @@ namespace Azure { namespace Security { namespace Attestation {
   {
     /** @brief Version to use when communicating with the attestation service.
      */
-    ServiceVersion Version;
+    ServiceVersion Version{ServiceVersion::V2020_10_01};
```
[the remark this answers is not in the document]
LarryOsterman: API Review mentioned that the value wasn't discoverable - this might be a limitation of the current ApiView implementation but I got explicit feedback to add a default constructor so it can be discoverable.

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
LarryOsterman: `bodyToSet` is a raw pointer, not a smart pointer. IMHO when checking pointers, it's best not be explicit about null pointer checks and not rely on the implicit conversion from pointer to integer and then from integer to boolean.

`std::xxxx_ptr` provides an `operator bool` which allows customers to use the `if (bodyToSet)` construct safely. THe same is true for `Azure::Nullable`.

And you're right, it's not strictly necessary - the relevant pointer conversions are safe. But this is a coding style convention I have.

Also, to be clear: I *really* don't like the whole raw pointer thing. But Nullable<T> cannot be specialized on <void>.

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
LarryOsterman: Comment added. Basically this allows a creator of an AttestationTokenInternal to create an AttestationToken whose body doesn't match the object contained in the JWT.

The primary use for this is in the GetPolicy API - the JWT contains a StoredAttestationPolicy object, which has a single value AttestationPolicy, which is a Base64Url encoded version of the policy document. That complexity is necessary for the wire protocol but is unnecessary for SDK customers, so the GetPolicy API returns an AttestationToken<std::string>.

To make that happen, `preferredBody` is set to the std::string value to be returned to the customer.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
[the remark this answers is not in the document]
vhvb1989: It should be enough to add a comment to the field to say that the value is set with the constructor.
Otherwise, we can have a sync bug in the future where we believe we updated the apiVersion by changing the default value for the field, but the constructor would still be setting the old value.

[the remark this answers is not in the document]
vhvb1989: Or, you could also remove the parameters from the constructor.  Let the constructor only call base-class constructor but let the field's default values.  It would be different from the other SDK clients (storage, key vault) but I think it makes sense

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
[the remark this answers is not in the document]
vhvb1989: it's been a debate in the past, but so far there is no API public operation which takes Azure::Core::Url as a input.

When creating an SDK client takes an URI, it is represented as an std::string so it is more idiomatic to the language and in a way customer don't need to learn about the SDK's URI representation.

The Azure::Core::Url is not a general purpose component.  People can't expect it to work for other than Azure URIs.   

I couldn't convince @ahsonkhan in the past to let Azure::Core::Url within the Azure::_internal namespace and it ended up in the public surface.  However, it is still sort of invisible to customers (not output or input anywhere)

sdk/attestation/azure-security-attestation/src/private/attestation_client_private.hpp
vhvb1989: makes sense. Thank you

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -139,28 +132,78 @@ namespace Azure { namespace Security { namespace Attestation {
      *
      * @note TPM attestation requires an authenticated attestation client.
      */
-    explicit AttestationClient(
```
vhvb1989, line 142: why is explicit gone?  We should not allow implicit conversions,

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -139,28 +132,78 @@ namespace Azure { namespace Security { namespace Attestation {
...
 
-    /**
-     * @brief Returns the API version the client was configured with.
+    /** @brief Construct a new Attestation Client object
+     *
+     * @param endpoint The URL address where the client will send the requests to.
+     * @param credential The authentication method to use (required for TPM attestation).
+     * @param options The options to customize the client behavior.
+     */
+    static AttestationClient const Create(
```
vhvb1989, line 164: Is this part of the API review?   Why do we want a factory here?  I haven't seen that elsewhere (on C++)

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -139,28 +132,78 @@ namespace Azure { namespace Security { namespace Attestation {
...
+        Azure::Core::Context const& constext = Azure::Core::Context{});
+
+    /** @brief Construct a new anonymous Attestation Client object
+     *
+     * @param endpoint The URL address where the client will send the requests to.
+     * @param options The options to customize the client behavior.
+     *
+     * @note TPM attestation requires an authenticated attestation client.
+     */
+    static AttestationClient* const CreatePointer(
```
vhvb1989, line 201: should it be at least `std::unique_ptr<AttestationClient>` ?  Who's going to delete this?

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -215,11 +211,68 @@ void AttestationClient::RetrieveResponseValidationCollateral(
...
 }
 
-void AttestationClient::CheckAttestationSigners() const
+/** @brief Construct a new Attestation Client object
+ *
+ * @param endpoint The URL address where the client will send the requests to.
+ * @param credential The authentication method to use (required for TPM attestation).
+ * @param options The options to customize the client behavior.
+ */
+AttestationClient const AttestationClient::Create(
```
vhvb1989, line 220: The `not-so-good` about factory methods is that they are accessible from class instances as well.
So, you attestation client will also list the `client.Create()` method just like the class `AttestationClient::Create()`.

What I usually prefer is to create the factory as an additional class with no constructor (real factory). Something like `AttestationClientFactory`.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
```diff
@@ -215,11 +211,68 @@ void AttestationClient::RetrieveResponseValidationCollateral(
...
+    std::string const& endpoint,
+    std::shared_ptr<Core::Credentials::TokenCredential const> credential,
+    AttestationClientOptions options,
+    Azure::Core::Context const& context)
+{
+  std::unique_ptr<AttestationClient> returnValue(
+      new AttestationClient(endpoint, credential, options));
+  returnValue->RetrieveResponseValidationCollateral(context);
+  // Release the client pointer from the unique pointer to let the parent manage it.
+  return returnValue.release();
```
vhvb1989, line 262: return the `std::unique_ptr`.  It's better usability for customers

vhvb1989, line 262:
```suggestion
  auto returnValue = std::make_unique<AttestationClient>(endpoint, credential, options);
  returnValue->RetrieveResponseValidationCollateral(context);
  return returnValue;
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: What value does explicit have on a private constructor?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: This was a compromise worked out between @tg-msft, @JeffreyRichter  and I - instead of allowing the clients to be explicitly constructed, we use a factory method and issue the magic I/O in the factory method. That was seen as a significant usability barrier during API review (especially since we know exactly what the developer mistake was).

Moving to a factory pattern removes the super-undiscoverable API that you have to call before the service will work. And in addition, since the factory takes a Context, it makes it clear that the factory does I/O.

While the factory pattern is novel to the Attestation service, it's not an unreasonable pattern for C++ classes (for instance, the [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c50-use-a-factory-function-if-you-need-virtual-behavior-during-initialization) recommend it for certain constructs.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: I was concerned about making a choice about what to return to the caller and instead was relying on documentation. However reading C.50, I've changed it to always return a std::unique_ptr - callers that want a shared_ptr can transfer it themselves.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
vhvb1989: Ah,  I was missing the constructor is now private.  Then the issue is not the std::unique_ptr, but calling `std::make_unique<T>` (It has not access to private).

So, instead of calling `std::make_unique<T>`, you can just do:
```cpp
static std::unique_ptr<Factory> Create() {
        auto instance = std::unique_ptr<Factory>(new Factory());
        return instance;
    }
```
Then you can return the unique_ptr.  See this https://godbolt.org/z/obs3b3zae

LarryOsterman: To my eyes, that is almost identical to the original code with the exception of removing the .release, which I removed earlier.  What is the difference here?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -114,53 +114,96 @@ namespace Azure { namespace Security { namespace Attestation {
    */
 
   class AttestationClient final {
+    /** @brief Construct a new Attestation Client object
```
vhvb1989, line 117: please don't hate me for this, but, can you explicitly add `private` here?

I know that's the default for `class`, but there's like a quick readability improvement when the eyes look at it w/o checking if this is a class or a struct.

I can live with it if you don't want to add it, but if it is not such a big deal for you, I would appreciate it :)

For example, I was not initially realizing that the constructors were becoming private because I usually look for either private or public at the first place (I know, that's my bad)

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -114,53 +114,96 @@ namespace Azure { namespace Security { namespace Attestation {
...
     /** @brief Construct a new anonymous Attestation Client object
      *
      * @param endpoint The URL address where the client will send the requests to.
      * @param options The options to customize the client behavior.
+     * @return AttestationClient The newly created client.
      *
      * @note TPM attestation requires an authenticated attestation client.
      */
-    explicit AttestationClient(
+    static AttestationClient const Create(
```
vhvb1989, line 165: I have some feelings here about returning a `const` object.  Why did you decide this?

To me, as a user, it means I am getting an instance that I can't mutate or call any non-constant method from it.  Is that expected?

I would typically expect to be the customer who would decide if calling:
`auto client = AttestationClient::Create();`
or
`auto const client = AttestationClient::Create();`

depending on my intentions for the client.  But it might be an explanation

LarryOsterman: The C++ SDK guidelines say that clients are immutable.

The way you express immutability in C++ is to mark the class as `const`. I'm leveraging language features to make the contract specified in the guidelines explicit.

The downside of that is const poisoning. But there is no ambiguity on the part of customers that service clients are immutable.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -44,57 +44,31 @@ namespace Azure { namespace Security { namespace Attestation {
...
-    std::string const& ClientVersion() const { return m_apiVersion; }
-
     /**
      * @brief Construct a new Attestation Administration Client object from another attestation
      * administration client.
      *
      * @param attestationClient An existing attestation client.
      */
-    explicit AttestationAdministrationClient(
+    AttestationAdministrationClient(
```
vhvb1989, line 54: this is still public right?  so do we want to keep the `explicit` ?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -121,46 +121,16 @@ namespace Azure { namespace Security { namespace Attestation {
...
-        : AttestationClient(endpoint, nullptr, options)
-    {
-    }
-
     /** @brief Construct a new Attestation Client object from an existing attestation client.
      *
      * @param attestationClient An existing attestation client.
      */
-    explicit AttestationClient(AttestationClient const& attestationClient)
+    AttestationClient(AttestationClient const& attestationClient)
```
vhvb1989, line 128: public, keep `explicit` ?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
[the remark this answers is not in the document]
LarryOsterman: I went back to the ApiReview feedback, I think this is primarily and ApiView issue and not worth fixing.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
LarryOsterman: It's not public any more. And it can't be explicit otherwise you can't return a std::unique_ptr to it :(.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: It's the copy constructor and it can't be explicit (otherwise you can't return a `std::unique_ptr<AttestationClient>` from a method). I'm not sure why that is, but it is.

LarryOsterman: Ok, this is weird. If I make the copy constructor explicit, I get:
```
  C:\PROGRA~1\MIB055~1\2022\ENTERP~1\VC\Tools\MSVC\1431~1.311\bin\Hostx64\x64\cl.exe  /nologo /TP -DAZ_RTTI -DBUILD_CURL_HTTP_TRANSPORT_ADAPTER -DCURL_STATICLIB -IG:\Az\LarryO\azure-sdk-for-cpp\sdk\attestation\azure-security-attestation\inc -IG:\Az\LarryO\azure-sdk-for-cpp\sdk\core\azure-core\inc -external:I G:\Az\LarryO\azure-sdk-for-cpp\out\build\x64-DebugWithTests\vcpkg_installed\x64-windows-static\include -external:W0 /DWIN32 /D_WINDOWS /W4 /GR /EHsc /MDd /Zi /Ob0 /Od /RTC1 /permissive- /W4 /WX /wd5031 /wd4668 /wd4820 /wd4255 /wd4710 -std:c++14 /showIncludes /Fosdk\attestation\azure-security-attestation\CMakeFiles\azure-security-attestation.dir\src\attestation_client.cpp.obj /Fdsdk\attestation\azure-security-attestation\CMakeFiles\azure-security-attestation.dir\azure-security-attestation.pdb /FS -c G:\Az\LarryO\azure-sdk-for-cpp\sdk\attestation\azure-security-attestation\src\attestation_client.cpp
G:\Az\LarryO\azure-sdk-for-cpp\sdk\attestation\azure-security-attestation\src\attestation_client.cpp(228): error C2440: 'return': cannot convert from 'Azure::Security::Attestation::AttestationClient' to 'Azure::Security::Attestation::AttestationClient'
  G:\Az\LarryO\azure-sdk-for-cpp\sdk\attestation\azure-security-attestation\src\attestation_client.cpp(228): note: Constructor for class 'Azure::Security::Attestation::AttestationClient' is declared 'explicit'
G:\Az\LarryO\azure-sdk-for-cpp\sdk\attestation\azure-security-attestation\src\attestation_client.cpp(243): error C2440: 'return': cannot convert from 'const Azure::Security::Attestation::AttestationClient' to 'Azure::Security::Attestation::AttestationClient'
  G:\Az\LarryO\azure-sdk-for-cpp\sdk\attestation\azure-security-attestation\src\attestation_client.cpp(243): note: Constructor for class 'Azure::Security::Attestation::AttestationClient' is declared 'explicit'
```
However if I put a breakpoint, the copy constructor is never called.

LarryOsterman: Digging more. [Making a copy constructor explicit prevents it from being used as a parameter to a method](https://en.wikipedia.org/wiki/Copy_constructor_(C++)#Explicit_copy_constructor). And it also means you can never return an instance of the object from a function.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -114,53 +114,96 @@ namespace Azure { namespace Security { namespace Attestation {
...
+     * @param credential The authentication method to use (required for TPM attestation). If the
+     * credential parameter is not supplied, the connection will be unauthenticated.
+     * @param options The options to customize the client behavior.
+     * @return std::unique_ptr<AttestationClient> The newly created client.
      */
-    explicit AttestationClient(AttestationClient const& attestationClient)
-        : m_endpoint(attestationClient.m_endpoint), m_apiVersion(attestationClient.m_apiVersion),
-          m_pipeline(attestationClient.m_pipeline),
-          m_tokenValidationOptions(attestationClient.m_tokenValidationOptions){};
+    static std::unique_ptr<AttestationClient> CreatePointer(
```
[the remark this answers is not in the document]
LarryOsterman: That was a mistake - they all should be const.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: Actually I was wrong. Most of the other languages guidelines say clients are immutable, C++ is mute on the subject. Given @JeffreyRichter's clarification that the real requirement is thread safe and not immutable, I'm reverting this change.

sdk/attestation/azure-security-attestation/src/attestation_client.cpp
LarryOsterman: Moved to a separate issue so this commit can be completed.

vhvb1989 (approved):
LGTM

sdk/core/azure-core-test/CMakeLists.txt
```diff
@@ -45,5 +45,5 @@ target_include_directories (azure-core-test-fw
 
 # make sure that users can consume the project as a library.
 add_library (Azure::Core::Test ALIAS azure-core-test-fw)
-target_link_libraries(azure-core-test-fw PRIVATE azure-core gtest)
+target_link_libraries(azure-core-test-fw PRIVATE azure-core Azure::azure-identity gtest)
```
ahsonkhan, line 48: Why does azure-core-test depend on azure-identity? This means we can't consume this project as a standalone, or exclude building azure-identity and still have core/core-tests work.

It looks like we introduced this for test_base.hpp adding `CreateClientSecretCredential`, which we then use in Attestation unit tests (and only there). I think this dependency should move there and removed from `test_base.hpp` in azure-core. OR, we find some other mechanism of having a "core" test, but one that doesn't build with azure-core. 

Having azure-core-test depend on something upstream like azure-identity is inverted to me.

PR #3561, opened 2022-04-15
Initial implementation of OpenTelemetry APIs.

### LINKED ISSUE #3554: Create Distributed Tracing API shims and azure-core-opentelemetry VCPKG package for consumption

LarryOsterman:
# Implement OpenTelemetry APIs as the foundation for implementing distributed tracing.

This pull request implements a set of APIs to implement OpenTelemetry APIs as the foundation for distributed tracing. Future pull
requests will integrate these APIs into the core azure functionality.

As per discussions with the OpenTelemetry team, OpenTelemetry should be implemented as a package outside azure core, that enables flexibility w.r.t. telemetry technologies. 

As such, Azure::Core::Tracing defines a set of abstract base classes which define the tracing API contract expected by Azure Core. The new azure-core-opentelemetry-cpp package contains an implementation of those abstract base classes. 

A customer who wishes to enable tracing in their application would do something like this:
```cpp
// <Create a new OpenTelemetry tracing provider and set that as the default trace provider:
opentelemetry::trace::Provider::SetTracerProvider(myTracerProvider);

// Next create an Azure Core Tracing OpenTelemetry Provider using the
// default OpenTelemetry tracer provider which was just set.
std::shared_ptr<Azure::Core::Tracing::TracerProvider> tracerProvider = std::make_shared<Azure::Core::Tracing::OpenTelemetry::TracerProvider>();

// Connect the tracerProvider to the current application context.
ApplicationContext().SetTracerProvider(tracerProvider);
```

This pattern is somewhat different from the pattern normally used in Azure Core to implement plugins, that was a consequence of the desire to keep the entire OpenTelemetry implementation abstracted from the core abstractions.

This pull request also implements functionality to integrate Service Method tracing into the OpenTelemetry APIs.

To integrate OpenTelemetry into your service method implementation, you need to add a new `ServiceTracing` object to
your service client:

```cpp
 class ServiceClient {
  private:
    ServiceClientOptions m_clientOptions;
    Azure::Core::Tracing::_internal::DiagnosticTracingFactory m_diagnosticFactory;

  public:
    explicit ServiceClient(ServiceClientOptions const& clientOptions = ServiceClientOptions{})
        : m_diagnosticFactory(clientOptions, "<Telemetry Service Name>", PackageVersion::ToString())
    {
    }
```

Then, on each Service Method, you ask the service tracing object to create a "span" for the API:
```cpp
    Azure::Response<std::string> MyServiceMethod(
        std::string const& firstParameter,
        Azure::Core::Context const& context = Azure::Core::Context{})
    {
      // `CreateSpan` returns a std::pair, the `first` value is an updated context which MUST be used
      // for subsequent operations, the `second` value is a Span object that will be used to reflect the status
      // of the service method to tracing.
      auto contextAndSpan = m_diagnosticFactory.CreateSpan("MyServiceMethod", context);

      // <Call Into Service via an HTTP pipeline using `contextAndSpan.first` as the `context` value.

      // Reflect that the operation was successful.
      contextAndSpan.second.SetStatus(Azure::Core::Tracing::_internal::SpanStatus::Ok);
      <Return a Response<std::string> to the caller.

      // When contextAndSpan goes out of scope, it ends the span, which will record it.
    }
```

## Pull Request Checklist

- [x] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [X] Comments in source
- [X] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [X] Self review of PR done
- [X] Any breaking changes?

Fixes #3554


sdk/core/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 
 add_subdirectory(azure-core)
+if (NOT DISABLE_AZURE_CORE_OPENTELEMETRY)
+add_subdirectory(azure-core-opentelemetry)
```
gearama, line 13: Nit : indenting on line 13

sdk/core/azure-core-opentelemetry/NOTICE.txt
```diff
@@ -0,0 +1,135 @@
+azure-core
+
+NOTICES AND INFORMATION
+Do Not Translate or Localize
```
gearama, line 4: why? probably some legal thing .. still .. sad

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * @brief Azure SDK abstractions.
+ *
+ */
+namespace Azure {
+
+/**
+ * @brief Azure Identity SDK abstractions.
+ *
+ */
+namespace Core {
```
gearama, line 51: this might not be properly formatted here, it should be indented since it's under Azure namespace

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
```diff
@@ -0,0 +1,61 @@
...
+  /** @brief Azure Tracing Abstractions
+   */
+  namespace Tracing {
+    /** @brief OpenTelemetry Tracing Abstractions
+     */
+    namespace OpenTelemetry {
+    }
+  }
+
+} // namespace Azure
```
gearama, line 61: Also you are missing a closing bracket , 4 are open , 3 are closing

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/opentelemetry.hpp
```diff
@@ -0,0 +1,150 @@
...
+  public:
+    OpenTelemetryProvider(
+        opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider> tracerProvider);
+    OpenTelemetryProvider();
+
+    virtual std::shared_ptr<Azure::Core::Tracing::Tracer> CreateTracer(
+        std::string const& name,
+        std::string const& version = "") const override;
+  };
+}}}} // namespace Azure::Core::Tracing::OpenTelemetry
```
gearama, line 150: missing new line at EOF

sdk/core/azure-core-opentelemetry/src/opentelemetry.cpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  void OpenTelemetrySpan::AddEvent(std::string const& eventName) { m_span->AddEvent(eventName); }
+
+  void OpenTelemetrySpan::AddEvent(std::exception const& ex) { m_span->AddEvent(ex.what()); }
+
+  void OpenTelemetrySpan::SetStatus(SpanStatus const& status, std::string const& statusMessage)
+  {
+    opentelemetry::trace::StatusCode statusCode = opentelemetry::trace::StatusCode::kUnset;
+    if (status == Azure::Core::Tracing::SpanStatus::Error)
```
gearama, line 159: switch maybe ?

sdk/core/azure-core-opentelemetry/src/opentelemetry.cpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  std::shared_ptr<Azure::Core::Tracing::Span> OpenTelemetryTracer::CreateSpan(
+      std::string const& spanName,
+      CreateSpanOptions const& options = {}) const
+  {
+    opentelemetry::trace::StartSpanOptions spanOptions;
+    spanOptions.kind = opentelemetry::trace::SpanKind::kInternal;
+    if (options.Kind == Azure::Core::Tracing::SpanKind::Client)
+    {
```
gearama, line 53: switch maybe?

sdk/core/azure-core-opentelemetry/src/opentelemetry.cpp
```diff
@@ -0,0 +1,179 @@
+
```
gearama, line 1: file header

sdk/core/azure-core-opentelemetry/src/private/package_version.hpp
```diff
@@ -0,0 +1,62 @@
...
+              AZURE_CORE_OPENTELEMETRY_VERSION_MINOR) "." AZURE_CORE_OPENTELEMETRY_VERSION_ITOA(AZURE_CORE_OPENTELEMETRY_VERSION_PATCH) "-" AZURE_CORE_OPENTELEMETRY_VERSION_PRERELEASE
+          : AZURE_CORE_OPENTELEMETRY_VERSION_ITOA(AZURE_CORE_OPENTELEMETRY_VERSION_MAJOR) "." AZURE_CORE_OPENTELEMETRY_VERSION_ITOA(
+              AZURE_CORE_OPENTELEMETRY_VERSION_MINOR) "." AZURE_CORE_OPENTELEMETRY_VERSION_ITOA(AZURE_CORE_OPENTELEMETRY_VERSION_PATCH);
+    }
+  };
+}}}} // namespace Azure::Core::OpenTelemetry::_detail
+
+#undef AZURE_CORE_OPENTELEMETRY_VERSION_ITOA_HELPER
+#undef AZURE_CORE_OPENTELEMETRY_VERSION_ITOA
+
```
gearama, line 58: remove empty line

sdk/core/azure-core-opentelemetry/src/private/package_version.hpp
```diff
@@ -0,0 +1,62 @@
...
+#define AZURE_CORE_OPENTELEMETRY_VERSION_ITOA(i) AZURE_CORE_OPENTELEMETRY_VERSION_ITOA_HELPER(i)
+
+namespace Azure { namespace Core { namespace OpenTelemetry { namespace _detail {
+  /**
+   * @brief Provides version information.
+   *
+   */
+  class PackageVersion final {
+  public:
+    /// Major numeric identifier.
```
gearama, line 28: not the right comment type 
use 
/**
* @brief
*/

sdk/core/azure-core-opentelemetry/test/ut/azure_core_otel_test.cpp
```diff
@@ -0,0 +1,690 @@
...
+#if defined(_MSC_VER)
+#pragma warning(pop)
+#endif
+
+#include <chrono>
+#include <gtest/gtest.h>
+
+class OpenTelemetryTests : public Azure::Core::Test::TestBase {
+private:
+  class CustomLogHandler : public opentelemetry::sdk::common::internal_log::LogHandler {
```
gearama, line 31: should this class definition be pulled out ?

sdk/core/azure-core-opentelemetry/test/ut/azure_core_otel_test.cpp
```diff
@@ -0,0 +1,690 @@
...
+  }
+
+  // Create
+  virtual void SetUp() override
+  {
+    Azure::Core::Test::TestBase::SetUpTestBase(AZURE_TEST_RECORDING_DIR);
+
+    opentelemetry::sdk::common::internal_log::GlobalLogHandler::SetLogHandler(
+        opentelemetry::nostd::shared_ptr<opentelemetry::sdk::common::internal_log::LogHandler>(
+            new CustomLogHandler()));
```
gearama, line 124: having the class definition outside doea not make life harder here

gearama (changes requested):

sdk/core/azure-core-opentelemetry/NOTICE.txt
LarryOsterman: Corporate Governance thingy. Controlled by the cgmanifest.json file.

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
LarryOsterman: THis is what clang-format came up with :(. I don't know why but it's what it does. I tried changing it to add the indentation but clang-format removed it :(.

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
LarryOsterman: Surprised that compiles... But fixed.

sdk/core/azure-core-opentelemetry/src/opentelemetry.cpp
LarryOsterman: C++ doesn't support using "switch" on strings :(. It would be nice if it did, but that's not an option :(.

sdk/core/azure-core-opentelemetry/src/opentelemetry.cpp
LarryOsterman: switch doesn't work on std::string :(.

sdk/core/azure-core-opentelemetry/src/opentelemetry.cpp
LarryOsterman: Surprised this isn't in the CI pipeline.

sdk/core/azure-core-opentelemetry/src/private/package_version.hpp
LarryOsterman: I'm following the example used in template. If this changes, we should change all the packages that follow the template's example.

sdk/core/azure-core-opentelemetry/test/ut/azure_core_otel_test.cpp
LarryOsterman: Why? I don't think it should be global, why do you think it should be global?

sdk/core/azure-core-opentelemetry/src/private/package_version.hpp
gearama: yet somehow it is incosistent with itself the class comment is one way, the public members is another. also if i remeber corectly there is a guidance for cpp regarding comments on public members https://azure.github.io/azure-sdk/cpp_introduction.html#documentation-style

sdk/core/azure-core-opentelemetry/test/ut/azure_core_otel_test.cpp
gearama: nested private classes seems somehow dirty, but given that it's for testing purposes ... fine

gearama (approved):
i'm not going to block it for a comment formatting

sdk/core/azure-core-opentelemetry/src/private/package_version.hpp
LarryOsterman: I agree that this should be fixed. But fixing it involves changing:
* Azure::Attestation
* Azure::Core
* Azure::Identity
* Azure::Keyvault::Secrets
* Azure::Keyvault::Cryptography
* Azure::Keyvault::Certificates
* Azure::Storage::Blobs
* Azure::Storage::Common
* Azure::Storage::Files::Datalake
* Azure::Storage::Files::Shares
* Azure::Storage::Queues
* Azure::Template

Making that change is beyond the scope of this PR. However it makes a lot of sense long-term. I'll raise an issue.

sdk/core/azure-core-opentelemetry/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Release History
```
ahsonkhan, line 1: Question on naming and folder structure: How do other languages name this (and the namespace)?

For example, I am seeing tracing in there as well, like core-tracing-opentelemetry:
https://docs.microsoft.com/en-us/python/api/overview/azure/core-tracing-opentelemetry-readme?view=azure-python-preview

sdk/core/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 
 add_subdirectory(azure-core)
+if (NOT DISABLE_AZURE_CORE_OPENTELEMETRY)
```
ahsonkhan, line 12: I am assuming other SDKs have this on by default. So, am I right to assume we don't want this to be opt-in, something like `ENABLE_AZURE_CORE_OPENTELEMETRY`?

CMakeLists.txt
```diff
@@ -55,21 +55,31 @@ if(MSVC_USE_STATIC_CRT AND MSVC)
   #
   # 5. We "replace with empty string" (i.e. remove) first, then add, so that '/MT'
   # will be present (and present once) even if '/MD' was not.
-
+  message(STATUS "Configuring Static Runtime Library.")
+  if(${CMAKE_CXX_FLAGS} MATCHES ".*/MD.*")
   string(REGEX REPLACE "/MD" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MT")
+  endif()
```
ahsonkhan, line 62: nit: add indentation for the if-blocks for readability

Also, just wondering, what motivated adding these if-guards?

sdk/core/azure-core-opentelemetry/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    include(AddGoogleTest)
+    enable_testing ()
+  endif()
+  
+  add_subdirectory(test/ut)
+endif()
+
+#if (BUILD_PERFORMANCE_TESTS)
+#  add_subdirectory(test/perf)
+#endif()
```
ahsonkhan, line 99: nit: remove commented out snippet if not relevant

sdk/core/azure-core-opentelemetry/NOTICE.txt
```diff
@@ -0,0 +1,135 @@
...
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+License notice for libcurl
```
ahsonkhan, line 67: Looking at the cgmanifest.json, if this package doesn't need/depend on libcurl, json, etc. we should remove them from the notice, and include the open telemetry one (and GoogleTest).

sdk/core/azure-core/src/context.cpp
```diff
@@ -12,7 +12,7 @@ Azure::DateTime Azure::Core::Context::GetDeadline() const
   // Contexts form a tree. Here, we walk from a node all the way back to the root in order to find
   // the earliest deadline value.
   auto result = DateTime::max();
-  for (auto ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
+  for (std::shared_ptr<ContextSharedState> ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
```
ahsonkhan, line 15: Question: What are your thoughts on such a change? Is `auto` not appropriate here.

LarryOsterman: THis change was to fix a bug reported by the code analysis tools - apparently auto cannot properly deduce the type of the destination, causing extra copies. Having said that, I think the right fix is to say `auto & ptr` or `const auto& ptr`.

sdk/core/azure-core-opentelemetry/NOTICE.txt
LarryOsterman: I thought the NOTICE.txt file was auto-generated by the cgmanifest processing code, is that not the case?

sdk/core/azure-core-opentelemetry/CMakeLists.txt
LarryOsterman: It will become relevent eventually, I believe. Not 100% sure though.

CMakeLists.txt
LarryOsterman: Adding the if guards prevents build warnings caused by running the REGEX REPLACE multiple times (the C compiler complains if the same switches are specified multiple times with different values. Without the if guards, you get /MD and /MDd defined, which triggers a compiler command line warning.

I don't like build warnings, they upset me, so I got rid of them.

sdk/core/CMakeLists.txt
LarryOsterman: This was added when OpenTelemetry had a bug that prevented it from compiling. This allowed the CI system to make forward progress even if the opentelemetry-cpp package was broken.

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#pragma once
+
+/**
+ * @def AZ_IDENTITY_DLLEXPORT
+ * @brief Applies DLL export attribute, when applicable.
+ * @note See https://docs.microsoft.com/cpp/cpp/dllexport-dllimport?view=msvc-160.
+ */
+
+#if defined(AZ_CORE_TRACING_OPENTELEMETRY_DLL) || (0 /*@AZ_IDENTITY_DLL_INSTALLED_AS_PACKAGE@*/)
```
antkmsft, line 19:
```suggestion
#if defined(AZ_CORE_TRACING_OPENTELEMETRY_DLL) || (0 /*@AZ_CORE_TRACING_OPENTELEMETRY_DLL_INSTALLED_AS_PACKAGE@*/)
```

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * @file
+ * @brief DLL export macro.
+ */
+
+// For explanation, see the comment in azure/core/dll_import_export.hpp
+
+#pragma once
+
+/**
+ * @def AZ_IDENTITY_DLLEXPORT
```
antkmsft, line 14:
```suggestion
 * @def AZ_CORE_TRACING_OPENTELEMETRY_DLLEXPORT
```

sdk/core/azure-core-opentelemetry/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    opentelemetry-cpp::sdk 
+    opentelemetry-cpp::trace)
+
+
+get_az_version("${CMAKE_CURRENT_SOURCE_DIR}/src/private/package_version.hpp")
+generate_documentation(azure-core-opentelemetry ${AZ_LIBRARY_VERSION})
+
+az_vcpkg_export(
+    azure-core-opentelemetry
+    CORE
```
antkmsft, line 75:
```suggestion
    CORE_TRACING_OPENTELEMETRY
```

sdk/core/azure-core-opentelemetry/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    "azure/core-opentelemetry/dll_import_export.hpp"
+  )
+
+az_rtti_setup(
+  azure-core-opentelemetry
+  CORE
```
antkmsft, line 81:
```suggestion
  CORE_TRACING_OPENTELEMETRY
```

sdk/core/azure-core-opentelemetry/CHANGELOG.md
LarryOsterman: For .Net, opentelemetry is built into the runtime.
For Java, it is azure-core-tracing-opentelemetry,
For JS/TS, it is @Azure/core-tracing
For Python, it is azure-core-tracing-opentelemetry

So it makes sense to rename the directory to azure-core-tracing-opentelemetry

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/rtti.hpp
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ * @brief Run-time type info enable or disable.
+ *
+ * @details When RTTI is enabled, defines a macro `AZ_CORE_RTTI`. When
```
antkmsft, line 8:
```suggestion
 * @details When RTTI is enabled, defines a macro `AZ_CORE_TRACING_OPENTELEMETRY_RTTI`. When
```

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/rtti.hpp
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @details Each library has this header file. These headers are being configured by
+ * `az_rtti_setup()` CMake macro. CMake install will patch this file during installation, depending
+ * on the build flags.
+ */
+
+#pragma once
+
+/**
+ * @def AZ_CORE_RTTI
```
antkmsft, line 19:
```suggestion
 * @def AZ_CORE_TRACING_OPENTELEMETRY_RTTI
```

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/rtti.hpp
```diff
@@ -0,0 +1,36 @@
...
+ * this issue, CMake install would patch the header it places in the installation directory, so that
+ * condition:
+ * `#if defined(AZ_RTTI) || (0)`
+ * becomes, effectively,
+ * `#if defined(AZ_RTTI) || (0 + 1)`
+ * when the library was built with RTTI support, and will make no changes to the
+ * condition when it was not.
+ */
+
+#if defined(AZ_RTTI) || (0 /*@AZ_CORE_RTTI@*/)
```
antkmsft, line 34:
```suggestion
#if defined(AZ_RTTI) || (0 /*@AZ_CORE_TRACING_OPENTELEMETRY_RTTI@*/)
```

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/rtti.hpp
```diff
@@ -0,0 +1,36 @@
...
+#define AZ_CORE_RTTI
```
antkmsft, line 35:
```suggestion
#define AZ_CORE_TRACING_OPENTELEMETRY_RTTI
```

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#undef AZ_CORE_TRACING_OPENTELEMETRY_BUILT_AS_DLL
+
+/**
+ * @brief Azure SDK abstractions.
+ *
+ */
+namespace Azure {
+/**
+ * @brief Azure Identity SDK abstractions.
```
antkmsft, line 47:
```suggestion
 * @brief Abstractions commonly used by Azure SDK client libraries.
```

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
```
antkmsft, line 5: Why the include directory is `azure/core-opentelemetry/`, and not `azure/core/opentelemetry/`?

LarryOsterman: azure-core-opentelemetry is a separate package from azure-core. We cannot put any opentelemetry related files under azure/core/azure-core.

The distributed tracing folks require that the actual tracing implementation be implemented as a separate package from azure core.

sdk/core/CMakeLists.txt
LarryOsterman: The opentelemetry-cpp vcpkg cannot compilie for UWP, so we need a mechanism to disable opentelemetry on UWP builds.

It stinks, but it's what we've got.

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
LarryOsterman: azure-core-tracing-opentelemetry is a separate package from azure-core. 

I don't know how to put folders in azure/core/internal into a package that isn't a part of azure-core.

LarryOsterman:
Hey @ahsonkhan could you look?

sdk/core/azure-core-opentelemetry/inc/azure/core-opentelemetry/dll_import_export.hpp
antkmsft: It can be a separate package and have include folder `inc/azure/core/opentelemetry`: multiple packages do have `inc/azure` folder, and they are combined into a single tree, none of them replaces other.

sdk/core/azure-core-tracing-opentelemetry/inc/azure/core-tracing-opentelemetry/opentelemetry.hpp
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
antkmsft, line 1: Is this header needed at all? I don't see any package using it. And everything it has is in internal namespace.
Maybe, we need a package without any public headers?

sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -203,7 +204,7 @@ namespace Azure { namespace Core {
      */
     template <class T> bool TryGetValue(Key const& key, T& outputValue) const
     {
-      for (auto ptr = m_contextSharedState; ptr; ptr = ptr->Parent)
```
antkmsft, line 206: Unnecessary change

sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
```diff
@@ -1,5 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
+#if !defined(INCLUDE_AZURE_CORE_EXTENDABLE_ENUMERATION_HPP)
```
antkmsft, line 3: why do we need include guard - isn't `#pragma once` enough? We use `#pragma once` everywhere, we decided to use it and not include guards.

sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
```diff
@@ -1,5 +1,9 @@
...
+#pragma once
+
+#define INCLUDE_AZURE_CORE_EXTENDABLE_ENUMERATION_HPP
 
 /**
  * @file
```
antkmsft, line 9: BTW, in the code below, code snippet is incorrect (incomplete): statics must be marked with `..._DLLEXPORT`, or they won't link when Azure SDK is built as DLL.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -0,0 +1,176 @@
...
+    explicit TracingAttributes(std::string const& that) : ExtendableEnumeration(that) {}
+
+    /**
+     * @brief Represents an "Internal" operation.
+     *
+     */
+    AZ_CORE_DLLEXPORT const static TracingAttributes AzNamespace;
+  };
+
+}}}} // namespace Azure::Core::Tracing::_internal
```
antkmsft, line 176:
```suggestion
}}}} // namespace Azure::Core::Tracing::_internal

```

antkmsft (changes requested):
I reactivated my previous comment.
I think opentelemetry should have its public headers in `inc/azure/core/tracing/opentelemetry` and not `inc/azure/core-tracing-opentelemetry`. Maybe it even doesn't need to have any public headers, because I don't see any package using it.
If it is internal-only, the include path should be `inc/azure/core/internal/tracing/opentelemetry`.

What I don't want users to see it when they type
```
#include <azure/c
```
is to see
```
[] core/
[] core-tracing-opentelemetry/
```

And when they type
```
#inlclude <azure/core/tracing/
```
to see
```
[] opentelemetry.hpp
[] tracing.hpp
```
which does not bring anything new in `Azure::Core::Telemetry` namespace (only in `Azure::COre::Telemetry::_internal` which is not intended for customer use).

vcpkg.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "azure-sdk-for-cpp",
   "version": "1.5.0",
+  "builtin-baseline": "f0aa678b7471497f1adedcc99f40e1599ad22f69",
```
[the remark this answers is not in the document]
LarryOsterman: You cannot specify a version without having a baseline commit.

And opentelemetry-cpp needs to be greater or equal to version 1.3.0 to fix a bug in opentelemetry-cpp

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -77,8 +78,7 @@ add_executable (
     transport_adapter_implementation_test.cpp
     url_test.cpp
     uuid_test.cpp
-    exception_test.cpp
-)
```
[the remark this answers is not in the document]
LarryOsterman: Not a bad idea. Not sure why the macro changed, to be honest.

sdk/core/azure-core-tracing-opentelemetry/inc/azure/core-tracing-opentelemetry/opentelemetry.hpp
LarryOsterman: No package will ever consume this (unless the package is testing its open telemetry integration).

This file defines a single public class: OpenTelemetryProvider, and end-user developers will absolutely use it. All the _internal types are in the header because C++ gets really annoyed when you try to reference a class that is never defined.

sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
LarryOsterman: [C++ Core Guidelines SF.8](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-guards) says that `#pragma` is a non-standard extension and should be avoided in favor of using the language standard.

If the Azure C++ guidelines state that we're intentionally ignoring SF.8, then we can remove it, but otherwise IMHO we should follow the best practices of the language.

sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
LarryOsterman: If I added in the DLLEXPORT goo, it would unnecessarily complicate the example. Examples should demonstrate the pattern without any complicated goo.

And the example works just fine if it's not used in a header.

sdk/core/azure-core/inc/azure/core/context.hpp
LarryOsterman: This change was made to silence "Severity	Code	Description	Project	File	Line	Suppression State
Message	lnt-accidental-copy	'auto' doesn't deduce references. A possibly unintended copy is being made"

Is the bar that changes which silence warnings are unnecessary?

LarryOsterman:
> I reactivated my previous comment. I think opentelemetry should have its public headers in `inc/azure/core/tracing/opentelemetry` and not `inc/azure/core-opentelemetry`. Maybe it even doesn't need to have any public headers, because I don't see any package using it. If it is internal-only, the include path should be `inc/azure/core/internal/tracing/opentelemetry`.
> 
> What I don't want users to see it when they type
> 
> ```
> #include <azure/c
> ```
> 
> is to see
> 
> ```
> [] core/
> [] core-opentelemetry/
> ```
> 
> And when they type
> 
> ```
> #inlclude <azure/core/tracing/
> ```
> 
> to see
> 
> ```
> [] opentelemetry.hpp
> [] tracing.hpp
> ```
> 
> of which `opentelemetry.hpp` does not bring anything new in `Azure::Core::Telemetry` namespace (only in `Azure::Core::Telemetry::_internal` which is not intended for customer use).

As a point of clarification: opentelemetry.hpp DOES bring a new public class which customers are expected to use: `Azure::Core::Tracing::OpenTelemetry::TracerProvider`. And the rest of the types in opentelemetry.hpp exist because the C++ compiler gets quite upset if you attempt to reference classes which have not been defined.

sdk/core/azure-core-tracing-opentelemetry/test/ut/service_support_test.cpp
```diff
@@ -0,0 +1,601 @@
...
+      Azure::Core::Context const& context = Azure::Core::Context{})
+  {
+    auto contextAndSpan = m_serviceTrace.CreateSpan("ApiWhichThrows", context);
+
+    try
+    {
+      auto rawResponse = SendHttpRequest(false, contextAndSpan.first);
+      return Azure::Response<std::string>("", std::move(rawResponse));
+    }
+    catch (std::exception& ex)
```
antkmsft, line 513:
```suggestion
    catch (std::exception const& ex)
```
```suggestion
    catch (std::exception& ex)
```

antkmsft, line 513:
```suggestion
    catch (std::exception const& ex)
```
```suggestion
    catch (std::exception& ex)
```

antkmsft:
If we can put `azure/core/tracing/opentelemetry/opentelemetry.hpp`, and it compiles on all platforms in CI, I think we can say it is supported. We can also check with Casey.

antkmsft:
If we can put `azure/core/tracing/opentelemetry/opentelemetry.hpp`, and it compiles on all platforms in CI, I think we can say it is supported. We can also check with Casey.

sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
antkmsft: I had the same argument, but when deciding on guidelines, we decided to not follow strictly standard for `#pragma once`.

antkmsft: I had the same argument, but when deciding on guidelines, we decided to not follow strictly standard for `#pragma once`.

antkmsft: https://azure.github.io/azure-sdk/cpp_introduction.html#cpp-style-headerguards - it does not explicitly say "do not use include guards" [but use pragma once instead], but I remember that was the intent. We can clarify during our Tuesday meeting.

antkmsft: https://azure.github.io/azure-sdk/cpp_introduction.html#cpp-style-headerguards - it does not explicitly say "do not use include guards" [but use pragma once instead], but I remember that was the intent. We can clarify during our Tuesday meeting.

LarryOsterman:
Ping. The only approval is from George, but there have been many comments from others. I'm not comfortable checking in with this level of review.

antkmsft (approved):

sdk/core/azure-core-tracing-opentelemetry/src/opentelemetry.cpp
```diff
@@ -0,0 +1,189 @@
+
```
ahsonkhan, line 1: Missing license comment on top of file.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.7.0-beta.1 (Unreleased)
 
 ### Features Added
+Added implementation for Distributed Tracing.
```
ahsonkhan, line 6:
```suggestion
- Added implementation for Distributed Tracing.
```

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -132,7 +133,7 @@ set(
     src/operation_status.cpp
     src/strings.cpp
     src/uuid.cpp
-)
+    src/tracing/tracing.cpp)
```
ahsonkhan, line 136: Nit: sort alpha ordering

ahsonkhan (commented):
Some lingering (admittedly late) feedback.

sdk/core/azure-core-tracing-opentelemetry/src/opentelemetry.cpp
LarryOsterman: W$W#$#..

This should be covered in the CI pipeline.

LarryOsterman: This is now: https://github.com/Azure/azure-sdk-for-cpp/issues/3675

sdk/core/azure-core/CMakeLists.txt
LarryOsterman: I'll make the change, but it's a pretty big change since nothing else in that list is alphabetical.

vcpkg.json
```diff
@@ -11,6 +12,11 @@
     },
     {
       "name": "openssl"
+    },
+    {
+      "name": "opentelemetry-cpp",
```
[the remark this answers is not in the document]
LarryOsterman: There are 2 vcpkg.json files for each package in the repo, and there's a vcpkg.json file in the root, for a total of 183 vcpkg.json files.

The vcpkg.json file in the root is used when building the entire repo - it specifies the union of all package dependencies.
For each package, there are two vcpkg files - one in the root of the package and a second in the vcpkg directory under the package.

The vcpkg.json file in the root of the package specifies the dependencies when building just the package.
The vcpkg\vcpkg.json file specifies the dependencies for the package when it is published to vcpkg.

So the vcpkg.json in the root specifies the dependencies when you run vcpkg install in the root.

PR #3632, opened 2022-05-10
Fix CI handling of ClientCertCred sample failure

antkmsft:
This is to prevent CI failures like this: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1555081&view=logs&j=954b0509-85cd-56e3-2fbc-6a81ffc23f75&t=5c4ca90a-fea4-57c0-8491-790045602194


RickWinter:
Is this in line with the other languages and how they handle certificate client testing?

LarryOsterman:
> @RickWinter, I can't find .NET SDK, for example, running samples in the same way as we do. This is not testing, this is running the sample in CI. Our test does run with the certificate file, so we can't say we are not testing it - we do. The reason we would want to run the sample is to, primarily, ensure that the sample code gets updated along with the SDK code. And here, even if we do ignore the return code, we do compile the sample. It is not possible to break the API and not notice that the sample no longer compiles. So I think this fix is sufficient.

The other SDKs have the ability to run samples in their nightly CI system, usually it's an option in the ci.yaml or tests.yml file.

For instance, in Python it's triggered by setting "TestSamples=true" in the tests.yml file. I might ask @weshaggard how it's done in the various languages.

eng/pipelines/templates/jobs/live.tests.yml
```diff
@@ -165,7 +165,7 @@ jobs:
           echo "**********Running sample: ${sample}"
           bash -c "$sample"
           status=$?
-          if [[ $status -eq 0 ]]; then
+          if [[ $status -eq 0 || "$sample" == *"client_certificate_credential_sample"* ]]; then
```
LarryOsterman, line 168: I'm *really* uncomfortable putting a carveout for a specific sample inside the global live.tests.yml file.

This seems like it should be configurable in the ci.yml file.

LarryOsterman (changes requested):

antkmsft:
@LarryOsterman, even if we manage to create a certificate, we do not hardcode path to a cert file in the sample. It is empty by default, so the sample, without modifications, will be returning 1.
https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/identity/azure-identity/samples/client_certificate_credential.cpp#L14
We do have `return std::string()` in all other samples as well.

LarryOsterman:
> @LarryOsterman, even if we manage to create a certificate, we do not hardcode path to a cert file in the sample. It is empty by default, so the sample, without modifications, will be returning 1. https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/identity/azure-identity/samples/client_certificate_credential.cpp#L14 We do have `return std::string()` in all other samples as well.

Oh, fascinating. FWIW, in the other languages, the samples depend on external environment variables (AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET) for Client ID and Tenant ID, it wouldn't be a stretch to include a certificate to that set of environment variables (if you want an example, the attestation SDK's samples do this).

eng/pipelines/templates/jobs/live.tests.yml
antkmsft: We can create a work item for the EngSys team to create a mechanism for having that list separated and stored somewhere else, or find some other way for the sample to communicate that it will return 1 and it is ok.
Right now, we need to release Identity, and this blocks the CI.
This functionality is tested. It is not the test, it is sample.
This is the best way to fix this at the moment. I do not want to modify the sample to return 0. I do not want to hardcode path to a certificate.

LarryOsterman: Is there a way to remove the sample from the CI pipeline (remove it from the cmakelists.txt in the overall samples directory)?

Forgive me, but you can't get the sample to work in the CI pipeline, I'd rather see the sample removed from the CI pipeline rather than special case it in the CI pipeline. I believe you can do that by removing the `create_per_service_target_build_for_sample(identity client_certificate_credential_sample)` line from the cmakelists.txt.

And since the sample failure is going to be ignored anyway, executing it in the CI pipeline doesn't really provide any value.

If, in the future you make changes to the sample to get it to run in the CI pipeline, you can add it back.

antkmsft: We for sure want to build and fail the CI if build fails. We may run it, but if the sample returns 1, it is ok.

LarryOsterman: I agree that we want to build the sample. 

But we don't want to run it if it is guaranteed to fail to work. And I would MUCH prefer to see us change the AzureBuildTargetsForCI.cmake file (in cmakemodules) to allow a parameter that prevents the file from being added to the ${service}-samples.txt file rather than see the CI pipeline modified.

RickWinter:
Could we pre-generate the cert as larry suggested and then use an environment variable to read it in?

vhvb1989 (approved):

PR #3635, opened 2022-05-10
Update samples to use EnvironmentCredential

antkmsft:


Jinming-Hu (approved):
So `get_env.h` doesn't belong to a specific package. We just include it whenever we need to use it. Is that correct?

antkmsft:
@Jinming-Hu, it is only for samples, to simplify them, so that we don't draw too much of reader's attention with disabling warnings on Windows and working around it on UWP.
Technically, it is implemented as an `OBJECT` library in CMake terms i.e. it does create an `.obj/.o` file, but not a `.lib/.so`, and that object file gets passed to a linker when linking the sample file.
When transforming samples to a real code, users can simply drop `#include <get_env.hpp>` (replace with `#include <cstdlib>`), and everything (`std::getenv()`) will mot likely still work - unless they also compile on the strictest warning level on Windows, or if they are on UWP, they may want to use `GetEnvironmentVariable()` WinAPI, take our solution, or replace with some alternative, such as reading from config file.

In SDK code, we should use `Azure::Core::_internal::Environment::GetVariable()`.
And in tests, we can use `Azure::Core::_internal::Environment::SetVariable()`.
https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/inc/azure/core/internal/environment.hpp

Jinming-Hu:
@antkmsft Thanks for the detailed explanation. I have no concern regarding this, just wanted to confirm.

samples/helpers/get-env/inc/get_env.hpp
```diff
@@ -34,18 +34,3 @@ char* getenv(const char* name);
 #endif
 
 #endif
-#include <stdexcept>
```
LarryOsterman, line 37: There is a fairly significant behavior difference between `GetEnvHelper::GetEnv` and `std::getenv` which means that one is not a plugin replacement for the other.

`GetEnvHelper::GetEnv` always returns a valid `std::string`, but `std::getenv` does not. That means that if the environment variables are not defined, the sample will crash (the `std::string` constructor crashes if the value passed in is null), but `GetEnvHelper::GetEnv` will emit a diagnostic message to the developer if the environment variable is not present.

LarryOsterman (changes requested):

LarryOsterman:
In general, I'm highly supportive of this pull request (it's REALLY good stuff), however I'm concerned about the loss of diagnostic information if the environment variables aren't present.

antkmsft:
@LarryOsterman, we used to have a helper like this, but Jeffrey asked us no to have it and use as much standard syntax as possible in *samples*. If we are relying on diagnostics, the sample code should handle null values from `std::getenv()`.

LarryOsterman:
> @LarryOsterman, we used to have a helper like this, but Jeffrey asked us no to have it and use as much standard syntax as possible in _samples_. If we are relying on diagnostics, the sample code should handle null values from `std::getenv()`.

That makes sense. But it also implies that the next iteration of this pull request will add checks around all the calls to `std::getenv()` to ensure that the sample code handles null values from `std::getenv()`.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -404,7 +400,7 @@ ignored (this possibly surprising behavior is there because retries could cause
     {
       // Create a PEM encoded X.509 certificate to add based on the POLICY_SIGNING_CERTIFICATE_0
       // certificate.
-      std::string const certToAdd(GetEnvHelper::GetEnv("POLICY_SIGNING_CERTIFICATE_0"));
+      std::string const certToAdd(std::getenv("POLICY_SIGNING_CERTIFICATE_0"));
```
LarryOsterman, line 403: Check for null value from return from std::getenv before assigning to a std::string.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -429,7 +425,7 @@ ignored (this possibly surprising behavior is there because retries could cause
 ```cpp
 // Create a PEM encoded X.509 certificate to add based on the POLICY_SIGNING_CERTIFICATE_0
 // certificate.
-std::string const certToRemove(GetEnvHelper::GetEnv("POLICY_SIGNING_CERTIFICATE_0"));
+std::string const certToRemove(std::getenv("POLICY_SIGNING_CERTIFICATE_0"));
```
LarryOsterman, line 428: Check for null value before assigning to std::string.

sdk/attestation/azure-security-attestation/samples/attestation/attest_openenclave_shared.cpp
```diff
@@ -43,7 +35,7 @@ int main()
   {
     std::cout << "In function: SampleAttestSgxEnclaveSimple" << std::endl;
     // create client
-    std::string const shortLocation(GetEnvHelper::GetEnv("LOCATION_SHORT_NAME"));
+    std::string const shortLocation(std::getenv("LOCATION_SHORT_NAME"));
```
LarryOsterman, line 38: Check for null value before assigning to std::string.

LarryOsterman (changes requested):

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -115,7 +115,7 @@ authentication, the documentation for that API will reflect that the attestation
 
 To interact with the authenticated APIs supported by the Azure Attestation service, your client must present an Azure Active Directory bearer token to the service.
 
-The simplest way of providing a bearer token is to use the  `ClientSecretCredential` authentication method by providing client secret credentials is being used in this
+The simplest way of providing a bearer token is to use the `EnvironmentCredential` authentication method by providing client secret credentials is being used in this
```
ahsonkhan, line 118: This sentence reads weirdly (almost like two separate sentences are merged into one), especially "is being used in this getting started section".

Also, do we need to update `providing client secret credentials`, or is that good as-is?

sdk/attestation/azure-security-attestation/samples/attestation/attest_openenclave_shared.cpp
```diff
@@ -6,18 +6,10 @@
...
- * - ATTESTATION_ISOLATED_URL:  Points to an Attestation Service Instance in Isolated mode.
- * - LOCATION_SHORT_NAME:  Specifies the short name of an Azure region to use for shared mode
+ * - ATTESTATION_AAD_URL: Points to an Attestation Service Instance in AAD mode.
+ * - ATTESTATION_ISOLATED_URL: Points to an Attestation Service Instance in Isolated mode.
+ * - LOCATION_SHORT_NAME: Specifies the short name of an Azure region to use for shared mode
  * operations.
- * - AZURE_TENANT_ID:     Tenant ID for the Azure account. Used for authenticated calls to the
- * attestation service.
- * - AZURE_CLIENT_ID:     The Client ID to authenticate the request. Used for authenticated calls to
- * the attestation service.
```
ahsonkhan, line 16: Should we still call these out in docs? For `EnvCred` to work, don't these still need to be set by the user?

sdk/attestation/azure-security-attestation/samples/policy/reset_sealed_policy.cpp
```diff
@@ -56,15 +56,13 @@ int main()
     clientOptions.TokenValidationOptions.TimeValidationSlack = 10s;
 
     // create client
-    auto const credential = std::make_shared<Azure::Identity::ClientSecretCredential>(
-        GetEnvHelper::GetEnv("AZURE_TENANT_ID"),
-        GetEnvHelper::GetEnv("AZURE_CLIENT_ID"),
-        GetEnvHelper::GetEnv("AZURE_CLIENT_SECRET"));
+    auto const credential = std::make_shared<Azure::Identity::EnvironmentCredential>();
```
ahsonkhan, line 59: Just to make  sure I am not misunderstanding something, `EnvironmentCredential` just makes setting/passing the required env variables to create the credential easier, correct? And it doesn't get rid of the need to set these env variables on the user's machine, all together, right?

sdk/storage/azure-storage-blobs/samples/blob_list_operation.cpp
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-#include "get_env.hpp"
+#include <get_env.hpp>
```
ahsonkhan, line 4: Should we specify a more exact path in these, given `get_env.hpp` lives in the `samples` root folder and not under `sdk/storage`?

LarryOsterman: That's the core difference between `"foo"` and `<foo>`.

`<foo>` says "Search in the `INCLUDE` environment variable and in the directories referenced by the `-I` switches. 
`"foo"` says "Search in the current directory, in the directories of any currently open include files, the `INCLUDE` environment variable and any `-I` directories".

Most customers treat `<>` paths to be the paths they get from an SDK, and the `""` paths to be stuff they included in their application.

For samples, if the header comes from a vcpkg, IMHO it should be `<>`.

PR #3640, opened 2022-05-11
Update all-inclusive headers, and CMakeFiles for including all headers

antkmsft:


vhvb1989 (approved):
<3 amazing..
Is there a tool you use to find missing headers ?  Or is this all manual talent?

Thank you.

Please wait for @gearama  and @Jinming-Hu  approval before merging.

antkmsft:
@vhvb1989, thank you! As they say, I had time and inspiration :) - this time I was in a specific stubborn mood so it even felt satisfying to review such things manually.
Hence I rearranged the order of some headers in CMakeLists, so it visually looks like the tree structure in the Solution Explorer, and it is easier to compare.

sdk/storage/azure-storage-blobs/CMakeLists.txt
```diff
@@ -52,12 +52,13 @@ set(
...
     inc/azure/storage/blobs/page_blob_client.hpp
     inc/azure/storage/blobs/rest_client.hpp
+    inc/azure/storage/blobs/rtti.hpp
     inc/azure/storage/blobs.hpp
-    src/private/package_version.hpp
 )
 
 set(
   AZURE_STORAGE_BLOBS_SOURCE
+    src/private/package_version.hpp
```
Jinming-Hu, line 61: We keep the files in alphabetical order in storage.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs.hpp
```diff
@@ -12,9 +12,13 @@
 #include "azure/storage/blobs/blob_client.hpp"
 #include "azure/storage/blobs/blob_container_client.hpp"
 #include "azure/storage/blobs/blob_lease_client.hpp"
+#include "azure/storage/blobs/blob_options.hpp"
+#include "azure/storage/blobs/blob_responses.hpp"
```
Jinming-Hu, line 16: I won't expect people directly depend on these two files. But I assume it's ok to add them here.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs.hpp
```diff
@@ -12,9 +12,13 @@
...
 #include "azure/storage/blobs/blob_sas_builder.hpp"
 #include "azure/storage/blobs/blob_service_client.hpp"
 #include "azure/storage/blobs/block_blob_client.hpp"
 #include "azure/storage/blobs/dll_import_export.hpp"
 #include "azure/storage/blobs/page_blob_client.hpp"
-#include "azure/storage/common/storage_exception.hpp"
+#include "azure/storage/blobs/rest_client.hpp"
+#include "azure/storage/blobs/rtti.hpp"
```
Jinming-Hu, line 23: I don't think we should expose `rest_client.hpp`. It may be renamed in the future. I think we should only expose it indirectly, like via `blob_options.hpp`.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs.hpp
```diff
@@ -12,9 +12,13 @@
...
+#include <azure/storage/common/storage_exception.hpp>
```
Jinming-Hu, line 24: We divide include headers into groups in storage, following [this](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) rule. 

So please move this header to after `#pragma once` and make it a separate group.

sdk/storage/azure-storage-common/CMakeLists.txt
```diff
@@ -55,14 +55,15 @@ set(
...
+    inc/azure/storage/common/rtti.hpp
     inc/azure/storage/common/storage_common.hpp
     inc/azure/storage/common/storage_credential.hpp
     inc/azure/storage/common/storage_exception.hpp
-    src/private/package_version.hpp
 )
 
 set(
   AZURE_STORAGE_COMMON_SOURCE
+    src/private/package_version.hpp
```
Jinming-Hu, line 66: also, alphabetical order

sdk/storage/azure-storage-blobs/CMakeLists.txt
Jinming-Hu: This is a header, why is it put in SOURCE ?

Jinming-Hu (commented):
It basically looks good to me, except for a few small comments. 

So I have a question, if A depends on B, B depends on C. Do we need to include C in A?

sdk/keyvault/azure-security-keyvault-certificates/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Renamed `keyvault_certificates.hpp` to `certificates.hpp`.
```
ahsonkhan, line 9: What motivated this? For keyvault keys, our simplified header is called keyvault_keys.hpp.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keyvault_keys.hpp
```diff
@@ -4,14 +4,11 @@
...
 #pragma once
 
-#include "azure/keyvault/keys/cryptography/cryptography_client.hpp"
-#include "azure/keyvault/keys/cryptography/cryptography_client_models.hpp"
-#include "azure/keyvault/keys/cryptography/cryptography_client_options.hpp"
-#include "azure/keyvault/keys/dll_import_export.hpp"
-#include "azure/keyvault/keys/key_client.hpp"
-#include "azure/keyvault/keys/key_client_models.hpp"
-#include "azure/keyvault/keys/key_client_options.hpp"
+#include "azure/keyvault/keys.hpp"
```
ahsonkhan, line 14: If we really want to do this going forward, let's come up with a naming guideline to enforce this for future SDKs. We may want to leave already shipping SDKs like Keyvault Keys, as is. It's a small wart, imo, and not worth introducing multiple "top-level" header files that would require us to update samples. We now end up with multiple, which defeats the purpose and not worth the trade-off.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs.hpp
antkmsft: It is easier to include all, rather than judge individually. The goal of the file is to include all public APIs. The headers are in the public include directory, so I think it is easier to include.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs.hpp
antkmsft: If it gets renamed, the all-inclusive header should get updated too. As I said above, the goal is to include all public headers. If customer can see it intellisense, when typing `#include <azure/storage/blobs/`, it should be included in this file.

sdk/storage/azure-storage-common/CMakeLists.txt
antkmsft: I copied the visual order of the expanded tree I see in Solution explorer, so it is easier to check if I missed something. If some tool does rearrange the order to strictly alphabetical, I am ok with it.

sdk/keyvault/azure-security-keyvault-keys/inc/azure/keyvault/keyvault_keys.hpp
antkmsft: We do update all the samples to use new. The old one is quietly there, because we can't break it. I think it is the right way.

The rule should be, if the headers are in `azure/service/...` directory, the all-inclusive header is `azure/service.hpp`

sdk/keyvault/azure-security-keyvault-certificates/CHANGELOG.md
antkmsft: If keyvault include files are in `azure/keyvault/certificates/`, then the all-inclusive file is `azure/keyvault/certificates.hpp`.

sdk/storage/azure-storage-blobs/CMakeLists.txt
antkmsft: it is in `src` folder, i.e. not something discoverable by the code that has path to include directory. I don't think either thing makes a technical difference, but right now, it was useful for visual comparison against Solution Explorer.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs.hpp
antkmsft: Done - here, and 3 other occurrences.

gearama (approved):
hope it builds :)

antkmsft:
> So I have a question, if A depends on B, B depends on C. Do we need to include C in A?

Hey @Jinming-Hu, I view these files as such: if a file is present in `azure/service/...` (including subrirectories), it should be included in `azure/service.hpp`. Without too much analysis, because it is easier (excluding the headers in `azure/service/internal`).

I know, sometimes you can get away with just including `A` (because it should include `B`), but there could be confusion, if a user wants to use `B`, but for some reason (let's say `A` no longer includes it)  itis not available when they `#include <azure/service.hpp>`, so they need to write `#include <azure/service/b.hpp>`. That would be strange.

In general, I'd view it this way: as a library owner, feel free to have less files included if you know that `A` includes `B`. But it is much easier to verify and even automate, if it would include all files in the include directory (except `azure/service/internal`).
After all, it is harmless to include it explicitly, twice.

antkmsft:
@Jinming-Hu, is it looking good for you, can I check this in?
I also had to lower code coverage a bit - it looks like some other change has sneaked past the CI, and the current coverage in Storage is now slightly lower. But I don't think it is related with anything in this PR (although even if it is, I think it rather should be fixed in a separate PR).

Jinming-Hu (approved):

PR #3650, opened 2022-05-16
Storage Blob Stg78 Features

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -402,6 +402,10 @@ namespace Azure { namespace Storage { namespace Blobs {
        * container was deleted.
        */
       Nullable<int32_t> RemainingRetentionDays;
+      /**
+       * Indicates if version level worm is enabled on this container.
+       */
+      bool HasImmutableStorageWithVersioning = false;
```
katmsft, line 408: Any change log needed for this?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -2200,6 +2207,218 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+         */
+        std::string Expression;
+        Nullable<QuerySerialization> InputSerialization;
+        Nullable<QuerySerialization> OutputSerialization;
+      };
+    } // namespace _detail
+    /**
+     * @brief Response type for #Azure::Storage::Blobs::BlobClient::Query.
+     */
+    struct QueryBlobResult final
```
katmsft, line 2388: Seems to be missing some fields:
Content-Encoding
Content-Language
Cache-Control
Content-Disposition
Access-Control-Allow-Origin
Access-Control-Expose-Headers
Vary
Access-Control-Allow-Credentials
x-ms-blob-committed-block-count
x-ms-server-encrypted

sdk/storage/azure-storage-blobs/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ### Features Added
 
+- Bumped up API version to `2020-10-02`.
+- Added new API: `BlockBlobClient::Query()`.
+- Added `ContinuationToken` and `PageSizeHint` in `GetPageRangesOptions`.
+- Added support for listing system containers.
+- Added support for listing deleted root blob with active versions.
+
```
katmsft, line 12: Is source auth with OAuth for Copy Blob API supported? I only see the change in the protocol layer.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
Jinming-Hu: Where did you get these?
In .Net, they just reuse the struct for `Download()`. But apparently that's wrong.
I've tested, storage service doesn't return all those headers as stated in swagger file/API reference.

So the situation is, we don't know which headers will be returned and which won't. I don't think customers will expect to get those headers from `Query()` API anyway.

sdk/storage/azure-storage-blobs/test/ut/blob_query_test.cpp
```diff
@@ -0,0 +1,437 @@
...
+    {
+      queryOptions.OutputTextConfiguration
+          = Blobs::BlobQueryOutputTextOptions::CreateCsvTextOptions();
+      auto queryResponse
+          = client.Query("SELECT * from BlobStorage WHERE id > 101 AND price < 100;", queryOptions);
+      auto data = queryResponse.Value.BodyStream->ReadToEnd();
+      EXPECT_EQ(
+          std::string(data.begin(), data.end()),
+          R"csv(103,apples,99
+106,lemons,69
```
katmsft, line 92: nic: indent.

Jinming-Hu: This is not indentation issue. It's called raw string literal.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
katmsft: Got it from the official doc:
https://docs.microsoft.com/en-us/rest/api/storageservices/query-blob-contents
Might be a doc issue, please double check.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -2200,6 +2207,218 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+      {
+        /**
+         * Array of BlobQueryArrowField.
+         */
+        std::vector<BlobQueryArrowField> Schema;
+      };
+      /**
+       * @brief Parquet configuration.
+       */
+      struct ParquetConfiguration final
```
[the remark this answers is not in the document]
katmsft: I mean, are there more configurable fields for `ParquetConfiguration` other than null/not null? If not, should we just expose a boolean in the option?
Is this preserved for forward compatibility?

sdk/storage/azure-storage-blobs/test/ut/blob_query_test.cpp
katmsft: line 49 to line 69 in this file is handling raw string literal well.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
katmsft: According to the doc, most of the headers listed above only return under a certain condition. I am fine with you adding it later but from a feature completeness perspective, it would be nice to have them in version 1.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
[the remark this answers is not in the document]
katmsft: > In the format field we must mention the string parquet and the ParquetTextConfiguration doesn’t have any child element in it

Found this in the design doc. I must say that changing to a boolean is much less confusing here.

sdk/storage/azure-storage-blobs/src/private/avro_parser.cpp
```diff
@@ -0,0 +1,680 @@
...
+    {
+      uint64_t r = 0;
+      int nb = 0;
+      while (true)
+      {
+        uint8_t c = (*data.BufferPtr)[data.Offset++];
+        r = r | ((static_cast<uint64_t>(c) & 0x7f) << (nb * 7));
+        if (c & 0x80)
+        {
+          ++nb;
```
katmsft, line 25: I am confused as why we are moving 7 digits per iteration, e.g. for -128, it has only 1 byte, however, it is reading to second iteration, which reads another byte from the buffer.

could you add some test cases for negative numbers?

sdk/storage/azure-storage-blobs/test/ut/blob_query_test.cpp
Jinming-Hu: The grammar is different.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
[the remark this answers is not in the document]
Jinming-Hu: Generator doesn't work that way. If it was a boolean, the generated xml would be:

```xml
<ParquetTextConfiguration>true</ParquetTextConfiguration>
```

sdk/storage/azure-storage-blobs/src/private/avro_parser.cpp
Jinming-Hu: Look at the binary encoding section at this doc https://avro.apache.org/docs/current/spec.html

Jinming-Hu: > could you add some test cases for negative numbers?

We don't test internal functions.

Integer parsing is a very basic functionality. If the whole avro parser works, I assume this function is good.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
Jinming-Hu: I decided to hold it off until we have customer request.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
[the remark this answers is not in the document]
katmsft: I'm not referring to the protocol layer code. In convenience layer options we can hide this detail.

sdk/storage/azure-storage-blobs/src/private/avro_parser.cpp
katmsft: My concern is that this is actually tested for positive numbers but not negative numbers. If you have confidence that's fine, but wouldn't hurt for some extra coverage.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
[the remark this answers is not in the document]
Jinming-Hu: [Here](https://github.com/Azure/azure-sdk-for-cpp/blob/a798e02887fc200c3f66c847f1b5c93fd03a921d/sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_options.hpp#L1079)'s convenience layer code. Does it look good to you?

katmsft (approved):
Approved with some suggestions.

PR #3654, opened 2022-05-17
Split out attestation client factory into separate class

LarryOsterman:
# Pull Request Checklist

- [X] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [X] Comments in source
- [X] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
+++ b/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
@@ -242,7 +242,8 @@ namespace Azure { namespace Security { namespace Attestation {
     RemoveIsolatedModeCertificate(
         std::string const& pemEncodedCertificateToAdd,
         AttestationSigningKey const& signerForRequest,
-        RemoveIsolatedModeCertificateOptions const& options = RemoveIsolatedModeCertificateOptions{},
+        RemoveIsolatedModeCertificateOptions const& options
+        = RemoveIsolatedModeCertificateOptions{},
         Azure::Core::Context const& context = Azure::Core::Context{}) const;
 
   private:
--- a/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
+++ b/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
@@ -253,7 +253,7 @@ namespace Azure { namespace Security { namespace Attestation {
     std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
     AttestationTokenValidationOptions m_tokenValidationOptions;
 
-    mutable std::vector<Models::AttestationSigner> m_attestationSigners;
+    std::vector<Models::AttestationSigner> m_attestationSigners;
 
     /**
      * @brief Construct a new Attestation Administration Client object.
@@ -287,7 +287,7 @@ namespace Azure { namespace Security { namespace Attestation {
      * @param context Client context for the request to the service.
      */
     void RetrieveResponseValidationCollateral(
-        Azure::Core::Context const& context = Azure::Core::Context{}) const;
+        Azure::Core::Context const& context = Azure::Core::Context{});
   };
 
   /** @brief Construct a new AttestationAdministrationClient object.
--- a/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
+++ b/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
@@ -44,6 +44,8 @@ namespace Azure { namespace Security { namespace Attestation {
    *
    */
   class AttestationAdministrationClient final {
+    friend class AttestationAdministrationClientFactory;
+
   public:
     /**
      * @brief Construct a new Attestation Administration Client object from another attestation
@@ -243,38 +245,6 @@ namespace Azure { namespace Security { namespace Attestation {
         AddIsolatedModeCertificatesOptions const& options = AddIsolatedModeCertificatesOptions{},
         Azure::Core::Context const& context = Azure::Core::Context{}) const;
 
-    /**
-     * @brief Construct a new Attestation Administration Client object.
-     *
-     * @param endpoint The URL address where the client will send the requests to.
-     * @param credential The authentication token to use.
-     * @param options The options to customize the client behavior.
-     * @return std::unique_ptr<AttestationAdministrationClient> The newly created client.
-     */
-    static AttestationAdministrationClient Create(
-        std::string const& endpoint,
-        std::shared_ptr<Core::Credentials::TokenCredential const> credential,
-        AttestationAdministrationClientOptions const& options
-        = AttestationAdministrationClientOptions(),
-        Azure::Core::Context const& context = Azure::Core::Context{});
-    /**
-     * @brief Construct a pointer to a new Attestation Administration Client object.
-     *
-     * @note It is the responsibility of the caller to manage the lifetime of the returned
-     * AttestationAdministrationClient object, typically by constructing a std::unique_ptr or
-     * std::shared_ptr from this pointer.
-     *
-     * @param endpoint The URL address where the client will send the requests to.
-     * @param credential The authentication token to use.
-     * @param options The options to customize the client behavior.
-     */
-    static std::unique_ptr<AttestationAdministrationClient> CreatePointer(
-        std::string const& endpoint,
-        std::shared_ptr<Core::Credentials::TokenCredential const> credential,
-        AttestationAdministrationClientOptions const& options
-        = AttestationAdministrationClientOptions(),
-        Azure::Core::Context const& context = Azure::Core::Context{});
-
   private:
     Azure::Core::Url m_endpoint;
     std::string m_apiVersion;
@@ -317,6 +287,41 @@ namespace Azure { namespace Security { namespace Attestation {
      */
     void RetrieveResponseValidationCollateral(
         Azure::Core::Context const& context = Azure::Core::Context{}) const;
+  };
+
+  class AttestationAdministrationClientFactory final {
+  public:
+    /**
+     * @brief Construct a new Attestation Administration Client object.
+     *
+     * @param endpoint The URL address where the client will send the requests to.
+     * @param credential The authentication token to use.
+     * @param options The options to customize the client behavior.
+     * @return std::unique_ptr<AttestationAdministrationClient> The newly created client.
+     */
+    static AttestationAdministrationClient Create(
+        std::string const& endpoint,
+        std::shared_ptr<Core::Credentials::TokenCredential const> credential,
+        AttestationAdministrationClientOptions const& options
+        = AttestationAdministrationClientOptions(),
+        Azure::Core::Context const& context = Azure::Core::Context{});
+    /**
+     * @brief Construct a pointer to a new Attestation Administration Client object.
+     *
+     * @note It is the responsibility of the caller to manage the lifetime of the returned
+     * AttestationAdministrationClient object, typically by constructing a std::unique_ptr or
+     * std::shared_ptr from this pointer.
+     *
+     * @param endpoint The URL address where the client will send the requests to.
+     * @param credential The authentication token to use.
+     * @param options The options to customize the client behavior.
+     */
+    static std::unique_ptr<AttestationAdministrationClient> CreatePointer(
+        std::string const& endpoint,
+        std::shared_ptr<Core::Credentials::TokenCredential const> credential,
+        AttestationAdministrationClientOptions const& options
+        = AttestationAdministrationClientOptions(),
+        Azure::Core::Context const& context = Azure::Core::Context{});
   };
 
 }}} // namespace Azure::Security::Attestation
--- a/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
+++ b/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
@@ -215,7 +215,7 @@ namespace Azure { namespace Security { namespace Attestation {
     AddIsolatedModeCertificate(
         std::string const& pemEncodedCertificateToAdd,
         AttestationSigningKey const& signerForRequest,
-        AddIsolatedModeCertificatesOptions const& options = AddIsolatedModeCertificatesOptions{},
+        AddIsolatedModeCertificateOptions const& options = AddIsolatedModeCertificateOptions{},
         Azure::Core::Context const& context = Azure::Core::Context{}) const;
 
     /**
@@ -242,7 +242,7 @@ namespace Azure { namespace Security { namespace Attestation {
     RemoveIsolatedModeCertificate(
         std::string const& pemEncodedCertificateToAdd,
         AttestationSigningKey const& signerForRequest,
-        AddIsolatedModeCertificatesOptions const& options = AddIsolatedModeCertificatesOptions{},
+        RemoveIsolatedModeCertificateOptions const& options = RemoveIsolatedModeCertificateOptions{},
         Azure::Core::Context const& context = Azure::Core::Context{}) const;
 
   private:
--- a/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
+++ b/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
@@ -306,24 +306,7 @@ namespace Azure { namespace Security { namespace Attestation {
      * @param options The options to customize the client behavior.
      * @return std::unique_ptr<AttestationAdministrationClient> The newly created client.
      */
-    static AttestationAdministrationClient Create(
-        std::string const& endpoint,
-        std::shared_ptr<Core::Credentials::TokenCredential const> credential,
-        AttestationAdministrationClientOptions const& options
-        = AttestationAdministrationClientOptions{},
-        Azure::Core::Context const& context = Azure::Core::Context{});
-    /**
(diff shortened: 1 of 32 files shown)


LarryOsterman:
Ping. I would appreciate it if I could get this in soon.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -319,4 +290,45 @@ namespace Azure { namespace Security { namespace Attestation {
...
+     *
+     * @note It is the responsibility of the caller to manage the lifetime of the returned
+     * AttestationAdministrationClient object, typically by constructing a std::unique_ptr or
+     * std::shared_ptr from this pointer.
+     *
+     * @param endpoint The URL address where the client will send the requests to.
+     * @param credential The authentication token to use.
+     * @param options The options to customize the client behavior.
+     */
+    static std::unique_ptr<AttestationAdministrationClient> CreatePointer(
```
antkmsft, line 326: Can we have just one factory method - for example, with the `unique_ptr` (more flexible), and maybe add another if there's a request? Unless it is very common to worry about dynamic memory fragmentation when creating and destroying hordes of clients, I think it is more confusing to discover both methods in IntelliSense, and then be thinking about benefits of each? `unique_ptr` privides RAII, the same as byvalue return, and ownership can be transferred to other smart pointers. I'd name it `Create()`, and make it return `unique_ptr`.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -44,6 +44,8 @@ namespace Azure { namespace Security { namespace Attestation {
    *
    */
   class AttestationAdministrationClient final {
```
antkmsft, line 46: BTW, this class doesn't need virtual destructor so far - it does not inherit anything, and is not inheritable. And if it becomes that, most likely we'll have to restrict copy constructor and assignment, so by-value return from the factory will stop working as well.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
LarryOsterman: Every service client in the Azure SDK allows you to create a client as a value type. If we only allowed the pointer version of `Create`, the attestation service would be unique because it did not allow you to create a local service client.

I thought that was too great a loss of functionality to remove the ability to create a local client.

And I *hate* inconsistency.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -114,6 +114,9 @@ namespace Azure { namespace Security { namespace Attestation {
    */
 
   class AttestationClient final {
```
antkmsft, line 116: This class also doesn't technically need a virtual destructor. But as above, I cannot see how it can hurt in practice, and if choosing between two, virtual is better.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -259,7 +294,7 @@ namespace Azure { namespace Security { namespace Attestation {
      * @details The TPM attestation protocol is defined
      * [here](https://docs.microsoft.com/azure/attestation/virtualization-based-security-protocol')
      */
-    std::string ValueToSend;
+    std::string Payload;
```
antkmsft, line 297: I like some naming changes here, including this one!

antkmsft (approved):
I don't see anything blocking, what I see looks good to me, I'd recommend to take all my feedback, but the final decision is up to you, all would look good either way.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -169,7 +172,7 @@ namespace Azure { namespace Security { namespace Attestation {
      */
     Response<Models::AttestationToken<Models::AttestationResult>> AttestSgxEnclave(
         std::vector<uint8_t> const& sgxQuoteToAttest,
-        AttestEnclaveOptions options = AttestEnclaveOptions(),
+        AttestSgxEnclaveOptions options = AttestSgxEnclaveOptions(),
```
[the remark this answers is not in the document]
LarryOsterman: The C++ core guidelines are unambiguous about preferring the `{}` syntax over the `()` syntax. That's because `()` is easily confused with a method call.

I'll fix these.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
antkmsft: `unique_ptr` is local, you can have it in function scope and RAII will behave the same. Most of the time, no one would think too much about comparing local value vs local smart pointer (unless it is perf-critical or you don't want to fragment dynamic memory), you just get something, and you work with it.
If other factories also have by-value returns, or even both, I'd recommend to also update them to only have one method returning smart pointer.

LarryOsterman: The pattern I see THROUGHOUT the Azure SDK for C++ is:

```c++
auto containerClient
      = BlobContainerClient::CreateFromConnectionString(GetConnectionString(), containerName);

  containerClient.CreateIfNotExists();
```

Literally every single sample I could find follows this pattern.


If Create returned a `std::unique_ptr`, that semantic would not be possible. You would ONLY be able to refer to an `AttestationClient` by reference (`std::unique_ptr` is NOT a value type - it is a container for a reference type).

Unfortunately the ship has sailed w.r.t. the pattern of Azure C++ SDK clients returning clients by value (see all the storage APIs which have a `CreateFromConnectionString` factory method. Every one of those returns a value type, none of them return pointers.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
```diff
@@ -44,6 +44,8 @@ namespace Azure { namespace Security { namespace Attestation {
...
+    friend class AttestationAdministrationClientFactory;
```
ahsonkhan, line 47: Given this pattern will set a precedence of what factory-based client creation looks like, can you help me understand why the client class needs to have the factory be a friend?

ahsonkhan: We'd want to add guidance in the guidelines on when/why to use this pattern for other SDKs and the client/factory class relationship.

LarryOsterman: In the factory pattern, the client class's constructor is private. As long as the factory is a separate class, the factory must be a friend.

And the user experience for a separate factory class is vastly better than the experience for a factory method.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -214,8 +217,45 @@ namespace Azure { namespace Security { namespace Attestation {
...
+
+  private:
+    Azure::Core::Url m_endpoint;
+    std::string m_apiVersion;
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential const> m_credentials;
+    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
+    AttestationTokenValidationOptions m_tokenValidationOptions;
+
+    mutable std::vector<Models::AttestationSigner> m_attestationSigners;
+    /** @brief Construct a new Attestation Client object
```
ahsonkhan, line 230: nit: add new line
```suggestion
    mutable std::vector<Models::AttestationSigner> m_attestationSigners;
    
    /** @brief Construct a new Attestation Client object
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -226,7 +266,7 @@ namespace Azure { namespace Security { namespace Attestation {
     static AttestationClient Create(
         std::string const& endpoint,
         std::shared_ptr<Core::Credentials::TokenCredential const> credential,
-        AttestationClientOptions options = AttestationClientOptions(),
+        AttestationClientOptions options = AttestationClientOptions{},
```
ahsonkhan, line 269: Just so I understand such a change, when do we use one vs the other?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
```diff
@@ -214,8 +217,45 @@ namespace Azure { namespace Security { namespace Attestation {
...
+        Azure::Core::Context const& context = Azure::Core::Context{}) const;
+
+  private:
+    Azure::Core::Url m_endpoint;
+    std::string m_apiVersion;
+    std::shared_ptr<Azure::Core::Credentials::TokenCredential const> m_credentials;
+    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
+    AttestationTokenValidationOptions m_tokenValidationOptions;
+
+    mutable std::vector<Models::AttestationSigner> m_attestationSigners;
```
ahsonkhan, line 229: @CaseyCarter what's our guideline on using `mutable` vs avoiding it in private C++ members? I thought we'd avoid using `const` whenever we have APIs that mutate internal state, rather than using `mutable`.

@antkmsft, @LarryOsterman, @JeffreyRichter - what are your thoughts on this as well?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -216,9 +216,44 @@ namespace Azure { namespace Security { namespace Attestation {
...
+    Azure::Nullable<std::string> DraftPolicyForAttestation{};
+
+    /** @brief Specifies the options which should be used to validate the attestation token returned
+     * by the attestation service. Overrides the value specified in the AttestationClient.
+     * @details If not provided by the caller, the token validation options
+     * specified when the @{link AttestationClient} was created will be used.
+     */
+    Azure::Nullable<AttestationTokenValidationOptions> TokenValidationOptionsOverride{};
+  };
+  /** @brief Parameters sent to the attestation service for the AttestationClient::AttestOpenEnclave
```
ahsonkhan, line 253: nit: spacing :)

```suggestion
  };
  
  /** @brief Parameters sent to the attestation service for the AttestationClient::AttestOpenEnclave
```

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client_options.hpp
```diff
@@ -216,9 +216,44 @@ namespace Azure { namespace Security { namespace Attestation {
     AttestationDataType DataType;
   };
 
-  /** @brief Parameters sent to the attestation service to be consumed in an attestation operation.
+  /** @brief Parameters sent to the attestation service for the AttestationClient::AttestSgxEnclave
+   * API.
    */
-  struct AttestEnclaveOptions final
+  struct AttestSgxEnclaveOptions final
```
ahsonkhan, line 222: Does this option rename need to be called out in CL?

ahsonkhan (approved):
Other than that, looks good.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
ahsonkhan: > And the user experience for a separate factory class is vastly better than the experience for a factory method.

I am curious to know why a separate class is a clear winner here, from a usability perspective.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
ahsonkhan: Oh, I see the previous discussion here:
https://github.com/Azure/azure-sdk-for-cpp/pull/3654#discussion_r877638614

@antkmsft, @CaseyCarter, and @LarryOsterman - for the uninitiated, can we please state in the SDK guidelines exactly when and where to use `()` vs `{}` and what (if any) subtle semantic differences exist between the two.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_administration_client.hpp
ahsonkhan: Why do we need both `Create()` and `CreatePointer()`? Can we just have `Create()` that returns the client as a value-type, or is there some strong customer need for `CreatePointer()` (maybe specific to attestation)?

Using that is easier and simpler, like the `CreateFromConnectionString()` sample snippet you should above from storage.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
CaseyCarter: A `mutable` member in a class without a comment to indicate how that member is accessed thread-safely from simultaneous calls to `const` member functions is a huge red flag.

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
LarryOsterman: You always use {}. [See ES.23](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list).

sdk/attestation/azure-security-attestation/samples/attestation/attest_sgxenclave_with_runtime_json.cpp
```diff
@@ -44,19 +44,19 @@ int main()
...
     // RunTimeData as a JSON object when it is returned in the resulting token.
-    AttestEnclaveOptions attestOptions;
+    AttestSgxEnclaveOptions attestOptions;
 
     attestOptions.RunTimeData
         = AttestationData{AttestationCollateral::RunTimeData(), AttestationDataType::Json};
 
     Azure::Response<AttestationToken<AttestationResult>> const sgxResult
-        = attestationClient.AttestSgxEnclave(sgxEnclaveQuote, attestOptions);
+        = attestationClient->AttestSgxEnclave(sgxEnclaveQuote, attestOptions);
```
ahsonkhan, line 59: I might be missing something that requires us to use this pattern, possibly specific to attestation, but we have taught users about the `client(dot)operation` experience/calling convention. As in, the client instances they use are value types and not pointers.

What's motivating us to enable and showcase this pattern to users instead for our factory create APIs?

sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp
ahsonkhan: @CaseyCarter how do we guarantee (and document) thread-safety for other non-const methods, that do end up modifying the private members?
Does `mutable` introduce an extra "burden-of-proof"?

LarryOsterman: Service client objects are supposed to be thread safe, which in practice means they need to be immutable. That in turn means that all service methods need to be marked as `const` (so they can be called from 'const` clients).

Having said that, the only method which modifies the `m_attestationSigners` field is the `RetrieveResponseValidationCollateral` and that is called from the factory on a non-const instance of the client. So `RetrieveResponseValidationCollateral` doesn't need to be marked as `const`.

sdk/attestation/azure-security-attestation/samples/attestation/attest_sgxenclave_with_runtime_json.cpp
LarryOsterman: Anton felt STRONGLY that the factory pattern should always return a pointer and never a concrete class.

And I can understand his motivation - it adds significant flexibility w.r.t. how we express factory created classes to the customer. 

It also dramatically simplifies the factory pattern (you only have a single `Create` and not a separate `CreatePointer` method).

Ultimately, I didn't feel strong enough about the ability to create concrete class instances to insist on the ability to create concrete class instances.

LarryOsterman: Btw, returning a `std::unique_ptr` from the factory is consistent with [C.50](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-factory).

ahsonkhan: I am still not sure about the usability of the clients returned from the factory pattern, in how we are defining it here. I want to understand the flexibility benefit for the customer better, to add if it's worth the change in coding pattern.

@JeffreyRichter what do you think?

Here's the alternative pattern based on what the end user writes:
https://github.com/Azure/azure-sdk-for-cpp/blob/2aae6be7a36de2135a11f8c2552eb2d4143f0536/sdk/storage/azure-storage-queues/samples/queue_getting_started.cpp#L34-L39

LarryOsterman: We *must* have the ability to create reference values in a factory function - our test framework depends on it.

@antkmsft feels strongly that factories MUST return pointers to objects and MUST NOT return concrete objects.
@ahsonkhan feels strongly that factories MUST have the ability to return concrete objects.

@ahsonkhan, I think you and @antkmsft need to get together and duke this out.  I don't have a strong preference between the two.

PR #3657, opened 2022-05-19
Remove noexcept specifier from Context::IsCancelled()

### LINKED ISSUE #3567: Azure::Core::Context::IsCancelled is marked as noexcept, but it calls Azure::Core::Context::GetDeadline which may throw

antkmsft:
Closes #3567

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Removed `noexcept` specification from `Azure::Core::Context::IsCancelled()`.
+
 ### Bugs Fixed
 
 ### Other Changes
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -233,7 +233,7 @@ namespace Azure { namespace Core {
      * @brief Checks if the context is cancelled.
      * @return `true` if this context is cancelled; otherwise, `false`.
      */
-    bool IsCancelled() const noexcept { return GetDeadline() < std::chrono::system_clock::now(); }
+    bool IsCancelled() const { return GetDeadline() < std::chrono::system_clock::now(); }
 
     /**
      * @brief Checks if the context is cancelled.


Jinming-Hu:
Is this a breaking change?

antkmsft:
@Jinming-Hu, technically it is, although it is practically very unlikely that it will throw an exception (only if `system_clock::now()` returns a year below 0001, or above 58,494. Plus, if it did in the past, user app would've terminate immediately, and if you look from that standpoint, this is even an improvement, because it gives more control to handle.

Jinming-Hu:
> @Jinming-Hu, technically it is, although it is practically very unlikely that it will throw an exception (only if `system_clock::now()` returns a year below 0001, or above 58,494. Plus, if it did in the past, user app would've terminate immediately, and if you look from that standpoint, this is even an improvement, because it gives more control to handle.

What I'm thinking is what if a customer has used this noexcept function in another noexcept function he defined, and we remove `noexcept` directive, what would happen?

LarryOsterman:
>> @Jinming-Hu, technically it is, although it is practically very unlikely that it will throw an exception (only if system_clock::now() returns a year below 0001, or above 58,494. Plus, if it did in the past, user app would've terminate immediately, and if you look from that standpoint, this is even an improvement, because it gives more control to handle.
> What I'm thinking is what if a customer has used this noexcept function in another noexcept function he defined, and we remove noexcept directive, what would happen?

It means that `std::terminate` is called in the client's `noexcept` function instead of `std::terminate` being called in `IsCancelled`.

JeffreyRichter (approved):

Jinming-Hu (approved):

PR #3662, opened 2022-05-23
Storage DataMovement: Job suspension/resumption/cancellation Error/Progress handling

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
+++ b/sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
@@ -3,22 +3,64 @@
 
 #pragma once
 
+#include <functional>
+#include <memory>
 #include <string>
+
+#include <azure/core/credentials/credentials.hpp>
+#include <azure/storage/common/storage_credential.hpp>
 
 namespace Azure { namespace Storage {
   struct StorageTransferManagerOptions final
   {
     std::string TransferStateDirectoryPath;
+    Nullable<int> NumThreads;
+    Nullable<size_t> MaxMemorySize;
+  };
+
+  struct TransferCredential final
+  {
+    std::string ConnectionString;
+    std::string SasCredential; // TODO: or Azure SAS credential?
+    std::shared_ptr<StorageSharedKeyCredential> SharedKeyCredential;
+    std::shared_ptr<Core::Credentials::TokenCredential> TokenCredential;
+  };
+
+  struct TransferProgress final
+  {
+    int64_t NumFilesTransferred = 0;
+    int64_t NumFilesSkipped = 0;
+    int64_t NumFilesFailed = 0;
+    int64_t TotalBytesTransferred = 0;
+  };
+
+  struct TransferError final
+  {
+    std::string JobId;
+    std::string SourceUrl;
+    std::string DestinationUrl;
+    // TODO: error information: std::exception_ptr or descriptive string?
+  };
+
+  struct ResumeJobOptions final
+  {
+    TransferCredential SourceCredential;
+    TransferCredential DestinationCredential;
+    std::function<void(const TransferProgress&)> ProgressHandler;
+    std::function<void(TransferError&)> ErrorHandler;
   };
 
   namespace Blobs {
-
     struct ScheduleUploadBlobOptions final
     {
+      std::function<void(const TransferProgress&)> ProgressHandler;
+      std::function<void(TransferError&)> ErrorHandler;
     };
 
     struct ScheduleDownloadBlobOptions final
     {
+      std::function<void(const TransferProgress&)> ProgressHandler;
+      std::function<void(TransferError&)> ErrorHandler;
     };
   } // namespace Blobs
 }} // namespace Azure::Storage
--- a/sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
+++ b/sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
@@ -3,8 +3,17 @@
 
 #pragma once
 
-namespace Azure { namespace Storage {
+#include <atomic>
+#include <functional>
+#include <future>
+#include <memory>
+#include <string>
+#include <type_traits>
 
+#include "azure/storage/datamovement/datamovement_options.hpp"
+#include "azure/storage/datamovement/scheduler.hpp"
+
+namespace Azure { namespace Storage {
   enum class TransferType
   {
     SingleUpload = 0,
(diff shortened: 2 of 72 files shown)


sdk/storage/azure-storage-datamovement/samples/datamovement_getting_started.cpp
```diff
@@ -39,17 +40,25 @@ int main()
...
 
+  auto job1Status = job1.WaitHandle.get();
+  if (job1Status == JobStatus::Successful)
+  {
+    std::cout << "job1 successful" << std::endl;
+  }
+  auto job2Status = job2.WaitHandle.get();
+  if (job2Status == JobStatus::Successful)
+  {
+    std::cout << "job1 successful" << std::endl;
```
Jinming-Hu, line 61:
```suggestion
    std::cout << "job2 successful" << std::endl;
```

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
```diff
@@ -3,22 +3,64 @@
...
   {
     std::string TransferStateDirectoryPath;
+    Nullable<int> NumThreads;
+    Nullable<size_t> MaxMemorySize;
   };
 
-  namespace Blobs {
+  struct TransferCredential final
+  {
+    std::string ConnectionString;
```
EmmaZhu, line 23: Minor: Is connection string still necessary?  The other three should be able to represent all credentials we support for now.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
```diff
@@ -3,22 +3,64 @@
...
+    int64_t NumFilesFailed = 0;
+    int64_t TotalBytesTransferred = 0;
+  };
 
+  struct TransferError final
+  {
+    std::string JobId;
+    std::string SourceUrl;
+    std::string DestinationUrl;
+    // TODO: error information: std::exception_ptr or descriptive string?
```
EmmaZhu, line 42: An exception instance should be better. We may also need to report to customer about details of requests.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
```diff
@@ -13,12 +22,116 @@ namespace Azure { namespace Storage {
     DirectoryDownload = 3,
   };
 
+  enum class JobStatus
+  {
+    InProgress,
+    Paused,
+    Successful,
```
EmmaZhu, line 29: Succeeded?

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
```diff
@@ -13,12 +22,116 @@ namespace Azure { namespace Storage {
...
+
   struct JobProperties final
   {
-    std::string JobId;
+    std::string Id;
     std::string SourceUrl;
     std::string DestinationUrl;
-    TransferType Type;
+    TransferType Type = static_cast<TransferType>(std::underlying_type_t<TransferType>(0));
+    std::shared_future<JobStatus> WaitHandle;
```
EmmaZhu, line 41: I'm fine with std::shared_future for waiting on completion in C++, but it's not support in C or P/Invoke. 

With JobStatus as parameter, if the job is failed, customer won't be able to get the fail message from here.  It's just feeling strange to get completion and detailed error message in two different places.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/storage_transfer_manager.hpp
```diff
@@ -16,16 +20,33 @@ namespace Azure { namespace Storage {
     explicit StorageTransferManager(const StorageTransferManagerOptions& options);
     virtual ~StorageTransferManager() = 0;
 
-    void PauseJob(const std::string& jobId);
-    void PauseAllJobs();
-    void ResumeJob(const std::string& jobId);
-    void ResumeAllJobs();
     void CancelJob(const std::string& jobId);
     void CancelAllJobs();
```
EmmaZhu, line 24: We may need to discuss about concept of cancel/pause.. 
Do we really need to provide a method to cancel all?

sdk/storage/azure-storage-datamovement/src/tasks/upload_blob_from_file_task.cpp
```diff
@@ -1,16 +1,23 @@
...
 
+#include "azure/storage/datamovement/job_properties.hpp"
 #include "azure/storage/datamovement/scheduler.hpp"
+#include "azure/storage/datamovement/utilities.hpp"
 
 namespace Azure { namespace Storage { namespace Blobs { namespace _detail {
 
   namespace {
-    constexpr uint64_t SingleUploadThreshold = 4 * 1024 * 1024;
+    constexpr uint64_t SingleUploadThreshold = 128 * 1024;
```
EmmaZhu, line 20: Why change this? I think 4MB is good.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
Jinming-Hu: Just like customers can create clients from connection string, I think they can do it here. I'm considering making factory methods here, so that they can only provide one type of authorization.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
Jinming-Hu: I'm thinking the opposite. As you may have noticed, task must not throw exceptions when it's running. Considering the fact that error handler is executed synchronously in tasks and exception will be triggered there when customer tries to observe the exception ptr, I'm worried that the whole scheduling system will be messed up if they don't properly handle the exception.

If we use exception ptr here, we may need another layer of security to ensure exception won't escape from task scope. 
Let's keep the TODO here and revisit this later.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
Jinming-Hu: > I'm fine with std::shared_future for waiting on completion in C++, but it's not support in C or P/Invoke.

I'd like to focus on C++ ATM. We couldn't use the same API interface anyway in we were going to work on C, like `std::string` isn't available there either.

> With JobStatus as parameter, if the job is failed, customer won't be able to get the fail message from here. It's just feeling strange to get completion and detailed error message in two different places.

There could be multiple error messages for one job. In the worst case, errors cannot fit into memory. We need to be able to handle this case. I'm guessing callback is the best choice.

Do you have other suggestions here?

sdk/storage/azure-storage-datamovement/src/tasks/upload_blob_from_file_task.cpp
Jinming-Hu: I don't know. We can use 4MB. This value isn't used anyway. Let's revisit when we actually start working on this.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/storage_transfer_manager.hpp
Jinming-Hu: Let's discuss this in weekly meeting. I can remove this if we agree it's not necessary.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
Jinming-Hu: fixed

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/job_properties.hpp
EmmaZhu: We can leave it here for now. It's an interface design, we'd need a  principle for this.  I agree that the callback is a good choice to report all detailed errors, may be good to have a summary error here as the transfer option is optional and for some simple transfer customer may not bother to specify a callback. Anyway, we can keep it here for now.

sdk/storage/azure-storage-datamovement/inc/azure/storage/datamovement/datamovement_options.hpp
EmmaZhu: We can catch all exception thrown during execution and report them to customer through the callback. Scheduler doesn't need to know anything about this. Anyway, let's a TODO here for now.

EmmaZhu (approved):

PR #3685, opened 2022-05-31
Identity 1.3.0 Release

antkmsft:


vhvb1989 (approved):
Awesome!

Jinming-Hu (approved):

sdk/identity/azure-identity/CHANGELOG.md
```diff
@@ -1,14 +1,11 @@
...
 
 ### Features Added
 
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+- Added `ClientCertificateCredential`, and updated `EnvironmentCredential` to support client certificate authentication.
+- Added `ChainedTokenCredential`.
```
ahsonkhan, line 8: Do we typically highlight (by duplicating) the features from previous betas, when making a GA release, by copying verbatims? Any reason we'd want to this time?

Is this so they get added to the release notes again to make it more visible to the end user? If so, maybe we should word it as a summary to convey the intent.

I see the following for the previous release:
> No changes since `1.2.0-beta.1`.

ahsonkhan: Would we want to always do something like it for other packages?
It makes it seem like these things got added (or fixed), since the last beta release.

@vhvb1989 - thoughts?

ahsonkhan (approved):

sdk/identity/azure-identity/CHANGELOG.md
antkmsft: I think it might be better for release notes, but we haven't formalized this.
Given that we run most of the new features Betas first, most of the releases will be saying "No changes since Beta".

antkmsft: Let me go with this, and let's discuss during a team meeting.

RickWinter (approved):

PR #3691, opened 2022-06-02
Additional OpenTelemetry Feedback...

LarryOsterman:
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -193,27 +193,48 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
...
         std::string const& spanName,
         Azure::Core::Tracing::_internal::CreateSpanOptions& spanOptions,
-        Azure::Core::Context const& clientContext);
+        Azure::Core::Context const& clientContext) const;
 
-    std::unique_ptr<Azure::Core::Tracing::_internal::AttributeSet> CreateAttributeSet();
+    std::unique_ptr<Azure::Core::Tracing::_internal::AttributeSet> CreateAttributeSet() const;
 
-    static std::unique_ptr<DiagnosticTracingFactory> DiagnosticFactoryFromContext(
+    static std::unique_ptr<ContextAndSpanFactory> ContextAndSpanFactoryFromContext(
```
ahsonkhan, line 237: I am glad this class is internal. That said, a method name like this makes me think the factory class name might need another look.

ahsonkhan: Hmm, so context and span is literally an alias for a pair of context and span. I see.

LarryOsterman: Could you explain further? The ContextAndSpanFactory creates ContextAndSpan objects. This method creates a ContextAndSpanFactory from a Context object (we have situations where we need to be able to retrieve the ContextAndSpanFactory for the current client from a pipeline policy and the Context object is the only thing that ties the client and the policies together).

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -193,27 +193,48 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
...
+     * @param clientContext parent context object.
+     *
+     * @returns Newly allocated context and Span object.
+     *
+     */
     ContextAndSpan CreateSpan(
         std::string const& spanName,
-        Azure::Core::Tracing::_internal::SpanKind const& spanKind,
-        Azure::Core::Context const& clientContext);
+        Azure::Core::Context const& clientContext) const;
```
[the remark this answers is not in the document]
LarryOsterman: You need to pass the context passed into the method which called CreateSpan.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -193,27 +193,48 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
...
+    /**
+     * @brief Create a span with the specified span name.
+     *
+     * @details This method is a convenience method intended for use by service clients, it creates
+     * a SpanKind::Internal span and context.
+     *
+     * @param spanName Name for the span to be created.
+     * @param clientContext parent context object.
+     *
+     * @returns Newly allocated context and Span object.
```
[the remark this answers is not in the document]
LarryOsterman: "span" is an object noun - in this case, the concept of a opentelemetry tracing span. "Span" is a proper noun which is the name of a type.

When referring to the concept, I try to use "span", when referring to the type, I try to use "Span". If I got it wrong, please let me know.

antkmsft (approved):

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
ahsonkhan: We discussed and brainstormed some ideas offline.

**Conclusion:**
- Name `DiagnosticTracingFactory`​ as: ​`TracingContextFactory`​.
- Rename the static method ​`DiagnosticFactoryFromContext`​ to ​`CreateFromContext​`.
- Change ​`std::pair`​ ​`ContextAndSpan`​ into its own more extensible/future-proof ​`TracingContext​` type.
- Rename factory methods from `​CreateSpan​` to ​`CreateTracingContext​`.
- It is fine to have an internal returning type method on ​`TracingProvider​`, it is the exception to the rule. We can't hide `CreateTracer` from the user and no C++ keyword is sufficient here (private or protected virtual).
- For later: Some doc comment improvement suggestions (note about the ​`CreateTracer​` method intended for internal use, file comment at the top of `​tracing.hpp`​, why we delete copy ctors on ​`ServiceSpan`​​, etc.).
- Leave the header structure as-is (no need to split-off tracing.hpp internal types into its own header, used by an open telemetry provider implementer).

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -29,7 +29,7 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
   private:
     std::shared_ptr<Span> m_span;
 
-    friend class DiagnosticTracingFactory;
+    friend class ContextAndSpanFactory;
```
ahsonkhan, line 32: There are some readme/docs that we'd want to update based on the renames in this PR. For example. Though that's not release blocking and can be done separately.
https://github.com/Azure/azure-sdk-for-cpp/blob/main/doc/DistributedTracing.md

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
ahsonkhan: We'd want to update this alias to no longer be called `TracingContext` to avoid confusion with the Context and Span type now named `TracingContext`:
https://github.com/Azure/azure-sdk-for-cpp/blob/48ef687a8f941a6ab755b1d4e56701edcb8e29cf/sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp#L180-L181

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
LarryOsterman: Already done.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -193,28 +191,52 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
...
 
     /** @brief A ContextAndSpan provides an updated Context object and a new span object
      * which can be used to add events and attributes to the span.
      */
-    using ContextAndSpan = std::pair<Azure::Core::Context, ServiceSpan>;
+    struct TracingContext
+    {
+      Azure::Core::Context Context;
+      ServiceSpan Span;
+    };
```
ahsonkhan, line 204: nit: We might add methods to these in the future, correct? Make it a class (and add `public:`).

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -193,28 +191,52 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
...
         Azure::Core::Tracing::_internal::CreateSpanOptions& spanOptions,
-        Azure::Core::Context const& clientContext);
+        Azure::Core::Context const& context) const;
 
-    std::unique_ptr<Azure::Core::Tracing::_internal::AttributeSet> CreateAttributeSet();
+    std::unique_ptr<Azure::Core::Tracing::_internal::AttributeSet> CreateAttributeSet() const;
 
-    static std::unique_ptr<DiagnosticTracingFactory> DiagnosticFactoryFromContext(
-        Azure::Core::Context const& context);
+    static std::unique_ptr<TracingContextFactory> FromContext(Azure::Core::Context const& context);
```
ahsonkhan, line 239: We were thinking `CreateFromContext`, I would consider adding the `Create` verb. But maybe that is a bit confusing compared to the other `CreateX` methods.
What do you think, @LarryOsterman ?

Alternatively, we could be super explicit and call it `CreateFactoryFromContext`. I think either is fine (but I lean towards `CreateFromContext`).

LarryOsterman: You're right and it's easy to fix.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
LarryOsterman: The pattern we are using is that if a type needs a constructor or methods, it becomes a class, otherwise it's a struct and yes, we can add fields to a struct in the future (we can't remove them or change their type, but we can add new ones). This is the same rules for ServiceModels.

PR #3698, opened 2022-06-02
Added ability to create instance with pointer

LarryOsterman:
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/sdk/attestation/azure-security-attestation/samples/attestation/attest_openenclave_shared.cpp
+++ b/sdk/attestation/azure-security-attestation/samples/attestation/attest_openenclave_shared.cpp
@@ -47,7 +47,7 @@ int main()
     std::string const endpoint
         = "https://shared" + shortLocation + "." + shortLocation + ".attest.azure.net";
 
-    AttestationClient const attestationClient(AttestationClient::Create(endpoint));
+    AttestationClient attestationClient(AttestationClient::Create(endpoint));
 
     std::vector<uint8_t> const sgxEnclaveQuote = AttestationCollateral::SgxQuote();
 
--- a/sdk/attestation/azure-security-attestation/samples/attestation/attest_sgxenclave_with_runtime_binary.cpp
+++ b/sdk/attestation/azure-security-attestation/samples/attestation/attest_sgxenclave_with_runtime_binary.cpp
@@ -44,7 +44,7 @@ int main()
 
     // create client
     std::string endpoint(GetEnvHelper::GetEnv("ATTESTATION_AAD_URL"));
-    AttestationClient attestationClient(AttestationClient::Create(endpoint));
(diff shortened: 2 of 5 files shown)


sdk/attestation/azure-security-attestation/samples/attestation/attest_sgxenclave_with_runtime_binary.cpp
```diff
@@ -44,7 +44,8 @@ int main()
 
     // create client
     std::string endpoint(GetEnvHelper::GetEnv("ATTESTATION_AAD_URL"));
-    AttestationClient attestationClient(AttestationClient::Create(endpoint));
+    std::shared_ptr<AttestationClient> attestationClient(
+        std::make_unique<AttestationClient>(AttestationClient::Create(endpoint)));
```
antkmsft, line 48: If I read in a sample, that shared pointer is initialized with unique pointer, I start asking unnecessary questions, why it is written this way, and if there is a reason behind this that I don't see at first glance - and all that drives my attention and time from reading the sample.

I'd either use local variable (most common and simple case), or shared_ptr with make_shared / unique_ptr with make_unique, or even auto with either make_shared or make_unique.

antkmsft (approved):

sdk/attestation/azure-security-attestation/samples/attestation/attest_openenclave_shared.cpp
```diff
@@ -47,7 +47,7 @@ int main()
     std::string const endpoint
         = "https://shared" + shortLocation + "." + shortLocation + ".attest.azure.net";
 
-    AttestationClient const attestationClient(AttestationClient::Create(endpoint));
```
antkmsft, line 50: We have const in other places, I thik we should keep it here, if it compiles.

sdk/attestation/azure-security-attestation/samples/attestation/attest_sgxenclave_with_runtime_binary.cpp
LarryOsterman: You're right, `std::make_shared` makes more sense here.

sdk/attestation/azure-security-attestation/samples/attestation/attest_openenclave_shared.cpp
LarryOsterman: For the samples, I went back to the original state of the various samples before I started tweaking the factory methods. This is the way the sample was when it was originally written.

And honestly, I expect customers will usually use the non-const form - that seems to be the norm throughout most of the Azure SDK.

PR #3701, opened 2022-06-03
Format vcpkg.json

antkmsft:


vhvb1989 (approved):

sdk/storage/azure-storage-common/vcpkg.json
```diff
@@ -20,7 +20,7 @@
     {
       "name": "openssl",
       "platform": "!windows",
-      "version>=" : "1.1.1n"
```
LarryOsterman, line 23: Note that `version>=` doesn't actually *work* when referring to non semver versions. It's a bug in vcpkg that this is allowed.

Also, we're at OpenSSL 3.0.2, I believe, not 1.1.1n.

LarryOsterman: We just require OpenSSL 1.1.1 or greater.

The vcpkg baseline that we're using has OpenSSL 3.0.2 (or 3.0.3, not sure which) as its value.

I don't think storage has a direct dependency on OpenSSL any more.

PR #3705, opened 2022-06-03
Removed version>= fields for openssl in vcpkg.json files

### LINKED ISSUE #3703: If we need OpenSSL >= 3.0.2, say so instead of 1.1.1n
"[W]e're at OpenSSL 3.0.2, I believe, not 1.1.1n."

_Originally posted by @LarryOsterman in https://github.com/Azure/azure-sdk-for-cpp/pull/3701#discussion_r888611927_

LarryOsterman:
# Remove `version>=` fields from vcpkg.json files.

Fixes #3703 

These fields are broken because:

1.  These fields are illegal in the vcpkg schema because `version>=` only works with semver version numbers and 1.1.1n is not a legal semver version number.
2. We rely on the vcpkg baseline to determine which version of openssl we use, currently 3.0.2.

## Pull Request Checklist

- [X] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [X] Comments in source
- [X] No typos
- [X] Update changelog
- [X] Not work-in-progress
- [X] Self review of PR done
- [X] Any breaking changes?

### DIFF
--- a/sdk/attestation/azure-security-attestation/vcpkg/vcpkg.json
+++ b/sdk/attestation/azure-security-attestation/vcpkg/vcpkg.json
@@ -25,8 +25,7 @@
       "host": true
     },
     {
-      "name": "openssl",
-      "version>=": "1.1.1n"
+      "name": "openssl"
     }
   ]
 }
--- a/sdk/core/azure-core/vcpkg/vcpkg.json
+++ b/sdk/core/azure-core/vcpkg/vcpkg.json
@@ -13,8 +13,7 @@
   "dependencies": [
     {
       "name": "openssl",
-      "platform": "!windows & !uwp",
-      "version>=": "1.1.1n"
+      "platform": "!windows & !uwp"
     },
     {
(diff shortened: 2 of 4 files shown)


antkmsft (approved):
Let's also get @Jinming-Hu's approval.

RickWinter (approved):

RickWinter:
Would it be best to have version >= 3.0.2 to ensure its at least of that version?

LarryOsterman:
> Would it be best to have version >= 3.0.2 to ensure its at least of that version?

A great question, not 100% sure to be honest. We define the version we take from our vcpkg baseline (eng\vcpkg-commit.txt) and we'll pick the version of whatever vcpkg has at that point. As I understand it, the version overrides are there if we specifically require functionality which might be different from the baseline (typically older than the baseline). 

If we just want to use whatever the baseline has, we should leave it blank.

Jinming-Hu (approved):

Issue #3508, opened 2022-04-06
Template client is out-dated and needs to be updated, removing or updating `ClientVersion`


ahsonkhan:
SDK authors start with the template client and might incorrectly assume we want a ClientVersion() method.

This is not what we want for SDKs, and hence the template client needs to be updated to reflect the correct pattern here.

https://github.com/Azure/azure-sdk-for-cpp/blob/487c7ce5425cbb44c68d0250c799fe0388b00f47/sdk/template/azure-template/inc/azure/template/template_client.hpp#L12

cc @LarryOsterman, @RickWinter 

It showed up in Attestation APIView because it started off as a template client:
https://github.com/Azure/azure-sdk-for-cpp/blob/af7281ef1a273404abb9c79f4de775b5b32f089b/sdk/attestation/azure-security-attestation/inc/azure/attestation/attestation_client.hpp#L163

https://apiview.dev/Assemblies/Review/5b2efd771af642728e4796813564404b#Azure::Security::Attestation::AttestationAdministrationClient

![image](https://user-images.githubusercontent.com/6527137/162083239-73c7af61-244a-46fa-bda7-32b315cfd036.png)

LarryOsterman:
Removing ClientVersion from Template as a part of updating Template to match service requirements.

PR #3706, opened 2022-06-03
Enable Distributed Tracing for Attestation SDK client.

### LINKED ISSUE #3555: Integrate Distributed Tracing API into Service Client methods - Attestation

LarryOsterman:
# Enable Distributed Tracing in Attestation and Template SDK Clients.

While working on integrating tracing into Attestation and Template, I realized that it was fragile to require specifying the Service name and package version for both the pipeline constructor and the `TracingContextFactory` constructor. To resolve the duplication, I added a new constructor to the HTTP pipeline which doesn't require the service name and package version. I copied the code in the `TelemetryPolicy` which creates the User-Agent header into the `TracingContextFactory` and added support to set the User-Agent header from the `RequestActivityPolicy` (if it was not previously set).

That required a bit of refactoring how the `TracingContextFactory` class worked and was hooked up into the Context chain.

I also added test cases to catch a regression in the `InputSanitizer` class which was introduced in a previous checkin.

## Pull Request Checklist

- [X] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [X] Comments in source
- [X] No typos
- [X] Update changelog
- [X] Not work-in-progress
- [X] Self review of PR done
- [X] Any breaking changes?

Fixes #3555 #3508


antkmsft:
I will review a bit later. Just wanted to say, you can't add it for the GA version yet, since it depends on beta core. Beta core is not available in the cental vcpkg repo.

LarryOsterman:
> I will review a bit later. Just wanted to say, you can't add it for the GA version yet, since it depends on beta core. Beta core is not available in the cental vcpkg repo.

That's fine - Attestation isn't going to GA in June.

antkmsft:
@LarryOsterman, sounds good, I just saw that you updated the release work item, I'll additionally notify Cala (to confirm there's no mistake).

sdk/core/azure-core/src/tracing/tracing.cpp
```diff
@@ -21,41 +154,61 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
   const TracingAttributes TracingAttributes::RequestId("requestId");
   const TracingAttributes TracingAttributes::HttpStatusCode("http.status_code");
 
+  std::string TracingContextFactory::BuildUserAgent(
```
antkmsft, line 157: Please create a work item to combine this logic, and the logic we have for TelemetryPolicy, so there is minimum code repetition.
We are updating core anyways, either we re-ship now, or next month, so please have this code in one place, and make both TelemetryPolicy and this place to call the same function.

sdk/template/ci.yml
```diff
@@ -27,6 +27,8 @@ stages:
     parameters:
       ServiceDirectory: template
       CtestRegex: azure-template
+      LineCoverageTarget: 99
```
antkmsft, line 30: It is bad to have template showing a bad example of dropping the default coverage. Plus, this value will be copied by other authors. Improve the coverage so that it will pass on default settings.

sdk/core/azure-core/test/ut/log_policy_test.cpp
```diff
@@ -61,7 +61,33 @@ void SendRequest(LogOptions const& logOptions, std::string const& portAndPath =
...
+  request.SetHeader("Request-Id", "Request-Id");
+  request.SetHeader("Retry-After", "Retry-After");
+  request.SetHeader("Server", "Server");
+  request.SetHeader("traceparent", "traceparent");
+  request.SetHeader("tracestate", "tracestate");
+  request.SetHeader("Transfer-Encoding", "Transfer-Encoding");
+  request.SetHeader("User-Agent", "User-Agent");
+  request.SetHeader("x-ms-client-request-id", "x-ms-client-request-id");
+  request.SetHeader("x-ms-request-id", "x-ms-request-id");
+  request.SetHeader("x-ms-return-client-request-id", "x-ms-return-client-request-id");
```
antkmsft, line 90: This clutters the test case a lot. Please have a boolean constructor parameter whether to add all headers, false by default, and only add all these headers if that parameter is true. Then add one test case that tests the default headers and nothing else. Leave the existing test cases without changing.

sdk/core/azure-core/test/ut/service_tracing_test.cpp
```diff
@@ -32,6 +32,82 @@ TEST(TracingContextFactory, ServiceTraceEnums)
...
+          "11.0.0",
+          "AzCopy / 10.0.4-Preview ",
+          "AzCopy / 10.0.4-Preview azsdk-cpp-storage-blob/11.0.0 ("},
+         {"storage-blob",
+          "11.0.0",
+          "  01234567890123456789abcde  ",
+          "01234567890123456789abcd azsdk-cpp-storage-blob/11.0.0 ("}};
+
+  constexpr auto UserAgentEnd = ')';
+  constexpr auto OSInfoMinLength = 10;
```
antkmsft, line 93: This will currently fail on UWP, where we currently put just `"UWP"` (because we haven't figured out how to get OS info yet, and left it for the future)

sdk/core/azure-core/src/tracing/tracing.cpp
```diff
@@ -1,6 +1,139 @@
...
+#if !defined(WINAPI_PARTITION_DESKTOP) \
+    || WINAPI_PARTITION_DESKTOP // See azure/core/platform.hpp for explanation.
 
+namespace Azure { namespace Core { namespace Tracing { namespace _internal {
+
+  /**
+   * @brief HkeyHolder ensures native handle resource is released.
+   *
+   */
+  class HkeyHolder final {
```
antkmsft, line 31: Please reuse the code from two places, don't duplicate.

sdk/core/azure-core/src/http/log_policy.cpp
```diff
@@ -89,7 +89,7 @@ Azure::Core::CaseInsensitiveSet const
         "traceparent",
         "tracestate",
         "Transfer-Encoding",
-        "User-Agent"
+        "User-Agent",
```
antkmsft, line 92: If we ship `beta.2` out of band (which we may if we really want all these changes for Attestation this month - let's discuss offline), then this fix needs to be mentioned in the changelog.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -178,12 +179,18 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
     static Azure::Core::Context::Key ContextSpanKey;
     static Azure::Core::Context::Key TracingFactoryContextKey;
 
+    std::string BuildUserAgent(
```
antkmsft, line 182: This, again, is not available in `1.7.0-beta.1` that already shipped.

antkmsft (changes requested):
1. This can't be shipped until the next version of core is shipped.
2. Maximize code reuse.
3. Do not drop the default coverage targets for Template.
4. Do not clutter the log policy test, make a new testcase that specifically targets the default headers.

sdk/core/azure-core/inc/azure/core/internal/http/pipeline.hpp
```diff
@@ -145,6 +145,88 @@ namespace Azure { namespace Core { namespace Http { namespace _internal {
...
+     * @remark This specialization of the HttpPipeline constructor constructs an HTTP pipeline
+     * *without* a telemetry policy. It is intended for use by service clients which have converted
+     * to use distributed tracing - the distributed tracing policy adds the User-Agent header to the
+     * request.
+     *
+     * @param clientOptions The SDK client options.
+     * @param perRetryPolicies The service-specific per retry policies.
+     * @param perCallPolicies The service-specific per call policies.
+     */
+    explicit HttpPipeline(
```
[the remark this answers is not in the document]
LarryOsterman: This was not going to be in Attestation for June, I'll not be completing the pull request until after tomorrows Beta-3 release.

sdk/core/azure-core/src/tracing/tracing.cpp
LarryOsterman: Fair, but functionally TelemetryPolicy is dead code waiting to be removed.

antkmsft: Until we ship core 2.0, it will have to be there, we are forever backwards compatible within major versions.

sdk/template/ci.yml
LarryOsterman: This change increases the line coverage (the previous bar was 95%) and reduces the branch coverage.

Increasing the branch coverage will require that I add back the functionality that we have all agreed is an anti-pattern.

I'd rather reduce the branch coverage in the ci.yml than add code that we KNOW is incorrect.

sdk/core/azure-core/test/ut/service_tracing_test.cpp
LarryOsterman: This is the existing test collateral, just framed in a slightly different way. I rewrote it to be more data driven, but it is exactly the same test collateral we already have in the TelemetryPolicy tests. I'm not comfortable with changing existing test collateral.

sdk/template/ci.yml
antkmsft: We don't want overrides for template in either direction, because the code will be cloned for new projects.
There must be a way to boost branch coverage so that it passes, without adding the code that is incorrect.

antkmsft: I personally think it will be ok if you put `// LCOV_EXCL_LINE` comment after the yellow lines. I don't really see how you can cover them, and library authors will have the same issue. But maybe there's still a way. This is just a shortcut, but I think it is still much better for Template than dropping the coverage.
image

antkmsft: Other option is if you make some coverable branches and cover them, so percent-wise the branch coverage will be ok.

antkmsft: Just an idea, if you make the method to calculate how many days in a month in a given year and month, with leap year handling, this may give the code enough branches to cover so that we can reach 70% branch coverage easily.

sdk/core/azure-core/test/ut/log_policy_test.cpp
```diff
@@ -216,6 +252,74 @@ TEST(LogPolicy, Headers)
...
+  EXPECT_TRUE(StartsWith(entry2.Message, "HTTP Response ("));
+  EXPECT_TRUE(EndsWith(entry2.Message, "ms) : 200 OKAY"));
+
+  // Ensure that the entire list of allowed headers is in the list of headers.
+  // This ensures that if a new header is added to the default allow list, we have a test case
+  // covering it.
+  for (auto const& allowedHeader :
+       Azure::Core::Http::Policies::_detail::g_defaultAllowedHttpHeaders)
+  {
+    // NOTE: If this fails, it means that we need to update the SendRequest function
```
antkmsft, line 317: 👍

sdk/core/azure-core/CMakeLists.txt
```diff
@@ -138,7 +138,7 @@ set(
     src/strings.cpp
     src/tracing/tracing.cpp
     src/uuid.cpp
-)
+ "src/http/user_agent.cpp")
```
antkmsft, line 141: Also, add the `user_agent.hpp` file

sdk/core/azure-core/inc/azure/core/internal/http/user_agent.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Azure { namespace Core { namespace Http { namespace _internal {
+  class UserAgentGenerator {
+  public:
+    static std::string GenerateUserAgent(
+        std::string const& componentName,
+        std::string const& componentVersion,
+        std::string const& applicationId);
+  };
+}}}} // namespace Azure::Core::Http::_internal
```
antkmsft, line 22:
```suggestion
}}}} // namespace Azure::Core::Http::_internal

```

sdk/core/azure-core/src/http/user_agent.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+      telemetryId << TrimString(applicationId).substr(0, 24) << " ";
+    }
+
+    static std::string const osVer = GetOSVersion();
+    telemetryId << "azsdk-cpp-" << componentName << "/" << componentVersion << " (" << osVer << ")";
+
+    return telemetryId.str();
+  }
+}}}} // namespace Azure::Core::Http::_internal
```
antkmsft, line 166:
```suggestion
}}}} // namespace Azure::Core::Http::_internal

```

sdk/template/ci.yml
LarryOsterman: Adding anything that involves calculating dates requires adding ICU support to the tracing library, and I honestly don't think we want that level of hassle (calculating dates in the future or past is one of the absolutely hardest things to get right in computer science). In general, I stay as far away as humanly possible from anything involving dates.

I'd rather implement case insensitive comparisons than deal with anything involving dates (case insensitive comparisons are only slightly less hard than calculating dates).

And for context: I spent 6 weeks adding support for the Reiwa Era in Japan to COM and even with all the testing I did (and it was a lot) we had to ship an emergency hotfix to fix a bug. Dates and times are evil constructs to be avoided always.

LarryOsterman: Btw, the coverage issues weren't due to the addition of the new branches, I believe that they're primarily because of removing the GetClientVersion() API.

antkmsft: I think we removed removed the `GetClientVersion()`, total coverage has dropped, plus we added some new lines, and it changed the picture.

As of date calculation, we do have one in our `DateTime.cpp`. I don't think it has to be prefect, just to have enough of if-elses. I.e. if month != 2, then it is 30 or 31 (another if-else). Else if year is divisible by 4, but not 100, or by 100 and also 400, it is 29, else 28.
But this is just an idea, I do not insist on `DaysInMonthYear()` solution, can be something else. Can be totally random code as well. Or Bse64, or URL encode. Something with ifs.

antkmsft: Currently, CI says: `"The code coverage value (58.3333%, 7 branches) is lower than the minimum value (70%)!"`
I.e. the total number of branches is 12, and 7 are covered. If you add 5 branches and cover them, then the coverage will be 12/17 branches, which will result in 70% coverage.

You can add 5 more ifs to `TemplateClient::GetValue()` (for example, this should add 6 extra branches
```cpp
// Blackjack basic strategy vs dealer 10, 6+ decks, H17.
if (key <= 0) { return 0; } // we were not dealt a hand
else if (key > 21) { return -100; } // we busted
else if (key == 21) { return 150; } // celebrate
else if (key == 11) { return 20; } // double down
else if (key < 11) { return 10; } // hit
else if (key > 11 && key < 17) { return 1; } // hit, but be less happy about it
else { return 0; } // >= 17 we always stay
})
```
)

sdk/core/azure-core/src/http/user_agent.cpp
```diff
@@ -0,0 +1,166 @@
...
+  auto const isSpace = [](int c) { return !std::isspace(c); };
+
+  s.erase(s.begin(), std::find_if(s.begin(), s.end(), isSpace));
+  s.erase(std::find_if(s.rbegin(), s.rend(), isSpace).base(), s.end());
+
+  return s;
+}
+} // namespace
+
+namespace Azure { namespace Core { namespace Http { namespace _internal {
```
antkmsft, line 146: This can actually live in namespace `_detail`, since nothing is using it from outside of Core.

sdk/core/azure-core/src/http/user_agent.cpp
```diff
@@ -0,0 +1,166 @@
...
+#if !defined(NOMINMAX)
+#define NOMINMAX
+#endif
+
+#include <windows.h>
+
+#if !defined(WINAPI_PARTITION_DESKTOP) \
+    || WINAPI_PARTITION_DESKTOP // See azure/core/platform.hpp for explanation.
+
+namespace Azure { namespace Core { namespace Http { namespace _internal {
```
antkmsft, line 33: this can live in an unnamed namespace

antkmsft:
@LarryOsterman, should there be not enough extra branches, I have a PR with 100 branches: https://github.com/Azure/azure-sdk-for-cpp/pull/3712

sdk/core/azure-core/inc/azure/core/internal/http/pipeline.hpp
antkmsft, line 163: Can you write it in a way that that delegates both constructors to one function, possibly a private constructor, or an init function that initializes m_policies, so that there is minimum code repetition.

antkmsft (approved):

PR #3711, opened 2022-06-06
ApiView command line generation script

### LINKED ISSUE #3666: Integrate ApiView with CI
Related: https://github.com/Azure/azure-sdk-tools/issues/3369

antkmsft:
This is a script that will be used for ApiView integration (https://github.com/Azure/azure-sdk-tools/issues/3369)
Closes #3666

You give it the path to azure-sdk-for-cpp repo, and a library name, such as azure-core, azure-identity, etc.

It generates the command line with all the header files for the given library - which is good because even if people forget to add some directory to the all-include file, the entities from it will still be present.

And it lists all the include directories for all libraries, which makes it a bit longer to run, but it is harmless - the stuff that was referenced stil ends up not being included in review. As example, I generated APIview for azure-identity which only depends on core. But I fed it all the include directories for all libraries, including, for example, Storage. But in the resulting AST tree, and in ApiView, no entity from Storage was ever listed. Which is great, because we don't have to implement the logic that figures out the dependencies, we simply list all the libraries and let clang++ to take care of including only the relevant ones.

### DIFF
--- a/eng/scripts/Get-ApiViewCommandLine.ps1
+++ b/eng/scripts/Get-ApiViewCommandLine.ps1
@@ -1,0 +1,53 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Usage: Get-ApiViewCommandLine.ps1 .\azure-sdk-for-cpp azure-identity
+# Or: ..\Get-ApiViewCommandLine.ps1 . azure-core
+# Or: Get-ApiViewCommandLine.ps1 ..\.. azure-security-attestation
+# Or: c:\src\azure-sdk-for-cpp\eng\scripts\Get-ApiViewCommandLine.ps1 c:\src\azure-sdk-for-cpp azure-identity
+
+param($RepoPath, $LibName)
+
+[String]$SdkRoot = Resolve-Path ($RepoPath + "\sdk")
+
+[String[]]$AllIncDirs = Get-ChildItem -Directory -Filter "inc" -Recurse $SdkRoot | Select-Object -ExpandProperty FullName
+
+[String[]]$AllIncDirsWithoutInc = $AllIncDirs | Select-Object @{ Label="Substr"; Expression = { $_.Substring(0, $_.Length - "inc".Length) } } | Select-Object -ExpandProperty Substr
+
+[String[]]$AllLibIncDirs = @()
+for($i = 0; $i -lt $AllIncDirsWithoutInc.Length; $i++) {
+    $isLibDir = $true
+    $libDir = $AllIncDirsWithoutInc[$i]
+    for($j = 0; $j -lt $AllIncDirsWithoutInc.Length; $j++) {
+        if ($i -eq $j) {
+            continue
+        }
+
+        $StartsWith = $AllIncDirsWithoutInc[$j] + "*"
+        if ($libDir -Like $StartsWith) {
+            $isLibDir = $false
+            break
+        }
+    }
+
+    if ($isLibDir){
+        $AllLibIncDirs += $libDir + "inc"
+    }
+}
+
+[String]$LibIncDir = $AllLibIncDirs | Where-Object {$_ -Match ("\\" + $LibName + "\\inc") } | Select-Object -First 1
+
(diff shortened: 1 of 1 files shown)


eng/scripts/Get-ApiViewCommandLine.ps1
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Usage: Get-ApiViewCommandLine.ps1 .\azure-sdk-for-cpp azure-identity
+# Or: ..\Get-ApiViewCommandLine.ps1 . azure-core
+# Or: Get-ApiViewCommandLine.ps1 ..\.. azure-security-attestation
+# Or: c:\src\azure-sdk-for-cpp\eng\scripts\Get-ApiViewCommandLine.ps1 c:\src\azure-sdk-for-cpp azure-identity
+
+param($RepoPath, $LibName)
```
benbp, line 9:
```suggestion
param([String]$RepoPath, [String]$LibName)
```

benbp (approved):

eng/scripts/Get-ApiViewCommandLine.ps1
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Usage: Get-ApiViewCommandLine.ps1 .\azure-sdk-for-cpp azure-identity
+# Or: ..\Get-ApiViewCommandLine.ps1 . azure-core
+# Or: Get-ApiViewCommandLine.ps1 ..\.. azure-security-attestation
+# Or: c:\src\azure-sdk-for-cpp\eng\scripts\Get-ApiViewCommandLine.ps1 c:\src\azure-sdk-for-cpp azure-identity
+
+param([String]$RepoPath, [String]$LibName)
```
weshaggard, line 9: Most of the other languages tie into their common discovery mechanism (i.e. https://github.com/Azure/azure-sdk-for-cpp/blob/main/eng/scripts/Language-Settings.ps1#L15) should we use that here and perhaps add extra information like the header files as another property?

eng/scripts/Get-ApiViewCommandLine.ps1
```diff
@@ -0,0 +1,53 @@
...
+    $CmdLine += " " + $header
+}
+
+$CmdLine += " -Xclang -ast-dump"
+
+foreach ($incDir in $AllLibIncDirs) {
+    $CmdLine += " -I " + $incDir
+}
+
+$CmdLine
```
weshaggard, line 53: Do we want this to just return a command line? Why not have it execute the command line?

antkmsft: This is more of a building block, of somewhat temporary shape, for us to sync with Praveen, and integrate ApiView with our repo one way or another. I do not know the exact output type that is expected, but it can be, and most probably will be changed the way you suggest. Maybe even to an extent that it would execute command line, write AST tree to a file, zip it, and rename the archive, i.e. will do all the ApiView's instructions for C++. The rest should be relatively easy, it is the right set of files and directories that is C++ specific and for which Praveen needed someone from the C++ team to figure out.
Currently, it prints the result, which is the easiest way to find out the ouput and, make sure it works, and debug if necessary.

eng/scripts/Get-ApiViewCommandLine.ps1
antkmsft: Hmm, this is interesting, especially from the point of having all the project information in one structure, but it would serve different purposes, and will be filled in differently. But the result would be easy to use.
But I'd say, let's get it to work in the current straightforward simple form, given my limited PowerShell knowledge (and slow PS development progress as a result), and given that Praveen will only have 1-2 days from now to integrate, until the next window comes up in about a month from now.
I am not rejecting the idea, but with all these factors, debugging complexity when making it all at once, I think the best way to go is to get it working, and then refactoring.

weshaggard: Sounds good. I'm just trying to get most of common project properties going through the same place instead of having the logic spread out in various places.

RickWinter (approved):

PR #3736, opened 2022-06-15
InputSanitizer: rename to HttpSanitizer, remove static member

antkmsft:
* `Azure::Core::_internal::InputSanitizer` => `Azure::Core::Http::_internal::InputSanitizer`.
* `azure/core/internal/input_sanitizer.hpp` => `azure/core/internal/http/http_sanitizer.hpp`.
* In .cpp: `using Azure::Core::_internal::InputSanitizer` instead of `namespace Azure { namespace Core {`
* `m_RedactedPlaceholder` => `namespace { std::string const RedactedPlaceholder = ... }`


sdk/core/azure-core/inc/azure/core/internal/http/http_sanitizer.hpp
```diff
@@ -18,12 +18,9 @@ namespace Azure { namespace Core { namespace _internal {
      */
     std::set<std::string> m_allowedHttpQueryParameters;
 
-    // Manifest constant indicating a field was redacted.
-    static const char* m_RedactedPlaceholder;
```
LarryOsterman, line 22: Nit: Why change from a static class member to a global variable in an anonymous namespace? As a static class member it is clear that this is associated with the class implementation.

LarryOsterman (approved):

sdk/core/azure-core/inc/azure/core/internal/http/http_sanitizer.hpp
antkmsft: Variables and functions from anonymous namespace are not visible outside of the .cpp file where it is visible, they are not even exposed via linker, so it is greater degree of encapsulation.
Since we mostly have one class per .cpp file, when we see unnamed namespace, they can be read as being associated with class implementation.

I'd have private static class variables only if there are inline functions in the header using them, then you can't do unnamed namespace. BTW, then such variables would need to be marked with `DLLEXPORT` attribute.

sdk/core/azure-core/src/http/http_sanitizer.cpp
```diff
@@ -0,0 +1,85 @@
+
```
ahsonkhan, line 1: This file is missing the license comment

ahsonkhan: Fixed in https://github.com/Azure/azure-sdk-for-cpp/pull/3739

PR #3762, opened 2022-06-23
Removed internal types from OpenTelemetry API surface; removed default parameters from OpenTelemetry API surface.

LarryOsterman:
# Pull Request Checklist

- [X] Doxygen docs
- [X] Unit tests
  - [X] PR is single purpose
  - [X] Related issue listed
- [X] Comments in source
- [X] No typos
- [X] Update changelog
- [X] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


LarryOsterman:
Ping.

gearama (approved):

ahsonkhan (commented):
Partial review, will continue shortly.

Are there any readme/docs we'd want to update to highlight the coding pattern change?

LarryOsterman:
> Partial review, will continue shortly.
> 
> Are there any readme/docs we'd want to update to highlight the coding pattern change?

To my knowledge, there is no coding pattern change. All I've done is hidden a ton of implementation details.

None of the public code changed, neither did the service implementation.

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -6,8 +6,20 @@
...
+```cpp
+
+auto otProvider = Azure::Core::Tracing::OpenTelemetry::OpenTelemetryProvider::Create();
+
+```
+
 ### Bugs Fixed
 
+- Removed `_internal` APIs from the public API surface. Also removed most of the
+`_internal` APIs from the public `opentelemetry.hpp` and `tracing.hpp` headers.
```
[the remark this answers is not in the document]
LarryOsterman: It's an open source project, we publish the entire repo, so of course end users will still see them. But the public headers for the telemetry stuff should NOT include any of the internal headers (and even if they did, it's not a huge deal: See also `Azure::Core::_internal::ClientOptions`).

The only `_internal` type which should be exposed to the customer is the `Azure::Core::_internal::TracerProvider`, and that is never exposed to the customer, instead it's inherited by `Azure::Core::TracerProvider` via private inheritance. 

The other `_internal` type referenced is an incomplete type and the definition of that type is not in the public header.

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -6,8 +6,20 @@
 
 ### Breaking Changes
 
+- The `Azure::Core::Tracing::OpenTelemetry::OpenTelemetryProvider` type can only
+be instantiated via a factory method: `OpenTelemetryProvider::Create`:
```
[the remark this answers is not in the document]
LarryOsterman: I'm not a huge fan of not being able to read all the text in my source code.  I tend to insert line breaks to improve the readability of source code because text scrolling off the right margin tends to disappear.

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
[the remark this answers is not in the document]
LarryOsterman: Could you point out where the public headers include the new impl header?

I tried very hard to ensure that the public headers don't reference the new impl header, if I screwed up, I'd love to be able to fix it.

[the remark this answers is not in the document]
LarryOsterman: Ah - found at least one reference. Removing it.

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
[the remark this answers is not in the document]
LarryOsterman: Based on my experience, VS code and VS don't do a particularly good job of line wrapping markdown lines.

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
```diff
@@ -3,309 +3,47 @@
...
-      virtual std::shared_ptr<Span> CreateSpan(
-          std::string const& spanName,
-          CreateSpanOptions const& options = {}) const = 0;
-
-      virtual std::unique_ptr<AttributeSet> CreateAttributeSet() const = 0;
-    };
   } // namespace _internal
 
   /**
    * @brief Trace Provider - factory for creating Tracer objects.
```
ahsonkhan, line 42: Is this still a factory for creating tracer objects? It's a method-less class, updated doc?

Also, the customer will likely find it confusing to see a class with no methods. We should add a note in the doc to explain the intent, to help usability (and address the surprise).

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -0,0 +1,291 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ * @brief Handling log messages from Azure SDK.
```
ahsonkhan, line 6: The doc comment needs fixing.

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
```diff
@@ -3,309 +3,47 @@
...
-     */
-    class AttributeSet {
-    public:
-      /**
-       * @brief Adds a Boolean attribute to the attribute set.
-       *
-       * @param attributeName Name of attribute to add.
-       * @param value Value of attribute.
-       */
-      virtual void AddAttribute(std::string const& attributeName, bool value) = 0;
```
ahsonkhan, line 67: I am going to assume all this is copy-pasted verbatim into `tracing_impl.hpp`, or at least isn't materially different, to warrant another review.

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -0,0 +1,291 @@
...
+#include "azure/core/url.hpp"
+#include <array>
+#include <memory>
+#include <string>
+#include <vector>
+
+// Forward declare Azure::Core::Http::Request to resolve an include file ordering problem.
+namespace Azure { namespace Core { namespace Http {
+  class Request;
+}}} // namespace Azure::Core::Http
```
ahsonkhan, line 24: Unrelated to this PR:
This sort of approach adds a bit of fragility in how our headers depend on each other, and might indicate a layering issue in the API.

We might want to re-visit why this was needed, and if there's an alternative approach:
> virtual void PropagateToHttpHeaders(Azure::Core::Http::Request& request) override;

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -0,0 +1,291 @@
...
+
+#include "azure/core/datetime.hpp"
+#include "azure/core/internal/extendable_enumeration.hpp"
+#include "azure/core/nullable.hpp"
+#include "azure/core/tracing/tracing.hpp"
+#include "azure/core/url.hpp"
+#include <array>
+#include <memory>
+#include <string>
+#include <vector>
```
ahsonkhan, line 19: Are all these include headers actually needed? It seems like at least `<vector>` isn't.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -4,7 +4,8 @@
 #include "azure/core/context.hpp"
 #include "azure/core/internal/client_options.hpp"
 #include "azure/core/internal/http/user_agent.hpp"
-#include "azure/core/tracing/tracing.hpp"
+#include "azure/core/internal/tracing/tracing_impl.hpp"
+#include <utility>
```
ahsonkhan, line 8: Just to learn, what's requiring this header to be included?

sdk/core/azure-core-tracing-opentelemetry/src/opentelemetry.cpp
```diff
@@ -2,9 +2,10 @@
 // SPDX-License-Identifier: MIT
 
 #include "azure/core/tracing/opentelemetry/opentelemetry.hpp"
+#include "opentelemetry_impl.hpp"
```
ahsonkhan, line 5: nit: We either add a `_private` suffix to such header file names within the /src/ directory, or put them in a `/private` directory.

For example:
https://github.com/Azure/azure-sdk-for-cpp/blob/f391e45bc144ce23488949aff9475f0ebfd199ba/sdk/core/azure-core/src/http/curl/curl_session_private.hpp
OR
https://github.com/Azure/azure-sdk-for-cpp/blob/f391e45bc144ce23488949aff9475f0ebfd199ba/sdk/core/azure-core/src/private/environment_log_level_listener.hpp

sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -6,10 +6,14 @@
 
 ### Breaking Changes
 
+- The `Azure::Core::Tracing::OpenTelemetry::OpenTelemetryProvider` type can only be instantiated via a factory method: `OpenTelemetryProvider::Create`.
```
antkmsft, line 9: If we're releasing this as GA, let's repeat the Beta 1 message for this release changelog instead of listing breaking changes.

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
ahsonkhan: **If** we decide to reword the doc comment, you'd want to update this markdown too:
https://github.com/Azure/azure-sdk-for-cpp/blob/f7a7d3e045290724a995f8c07d5773b99138bc50/doc/DistributedTracing.md#tracing-overview

> TracerProvider - this is a factory which creates Tracer objects.

ahsonkhan (commented):
Looks good overall. I had some questions and suggested a few minor fixes.

There's a potential design question on header layering, but that's unrelated to the intent of this PR.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 
 ### Other Changes
 
+- Removed `_internal` APIs from the public API surface. Also removed most of the `_internal` APIs from the public `tracing.hpp` header.
+
```
antkmsft, line 14: If we're releasing this as GA, let's repeat the Beta 1 message for this release changelog instead of listing breaking changes.
Stable releases do not normally have breaking changes, and this I think will create an unnecessary alert for to customer, especially that telemetry is mostly internal, few are using it, and most were using the GA anyway.
```suggestion
- Added prototypes and initial service support for Distributed Tracing.

```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
...
+
```
antkmsft, line 15: Please have only one blank line between version number and the last sentence here.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
```diff
@@ -4,7 +4,8 @@
 #include "azure/core/context.hpp"
 #include "azure/core/internal/client_options.hpp"
 #include "azure/core/internal/http/user_agent.hpp"
-#include "azure/core/tracing/tracing.hpp"
+#include "azure/core/internal/tracing/tracing_impl.hpp"
```
antkmsft, line 7:
```suggestion
#include "azure/core/internal/tracing/tracing_impl.hpp"

```

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -0,0 +1,291 @@
...
+
+  /**
+   * @brief Span - represents a span in tracing.
+   */
+  class Span {
+  public:
+    /**
+     * @brief Signals that the span has now ended.
+     */
+    virtual void End(Azure::Nullable<Azure::DateTime> endTime = {}) = 0;
```
antkmsft, line 182:
```suggestion
    virtual void End(Azure::Nullable<Azure::DateTime> endTime) = 0;
```

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
```diff
@@ -3,309 +3,47 @@
...
-     *
-     * @param name Name of the tracer object, typically the name of the Service client
-     * (Azure.Storage.Blobs, for example)
-     * @param version Version of the service client.
-     * @return std::shared_ptr<Azure::Core::Tracing::Tracer>
-     */
-    virtual std::shared_ptr<Azure::Core::Tracing::_internal::Tracer> CreateTracer(
-        std::string const& name,
-        std::string const& version = "") const = 0;
+  class TracerProvider : _internal::TracerProviderImpl {
```
antkmsft, line 44:
```suggestion
  class TracerProvider : private _internal::TracerProviderImpl {
```

doc/DistributedTracing.md
```diff
@@ -85,7 +85,7 @@ functions as an abstract class integration between OpenTelemetry and Azure Core:
 
 ```c++
 std::shared_ptr<Azure::Core::Tracing::TracerProvider> traceProvider
-      = std::make_shared<Azure::Core::Tracing::OpenTelemetry::OpenTelemetryProvider>(CreateOpenTelemetryProvider());
+      = Azure::Core::Tracing::OpenTelemetry::OpenTelemetryProvider::Create(CreateOpenTelemetryProvider());
```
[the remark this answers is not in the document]
LarryOsterman: @#$@#$ I thought I caught all of those. Fixing.

sdk/core/azure-core/src/tracing/tracing.cpp
```diff
@@ -30,6 +30,13 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
 
   using Azure::Core::Context;
 
+  std::shared_ptr<TracerProviderImpl> TracerImplFromTracer(
```
antkmsft, line 33: I don't remember why, but we don't have free-standing functions in the SDK, we put them all in "static classes".
I would ask to change that, but so far this is the current state of things, if I am not mistaking.

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     * @brief destroys an AttributeSet - virtual destructor to enable base class users to
+     * destroy derived classes.
+     */
+    virtual ~AttributeSet() = default;
+  };
+
+  /** @brief The Type of Span.
+   */
+  class SpanKind final : public Azure::Core::_internal::ExtendableEnumeration<SpanKind> {
```
antkmsft, line 119: DOes this really needs to be an extensible enum instead of the regular one? We use the extensible ones for the stuff that gets relayed over network, gets received from REST API calls etc.

antkmsft: The same question for other extensible enums.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
antkmsft: (new line before stdlib include)

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_otel_test.cpp
```diff
@@ -542,15 +558,15 @@ TEST_F(OpenTelemetryTests, SetStatus)
...
 
-    auto tracer = traceProvider->CreateTracer("TracerName");
-    auto span = tracer->CreateSpan("StatusSpan");
+    auto tracer = Azure::Core::Tracing::_internal::TracerImplFromTracer(traceProvider)
+                      ->CreateTracer("TracerName", "");
+    auto span = tracer->CreateSpan("StatusSpan", {});
     EXPECT_TRUE(span);
 
-    span->SetStatus(Azure::Core::Tracing::_internal::SpanStatus::Unset);
+    span->SetStatus(Azure::Core::Tracing::_internal::SpanStatus::Unset, "");
```
antkmsft, line 569: Super nit, but I think `{}` is better than `""` (there is more than one occurrence), because it does not require to statically allocate `[ ``\0`` ]` and to call `std::string(const char*)`.

antkmsft: Ah, sorry, nevermind, I did not notice this was in test code.

sdk/core/azure-core/src/tracing/tracing.cpp
```diff
@@ -30,6 +30,13 @@ namespace Azure { namespace Core { namespace Tracing { namespace _internal {
...
+      std::shared_ptr<TracerProvider> const& provider)
+  {
+    const auto pointer = static_cast<TracerProvider*>(provider.get());
+    return std::shared_ptr<TracerProviderImpl>(provider, pointer);
+  }
```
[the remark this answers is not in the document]
LarryOsterman: This is `std::shared_ptr` magic stuff, I strongly doubt it has a name because it's not a pattern (or rather the pattern's name is `std::static_pointer_cast<T>`).

I couldn't get `std::static_pointer_cast<T>` to work here, so I pulled the relevant guts out of it and used them.

It relies on the ctor for std::shared_ptr which takes a pointer to the _Ref part of the std::shared_ptr and the _Ptr part of the std::shared_ptr. And yes, this is all part of the design for std::shared_ptr. This uses case 8 of [the std::shared_ptr ctor](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr). 

And yes, it's not at all obvious - I got this trick from STL, if it matters.

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
```diff
@@ -3,309 +3,47 @@
...
+    friend std::shared_ptr<TracerProviderImpl> _internal::TracerImplFromTracer(
```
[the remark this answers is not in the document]
LarryOsterman: Do you mean something like "Marked as friend so it can access private members"? 

Isn't that exactly why you mark something as a friend?

sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/opentelemetry.hpp
```diff
@@ -165,24 +28,24 @@ namespace Azure { namespace Core { namespace Tracing { namespace OpenTelemetry {
...
 
+    explicit OpenTelemetryProvider(
+        opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider> tracerProvider
+        = opentelemetry::trace::Provider::GetTracerProvider());
+
   public:
-    OpenTelemetryProvider(
-        opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider> tracerProvider);
-    OpenTelemetryProvider();
+    static std::shared_ptr<OpenTelemetryProvider> Create(
```
antkmsft, line 44: Since this is public API, it needs doxygen comments

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
LarryOsterman: Is what a factory for creating tracer objects?  The `__internal::TracerProviderImpl` object? If so, yes.

However customers should never interact with this object. Non-service developers cannot see this type. Service developers should be interacting with the `ServiceSpan` type. For the extremely small set of service developers who are using the raw OpenTelemetry abstractions, they really should know what they're doing, so I'm not worried about confusing them.

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
LarryOsterman: Yes.

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
```diff
@@ -3,309 +3,47 @@
...
-    };
-
+    class Tracer;
     /**
-     * @brief Represents the status of a span.
+     * @brief Trace Provider - factory for creating Tracer objects.
      */
-    class SpanStatus final : public Azure::Core::_internal::ExtendableEnumeration<SpanStatus> {
-
+    class TracerProviderImpl {
```
antkmsft, line 21: Please add virtual destructor

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     * @brief The operation contains an error.
+     */
+    AZ_CORE_DLLEXPORT const static SpanStatus Error;
+  };
+
+  /**
+   * @brief Span - represents a span in tracing.
+   */
+  class Span {
```
antkmsft, line 177: This one laso needs virtual destructor

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
LarryOsterman: IMHO, this change was explicitly made to *remove* fragility in our header ordering and to fix a circular dependency issue.

`PropagateToHttpHeaders` was needed because there are OpenTelemetry headers which need to be injected into each HTTP request. 

Without significant refactoring of the `Request` object to change how headers are managed, this was the simplest change to resolve the issue.

The *implementation* of `PropagateToHttpHeaders` needs to know the definition of `Request`. The declaration of `PropagateToHttpHeaders` does not need to know the definition.

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
LarryOsterman: url, array, and vector didn't need to be there, map did.

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
LarryOsterman: `std::static_pointer_cast<T>`

But it's not actually used.

sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/opentelemetry.hpp
```diff
@@ -165,24 +28,24 @@ namespace Azure { namespace Core { namespace Tracing { namespace OpenTelemetry {
    * other Azure services.
    *
    */
-  class OpenTelemetryProvider final : public Azure::Core::Tracing::TracerProvider {
+  class OpenTelemetryProvider : public Azure::Core::Tracing::TracerProvider {
```
antkmsft, line 31: Can it continue being `final`?

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
LarryOsterman: I prefer that there NOT be newlines in the #includes - newlines function as context breaks for clang-format and IMHO should only ever be used when you absolutely need to specify the order of header inclusion.

So why should we have a break between the standard headers and the internal headers?

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
```diff
@@ -3,309 +3,47 @@
...
-       * @brief The kind of span to be created.
-       *
-       */
-      SpanKind Kind{SpanKind::Internal};
-      /**
-       * @brief Attributes associated with the span.
-       *
-       */
-      std::unique_ptr<AttributeSet> Attributes;
+    std::shared_ptr<TracerProviderImpl> TracerImplFromTracer(
```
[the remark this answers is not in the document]
antkmsft: I also commented about it, I don't think there's anything wrong with them, but for some reason we decided on it in the past, and I don't think we currently have free standing functions in public or `_internal` space.

sdk/core/azure-core/src/tracing/tracing.cpp
LarryOsterman: So be it. Annoying, but if those are the rules, they're the rules.

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
LarryOsterman: Strictly speaking, no. Changing.

doc/DistributedTracing.md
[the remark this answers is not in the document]
LarryOsterman: There is no scenario where an instance of an `OpenTelemetryProvider` should ever exist that is *not* wrapped in a `std::shared_pointer `.

Given that there is no scenario where customers should ever create an `OpenTelemetryProvider` *without* wrapping it in a `std::shared_pointer`,  I wanted to force customers to use the construct that is going to ensure that they are always successful.

sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/opentelemetry.hpp
LarryOsterman: Yes.

sdk/core/azure-core/inc/azure/core/internal/tracing/tracing_impl.hpp
```diff
@@ -3,20 +3,17 @@
...
 
 #pragma once
 
 #include "azure/core/datetime.hpp"
-#include "azure/core/internal/extendable_enumeration.hpp"
 #include "azure/core/nullable.hpp"
 #include "azure/core/tracing/tracing.hpp"
-#include "azure/core/url.hpp"
-#include <array>
+#include <map>
```
antkmsft, line 14:
```suggestion

#include <map>
```

sdk/core/azure-core/inc/azure/core/internal/tracing/service_tracing.hpp
antkmsft: To separate vsually, I think all other code has it, but also from practical standpoint so that after applying auto-format, include section does not look like

```
#include <algorithm>
#include <azure/...>
#include <vector>
```

antkmsft (approved):

sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp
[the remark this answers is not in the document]
ahsonkhan: I meant more about a comment explaining why we decided to design the class in this way, to be an empty type, which happens to be a friend of `TracerProviderImpl` (essentially what's the rationale behind this approach).

Also, what exactly does `TracerImplFromTracer` need to access from `TracerProvider` and why?

PR #3784, opened 2022-06-30
OpenTelemetry ApiView fix

antkmsft:
OpenTelemetry is the first and so far the only lib that includes 3rd party headers in its public headers, and that is by design.
If we provide the path to the 3rd party include dirs for clang++, that is not going to work for us -- because then ApiView will have all these 3rd party lib types for the review.

Using the command line it generated, I cleated this ApiView PR, LGTM: https://apiview.dev/Assemblies/Review/d76ee29956e840cdbfcbff340ab62678

### DIFF
--- a/eng/scripts/Get-ApiViewCommandLine.ps1
+++ b/eng/scripts/Get-ApiViewCommandLine.ps1
@@ -50,6 +50,9 @@ foreach ($incDir in $AllLibIncDirs) {
     $CmdLine += " -I " + $incDir
 }
 
+# Define an _azure_APIVIEW macro so that the public headers that depend on 3rd party headers apply a work around
+# for declaring 3rd party dependencies in a minimalistic way that is sufficient for the ApiView generation.
+# Otherwise, ApiView would've been including 3rd party types for review, which is not what we want.
 $CmdLine += " -D_azure_APIVIEW"
 
 $CmdLine
--- a/sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/internal/apiview.hpp
+++ b/sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/internal/apiview.hpp
@@ -3,6 +3,9 @@
 
 #pragma once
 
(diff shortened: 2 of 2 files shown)


sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/internal/apiview.hpp
```diff
@@ -0,0 +1,17 @@
...
+  template <typename> struct shared_ptr;
+}
+namespace trace {
+  struct TracerProvider;
+  struct Provider
+  {
+    nostd::shared_ptr<TracerProvider> GetTracerProvider();
+  };
+} // namespace trace
+} // namespace opentelemetry
```
antkmsft, line 17:
```suggestion
} // namespace opentelemetry

```

LarryOsterman (approved):

eng/scripts/Get-ApiViewCommandLine.ps1
```diff
@@ -50,4 +50,6 @@ foreach ($incDir in $AllLibIncDirs) {
     $CmdLine += " -I " + $incDir
 }
 
+$CmdLine += " -D_azure_APIVIEW"
```
ahsonkhan, line 53: I'd recommend adding doc comments in the source to describe why we need this (similar to your explanation in the PR description).

sdk/core/azure-core-tracing-opentelemetry/inc/azure/core/tracing/opentelemetry/internal/apiview.hpp
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+namespace opentelemetry {
```
ahsonkhan, line 6: For future maintenance, add a comment explaining the purpose of this file and the types in it. For an observer, it isn't clear why we need it.

PR #3841, opened 2022-07-19
Helm charts

gearama:
WORK IN PROGRESS, prototype example of what a stress test might look like, subject to change. got to a place where it's failrly stablenad stuff seems to work. Bugs are still present ( not found ) , extensive testing has not yet been performed. 

# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


sdk/core/azure-core/test/libcurl-stress-test/libcurl_stress_test.cpp
```diff
@@ -63,8 +55,9 @@ void Operation(int repetitions)
 int main(int argc, char** argv) // i can have either 0 or 2 params here
 {
   (void)argv; // to get rid of the unused warning
```
LarryOsterman, line 57: If you're not going to be using `argv`, then just eliminate it from the parameters:

```cpp
int main(int argc, char**)
```

sdk/core/azure-core/test/libcurl-stress-test/README.md
```diff
@@ -0,0 +1,29 @@
+# Stress test prototype
```
LarryOsterman, line 1: QQ: Is this a prototype or is it a template/example?

Prototype implies that it's an experiment, and this seems too complete to be a prototype.

sdk/core/azure-core/test/libcurl-stress-test/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
```
LarryOsterman, line 2: Would it make sense to include a link to a definition of what a Helm Chart is? For those ignorant souls who might be reading it?

LarryOsterman (changes requested):

sdk/core/azure-core/test/libcurl-stress-test/README.md
gearama: prototype of

sdk/core/azure-core/test/libcurl-stress-test/Chart.yaml
ahsonkhan: I agree with @LarryOsterman, I am not familiar with helm charts at all. Package manager for kubernetes?

Can you also share some context, maybe in the PR description, what the purpose of this PR is? What are we trying to accomplish and why (what's motivating this)?

sdk/core/azure-core/test/libcurl-stress-test/libcurl_stress_test.cpp
LarryOsterman: I'm confused. Godbolt says [this compiles](https://www.godbolt.org/z/qe1Tj6aTK):
```c++
int main(int argc, char**) {
    return argc+1;
}
```

sdk/core/azure-core/test/libcurl-stress-test/Chart.yaml
LarryOsterman: This is the C++ SDK. Everyone working in the C++ SDK is expected to have a working knowledge of C++. 

However we're *NOT* expected to have working knowledge of helm charts, so something that helps us understand them would be greatly appreciated.

gearama:
@larryosterman

sdk/core/azure-core/test/libcurl-stress-test/libcurl_stress_test.cpp
gearama: never seen that declaration before, :) seems hacky but if it works, why not?

sdk/core/azure-core/test/libcurl-stress-test/Chart.yaml
gearama: updated the readme

sdk/core/azure-core/test/libcurl-stress-test/libcurl_stress_test.cpp
LarryOsterman: As far as I know, it's not an uncommon C++ idiom (actually it might even be a C idiom). 

It's how you indicate to the C++ compiler (or more accurately C++ linters) that yes, there's a parameter to this method, but we're not going to be using it).

I use it a lot because it's so useful. And it's WAY less annoying than the "`<variable>;`" trick.

LarryOsterman (approved):

.vscode/cspell.json
```diff
@@ -11,6 +11,8 @@
     "*.exe",
     "*.a",
     "*.lib",
+    "*.yaml",  
+    "**/libcurl-stress-test/README.md",
```
ahsonkhan, line 15: We should try not be disable cspell on markdown text files, like this. Either we fix the spelling errors, or if there are false positives, please add them to the dictionary.

LarryOsterman: IIRC there may be a way to add cspell words to markdown files, which allows you to add them to a local dictionary instead of modifying the master dictionary.

ahsonkhan: Yep, there is! :)
You can add them as a comment on the top, as metadata (which isn't visible), like so:
https://raw.githubusercontent.com/Azure/azure-sdk-for-cpp/86aa9b4b0cfdef2af103ed2330064079ada68773/doc/DistributedTracing.md

PR #3850, opened 2022-07-26
STG 79 80 81 Features

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?


sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_container_client.hpp
```diff
@@ -285,6 +285,24 @@ namespace Azure { namespace Storage { namespace Blobs {
         const UploadBlockBlobOptions& options = UploadBlockBlobOptions(),
         const Azure::Core::Context& context = Azure::Core::Context()) const;
 
+    /**
+     * @brief The Filter Blobs operation enables callers to list blobs in a container whose
+     * tags match a given search expression.
+     *
+     * @param tagFilterSqlExpression The where parameter enables the caller to query blobs
+     * whose tags match a given expression. The given expression must evaluate to true for a blob to
+     * be returned in the results. The[OData - ABNF] filter syntax rule defines the formal grammar
```
ahsonkhan, line 294:
```suggestion
     * be returned in the results. The [OData - ABNF] filter syntax rule defines the formal grammar
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -1423,9 +1434,9 @@ namespace Azure { namespace Storage { namespace Blobs {
          */
         std::vector<BlobItem> Items;
         /**
-         * Array of ListBlobsHierarchySegmentResponseBlobPrefixesItem.
+         * Array of BlobName.
          */
-        std::vector<std::string> BlobPrefixes;
+        std::vector<BlobName> BlobPrefixes;
```
ahsonkhan, line 1439: Are we making sure this change isn't breaking in the public surface area?

sdk/storage/azure-storage-blobs/CHANGELOG.md
```diff
@@ -4,10 +4,19 @@
...
+- Added encryption scope support for `BlobClient::CopyFromUri()`.
+- Added support for filtering blobs on container level.
+- Added support for tags copy mode (replace or copy from source) when copying blobs from URL.
+- Added support for permanent delete permission in SAS.
+
 ### Breaking Changes
 
 ### Bugs Fixed
 
+- Fixed listing blobs failure when blob name contains invalid characters in xml.
```
ahsonkhan, line 18: How was this issue discovered?
Do we have a test for it. Also, can you please point me to where the fix is, in this PR, within the source?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
```diff
@@ -163,6 +163,49 @@ namespace Azure { namespace Storage {
...
+      struct BlobItem final
+      {
+        /**
+         * Blob name.
+         */
+        std::string Name;
+        /**
+         * Indicates whether this blob was deleted.
+         */
+        bool IsDeleted = bool();
```
LarryOsterman, line 178: Instead of using the function call form of the constructor, use the default value initializer.
```suggestion
        bool IsDeleted{};
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
```diff
@@ -163,6 +163,49 @@ namespace Azure { namespace Storage {
...
+         */
+        BlobItemDetails Details;
+        /**
+         * Indicates that this root blob has been deleted, but it has versions that are active.
+         */
+        Nullable<bool> HasVersionsOnly;
+        /**
+         * Size in bytes.
+         */
+        int64_t BlobSize = int64_t();
```
LarryOsterman, line 202: Initializer form instead of assigning to constructor.

```suggestion
        int64_t BlobSize{};
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_sas_builder.hpp
```diff
@@ -90,6 +90,11 @@ namespace Azure { namespace Storage { namespace Sas {
      */
     SetImmutabilityPolicy = 256,
```
LarryOsterman, line 91: In general, I'm not a fan of expressing bit values in decimal. It tends to be more error prone than `0x100`, `0x200`, `0x400`, etc.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -921,6 +919,17 @@ namespace Azure { namespace Storage { namespace Blobs {
          */
         std::string LeaseId;
       };
+      struct BlobName final
+      {
+        /**
+         * Indicates if the blob name is encoded.
+         */
+        bool Encoded = bool();
```
LarryOsterman, line 927:
```suggestion
        bool Encoded{};
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -921,6 +919,17 @@ namespace Azure { namespace Storage { namespace Blobs {
          */
         std::string LeaseId;
       };
+      struct BlobName final
+      {
+        /**
+         * Indicates if the blob name is encoded.
```
LarryOsterman, line 925: "encoded" implies that there is an encoding. What is the encoding for the blob name? UTF-8? UTF-16? Base64? Base64url, DER? Something else?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -921,6 +919,17 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+         */
+        bool Encoded = bool();
+        /**
+         * The name of the blob.
+         */
+        std::string Content;
```
LarryOsterman, line 931: If `Content` has the ability to be "encoded", then it shouldn't be represented as a `std::string` (which is either ASCII or UTF-8 text). Instead if `Content` can be encoded, this should probably be `std::vector<uint8_t>`.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -1314,48 +1323,50 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+      struct BlobItem final
+      {
+        /**
+         * Blob name.
+         */
+        BlobName Name;
+        /**
+         * Indicates whether this blob was deleted.
+         */
+        bool IsDeleted = bool();
```
LarryOsterman, line 1339:
```suggestion
        bool IsDeleted{};
```

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -1314,48 +1323,50 @@ namespace Azure { namespace Storage { namespace Blobs {
...
+         */
+        BlobItemDetails Details;
+        /**
+         * Indicates that this root blob has been deleted, but it has versions that are active.
+         */
+        Nullable<bool> HasVersionsOnly;
+        /**
+         * Size in bytes.
+         */
+        int64_t BlobSize = int64_t();
```
LarryOsterman, line 1363:
```suggestion
        int64_t BlobSize{};
```

You get the idea - I'm going to stop doing these at this point.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
```diff
@@ -2159,6 +2170,26 @@ namespace Azure { namespace Storage { namespace Blobs {
         Models::CopyStatus CopyStatus;
       };
     } // namespace _detail
+    /**
+     * @brief Optional, default 'replace'.  Indicates if source tags should be copied or replaced
+     * with the tags specified by x-ms-tags.
+     */
+    class BlobCopySourceTagsMode final {
```
LarryOsterman, line 2177: Use an `ExtendableEnumeration` here - that allows you to get rid of a LOT of boilerplate.

sdk/storage/azure-storage-blobs/src/blob_container_client.cpp
```diff
@@ -22,6 +22,78 @@
...
+      }
+      {
+        std::map<std::string, std::vector<Models::ObjectReplicationRule>> orPropertiesMap;
+        for (auto& policy : blobItem.Details.ObjectReplicationSourceProperties)
+        {
+          for (auto& rule : policy.Rules)
+          {
+            auto underscorePos = rule.RuleId.find('_', 3);
+            std::string policyId
+                = std::string(rule.RuleId.begin() + 3, rule.RuleId.begin() + underscorePos);
```
LarryOsterman, line 78: A comment here describing the format of rule.RuleId might be nice - there's an understanding about the format for this field that isn't obvious (the first 3 characters of the rule id are ignored (why?), then the policy ID is everything in the rule ID before the first underscore character.

That's not at all easy to glean from the code.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
LarryOsterman: Ok, reading further in the code, it appears that this is URL encoding. As such, IMHO the name should be "UrlEncoded" or something similar.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
LarryOsterman: Ok, reading further, this is URL encoded. Since URL encoded text is ASCII text, using std::string is probably reasonable.

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
```diff
@@ -1248,6 +1274,42 @@ namespace Azure { namespace Storage { namespace Test {
       auto blobItem = GetBlobItem(blobName);
       EXPECT_EQ(blobItem.Name, blobName);
     }
+    {
+      const std::string blobPrefix
+          = std::string("aaaaa\xEF\xBF\xBF") + "bbb/"; // UTF-8 0xEF, 0xBF, 0xBF is UTF-16 0xFFFF
```
LarryOsterman, line 1279: UTF-16 0xffff is [Unicode NOT A CHARACTER](https://unicode-table.com/en/FFFF/).

That is not supposed to ever be encoded in unicode text, is that intentional? If so, you might want to explain why you're using a character that should never be encoded.

LarryOsterman (changes requested):

sdk/storage/azure-storage-blobs/CHANGELOG.md
Jinming-Hu: It's a REST API layer change. You'll see the change if you diff the swagger with the previous version.

Jinming-Hu: `BlobName` used to be a `std::string`. Now it's a struct with a `bool` in it indicating if it's encoded.

We'll decode and return a `std::string` in convenience layer, so that it's not a breaking change.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
Jinming-Hu: It doesn't make any difference for primitive types, does it?

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_sas_builder.hpp
Jinming-Hu: I feel decimal is more intuitive. I can tell 256 is twice as 128 at one glance. But I need to pause for a moment to tell 0x100 is twice as 0x80.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
Jinming-Hu: It's in `_detail` namespace so it doesn't really matter. The name is defined in swagger file, which is not maintained by our team. Making a change to swagger will break all other languages.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
Jinming-Hu: This is in `_detail` namespace so it's non breaking.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
Jinming-Hu: The code gen was written before we have the base class. Apparently we haven't got time to adapt the extendable enum base class. Tracked in this issue. https://github.com/Azure/autorest.cpp/issues/109

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
Jinming-Hu: `0xfffe` and `0xffff` are valid code points, but not valid in XML. This feature is to add a way to list blobs with these unicode characters in blob names.

sdk/storage/azure-storage-blobs/src/blob_container_client.cpp
Jinming-Hu: Added some comments

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
Jinming-Hu: @ahsonkhan Do you have a preference on this one?

Most of the use cases are in generated code, I'd like to keep consistent. If we were to fix it, I'd like to fix them all. We'll need to update CodeGen first.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_sas_builder.hpp
LarryOsterman: Ok, I'm confused then. From the values, this looked like [flags enumeration](https://softwareengineering.stackexchange.com/questions/194412/using-scoped-enums-for-bit-flags-in-c) (where you would perform logical operators ("or", "and", "xor", etc) to values). 

If you are applying logical operators to these values, then the only thing that matters is their bit position. And if the only thing you care about is their bit position, then representing the values in hex makes much more sense than decimal because it's trivial to determine the bit positions of a hex digit.

However, if there are no circumstances where you would apply logical operators on the values, then using numeric values makes sense.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
LarryOsterman: That's fair.  I didn't realize that this was generated code.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/rest_client.hpp
LarryOsterman: Ah - I was unaware that rest_client was autogenerated.

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
LarryOsterman: Then document this please. Every reference I saw for 0xffff and 0xfffe indicated that they weren't supposed to appear in character strings, so this raised red flags.

sdk/storage/azure-storage-blobs/inc/azure/storage/blobs/blob_responses.hpp
LarryOsterman: There are C++ core guidelines that strongly recommend using initializer syntax (`{}`) over assignment initialization (`=xxx`) or function initialization (`()`). 

That's because assignment initialization is easily confused with assignment, and function initialization is easily confused with a function call. Initializer syntax is specific and unique.

However, if this is generated code, it's up to the generator

sdk/storage/azure-storage-blobs/test/ut/blob_container_client_test.cpp
Jinming-Hu: added

katmsft (approved):

sdk/storage/azure-storage-blobs/src/blob_sas_builder.cpp
```diff
@@ -154,8 +163,8 @@ namespace Azure { namespace Storage { namespace Sas {
     std::string stringToSign = Permissions + "\n" + startsOnStr + "\n" + expiresOnStr + "\n"
         + canonicalName + "\n" + Identifier + "\n" + (IPRange.HasValue() ? IPRange.Value() : "")
         + "\n" + protocol + "\n" + SasVersion + "\n" + resource + "\n" + snapshotVersion + "\n"
-        + CacheControl + "\n" + ContentDisposition + "\n" + ContentEncoding + "\n" + ContentLanguage
-        + "\n" + ContentType;
+        + EncryptionScope + "\n" + CacheControl + "\n" + ContentDisposition + "\n" + ContentEncoding
```
EmmaZhu, line 166: would cpp SDK support different SAS version? Sign string would be different for different version

sdk/storage/azure-storage-blobs/src/rest_client.cpp
```diff
@@ -4738,7 +4976,7 @@ namespace Azure { namespace Storage { namespace Blobs {
       {
         request.SetHeader("x-ms-lease-id", options.LeaseId.Value());
       }
-      request.SetHeader("x-ms-version", "2020-10-02");
+      request.SetHeader("x-ms-version", "2021-04-10");
```
EmmaZhu, line 4979: It needs to change API version in many places, why not define a constant?

sdk/storage/azure-storage-blobs/swagger/README.md
```diff
@@ -467,6 +469,54 @@ directive:
       $.ArchiveStatus.description = "For blob storage LRS accounts, valid values are rehydrate-pending-to-hot/rehydrate-pending-to-cool. If the blob is being rehydrated and is not complete then this value indicates that rehydrate is pending and also tells the destination tier.";
 ```
 
+### Striped Blob Support
```
EmmaZhu, line 472: Any document for this striped blob?

sdk/storage/azure-storage-blobs/src/blob_sas_builder.cpp
Jinming-Hu: No, we only support one specific SAS version. Customers won't be able to specify the version.

sdk/storage/azure-storage-blobs/src/rest_client.cpp
Jinming-Hu: That is a good point. I also ever considered that, many times. Unfortunately, codeGen is not maintained by our team now. So...

sdk/storage/azure-storage-blobs/swagger/README.md
Jinming-Hu: Nah, I prefer not. Striped blob is more of an internal only thing.

PR #3858, opened 2022-07-29
Retry poll calls on EINTR

### LINKED ISSUE #3692: Interrupted poll calls cause spurious HTTP request failures
To support cancellation, the curl client polls the underlying socket for read/writability with a short timeout (1 sec).  However, at least on POSIX systems, the call to `poll` may be interrupted by signals.  Currently, such an interruption returns an error and results in a spurious failure of the HTTP request.  Instead, the `poll` call should be retried.

Request retries mask this issue in many cases, but this issue surfaced because in an application I work on the process is sometimes signaled with a regular frequency, which resulted in operations failing at a very high rate.

This is hard to unit test, but to reproduce it should be sufficient to signal the process with a high frequency while making HTTP requests.  This is on a POSIX system (Linux, in my case).  I'm not familiar enough with Windows to know if there's a similar issue around `WSAPoll`.

Below is a patch I'm using to resolve the issue.  I'd send a pull request but this is for POSIX platforms only.
```
From 02a4b073a45ed5d54a3c04ac78aabf72f0a1a416 Mon Sep 17 00:00:00 2001
From: John Heffner <john@topofcenter.net>
Date: Wed, 1 Jun 2022 16:55:48 -0400
Subject: [PATCH] Retry poll calls on EINTR

When signals are delievered to the process, calls here to poll may be
interrupted and return with a spurious failure.  The call instead should be
restarted.
---
 sdk/core/azure-core/src/http/curl/curl.cpp | 26 ++++++++++++++++++++--
 1 file changed, 24 insertions(+), 2 deletions(-)

diff --git a/sdk/core/azure-core/src/http/curl/curl.cpp b/sdk/core/azure-core/src/http/curl/curl.cpp
index ea94d08c..fbbb1764 100644
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -16,6 +16,7 @@
 #include <poll.h> // for poll()
 #include <sys/socket.h> // for socket shutdown
+#include <chrono>
 #if !defined(WIN32_LEAN_AND_MEAN)
 #define WIN32_LEAN_AND_MEAN
@@ -103,17 +104,38 @@ int pollSocketUntilEventOrTimeout(
     interval = timeout;
   }
   int result = 0;
-  for (long counter = 0; counter < timeout && result == 0; counter = counter + interval)
+  bool poll_interrupted = false;  // deadline and now are only valid if poll_interrupted
+  long deadline;
+  long now;
+  for (long counter = 0; (poll_interrupted ? now < deadline : counter < timeout) && result == 0;
+       counter = counter + interval)
   {
     // check cancelation
     context.ThrowIfCancelled();
     result = poll(&poller, 1, interval);
+    if (result < 0 && EINTR == errno)
+    {
+      // Only make extra syscalls to get the time in the (rare) event that poll
+      // is interrupted. We jump through these hoops because the poll call does
+      // not tell us how long it waited before being interrupted.  For the first
+      // interruption only, assume no time has passed during the interrupted
+      // call.
+      now = std::chrono::duration_cast<std::chrono::milliseconds>(
+                std::chrono::steady_clock::now().time_since_epoch())
+                .count();
+      if (!poll_interrupted)
+      {
+        poll_interrupted = true;
+        deadline = now + timeout - counter;
+        result = 0;
+      }
+    }
     result = WSAPoll(&poller, 1, interval);
 #endif
   }
-  // result can be either 0 (timeout) or > 1 (socket ready)
+  // result can be 0 (timeout), > 0 (socket ready), or < 0 (error)
   return result;
 }
 
-- 
2.25.1

```

johnwheffner:
When signals are delievered to the process, calls here to poll may be
interrupted and return with a spurious failure.  The call instead should be
restarted.
 
Resolves #3692

### DIFF
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -1,12 +1,22 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
+#include "azure/core/platform.hpp"
+
+#if defined(AZ_PLATFORM_WINDOWS)
+#if !defined(WIN32_LEAN_AND_MEAN)
+#define WIN32_LEAN_AND_MEAN
+#endif
+#if !defined(NOMINMAX)
+#define NOMINMAX
+#endif
+#endif
+
 #include "azure/core/http/curl_transport.hpp"
 #include "azure/core/http/http.hpp"
 #include "azure/core/http/policies/policy.hpp"
 #include "azure/core/http/transport.hpp"
 #include "azure/core/internal/diagnostics/log.hpp"
-#include "azure/core/platform.hpp"
 
 // Private include
 #include "curl_connection_pool_private.hpp"
@@ -17,16 +27,11 @@
 #include <poll.h> // for poll()
 #include <sys/socket.h> // for socket shutdown
 #elif defined(AZ_PLATFORM_WINDOWS)
-#if !defined(WIN32_LEAN_AND_MEAN)
-#define WIN32_LEAN_AND_MEAN
-#endif
-#if !defined(NOMINMAX)
-#define NOMINMAX
-#endif
 #include <winsock2.h> // for WSAPoll();
 #endif
 
 #include <algorithm>
+#include <chrono>
 #include <string>
 #include <thread>
 
@@ -94,25 +99,35 @@ int pollSocketUntilEventOrTimeout(
   // we use 1 as arg.
 
   // Cancelation is possible by calling poll() with small time intervals instead of using the
-  // requested timeout. Default interval for calling poll() is 1 sec whenever arg timeout is
-  // greater than 1 sec. Otherwise the interval is set to timeout
-  long interval = 1000; // 1 second
-  if (timeout < interval)
-  {
-    interval = timeout;
-  }
+  // requested timeout. The polling interval is 1 second.
+  static constexpr std::chrono::milliseconds pollInterval(1000); // 1 second
   int result = 0;
-  for (long counter = 0; counter < timeout && result == 0; counter = counter + interval)
+  auto now = std::chrono::steady_clock::now();
+  auto deadline = now + std::chrono::milliseconds(timeout);
+  while (now < deadline)
   {
     // check cancelation
     context.ThrowIfCancelled();
+    int pollTimeoutMs = static_cast<int>(
+        std::min(
+            pollInterval, std::chrono::duration_cast<std::chrono::milliseconds>(deadline - now))
+            .count());
 #if defined(AZ_PLATFORM_POSIX)
-    result = poll(&poller, 1, interval);
+    result = poll(&poller, 1, pollTimeoutMs);
+    if (result < 0 && EINTR == errno)
+    {
+      continue;
+    }
 #elif defined(AZ_PLATFORM_WINDOWS)
-    result = WSAPoll(&poller, 1, interval);
+    result = WSAPoll(&poller, 1, pollTimeoutMs);
 #endif
+    if (result != 0)
+    {
+      return result;
+    }
+    now = std::chrono::steady_clock::now();
   }
-  // result can be either 0 (timeout) or > 1 (socket ready)
+  // result can be 0 (timeout), > 0 (socket ready), or < 0 (error)
   return result;
 }
 


sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -102,17 +103,38 @@ int pollSocketUntilEventOrTimeout(
     interval = timeout;
   }
   int result = 0;
-  for (long counter = 0; counter < timeout && result == 0; counter = counter + interval)
+  bool poll_interrupted = false; // deadline and now are only valid if poll_interrupted
+  long deadline{};
+  long now{};
+  for (long counter = 0; (poll_interrupted ? now < deadline : counter < timeout) && result == 0;
```
LarryOsterman, line 109: This for loop seems *really* awkward to me.

Essentially what seems to be happening here is that we start by iterating over the loop terminating when we have iterated for `timeout / interval` times.

However, if a signal is received, the loop termination condition changes to be "however many iterations are needed until the deadline is expired".

IMHO the loop termination condition should simply be whenever the deadline is exceeded.

One possible modification is to create a std::chrono::steady_clock variable named `deadline` and then change the loop from a for loop to something like:
```c++
long deadline = std::chrono::steady_clock::now() + std::chrono::duration<std::chrono::milliseconds>(timeout);
while (now < deadline)
{
  context.ThrowIfCancelled();
  result = poll(&poller, 1, interval);
  if (result < 0 && errno == EINTR)
  {
      continue;
  }
  result = WSAPoll(&poller, 1, interval);
#endif
  if (result != 0)
  {
    return result;
  }
  now = std::chrono::steady_clock::now();
}
return 0;
```

With a structure like that, you don't need to do any special handling for the EINTR case - you just ignore EINTR errors being returned from poll, and then the rest of the code "just works".

To be clear: this is just my scribbling code into the github editor. you can structure the code as you feel appropriate. But I feel strongly that there should just be a single termination condition for the loop rather than varying response conditions.

LarryOsterman (changes requested):

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -102,17 +103,38 @@ int pollSocketUntilEventOrTimeout(
...
 #if defined(AZ_PLATFORM_POSIX)
     result = poll(&poller, 1, interval);
+    if (result < 0 && EINTR == errno)
+    {
+      // Only make extra syscalls to get the time in the (rare) event that poll
+      // is interrupted. We jump through these hoops because the poll call does
+      // not tell us how long it waited before being interrupted.  For the first
+      // interruption only, assume no time has passed during the interrupted
+      // call.
+      now = std::chrono::duration_cast<std::chrono::milliseconds>(
```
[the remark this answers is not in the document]
johnwheffner: On a platform with a 32-bit `long`, I suppose so.  I'll change their types.  According to the standard, the `std::chrono::milliseconds` representation type is good for at least a couple hundred more years.

sdk/core/azure-core/src/http/curl/curl.cpp
johnwheffner: The comment indicates the motivation for the extra complexity, which is to avoid an extra syscall (to read the clock) in the common case.  Perhaps this is an unnecessary optimization but seems a shame to add overhead to what's got to be one of the most frequently executed codepaths in the sdk.  Even a single extra syscall in a poll/read or write loop can be significant in bulk data transfer, and the case where we actually need the clock here is quite rare.

LarryOsterman: My primary concern here is the added complexity of the loop termination conditional. There are functionally two separate termination cases, depending on whether a EINTR was thrown or not.

Also remember that this is polling a network socket. This function is almost never going to be CPU bound on any platform. You're looking at a *minimum* round trip time to the server measured in tens of milliseconds, and a polling interval of 1 second. I'm hard pressed to see how reading the steady clock is going to be a performance bottleneck (on Windows, reading the steady clock is literally reading 8 bytes of data from a piece of shared memory, I'm not sure how the corresponding functionalty works on Linux, but I'm pretty sure it's comparably fast).

IMHO reducing code complexity almost always trumps performance, unless you have evidence that the function is a performance bottleneck. And I'm hard pressed to see how this function is a performance bottleneck. 

Larry


PS: If you are *truly* worried about CPU consumption inside this loop, if you should look at context.ThrowIfCancelled(), it also calls std::chrono::system_clock::now() in addition to calling GetDeadline() which loops through a stack of context objects looking for a context object with a deadline. So you can return to the original number of calls to `std::chrono::steady_clock::now()` by calling `context.GetDeadLine()` to retrieve the context deadline outside the loop and then setting the `deadline` variable to the min of deadline from the parameter and deadline from the context.

johnwheffner: Fair enough, especially as the branches aren't easily tested.

johnwheffner:
This has exceeded my time box, so I'll leave it in your hands from here on @LarryOsterman.

LarryOsterman (approved):

sdk/core/azure-core/src/http/curl/curl.cpp
LarryOsterman: > This has exceeded my time box, so I'll leave it in your hands from here on @LarryOsterman.

That's completely fair. The fix for the CI issue is straightforward, but annoying. 

Basically at the top of the file, you need to move the NOMINMAX goo:
```c++
#if !defined(NOMINMAX)
#define NOMINMAX
#endif
#endif

#include "azure/core/http/curl_transport.hpp"
```

because the Windows headers are profoundly broken.

I understand your timebox limits and appreciate them, but I'd *LOVE* it if you could get contributor credit for this change (since you did ALL the work here and it's a good improvement to the SDK).

I'm going to let you make the call about whether you make this modification to your PR, or if you want me to drive it forward.

If you don't want to invest any more time in this PR(and I can understand that), I see two ways of moving forward with this:
 1. I can clone your fork and make the change there. That way this pull request completes with your PR.
 2. I can integrate your changes in my PR. The bad news is that I don't know how to give you credit.

Let me know how you want to proceed.

ahsonkhan: > The bad news is that I don't know how to give you credit.

Fwiw, at the very least, the changelog entry for the fix will have the acknowledgement of the contribution. :)

For example:
https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/CHANGELOG.md#acknowledgments

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -94,25 +95,33 @@ int pollSocketUntilEventOrTimeout(
...
   // Cancelation is possible by calling poll() with small time intervals instead of using the
-  // requested timeout. Default interval for calling poll() is 1 sec whenever arg timeout is
-  // greater than 1 sec. Otherwise the interval is set to timeout
-  long interval = 1000; // 1 second
-  if (timeout < interval)
-  {
-    interval = timeout;
-  }
+  // requested timeout. The polling interval is 1 second.
+  static constexpr std::chrono::milliseconds poll_interval(1000); // 1 second
```
Jinming-Hu, line 99:
```suggestion
  static constexpr std::chrono::milliseconds pollInterval(1000); // 1 second
```

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -94,25 +95,33 @@ int pollSocketUntilEventOrTimeout(
...
   int result = 0;
-  for (long counter = 0; counter < timeout && result == 0; counter = counter + interval)
+  auto now = std::chrono::steady_clock::now();
+  auto deadline = now + std::chrono::milliseconds(timeout);
+  while (now < deadline)
   {
     // check cancelation
     context.ThrowIfCancelled();
+    std::chrono::milliseconds poll_timeout = std::min(
```
Jinming-Hu, line 107:
```suggestion
    std::chrono::milliseconds pollTimeout = std::min(
```

Jinming-Hu (approved):
LGTM

ahsonkhan:
@LarryOsterman Do we suspect such a change could have any perf issues (maybe for example from referencing `EINTR`)? I am not familiar enough with that, is it a system call? Do you think the perf impact of such a change is minimal?

https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/core/perf#getting-started

This test might be worth checking before/after, unless we don't suspect much perf risk here:
https://github.com/Azure/azure-sdk-for-cpp/blob/9eae0dff88df72b19f6853709a30b3126e4464d0/sdk/core/perf/test/inc/azure/perf/test/http_client_get_test.hpp#L57

LarryOsterman:
> @LarryOsterman Do we suspect such a change could have any perf issues (maybe for example from referencing `EINTR`)? I am not familiar enough with that, is it a system call? Do you think the perf impact of such a change is minimal?
> 
> https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/core/perf#getting-started
> 
> This test might be worth checking before/after, unless we don't suspect much perf risk here:
> 
> https://github.com/Azure/azure-sdk-for-cpp/blob/9eae0dff88df72b19f6853709a30b3126e4464d0/sdk/core/perf/test/inc/azure/perf/test/http_client_get_test.hpp#L57

Calls to `errno` are intended to be incredibly fast - typically they either retrieve a thread-local variable or a global variable (depending on the threading support in the C runtime library). And `EINTR` is a manifest constant, so it's likely to be rather fast (since it's literally burned into the code :)).

The impact of adding the call to std::chrono::now() is going to be more impactful. On Windows, the GetTickCount64() API (which is what backs `std::chrono::steady_clock::now()`) reads a 64bit value from global shared memory so it's insanely efficient. On linux, `std::chrono::steady_clock::now()` is backed by [clock_gettime](https://www.educba.com/clock_gettime/) which similarly reads a value from kernel memory.

So none of the functionality added is likely to add more than a couple dozen instructions to the polling loop - it's possible that there are additional processor stalls introduced, but in the face of a 1 second polling interval, I don't htink it's going to matter that much.

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -1,12 +1,22 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
+#include "azure/core/platform.hpp"
```
ahsonkhan, line 4: Out of curiosity, what required this reordering of the platform.hpp header (along with `WIN32_LEAN_AND_MEAN`)?

If it's required, it would be useful to add that as a comment for future maintainers, and to avoid reordering it again.

LarryOsterman: The @!$@#$ windows headers #define min(x,y). 

And apparently something brings them in before the old code.

This wasn't a problem because nobody used std::min before, but this PR adds in a call to std::min. And thus chaos ensued. This fixes that problem.

PR #3889, opened 2022-08-17
Fix `azure-storage-blobs-cpp` beta install

### LINKED ISSUE #3884: Install via vcpkg private registry failed
I was trying to install storage-blobs 12.6.0.beta1 from vpckg private registry, which depends on storage-common 12.3.0.beta1. Both packages were released earlier this month. It failed with 

```
CMake Error at C:/Users/jinmhu/Desktop/vcpkg/scripts/buildsystems/vcpkg.cmake:824 (_find_package):
  _find_package called with invalid argument "12.3.0-beta.1"
Call Stack (most recent call first):
  CMakeLists.txt:35 (find_package)


CMake Error at C:/Users/jinmhu/Desktop/vcpkg/scripts/buildsystems/vcpkg.cmake:824 (_find_package):
  _find_package called with invalid argument "12.3.0-beta.1"
Call Stack (most recent call first):
  CMakeLists.txt:37 (find_package)
```

I think the issue is in this line of code.

https://github.com/Azure/azure-sdk-for-cpp/blob/c152c05d4a56074d20451b7efa483bbf17bef116/sdk/storage/azure-storage-blobs/vcpkg.json#L14

The command I was using: `vcpkg install azure-storage-blobs-cpp`.

Did I do something wrong here?

antkmsft:
Fixes #3884. In the issue, I explained a bit more about this type of version number limitation in Cmake.

@Jinming-Hu, the private registry is already fixed with this very fix, see two of the last commits for details:
https://github.com/Azure/azure-sdk-vcpkg-betas/commits/main

Once we release stable version, it may be added to `find_package()` again.


sdk/storage/azure-storage-blobs/CMakeLists.txt
```diff
@@ -32,9 +32,9 @@ if(FETCH_SOURCE_DEPS)
         add_subdirectory(${oneFolder} EXCLUDE_FROM_ALL)
     endforeach()
 elseif(NOT AZ_ALL_LIBRARIES)
-  find_package(azure-storage-common-cpp "12.3.0-beta.1" CONFIG QUIET)
+  find_package(azure-storage-common-cpp CONFIG QUIET)
```
Jinming-Hu, line 35: If we revert the latest change, shouldn't this be `12.2.4`?

antkmsft: It can be, but will it compile, does it have all the API that Blobs are using? If not, there's no much difference to having no version specified?

Jinming-Hu (approved):

PR #3922, opened 2022-09-01
Add github actions based check enforcer

benbp:
Adding check enforcer v2 action to repository. See https://github.com/Azure/azure-sdk-tools/pull/3287 for details.


LarryOsterman (approved):
Fundamentally I don't see anything challenging here, but I'm far from an expert.

PR #3941, opened 2022-09-19
Add NTFileTimeTimeConverter

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [ ] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/datetime.hpp
+++ b/sdk/core/azure-core/inc/azure/core/datetime.hpp
@@ -311,6 +311,51 @@ namespace Core { namespace _internal {
      */
     ~PosixTimeConverter() = delete;
   };
+
+  /**
+   * @brief Provides convertion methods for Windows NT filetime to an #Azure::DateTime.
+   *
+   */
+  class NTFileTimeTimeConverter final {
+  public:
+    /**
+     * @brief Converts Windows NT filetime to an #Azure::DateTime.
+     *
+     * @param ntFiletime The number of 100-nanoseconds since 1601-01-01.
+     * @return Calculated #Azure::DateTime.
+     */
+    static DateTime NTFileTimeToDateTime(int64_t ntFiletime)
+    {
+      auto t = DateTime(1601) + Azure::_detail::Clock::duration(ntFiletime);
+      return DateTime(t);
+    }
+
+    /**
+     * @brief Converts a DateTime to Windows NT filetime.
+     *
+     * @param dateTime The `%DateTime` to convert.
+     * @return The number of 100-nanoseconds since 1601-01-01.
+     */
+    static int64_t DateTimeToNTFileTime(DateTime const& dateTime)
+    {
+      return std::chrono::duration_cast<Azure::_detail::Clock::duration>(dateTime - DateTime(1601))
+          .count();
+    }
+
+  private:
+    /**
+     * @brief An instance of `%NTFileTimeTimeConverter` class cannot be created.
+     *
+     */
+    NTFileTimeTimeConverter() = delete;
+
+    /**
+     * @brief An instance of `%NTFileTimeTimeConverter` class cannot be destructed, because no
+     * instance can be created.
+     *
+     */
+    ~NTFileTimeTimeConverter() = delete;
+  };
 }} // namespace Core::_internal
 
 } // namespace Azure
--- a/sdk/core/azure-core/test/ut/datetime_test.cpp
+++ b/sdk/core/azure-core/test/ut/datetime_test.cpp
@@ -895,3 +895,44 @@ TEST(DateTime, LeapYear)
   EXPECT_NO_THROW(static_cast<void>(DateTime(2021, 2, 28)));
   EXPECT_THROW(static_cast<void>(DateTime(2021, 2, 29)), std::invalid_argument);
 }
+
+TEST(DateTime, WindowsNTFileTimeConverter)
+{
+  using namespace Azure::Core::_internal;
+
+  EXPECT_EQ(
+      NTFileTimeTimeConverter::DateTimeToNTFileTime(
+          NTFileTimeTimeConverter::NTFileTimeToDateTime(0)),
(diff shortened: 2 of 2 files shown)


Jinming-Hu:
Do we add changelog items for changes in `_internal` namespace?

ahsonkhan:
> Do we add changelog items for changes in `_internal` namespace?

No, since it isn't intended for end customers, it would otherwise be a bit noisy.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -311,6 +311,51 @@ namespace Core { namespace _internal {
      */
     ~PosixTimeConverter() = delete;
   };
+
+  /**
+   * @brief Provides convertion methods for Windows NT filetime to an #Azure::DateTime.
```
ahsonkhan, line 316:
```suggestion
   * @brief Provides conversion methods for Windows NT filetime to an #Azure::DateTime.
```

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -311,6 +311,51 @@ namespace Core { namespace _internal {
...
+   *
+   */
+  class NTFileTimeTimeConverter final {
```
ahsonkhan, line 319: `TimeTime`?

I think we should workshop the name a bit. Maybe we can avoid the window-ism, if this format is more general purpose.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -311,6 +311,51 @@ namespace Core { namespace _internal {
...
+    /**
+     * @brief Converts a DateTime to Windows NT filetime.
+     *
+     * @param dateTime The `%DateTime` to convert.
+     * @return The number of 100-nanoseconds since 1601-01-01.
+     */
+    static int64_t DateTimeToNTFileTime(DateTime const& dateTime)
+    {
+      return std::chrono::duration_cast<Azure::_detail::Clock::duration>(dateTime - DateTime(1601))
+          .count();
```
ahsonkhan, line 342: Will `count` always return 100-nanoseconds intervals? The doc doesn't seem to specify that:
https://cplusplus.com/reference/chrono/duration/count/

ahsonkhan: Oh never mind, our `Clock::duration` is 100-nanoseconds (10^-7), which is what we are casting to.

A nit outside the scope of this PR, but a comment stating 10000000 is 100-ns would be useful to understand at a quick glance for readers, here:
https://github.com/Azure/azure-sdk-for-cpp/blob/9b29538f8a608d185186eff35a16ebfccad9672b/sdk/core/azure-core/inc/azure/core/datetime.hpp#L23-L24

Similar to:
https://github.com/Azure/azure-sdk-for-cpp/blob/9b29538f8a608d185186eff35a16ebfccad9672b/sdk/core/azure-core/inc/azure/core/datetime.hpp#L48-L49

sdk/core/azure-core/inc/azure/core/datetime.hpp
LarryOsterman, line 316:
```suggestion
   * @brief Provides convertion methods for Windows NT FILETIME to an #Azure::DateTime.
```

LarryOsterman: The name of the type is "FILETIME", not "filetime".

sdk/core/azure-core/inc/azure/core/datetime.hpp
LarryOsterman: See @JeffreyRichter's email. The NT typename is "FILETIME", so this converter is a TimeConverter which converts NT FILETIMES.

Would you rather Win32FILETIME instead of NTFILETIME?

ahsonkhan: The documentation frequently uses the term `file time` on its own. And some search term hits show people refer to it as Windows filetime. Probably Win32/NT is implied by the context:
https://learn.microsoft.com/en-us/windows/win32/sysinfo/file-times
https://stackoverflow.com/questions/6161776/convert-windows-filetime-to-second-in-unix-linux

That said though, just `FileTimeConverter` would be too generic and take-up a good name.

> Would you rather Win32FILETIME instead of NTFILETIME?

Yeah, between Win32 vs NT, I would lean more towards Win32 to mirror Posix rather than NT. That seems more approachable and self-descriptive to me, but you are way more familiar with this space, so feel free to make the call here.

**Options:**
- `Win32FileTimeConverter` - my preference
- `NTFileTimeConverter`

LarryOsterman: I'm actually happier with WIn32 than I am with NT - Although it is a purely historical artifact, Win32 is actually similar to Posix - it's a description of an API surface which can be implemented on multiple platforms (originally there was something called Win32s which allowed Win32 applications to run on Win16 for example). And Win95 ran Win32 apps just fine.

ahsonkhan (approved):
Other than the type name change, looks good to me.

sdk/core/azure-core/inc/azure/core/datetime.hpp
Jinming-Hu: @JeffreyRichter Can you make the call for the class name?

JeffreyRichter:
Sure, let's go with Larry's suggestion of Win32

Jinming-Hu:
I think all the comments have been resolved. If there's no more comment, I'll merge this PR tomorrow.

RickWinter (approved):

PR #3957, opened 2022-09-21
fix build errors on macOS

Jinming-Hu:
fixes https://github.com/Azure/azure-sdk-for-cpp/issues/3956

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/context.hpp
+++ b/sdk/core/azure-core/inc/azure/core/context.hpp
@@ -147,6 +147,14 @@ namespace Azure { namespace Core {
     Context() : m_contextSharedState(std::make_shared<ContextSharedState>()) {}
 
     /**
+     * @brief Copy-constructs a new context from another `%Context` instance.
+     *
+     * @param other Another instance to copy.
+     *
+     */
+    Context(const Context& other) = default;
+
+    /**
      * @brief Assigns `%Context` to another `%Context` instance.
      *
      * @param other Another instance to copy.


sdk/core/azure-core/inc/azure/core/context.hpp
```diff
@@ -146,6 +146,14 @@ namespace Azure { namespace Core {
      */
     Context() : m_contextSharedState(std::make_shared<ContextSharedState>()) {}
 
+    /**
+     * @brief Copy-constructs a new context from another `%Context` instance.
+     *
+     * @param other Another instance to copy.
+     *
+     */
+    Context(const Context& other) = default;
```
ahsonkhan, line 155: @CaseyCarter: If we need a copy ctor and copy assignment, should we add an explicit `default` destructor, to follow the 0-3-5 rule (and get to three)?
https://en.cppreference.com/w/cpp/language/rule_of_three

Or, is the better alternative to remove the copy assignment operator, since it is defaulted anyway and go to **zero**?

ahsonkhan:
@Jinming-Hu how come this build error is showing up now? We'd want the same build environment in CI to catch such an error going forward.

Also, add a changelog entry please for the bug fix.

Jinming-Hu:
> @Jinming-Hu how come this build error is showing up now? We'd want the same build environment in CI to catch such an error going forward.

This issue has been existing for quite a long time on my Mac. I always got around it by applying a patch locally. I thought there was something wrong with my environment (compiler version etc). Until recently a customer came to us with the same issue.

sdk/core/azure-core/inc/azure/core/context.hpp
LarryOsterman: FWIW, this item is currently also included in #3954 since I also ran into the exact same issue.

CaseyCarter: I would remove the explicitly-defaulted copy assignment (the only effect of which was to keep the compiler from defining a copy constructor) instead of adding an explicitly-defaulted copy constructor. There will still be a defaulted copy assignment - now implicit instead of explicit - and there will be a new implicit copy constructor as well. I greatly prefer 0 to 3 or 5 when it works, because it's the simplest/default behavior.

ahsonkhan: Would removing the explicitly defaulted copy assignment be considered a source breaking change? Or will it be opaque to most callers?

CaseyCarter: > Would removing the explicitly defaulted copy assignment be considered a source breaking change? Or will it be opaque to most callers?

100% non-breaking. This class will be every bit as copy-assignable as before, I'd expect all codegen for existing programs to be bit-for-bit identical.

LarryOsterman: FWIW, I made Casey's suggested fix to the #3954 pull request and it worked like a champ.

LarryOsterman:
FWIW, the work for this PR is also included in #3954, IMHO since this PR has no other changes, I think it should be abandoned in favor of #3954.

antkmsft (approved):
I'm for explicitly declaring default operations with `= default` or `= delete` even when this is the default behavior. Especially in cases when it is not immediately clear, like this one. So, between @LarryOsterman's (#3954) approach to remove the `...operator=(...) = default` declaration and this one to add `Context(Context const&) = default` declaration, I pick this one. Let's have more of `= default` and `= delete` - they can't hurt, they'll only make things clearer.

@Jinming-Hu, I highly recommend to add TEST(Context, OpSemantic) test to `context_test.cpp`: test the default constructor, copy constructor, and assignment. The fact that the test variable is created on a stack will be the evidence that there is a desctructor.
This way, we can be sure that we don't introduce a breaking change when we touch these operations, and their behavior.

Next step for us can be to also assert these in tests using `std::is_move_assignable` and so on from `<type_traits>`. But I'd put separate effort in this, as we should think it out well. And add it not just to this class, but at least for some others, if not all. I'd do it together with explicitly declaring the 6 default operations for each class with `=default` and `=delete`.

ahsonkhan:
> I'm for explicitly declaring default operations with `= default` or `= delete` even when this is the default behavior. 

Imo, that would be more work dev-work for little customer value (particularly around `=default`). We should let the compiler define things for us, as appropriate defaults, as long as it is intentional, and conveys the same intent (and like you said, the intended behavior is tested). It would be a bit cumbersome to review/understand type definitions having to ignore a bunch of boiler-plate lines that are redundant.

PR #3966, opened 2022-09-28
Revert "Add NTFileTimeTimeConverter"

Jinming-Hu:
Reverts Azure/azure-sdk-for-cpp#3941


Jinming-Hu:
@ahsonkhan I was looking at .Net implementation. .Net has this support in system library. https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.fromfiletime?view=net-6.0#system-datetimeoffset-fromfiletime(system-int64).

C++ standard library doesn't have datatime support. Instead we use the one in core SDK.

Do you think it's fine to keep this converter in core SDK? Or you want me to move the converter to storage SDK anyway?

antkmsft (approved):

PR #4005, opened 2022-10-08
datalakestg82/change IsServerEncrypted not nullable

microzchang:
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-files-datalake/swagger/README.md
+++ b/sdk/storage/azure-storage-files-datalake/swagger/README.md
@@ -258,6 +258,8 @@ directive:
     transform: >
       $["201"].headers["Content-Length"]["x-ms-client-name"] = "FileSize";
       $["201"].headers["Content-Length"]["x-nullable"] = true;
+      $["201"].headers["x-ms-request-server-encrypted"]["x-nullable"] = true;
+      $["201"].headers["x-ms-request-server-encrypted"]["x-ms-client-default"] = "bool()";
       $["201"].headers["x-ms-encryption-key-sha256"]["x-nullable"] = true;
       delete $["201"].headers["x-ms-continuation"];
       $["201"].schema = {
--- a/sdk/storage/azure-storage-files-datalake/swagger/README.md
+++ b/sdk/storage/azure-storage-files-datalake/swagger/README.md
@@ -258,7 +258,6 @@ directive:
     transform: >
       $["201"].headers["Content-Length"]["x-ms-client-name"] = "FileSize";
       $["201"].headers["Content-Length"]["x-nullable"] = true;
-      $["201"].headers["x-ms-request-server-encrypted"]["x-nullable"] = true;
       $["201"].headers["x-ms-encryption-key-sha256"]["x-nullable"] = true;
       delete $["201"].headers["x-ms-continuation"];
       $["201"].schema = {
@@ -436,7 +435,6 @@ directive:
       $["Content-MD5"]["x-nullable"] = true;
       $["x-ms-content-crc64"]["x-ms-client-name"] = "TransactionalContentHash";
       $["x-ms-content-crc64"]["x-nullable"] = true;
-      $["x-ms-request-server-encrypted"]["x-nullable"] = true;
       $["x-ms-encryption-key-sha256"]["x-nullable"] = true;
       delete $["ETag"];
 ```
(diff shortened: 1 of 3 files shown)


Jinming-Hu (commented):
They should be nullable, right? Can you check blob SDK?

I think we should only revert the one for AppendFile to avoid a breaking change.

microzchang:
> 



> They should be nullable, right? Can you check blob SDK?
> 
> I think we should only revert the one for AppendFile to avoid a breaking change.

microzchang:
> They should be nullable, right? Can you check blob SDK?
> 
> I think we should only revert the one for AppendFile to avoid a breaking change.

I have checked blob sdk and all of them are not nullable.

Jinming-Hu:
> > They should be nullable, right? Can you check blob SDK?
> > I think we should only revert the one for AppendFile to avoid a breaking change.
> 
> I have checked blob sdk and all of them are not nullable.

This is not good. Non-nullable booleans are initialzed to `false`. If customer is running the SDK with an older API verison, the value would be `false`, which apparently doesn't reflect the real status of the blob. We're lying to customers.

microzchang:
> > > They should be nullable, right? Can you check blob SDK?
> > > I think we should only revert the one for AppendFile to avoid a breaking change.
> > 
> > 
> > I have checked blob sdk and all of them are not nullable.
> 
> This is not good. Non-nullable booleans are initialzed to `false`. If customer is running the SDK with an older API verison, the value would be `false`, which apparently doesn't reflect the real status of the blob. We're lying to customers.
> 
> Can you do some testing here? For `Create`, `Append` and `Flush`, which of them return `x-ms-request-server-encrypted` header in STG 74, which of them start returning this header only after STG74.

I also considered about this. But if it is nullable in datalake and non-nullable in blob, I think it will cause some inconvenience to users.  Is it better to make it non-nullable if we can confirm that `x-ms-request-server-encrypted` should be returned?

And I have tested that all of these three APIs return `x-ms-request-server-encrypted`  header in STG 74 (which API version is 2020-02-10).

sdk/storage/azure-storage-files-datalake/swagger/README.md
```diff
@@ -259,6 +259,7 @@ directive:
       $["201"].headers["Content-Length"]["x-ms-client-name"] = "FileSize";
       $["201"].headers["Content-Length"]["x-nullable"] = true;
       $["201"].headers["x-ms-request-server-encrypted"]["x-nullable"] = true;
+      $["201"].headers["x-ms-request-server-encrypted"]["x-ms-client-default"] = "bool()";
```
Jinming-Hu, line 262: Is this superfluous? What would the code look like without this statement?

microzchang: If without this statement,  the field will be nullable like this:
```
      Nullable<bool> IsServerEncrypted;
```
In the create API,it will include `Create File` |` Create Directory` | `Rename File` | `Rename Directory`.
`Create File` |` Create Directory` scenarios return `IsServerEncrypted` to customer.
`Rename File` | `Rename Directory` scenarios don't return `IsServerEncrypted` from REST API and in the convinence layer we just return renamedClient to customer.

Jinming-Hu: What if you remove Line 261 and 262?

microzchang: It will change the code of rest_client.cpp.
before removing  Line 261 and 262, the code is:
```
      if (pRawResponse->GetHeaders().count("x-ms-request-server-encrypted") != 0)
      {
        response.IsServerEncrypted
            = pRawResponse->GetHeaders().at("x-ms-request-server-encrypted") == std::string("true");
      }
```
after removing, the code is: 
```
 response.IsServerEncrypted
          = pRawResponse->GetHeaders().at("x-ms-request-server-encrypted") == std::string("true");
```
Because `Rename` scenario doesn't return `IsServerEncrypted`, removing these lines will cause an exception.

Jinming-Hu: Ok, got it.

Jinming-Hu (approved):

PR #4015, opened 2022-10-11
Converted WinHTTP to Async.

LarryOsterman:
# Converted WinHTTP to async operation.

Converted the HttpRequest object to be asynchronous. Split all the functionality related to an HttpRequest to a new type `WinHttpRequest` which encapsulates all the HTTP operations.

Along with the `WinHttpRequest` type, there's also a `WinHttpAction` which reflects a WinHTTP API call. 

## Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


sdk/core/azure-core/CMakeLists.txt
```diff
@@ -54,10 +54,11 @@ endif()
 if(BUILD_TRANSPORT_WINHTTP)
   SET(WIN_TRANSPORT_ADAPTER_SRC 
     src/http/winhttp/win_http_transport.cpp
+	src/http/winhttp/win_http_request.hpp
```
antkmsft, line 57:
```suggestion
    src/http/winhttp/win_http_transport.cpp
    src/http/winhttp/win_http_request.hpp
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,311 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
 
-  // We will only set the Status callback if a root certificate has been set.
-  AZURE_ASSERT(!m_options.ExpectedTlsRootCertificates.empty());
+    try
+    {
+      WinHttpAction* httpAction = reinterpret_cast<WinHttpAction*>(dwContext);
+      httpAction->OnHttpStatusOperation(
+          hInternet, internetStatus, statusInformation, statusInformationLength);
+    }
+    catch (Azure::Core::RequestFailedException& rfe)
```
antkmsft, line 502:
```suggestion
    catch (Azure::Core::RequestFailedException const& rfe)
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,311 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+    {
+      // If an exception is thrown in the handler, log the error and terminate the connection.
+      Log::Write(
+          Logger::Level::Error,
+          "Request Failed Exception Thrown: " + std::string(rfe.what()) + rfe.Message);
+      WinHttpCloseHandle(hInternet);
+    }
+    catch (std::exception& ex)
```
antkmsft, line 510:
```suggestion
    catch (std::exception const& ex)
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -766,76 +1003,107 @@ void WinHttpTransport::SendRequest(
...
+          "When uploading data, the body stream must have a known length.");
+    }
 
-  if (streamLength > 0)
+    if (streamLength > 0)
+    {
+      Upload(request, context);
+    }
+  }
+  catch (TransportException&)
```
antkmsft, line 1065:
```suggestion
  catch (TransportException const&)
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -563,6 +740,8 @@ WinHttpTransport::WinHttpTransport(
 {
 }
 
+WinHttpTransport::~WinHttpTransport() {}
```
antkmsft, line 743:
```suggestion
WinHttpTransport::~WinHttpTransport() = default;
```

antkmsft (approved):
I recommend to get a review from Ahson, he wrote the initial version of the WinHTTP adapter.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+#include <memory>
+#include <mutex>
+#include <wil\resource.h>
+#include <winhttp.h>
+
+namespace Azure { namespace Core { namespace Http { namespace _detail {
+
+  class WinHttpRequest;
+  /**
+   * @brief An outstanding WinHTTP action. This object is used process asynchronous WinHttp actions.
```
antkmsft, line 24:
```suggestion
   * @brief An outstanding WinHTTP action. This object is used process asynchronous WinHTTP actions.
```

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+   *
+   * The WinHttpRequest object has a WinHttpAction associated it to convert asynchronous winhttp
```
antkmsft, line 26:
```suggestion
   * The WinHttpRequest object has a WinHttpAction associated it to convert asynchronous WinHTTP
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,311 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+      std::function<void()> callback,
+      Azure::Core::Context const& context,
+      Azure::DateTime::duration const& pollInterval)
   {
-    WinHttpTransport* httpTransport = reinterpret_cast<WinHttpTransport*>(dwContext);
-    httpTransport->OnHttpStatusOperation(hInternet, dwInternetStatus);
+    context.ThrowIfCancelled();
+    ResetEvent(m_actionCompleteEvent.get());
+
+    // Call the provided callback to start the WinHttp action.
```
antkmsft, line 420:
```suggestion
    // Call the provided callback to start the WinHTTP action.
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: Is there any benefit to making this a const reference? Does it change the generated code in any way?

I'm not a huge fan of adding const to variable declarations unless there are specific semantic benefits (for instance, declaring a parameter as const means that the caller knows that the underlying function won't modify the value of the parameter so it can cache values across the method call). I'm not aware of a similar benefit for local variables.

LarryOsterman: Deleted previous comment after digging in a bit.

LarryOsterman:
@ahsonkhan Would you mind taking a look at this?

LarryOsterman:
Ping. I would appreciate @ahsonkhan's feedback on this one.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+        DWORD statusInformationLength);
+
+  public:
+    /**
+     * @brief Create a new WinHttpAction object associated with a specific WinHttpRequest.
+     *
+     * @param request Http Request associated with the action.
+     */
+    WinHttpAction(WinHttpRequest* request)
+        : m_httpRequest(request), m_actionCompleteEvent(CreateEvent(nullptr, TRUE, FALSE, nullptr))
```
ahsonkhan, line 70: maintainability nit: consider adding a comment describing why we are passing these options to `CreateEvent`

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+    {
+      if (!m_actionCompleteEvent)
+      {
+        throw std::runtime_error("Error creating Action Complete Event.");
```
ahsonkhan, line 74: What's an `Action Complete` event?

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+      }
+    }
+    /**
+     * @brief WaitForAction - Waits for an action to complete.
+     *
+     * @remarks The WaitForAction method waits until an action initiated by the `callback` function
+     * has completed. It polls with a timeout of milliseconds, checking to see if the provided
```
ahsonkhan, line 81: I read `It polls with a timeout of milliseconds` as, it keeps polling for some time/at a frequency of a particular interval and then stops due to the timeout value. How should I read this?

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+     *
+     * @returns true if the action completed normally, false if there was an error.
+
+     * @remarks If there is an error, the caller can determine the error code by calling
+     * GetStowedError() and GetStowedErrorInformation()
+     */
+    bool WaitForAction(
+        std::function<void()> callback,
+        Azure::Core::Context const& context,
+        Azure::DateTime::duration const& pingDuration = std::chrono::milliseconds(800));
```
ahsonkhan, line 96: Why is it defaulted to 800 ms? Is that supposed to be the same value as `RetryDelay`:
https://github.com/Azure/azure-sdk-for-cpp/blob/be4cb8f51639be2a36a3f47e014ff4508cc85f69/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp#L93

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * @brief Notify a caller that the action has completed successfully.
+     */
+    void CompleteAction();
```
ahsonkhan, line 101: Take this suggestion with a grain of salt (since I don't fully understand the implementations and caller/callee semantics for this function), but should it be renamed as the following, given the doc comment:
```suggestion
    void ActionCompleted();
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,314 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
-    DWORD dwInternetStatus,
-    LPVOID,
-    DWORD) noexcept
+namespace {
+
+#define APPEND_ENUM_STRING(id) \
+  if (internetStatus & (id)) \
+  { \
+    rv += std::string(#id) + " "; \
+  }
```
ahsonkhan, line 338: We try to avoid macros. Consider updating to a named (inline) function.

https://azure.github.io/azure-sdk/cpp_implementation.html#macros

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,314 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
-void WinHttpTransport::StatusCallback(
-    HINTERNET hInternet,
-    DWORD_PTR dwContext,
-    DWORD dwInternetStatus,
-    LPVOID,
-    DWORD) noexcept
+namespace {
+
+#define APPEND_ENUM_STRING(id) \
+  if (internetStatus & (id)) \
```
ahsonkhan, line 335: What is this line doing?
Is the type of `id` another "Windows integer type"? Looks like bitwise `&` of a DWORD and another DWORD. Why are we doing that?

Is that the type of the different `WINHTTP_STATUS_CALLBACK` values?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,314 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_DATA_AVAILABLE);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_READ_COMPLETE);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_WRITE_COMPLETE);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_REQUEST_ERROR);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_SENDREQUEST_COMPLETE);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_GETPROXYFORURL_COMPLETE);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_CLOSE_COMPLETE);
+  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_SHUTDOWN_COMPLETE);
+  // For <reasons> this is not defined on the Win2022 Azure DevOps image, so manually expand it.
+  //  APPEND_ENUM_STRING(WINHTTP_CALLBACK_STATUS_GETPROXYSETTINGS_COMPLETE);
```
ahsonkhan, line 368: Are we intentionally leaving this line commented out, even though line 371 is doing the "custom" append?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,314 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+  if (internetStatus & 0x08000000)
   {
-    return;
+    rv += "WINHTTP_CALLBACK_STATUS_GETPROXYSETTINGS_COMPLETE ";
```
ahsonkhan, line 371: nit: to be explicit/consistent with the macro coding pattern, and avoid someone thinking the leading space is a typo
```suggestion
    rv += "WINHTTP_CALLBACK_STATUS_GETPROXYSETTINGS_COMPLETE" + " ";
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,314 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
 
-  try
+std::string GetErrorMessage(DWORD error)
+{
+  std::string errorMessage = " Error Code: " + std::to_string(error);
+
+  char* errorMsg = nullptr;
+  if (FormatMessage(
+          FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_FROM_HMODULE | FORMAT_MESSAGE_ALLOCATE_BUFFER,
+          GetModuleHandle("winhttp.dll"),
```
ahsonkhan, line 386: Is it guaranteed that all error messages will come from `winhttp.dll`?

ahsonkhan (commented):
Some feedback and questions, especially around the header file restructuring, while trying to understand the async/callback approach.

Didn't get a chance to fully review of the `win_http_transport.cpp` implementation yet.

sdk/core/azure-core/test/ut/azure_core_test.cpp
```diff
@@ -21,7 +17,25 @@ int main(int argc, char** argv)
...
+#if defined(_DEBUG) && defined(_MSC_VER)
+  _CrtSetReportMode(_CRT_ERROR, _CRTDBG_MODE_DEBUG);
+#endif
+
+  signal(SIGABRT, [](int) {
+    try
+    {
+      throw;
+    }
+    catch (std::exception const& ex)
```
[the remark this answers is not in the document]
LarryOsterman: This rethrows and then catches whatever exception has not been unwound. It works around the MSVC behavior of popping up UI when an exception is thrown.

abort() is called from terminate(), which is called when an exception is not caught. This functions as a top-level exception handler. It's highly compiler dependent but it works for the compilers we support.

sdk/core/azure-core/test/ut/transport_adapter_base_test.cpp
```diff
@@ -311,7 +311,7 @@ namespace Azure { namespace Core { namespace Test {
     // Test that calling getValue again will return empty
     result = std::move(responseT.Value);
     EXPECT_STREQ(result.data(), expectedType.data());
-    result = responseT.Value;
+    result = responseT.Value; // Not 100% sure what this is testing - that std::move works?
```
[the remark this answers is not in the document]
LarryOsterman: As best as I can figure, this code is verifying that the compiler generated correct code, and IMHO we shouldn't have tests for the compiler generating correct code unless we have evidence that the compiler is NOT generating correct code.

sdk/core/azure-core/test/ut/transport_adapter_base_test.cpp
```diff
@@ -376,6 +376,18 @@ namespace Azure { namespace Core { namespace Test {
     t1.join();
   }
 
+  TEST_P(TransportAdapter, cancelRequest)
```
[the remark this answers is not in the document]
LarryOsterman: Literally the entire reason for this pull request.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: Yes, I intentionally left it commented out - the preferred incantation of this is the commented out version, this is a workaround for a header issue.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: No, many of them come from kernel32.dll. But the winhttp errors are in winhttp.dll, if we don't add it here, we won't find those errors

sdk/core/azure-core/inc/azure/core/http/win_http_transport.hpp
```diff
@@ -266,8 +171,8 @@ namespace Azure { namespace Core {
       WinHttpTransport(Azure::Core::Http::Policies::TransportOptions const& options);
 
       /**
-       * @brief Implements the HTTP transport interface to send an HTTP Request and produce an HTTP
-       * RawResponse.
+       * @brief Implements the HTTP transport interface to send an HTTP Request and produce an
+       * HTTP RawResponse.
```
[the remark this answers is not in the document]
LarryOsterman: IIRC this happened because the indentation in win_http_transport changed.

sdk/core/azure-core/inc/azure/core/http/win_http_transport.hpp
```diff
@@ -279,7 +184,7 @@ namespace Azure { namespace Core {
       // [Core Guidelines C.35: "A base class destructor should be either public
       // and virtual or protected and
       // non-virtual"](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-non-virtual)
-      virtual ~WinHttpTransport() = default;
+      virtual ~WinHttpTransport();
```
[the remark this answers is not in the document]
LarryOsterman: The constructor for the WinHTTP transport is no longer defaulted, so I had to remove the = default;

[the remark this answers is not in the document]
LarryOsterman: And the reason I didn't leave it inline is that it would bring in a TON of Win32 specific headers that had no business being in a public header file.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+// cspell:words PCCERT HCERTSTORE
+
+/**
+ * @file
+ * @brief #Azure::Core::Http::HttpTransport request support classes.
```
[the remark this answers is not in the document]
LarryOsterman: The reason these types were split out from the public header file is (a) because these are implementation details and are never used by clients of the sdk and (b) including them in the public header requires that I add a TON of WIn32 specific functionality to the public API surface, none of which is necessary.

Splitting them out in a separate PR would be extremely challenging because I would have to add a ton of Win32 specific code to this PR and then immediately remove it. 

It's easier to keep our public API surface clean.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,213 @@
...
+  class WinHttpRequest;
+  /**
+   * @brief An outstanding WinHTTP action. This object is used process asynchronous WinHTTP actions.
+   *
+   * The WinHttpRequest object has a WinHttpAction associated it to convert asynchronous WinHTTP
+   * operations to synchronous operations.
+   *
+   */
+  class WinHttpAction final {
+    friend class WinHttpRequest;
```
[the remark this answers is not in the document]
LarryOsterman: Added a comment explaining this.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
LarryOsterman: It's an event set to the signalled state when the current action completes.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
LarryOsterman: I've improved the comment - the word "pingDuration" was missing between "of" and "milliseconds".

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
LarryOsterman: I needed a polling duration as some number between 0 and 10 minutes (10 minutes is obviously too long, 0 is too short for HTTP), 800ms seemed like a reasonable value because it tracks the retry delay.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
LarryOsterman: `CompleteAction` is an active event - it tells the waiting caller that it's ok to proceed. It's not a passive notification (which is implied by `ActionCompleted`.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: The status callback for WinHTTP is one of many #defined values. rather than printing out 0x1000 to the debugger, I thought it was more readable to print out `WINHTTP_CALLBACK_STATUS_DETECTING_PROXY`. Technically these are bitfields, so I wrote the InternetStatusTOString function to treat them as a bitmap.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: Named functions don't have the tokenizing operator (`#`).  The macro depends on that functionality.

sdk/core/azure-core/test/ut/transport_adapter_base_test.hpp
```diff
@@ -56,6 +56,11 @@ namespace Azure { namespace Core { namespace Test {
       return std::string(_detail::AzureSdkHttpbinServerSchema) + "://"
           + std::string(_detail::AzureSdkHttpbinServer) + "/patch";
     }
+    inline static std::string Delay()
+    {
+      return std::string(_detail::AzureSdkHttpbinServerSchema) + "://"
+          + std::string(_detail::AzureSdkHttpbinServer) + "/delay";
```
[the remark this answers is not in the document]
LarryOsterman: Wow that's WAY better.

LarryOsterman:
Ping @ahsonkhan - I believe that I've addressed your concerns.

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
```diff
@@ -10,6 +11,26 @@ int main(int argc, char** argv)
   // End users need to decide if SIGPIPE should be ignored or not.
   signal(SIGPIPE, SIG_IGN);
 #endif
+  // Declare a signal handler to report unhandled exceptions on Windows - this is not needed for
+  // other
+// OS's as they will print the exception to stderr in their terminate() function.
```
RickWinter, line 16:
```suggestion
  // OS's as they will print the exception to stderr in their terminate() function.
```

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
```diff
@@ -10,6 +11,26 @@ int main(int argc, char** argv)
...
+#endif
+
+  signal(SIGABRT, [](int) {
+    try
+    {
+      throw;
+    }
+    catch (std::exception const& ex)
+    {
+      std::cout << "Exception thrown: " << ex.what() << std::endl;
```
RickWinter, line 30:
```suggestion
      std::cout << "SIGABRT RAISED: " << ex.what() << std::endl;
```

sdk/core/azure-core/test/ut/azure_core_test.cpp
```diff
@@ -21,7 +17,29 @@ int main(int argc, char** argv)
...
+    // thrown so we can catch them and report them to the caller. This is needed because the
+    // terminate() function on Windows calls abort() which normally pops up UI terminates without
+    // reporting the exception.
+    try
+    {
+      throw;
+    }
+    catch (std::exception const& ex)
+    {
+      std::cout << "Exception thrown: " << ex.what() << std::endl;
```
RickWinter, line 39:
```suggestion
      std::cout << "SIGABRT RAISED: " << ex.what() << std::endl;
```

sdk/core/azure-core/test/ut/azure_libcurl_core_main_test.cpp
```diff
@@ -57,6 +58,31 @@ namespace Azure { namespace Core { namespace Test {
...
+    {
+      // Rethrow any exceptions on the current stack - this will cause any pending exceptions to be
+      // thrown so we can catch them and report them to the caller. This is needed because the
+      // terminate() function on Windows calls abort() which normally pops up UI terminates without
+      // reporting the exception.
+      throw;
+    }
+    catch (std::exception const& ex)
+    {
+      std::cout << "Exception thrown: " << ex.what() << std::endl;
```
RickWinter, line 81: I'm rarely a fan of MACROS and our guidelines discourage the use. However, this code is copied in multiple different and could be consolidated into a single header that we include which could then have the macro inlined in each location.

```suggestion
      std::cout << "SIGABRT RAISED: " << ex.what() << std::endl;
```

sdk/core/azure-core/test/ut/transport_adapter_base_test.hpp
```diff
@@ -26,36 +26,13 @@ namespace Azure { namespace Core { namespace Test {
...
-          + std::string(_detail::AzureSdkHttpbinServer) + "/delete";
-    }
-    inline static std::string Patch()
-    {
-      return std::string(_detail::AzureSdkHttpbinServerSchema) + "://"
-          + std::string(_detail::AzureSdkHttpbinServer) + "/patch";
-    }
+    inline static std::string Get() { return Schema() + "://" + Host() + "/get"; }
+    inline static std::string Headers() { return Schema() + "://" + Host() + "/headers"; }
+    inline static std::string WithPort() { return Schema() + "://" + Host() + ":443/get"; }
```
RickWinter, line 31:
```suggestion
    inline static std::string GetWithPort() { return Schema() + "://" + Host() + ":443/get"; }
```

RickWinter, line 31: Is there a reason we care about the secure version of get and only get?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -12,7 +12,9 @@
 
 #if defined(BUILD_TRANSPORT_WINHTTP_ADAPTER)
 #include "azure/core/http/win_http_transport.hpp"
+#include "win_http_request.hpp"
```
RickWinter, line 15:
```suggestion
#include "azure/core/http/win_http_request.hpp"
```

LarryOsterman: win_http_request is in the same directory as win_http_transport.cpp. So that would have to be something different. Theoretically I could put it in a "private" subdirectory but it would be rather lonely there.

sdk/core/azure-core/test/ut/transport_adapter_base_test.hpp
LarryOsterman: I like the change, but it's bigger than this one line.

sdk/core/azure-core/test/ut/azure_libcurl_core_main_test.cpp
LarryOsterman: Let me see if I can come up with something.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,320 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+} // namespace
+
+namespace Azure { namespace Core { namespace Http { namespace _detail {
+
+  /**
+   * Wait for an action to complete.
+   *
+   */
+  bool WinHttpAction::WaitForAction(
+      std::function<void()> callback,
```
RickWinter, line 414: Can we specify the signature for the callback?  Would be nice to enforce we only get a callback to an expected signature?

Who can call this method?  Given we are executing random code via the callback, it would be good to lock down as much as possible.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
RickWinter: Oh I see, ignore my comment.  Didn't realize this is a private header.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,320 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+        // of data read.
+        CompleteActionWithData(statusInformationLength);
+      }
+      else if (internetStatus == WINHTTP_CALLBACK_STATUS_HANDLE_CLOSING)
+      {
+        // An HINTERNET handle is closing, complete the outstanding close request.
+        Log::Write(Logger::Level::Verbose, "Closing handle; completing outstanding Close request");
+        CompleteAction();
+      }
     }
```
RickWinter, line 593: Is it expected that we'd ever fall out of the else without completing the action?  At a minimum we can log the event.

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
LarryOsterman: Code is no longer there, so this change is unnecessary.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: The signature for the callback is "`void()`" It's in the `std::function` declaration. 

The header declaration for WaitForAction has more details about the callback function. And now that I think about it, the name is *horrible*. Let me change it to "initiateAction", which is more accurate.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: We log each event at the start of the `OnHttpStatusOperation` function. But you're right, there is a programming error where you set the expected status but don't actually handle the expected status - it makes sense to log this because it's almost certainly going to be a problem.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +840,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
-    std::wstring proxyWide{StringToWideString(m_options.ProxyInformation)};
+    std::wstring proxyWide{StringToWideString(options.ProxyInformation)};
     proxyInfo.dwAccessType = WINHTTP_ACCESS_TYPE_NAMED_PROXY;
     proxyInfo.lpszProxy = const_cast<LPWSTR>(proxyWide.c_str());
     proxyInfo.lpszProxyBypass = WINHTTP_NO_PROXY_BYPASS;
-    if (!WinHttpSetOption(request.get(), WINHTTP_OPTION_PROXY, &proxyInfo, sizeof(proxyInfo)))
+    if (!WinHttpSetOption(
+            m_requestHandle.get(), WINHTTP_OPTION_PROXY, &proxyInfo, sizeof(proxyInfo)))
     {
       GetErrorAndThrow("Error while setting Proxy information.");
```
RickWinter, line 869: Ditto on the `GetLastError()` as part of the exception message.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +840,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
+            m_requestHandle.get(),
             WINHTTP_AUTH_TARGET_PROXY,
             WINHTTP_AUTH_SCHEME_BASIC,
-            StringToWideString(m_options.ProxyUserName.Value()).c_str(),
-            StringToWideString(m_options.ProxyPassword.Value()).c_str(),
+            StringToWideString(options.ProxyUserName.Value()).c_str(),
+            StringToWideString(options.ProxyPassword.Value()).c_str(),
             0))
     {
       GetErrorAndThrow("Error while setting Proxy credentials.");
```
RickWinter, line 882: Ditto on the `GetLastError()` as part of the exception message.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +840,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
 
-  if (m_options.IgnoreUnknownCertificateAuthority || !m_options.ExpectedTlsRootCertificates.empty())
+  if (options.IgnoreUnknownCertificateAuthority || !options.ExpectedTlsRootCertificates.empty())
   {
     auto option = SECURITY_FLAG_IGNORE_UNKNOWN_CA;
-    if (!WinHttpSetOption(request.get(), WINHTTP_OPTION_SECURITY_FLAGS, &option, sizeof(option)))
+    if (!WinHttpSetOption(
+            m_requestHandle.get(), WINHTTP_OPTION_SECURITY_FLAGS, &option, sizeof(option)))
     {
       GetErrorAndThrow("Error while setting ignore unknown server certificate.");
```
RickWinter, line 892: Ditto on the `GetLastError()` as part of the exception message.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +840,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
 
-  if (m_options.EnableCertificateRevocationListCheck)
+  if (options.EnableCertificateRevocationListCheck)
   {
     DWORD value = WINHTTP_ENABLE_SSL_REVOCATION;
-    if (!WinHttpSetOption(request.get(), WINHTTP_OPTION_ENABLE_FEATURE, &value, sizeof(value)))
+    if (!WinHttpSetOption(
+            m_requestHandle.get(), WINHTTP_OPTION_ENABLE_FEATURE, &value, sizeof(value)))
     {
       GetErrorAndThrow("Error while enabling CRL validation.");
```
RickWinter, line 902: Ditto on the `GetLastError()` as part of the exception message.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +840,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
+  m_httpAction = std::make_unique<_detail::WinHttpAction>(this);
+
+  if (WinHttpSetStatusCallback(
+          m_requestHandle.get(),
+          &WinHttpAction::StatusCallback,
+          WINHTTP_CALLBACK_FLAG_ALL_NOTIFICATIONS,
+          0)
+      == WINHTTP_INVALID_STATUS_CALLBACK)
+  {
+    GetErrorAndThrow("Error while setting up the status callback.");
```
RickWinter, line 916: Ditto on the `GetLastError()` as part of the exception message.

sdk/core/azure-core/inc/azure/core/internal/diagnostics/global_exception.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <exception>
+#include <iostream>
+
+namespace Azure { namespace Core { namespace Diagnostics { namespace _internal {
+  /**
+   * @brief Global Exception handler used for test collateral. This is not intended to be used by
+   * any non-test code.
+   *
+   * The GlobalExceptionHandler class is used to catch any unhandled exceptions and report them as a
+   * part of test collateral, it is intended to be called as a SIGBRT handler set by:
```
RickWinter, line 14:
```suggestion
   * part of test collateral, it is intended to be called as a SIGABRT handler set by:
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -595,10 +783,24 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateConnecti
...
-Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestHandle(
+void _detail::WinHttpRequest::EnableWebSocketsSupport()
+{
+#pragma warning(push)
+  // warning C6387: _Param_(3) could be '0'.
+#pragma warning(disable : 6387)
+  if (!WinHttpSetOption(m_requestHandle.get(), WINHTTP_OPTION_UPGRADE_TO_WEB_SOCKET, nullptr, 0))
+#pragma warning(pop)
+  {
+    GetErrorAndThrow("Error while Enabling WebSocket upgrade.");
```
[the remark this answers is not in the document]
LarryOsterman: `GetErrorAndThrow` has a defaulted `errorCode` parameter which is defaulted to `GetLastError()`. Everywhere you see `GetErrorAndThrow`, you can assume that the error code is included in the output.

Originally `GetErrorAndThrow` used to call `GetLastError()` internally (which is stupidly fragile), for the initial websocket work, I changed it to make the error code parameter a new parameter defaulted to `GetLastError` (because I didn't want to change the existing signature).

sdk/core/azure-core/test/ut/transport_adapter_base_test.hpp
LarryOsterman: Oh, and `Schema()` is `https`, so it's secure. YOu could ask why don't we test the *non* secure version, but the test didn't do that earlier.

This was intended as a textual cleanup of the class suggested by Ahson, and IMHO it's a significant improvement.

ahsonkhan:
@LarryOsterman, if you haven't done so, this change warrants performance testing, especially some basic storage perf test runs. We should compare baseline (main) to this PR and curl vs win http to verify that the performance is comparable.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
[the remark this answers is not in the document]
ahsonkhan: > Originally `GetErrorAndThrow` used to call `GetLastError()` internally (which is stupidly fragile)...

For my edification, can you please explain why it is fragile?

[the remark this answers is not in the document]
LarryOsterman: When GetErrorAndThrow calls GetLastError to discover the error code, it means that it can only work if the caller has called SetLastError before calling GetErrorAndThrow.

That is extremely fragile because Win32 errors can come from functions which don't call `SetLastError`. By making the error code a parameter (defaulted to `GetLastError`), it means that GetErrorAndThrow can be used both when a Win32 API returns a status code via SetLastError AND when the Win32 API explicitly returns the status code.

See also [this](https://www.youtube.com/watch?v=1NqmvFq016Y) for more context on the `SetLastError` vs return and error code issue.

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
ahsonkhan: This would be worth stating as a comment :)

LarryOsterman:
> > @LarryOsterman, if you haven't done so, this change warrants performance testing, especially some basic storage perf test runs. We should compare baseline (main) to this PR and curl vs win http to verify that the performance is comparable.
> > A sanity check using steps from here would be good: [#1098 (comment)](https://github.com/Azure/azure-sdk-for-cpp/issues/1098#issuecomment-1075754237)
> 
> In the absence of a performance pipeline (which can provide more standardized results), I have done some ad-hoc performance testing on my machine and the differences in performance are basically in the noise (over the course of the unit tests, there is a slight performance degradation but it's not clear if that is variation within the tests or something else)).
> 
> When I run the existing HTTP performance tests, the new code is basically the same as the old.

And to confirm: I'm seeing a slight improvement when running the storage tests as per #1098.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
ahsonkhan: @CaseyCarter is there a way to write this implementation in a named-function (relatively cleanly), which doesn't rely on the stringizing operator which is only available in a macro?
https://learn.microsoft.com/en-us/cpp/preprocessor/stringizing-operator-hash?view=msvc-170

If not, we may need to update the guidelines, because this use-case of a macro isn't listed as one of the allowed exceptions in our guidelines.

LarryOsterman: This macro is intended for diagnostic purposes, which I believe makes it squarely within the guidelines. It is not used as a part of any public API contract.

LarryOsterman:
Pretty ping :).

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
ahsonkhan: Hmm, I don't think that was the intent of that guideline, but we can make it fit :)

@BillyONeal, since you wrote these guidelines initially (here: https://github.com/Azure/azure-sdk/pull/1146),  can you expand on this a bit? What do you mean by emission/suppression of diagnostic.

Btw Larry, the "Implementation" page of the guidelines are in fact about the Implementation and not just the public API contract. So, there are parts of it that mention coding patterns to adhere to, that is only implementation detail.

P.S. Unresolving only to make the conversation easier to find for others (not to block the merge), feel free to resolve and merge upon response/approval.

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
```diff
@@ -10,6 +14,17 @@ int main(int argc, char** argv)
   // End users need to decide if SIGPIPE should be ignored or not.
   signal(SIGPIPE, SIG_IGN);
 #endif
+  // Declare a signal handler to report unhandled exceptions on Windows - this is not needed for
+  // other
+// OS's as they will print the exception to stderr in their terminate() function.
```
ahsonkhan, line 19: The suggestion to indent the comment still applies. But also, the line above just has "// other" on it, so there is an explicit line break that needs to be removed.

```suggestion
  // other OS's as they will print the exception to stderr in their terminate() function.
```

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
ahsonkhan: Looks like the suggestion to indent the comment still applies.
See this comment:
https://github.com/Azure/azure-sdk-for-cpp/pull/4015#discussion_r1002037939

sdk/core/azure-core/inc/azure/core/internal/diagnostics/global_exception.hpp
```diff
@@ -0,0 +1,39 @@
...
+   * ```
+   *
+   */
+  struct GlobalExceptionHandler
+  {
+    static void HandleSigAbort(int)
+    {
+      // Rethrow any exceptions on the current stack - this will cause any pending exceptions to
+      // be thrown so we can catch them and report them to the caller. This is needed because the
+      // terminate() function on Windows calls abort() which normally pops up UI terminates
```
ahsonkhan, line 27: nit: wording?
```suggestion
      // terminate() function on Windows calls abort() which normally pops up a UI and terminates
```

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,223 @@
...
+      {
+        throw std::runtime_error("Error creating Action Complete Event.");
+      }
+    }
+    /**
+     * @brief WaitForAction - Waits for an action to complete.
+     *
+     * @remarks The WaitForAction method waits until an action initiated by the `callback` function
+     * has completed. Every pollDuration milliseconds, it checks to see if the context specified for
+     * therequest has been cancelled (or times out).
```
ahsonkhan, line 87:
```suggestion
     * the request has been cancelled (or times out).
```

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,223 @@
...
+   * @brief An outstanding WinHTTP action. This object is used to process asynchronous WinHTTP
+   * actions.
+   *
+   * The WinHttpRequest object has a WinHttpAction associated with it to convert asynchronous
+   * WinHTTP operations to synchronous operations.
+   *
+   */
+  class WinHttpAction final {
+    // The WinHttpRequest class is friend because it needs to have access to StatusCallback when it
+    // registers for notifications, but there's no other reason to make it public.
```
ahsonkhan, line 34: In that case, given this whole header is private, I would consider marking `StatusCallback()` public, so that `WinHttpRequest` only has access to that targeted method, and doesn't have access to all of the private methods and members of `WinHttpAction`. That seems like a much smaller access scope between the two types.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: github hides this conversation even if it's unresolved.

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
ahsonkhan, line 19: I am not sure what's happening with this comment, lol, in the GitHub view. You may want to fix it up manually.

It shouldn't have any indentation since it is not within any #if or other scope/block.

Please fix here/elsewhere it is duplicated.

sdk/core/azure-core/test/ut/azure_core_test.cpp
```diff
@@ -21,7 +16,16 @@ int main(int argc, char** argv)
   signal(SIGPIPE, SIG_IGN);
 #endif
 
-  signal(SIGABRT, AbortHandler);
+  // Declare a signal handler to report unhandled exceptions on Windows - this is not needed for
+  // other OS's as they will print the exception to stderr in their terminate() function.
```
ahsonkhan, line 20: nit, indentation not needed:
```suggestion
// Declare a signal handler to report unhandled exceptions on Windows - this is not needed for
// other OS's as they will print the exception to stderr in their terminate() function.
```

sdk/core/azure-core/test/ut/azure_libcurl_core_main_test.cpp
```diff
@@ -57,6 +57,18 @@ namespace Azure { namespace Core { namespace Test {
 
 int main(int argc, char** argv)
 {
+  // Declare a signal handler to report unhandled exceptions on Windows - this is not needed for
+  // other
+// OS's as they will print the exception to stderr in their terminate() function.
```
ahsonkhan, line 62: Same here, please fix.

sdk/core/azure-core/test/ut/transport_policy_options.cpp
```diff
@@ -527,10 +527,11 @@ namespace Azure { namespace Core { namespace Test {
   TEST_F(TransportAdapterOptions, MultipleCrlOperations)
   {
     std::vector<std::string> testUrls{
-        AzureSdkHttpbinServer::Get(),
-        "https://twitter.com/",
-        "https://www.example.com/",
-        "https://www.google.com/",
+        AzureSdkHttpbinServer::Get(), // Uses a Microsoft/DigiCert certificate.
+        //        "https://www.wikipedia.org", // Uses a LetsEncrypt Certificate
```
ahsonkhan, line 531: Why we are removing twitter? And why add a commented out Wikipedia url?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: I understand that heavy use of macros can be challenging, especially macros which hide significant amounts of code. And it can be tricky to write macros correctly. However, macros can also be used to great effect, especially when emitting diagnostics.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+   * Wait for an action to complete.
+   *
+   */
+  bool WinHttpAction::WaitForAction(
+      std::function<void()> initiateAction,
+      DWORD expectedCallbackStatus,
+      Azure::Core::Context const& context,
+      Azure::DateTime::duration const& pollInterval)
+  {
+    context.ThrowIfCancelled();
```
ahsonkhan, line 431: Why are we checking this here? Comment?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+              std::chrono::duration_cast<std::chrono::milliseconds>(pollInterval).count()));
+      if (waitResult == WAIT_TIMEOUT)
+      {
+        context.ThrowIfCancelled();
+      }
+      else if (waitResult != WAIT_OBJECT_0)
+      {
+        return false;
+      }
+    } while (waitResult != WAIT_OBJECT_0);
```
ahsonkhan, line 456: Interesting, how come the "else-if return" and "while-loop" conditions, identical?

Just a nit: The way the code is written here, I have to really think about how the control flow changes for different `waitResult` values. Consider writing differently.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+    if (m_stowedError != NO_ERROR)
```
ahsonkhan, line 457: Am I correct in assuming that `m_stowedError` is being modified by another method (only `CompleteActionWithError`), while we `WaitForSingleObject`?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
BillyONeal: I think I meant things like emission meaning `UNREFERENCED_PARAMETER` or similar, and suppression being warning disable pragmas or similar? It's been too long. This was also early enough that such guidelines were "if you want to not do this lets talk about it and we might make more exceptions" that sort of thing, so I think you should review this without regard for what the guidelines said.

If you do keep this I recommend doing:

```
rv += #id " "
```

instead, assuming `rv` is intended to be a `std::string` or similar. (That is, just do the concat inside the string literal rather than needing to construct a temporary std::string to do that)

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
ahsonkhan: Also, how is the code coverage for `win_http_transport.cpp`?

Our code coverage CI leg only runs on linux/curl, so we don't see the coverage number for Windows-specific transport adapter files:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1934177&view=codecoverage-tab

https://github.com/Azure/azure-sdk-for-cpp/blob/53c9da155244b72ac90ad27f515d4ea1e7505a01/eng/pipelines/templates/stages/platform-matrix.json#L143-L149

https://github.com/Azure/azure-sdk-for-cpp/blob/53c9da155244b72ac90ad27f515d4ea1e7505a01/eng/pipelines/templates/stages/platform-matrix-live.json#L37-L43

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+      return false;
+    }
+    return true;
+  }
+
+  void WinHttpAction::CompleteAction()
   {
-    WinHttpTransport* httpTransport = reinterpret_cast<WinHttpTransport*>(dwContext);
-    httpTransport->OnHttpStatusOperation(hInternet, dwInternetStatus);
+    auto scope_exit{m_actionCompleteEvent.SetEvent_scope_exit()};
```
ahsonkhan, line 466: Should we be logging here, and in places where other action event are triggered/changed?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
   {
-    // If an exception is thrown in the handler, log the error and terminate the connection.
-    Log::Write(
-        Logger::Level::Error,
-        "Request Failed Exception Thrown: " + std::string(rfe.what()) + rfe.Message);
-    WinHttpCloseHandle(hInternet);
+    // Note that the order of scope_exit and lock is important - this ensures that scope_exit is
+    // destroyed *after* lock is destroyed, ensuring that the event is not set to the signalled
+    // state before the lock is released.
+    auto scope_exit{m_actionCompleteEvent.SetEvent_scope_exit()};
```
ahsonkhan, line 473: Great comment! ❤️

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+      DWORD expectedCallbackStatus,
+      Azure::Core::Context const& context,
+      Azure::DateTime::duration const& pollInterval)
+  {
+    context.ThrowIfCancelled();
+    ResetEvent(m_actionCompleteEvent.get());
+    m_expectedStatus = expectedCallbackStatus;
+    m_stowedError = 0;
+    m_stowedErrorInformation = 0;
+    m_bytesAvailable = 0;
```
ahsonkhan, line 436: Should we be taking a lock here before "resetting" these fields?

sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
```diff
@@ -0,0 +1,223 @@
...
+    friend class WinHttpRequest;
+
+    // Containing HTTP request, used during the status operation callback.
+    WinHttpRequest* m_httpRequest{};
+    std::mutex m_actionCompleteMutex;
+    wil::unique_event m_actionCompleteEvent;
+    DWORD m_expectedStatus{};
+    DWORD m_stowedError{};
+    DWORD_PTR m_stowedErrorInformation{};
+    DWORD m_bytesAvailable{};
```
ahsonkhan, line 41: nit: add a comment around the fields that shouldn't be modified without taking a lock on `m_actionCompleteMutex`

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
     }
-    // Silently ignore if there's any statuses we get that we can't handle
-    return;
-  }
 
-  // We will only set the Status callback if a root certificate has been set.
-  AZURE_ASSERT(!m_options.ExpectedTlsRootCertificates.empty());
+    try
+    {
+      WinHttpAction* httpAction = reinterpret_cast<WinHttpAction*>(dwContext);
```
ahsonkhan, line 524: Is this cast safe to do?

sdk/core/azure-core-tracing-opentelemetry/test/ut/azure_core_test.cpp
LarryOsterman: It should be indented 2 spaces since it is a comment inside a function declaration.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
-  // Ask WinHTTP for the server certificate - this won't be valid outside a status callback.
-  wil::unique_cert_context serverCertificate;
+  /**
+   * @brief HTTP Callback to enable private certificate checks.
+   *
+   * This method is called by WinHTTP when a certificate is received. This method is called
+   * multiple times based on the state of the TLS connection.
+   *
+   * Special consideration for the WINHTTP_CALLBACK_STATUS_SENDING_REQUEST - this callback is
+   * called during the TLS connection - if a TLS roto certificate is configured, we verify that the
```
ahsonkhan, line 550:
```suggestion
   * called during the TLS connection - if a TLS root certificate is configured, we verify that the
```

sdk/core/azure-core/test/ut/azure_core_test.cpp
LarryOsterman: Indentation is needed because this is inside a function.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+      {
+        // Headers for an HTTP response are available, complete the current action.
+        CompleteAction();
+      }
+      else if (internetStatus == WINHTTP_CALLBACK_STATUS_READ_COMPLETE)
+      {
+        // A WinHttpReadData call has completed. Complete the current action, including the amount
+        // of data read.
+        CompleteActionWithData(statusInformationLength);
+      }
```
ahsonkhan, line 606: nit: Given most of these if-blocks are identical, other than the comment, would a switch be cleaner? Just a thought.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
-            WINHTTP_OPTION_SERVER_CERT_CONTEXT,
-            reinterpret_cast<void*>(serverCertificate.addressof()),
-            &bufferLength))
+    Log::Write(
+        Logger::Level::Informational,
+        "Status operation: " + std::to_string(internetStatus) + "("
+            + InternetStatusToString(internetStatus) + ")");
+    if (internetStatus == WINHTTP_CALLBACK_STATUS_SECURE_FAILURE)
+    {
+      Log::Write(Logger::Level::Error, "Security failure. :(");
```
ahsonkhan, line 567: Where is this parenthesis after the colon in the log message closed?

Or is this a sad face emoji?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+      }
     }
   }
 
-  if (!VerifyCertificatesInChain(m_options.ExpectedTlsRootCertificates, serverCertificate.get()))
+  void WinHttpRequest::HandleExpectedTlsRootCertificates(HINTERNET hInternet)
   {
-    Log::Write(Logger::Level::Error, "Server certificate is not trusted.  Aborting HTTP request");
+    if (!m_expectedTlsRootCertificates.empty())
+    {
```
ahsonkhan, line 627: nit: remove extra new line
```suggestion
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: @BillyONeal  D'oh, of course that's better - for some reason, I continually forget that I don't need to concatenate adjacent string constants.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +855,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
+}
+
+/*
+ * Explicit destructor for WinHttpRequest. This is declared outside
+ * the class definition because otherwise the WinHttpAction class needs to be
+ * fully declared in win_http_transport.hpp.
+ */
+_detail::WinHttpRequest::~WinHttpRequest()
+{
+
```
ahsonkhan, line 942: nit: remove extra line
```suggestion
```

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -638,73 +855,114 @@ Azure::Core::_internal::UniqueHandle<HINTERNET> WinHttpTransport::CreateRequestH
...
+      == WINHTTP_INVALID_STATUS_CALLBACK)
+  {
+    GetErrorAndThrow("Error while setting up the status callback.");
+  }
+}
+
+/*
+ * Explicit destructor for WinHttpRequest. This is declared outside
+ * the class definition because otherwise the WinHttpAction class needs to be
+ * fully declared in win_http_transport.hpp.
```
ahsonkhan, line 938: Why would this require `WinHttpAction` to be fully declared in win_http_transport, and why would we prefer not to do that?

cmake-modules/AddGoogleTest.cmake
```diff
@@ -78,7 +78,8 @@ macro(add_gtest TESTNAME)
         else()
             gtest_discover_tests(${TESTNAME}
                 TEST_PREFIX "${TESTNAME}."
-                PROPERTIES FOLDER "Tests")
+                PROPERTIES FOLDER "Tests"
+                DISCOVERY_TIMEOUT 600)
```
ahsonkhan, line 82: Looks like the default is 5. Is 600 seconds too log?

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
```diff
@@ -327,121 +329,335 @@ bool WinHttpTransport::VerifyCertificatesInChain(
...
+          error,
+          MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+          reinterpret_cast<LPSTR>(&errorMsg),
+          0,
+          nullptr)
+      != 0)
+  {
+    // Use a unique_ptr to manage the lifetime of errorMsg.
+    std::unique_ptr<char, decltype(&LocalFree)> errorString(errorMsg, &LocalFree);
+    errorMsg = nullptr;
```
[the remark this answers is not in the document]
LarryOsterman: For your first question: I'm not sure what the "Deleter" pattern you're referencing. What is here is bog-standard use of std::unique_ptr with a custom free function. These two lines transfer ownership of the memory in the `errorMsg` variable to the `errorString` variable, after which `errorMsg` is set to NULL to reflect that it no longer holds the memory. Now that we've included a dependency on wil, I can probably find another way of doing this if you find it icky.

For your second question:  Which error would you rather see in the exception?

`"WinHTTP Error 12044 was returned"`
or 

`"WinHTTP Error 12044 (Client Auth Certificate Needed)"`

Personally, I would prefer the second error message. And that's what you get from the call to FormatMessage. I *HATE* looking up magic error codes.

This also brings WinHTTP's error diagnostics to parity with the CURL error diagnostics.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: There is a common pattern across all the functions that do anything that might take a while of checking for cancellation before they do the expensive operation. 

I've annotated every call to ThrowIfCancelled to reflect this pattern.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: The `while` loop terminates when the event is set to the signalled state, otherwise it continues to run.
There are three possible return values from WaitForSingleObject:

1.  WAIT_OBJECT_0 - the event was set to the signalled state.
1. WAIT_TIMEOUT - the wait timed out
1. Anything else - There was an error.

The if/else clauses check for those three cases - if WaitForSingleObject returns WAIT_TIMEOUT, we check for cancellation and loop back to wait some more. If WaitForSingleObject returns WAIT_OBJECT_0, we return, if WaitForSingleObject returns anything else, we return false, which lets the caller know to throw an exception.

And I have literally no idea what the code coverage numbers are on Windows since we don't have the ability to measure it.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: `m_stowedError` is updated by the `CompleteActionWithError` method, which is called from the callback.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: By definition, there cannot be any actions outstanding at the time we call this because we have not yet called `initiateAction()`.

I'll add a comment.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: I sure hope so. It's only unsafe if someone changes the code in `WinHttpSendRequest` to something other than the current action.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: I guess? I want to keep each of these broken out separately because the message does not always reflect the API call which is being completed, but it does make the code a bit more compact.

cmake-modules/AddGoogleTest.cmake
LarryOsterman: Rick said make it 10 minutes. I'm making it 10 minutes.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: As I've mentioned before, WinHttpAction drags in a huge number of Windows-only headers which have no business being in our public API surface.

In general, I feel strongly that public headers should only declare public types. Implementation details confuse customers and it is not always clear what types are a part of the public contract and which are not.

But now that the destructor actually does something other than being an empty destructor, I'll edit the comment to make it more correct.

sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
LarryOsterman: That's a sad face emoji.

PR #3928, opened 2022-09-13
Update `AttestationClient::AttestTpm` API to match existing `AttestOpenEnclave` and `AttestSgxmEnclave` 

penglimsft:
# Pull Request Checklist

Please leverage this checklist as a reminder to address commonly occurring feedback when submitting a pull request to make sure your PR can be reviewed quickly:

See the detailed list in the [contributing guide](https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#pull-requests).

- [ ] [C++ Guidelines](https://azure.github.io/azure-sdk/cpp_introduction.html)
- [ ] Doxygen docs
- [ ] Unit tests
- [ ] No unwanted commits/changes
- [ ] Descriptive title/description
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] External references or docs updated
- [ ] Self review of PR done
- [ ] Any breaking changes?

PR #4016, opened 2022-10-11
Updated attestation change log to add periods for consistency

ahsonkhan:


sdk/attestation/azure-security-attestation/CHANGELOG.md
```diff
@@ -6,10 +6,10 @@
...
 ### Breaking Changes
 
-- Changed `AttestationClient::AttestTpm` to match `AttestOpenEnclave` and `AttestSgxEnclave` 
-- Added `std::vector<uint8_t>` dataToAttest parameter.
-- Removed `Payload` in `TpmAttestationOptions`
-- Changed `TpmResult` in `TpmAttestationResult` to type `std::vector<uint8_t>`
+- Changed `AttestationClient::AttestTpm` to match `AttestOpenEnclave` and `AttestSgxEnclave`:
+  - Added `std::vector<uint8_t>` dataToAttest parameter to the `AttestTpm()` client method.
+  - Removed `Payload` in `TpmAttestationOptions`.
+  - Changed `TpmResult` in `TpmAttestationResult` to type `std::vector<uint8_t>`.
```
ahsonkhan, line 12: Looking at https://github.com/Azure/azure-sdk-for-cpp/pull/3928/files#diff-f636729e2838c433532c5cf68755672ec6ccf9fce4b398b00a778e953fb10f5eR9, looks like the intent was to have 1 bullet for the breaking change (since there is actually *only one* method signature that is changing), and describing the break in 3 sub-bullets.

That makes the most sense to me, rather than listing 4 bullets to describe the same change. Otherwise, it makes the reader think there are multiple changes they have to react to.

Related to:
https://github.com/Azure/azure-sdk-for-cpp/pull/3992

@penglimsft does this look good to you as well?

antkmsft (approved):

LarryOsterman (approved):

PR #4024, opened 2022-10-13
In-memory Identity token cache

### LINKED ISSUE #3835: Token credential with local cache
We got a feature request from one of our customers.
Basically they're using token credential for authorization and they need different client options (like retry options) for different kind of requests so they need to initialize many clients for those requests.
The problem is the access token is cached in `BearerTokenAuthenticationPolicy`, which is created when initializing a blob client. So the access token cannot be shared among these blob clients even though requests are sent to the same storage account.

I'm looking for some way to fix this problem and while I'm looking at .Net code, I found [SharedTokenCacheCredential](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.sharedtokencachecredential?view=azure-dotnet). I think such a credential with local cache can work in this scenario.

@RickWinter @antkmsft What do you think? Can we provide the same credential in C++?

antkmsft:
Closes #3835.


sdk/identity/azure-identity/src/private/token_cache.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Azure { namespace Identity { namespace _detail {
+  /**
+   * @brief Implements an access token cache.
+   *
+   */
+  class TokenCache final {
+  private:
+    TokenCache() = delete;
+    ~TokenCache() = delete;
```
Jinming-Hu, line 24: Nit: there's no need to mark these as private  since we've deleted them.

sdk/identity/azure-identity/src/private/token_cache.hpp
```diff
@@ -0,0 +1,61 @@
...
+    class Internals;
+
+#if defined(TESTING_BUILD)
+    /**
+     * @brief Clears token cache. Intended to only be used in tests.
+     *
+     */
+    static void Clear();
+#endif
+  };
```
Jinming-Hu, line 60: I don't think this is a good idea. The code we test would be different from the code we ship.

Jinming-Hu: How about move this into `_detail` namespace?

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+    {
+      auto const compare = AuthorityHost.compare(other.AuthorityHost);
+      if (compare != 0)
+      {
+        return compare < 0;
+      }
+    }
+
+    return Scopes < other.Scopes;
```
Jinming-Hu, line 103:
```suggestion
    return std::tie(TenantId, ClientId, AuthorityHost, Scopes) < std::tie(other.TenantId, other.ClientId, other.AuthorityHost, other.Scopes);
```

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
```diff
@@ -0,0 +1,105 @@
...
+     */
+    static std::shared_timed_mutex Mutex;
+
+#if defined(TESTING_BUILD)
+    /**
+     * @brief A function that gets called before write lock is acquired.
+     *
+     */
+    static std::function<void()> OnBeforeWriteLock;
+#endif
```
Jinming-Hu, line 73: This scares me even more. We should not make the code we test and the code we ship too different.

sdk/identity/azure-identity/src/token_cache.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+#if defined(TESTING_BUILD)
+  if (TokenCache::Internals::OnBeforeWriteLock != nullptr)
+  {
+    TokenCache::Internals::OnBeforeWriteLock();
+  }
+#endif
+
+  std::unique_lock<std::shared_timed_mutex> writeLock(Internals::Mutex);
```
Jinming-Hu, line 55: We must not hold the write lock while refreshing the token, which usually takes hundreds of milliseconds.

This would make all the requests stall a period of time.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
```diff
@@ -0,0 +1,105 @@
...
+     * @brief The cache itself.
+     *
+     */
+    static std::map<CacheKey, Core::Credentials::AccessToken> Cache;
+
+    /**
+     * @brief Vector of expiration for the cache entries, from earliest to the latest.
+     *
+     */
+    static std::vector<std::pair<decltype(Core::Credentials::AccessToken::ExpiresOn), CacheKey>>
```
Jinming-Hu, line 58: I'm not sure we need this vector.

In most scenarios, there aren't too many distinct "Key" in a program. It doesn't matter if we just keep the cached entries forever in the map defined at L52.

Jinming-Hu: Removing this vector will also simplify the logic of `TokenCache::GetToken()`.

sdk/identity/azure-identity/src/token_cache.cpp
Jinming-Hu: I got an idea.

We add a boolean flag `TokenBeingRefreshed`. The first thread observed a token expires in 3 mins, it atomically set this value to `true`. Other threads will use the old token if it sees this value is `true`, because this old token is not expired yet. This can help us avoid too many threads refreshing the same token at the same time.

If other threads see this token is about to expire in 1 minute, it refreshes the token regardless the boolean flag.

sdk/identity/azure-identity/src/private/token_cache.hpp
antkmsft: This is already in `_detail` namespace, and it is a header under `/src`, so no way for the users to see it.
I agree with the idea as a whole, but realistically, this is the least of all evils.
This is commenting out a static function.
The function is supposed to be used only in tests. It is in a `_detail` class.
But if we comment it out, it is not available to SDK code, only to tests. And should one use it in the SDK code, it will be immediately clear that it should not have been used, and linker error will occur, forcing the SDK programmer to not use it, as intended.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
antkmsft: While agreeing on the principle, I respectfully disagree in this specific case - realistically this is the least of all evils. There is not a better way to achieve what we want, and if used carefully as we do, it is harmless. Even if half of the things would go wrong.

`OnBefore...` is a test hook. It is in `_detail` namespace in a header that is in `/src`, so not even available in includes. It is only available when the code is built with tests. SDK code should not be invoking it, only the test code. SDK code should not be setting the callback. If an SDK programmer tries to use it, a linker error will occur, the latest it can be notced is the nightly vcpkg build, but probably earlier.

Should you have built the SDK with test hooks, then you get one extra `nullptr` check. But it is between mutexes, so it is better to not do unnecessary things there. And it is not necessary to even check for null, in the ideal case. The worst thing that happens if the callback does not get invoked is that the test will fail / will fail to compile / link.

But it is not possible to test the multithreaded code with this quality of detail without these test hooks.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
antkmsft: Thank you, I didn't know about that. I sometimes have blind spots in some of the features that were added in C++11 or later. This is neat! I should have assumed something like this must exist by this time.

sdk/identity/azure-identity/src/token_cache.cpp
antkmsft: It was not possible for multiple threads to start refreshing the same token at the same time before, but no one during that update could read their token from cache even if their token was still good, you're right.

So I changed the code in the spirit of problems that you highlighted and partly in the spirit of your suggestions. Check it out. I'll summarize in a separate message.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
antkmsft: I updated the code mostly as you suggested. I'll summarize in a separate message.
I would prefer the cache to be cleaned up from time to time, but it is logistically harder now. But I agree it probably will never be a problem. The benefits we get in exchange (concurrent read access to one token while other threads update the other token) are greater. I was thinking about it, it should still be possible to clean up cache, if we find it is a problem. It would be a bit more expensive now, since we would need to hold the lock for the `map` and try (without waiting) to get a write lock for the expired elements, I think it could be done, especially if we tell it to run for every 50th insertion, for example (or we can go exponential, run after the cache reaches `32`, `64`, `128`, ... in size so that we clean up proportionally often to the cache caliber).

antkmsft (commented):
@Jinming-Hu, check it out:
No more bottleneck for everybody: if your token is still good, you are not blocked. If your token gets refreshed, you wait. So there are two levels of mutexes: for the cache container, and for each individual item. There is no cleaning up of expired items as you suggested this is likely will never be a problem, but I am still keeping that option open to add it later - for example, to clean up on every time the cache reaches `(2^n)` in size, where `n>=5` (for example).

antkmsft:
I added cache cleanup every time cache size is about to grow over `==2^n`, where `n>=5`. LMK what you think.

sdk/identity/azure-identity/src/private/token_cache.hpp
Jinming-Hu: ok, it's your call.

sdk/identity/azure-identity/src/client_secret_credential.cpp
```diff
@@ -62,28 +68,39 @@ Azure::Core::Credentials::AccessToken ClientSecretCredential::GetToken(
...
+    if (!scopes.empty())
     {
-      auto const& scopes = tokenRequestContext.Scopes;
-      if (!scopes.empty())
+      scopesStr = TokenCredentialImpl::FormatScopes(scopes, m_isAdfs);
+    }
+  }
+
+  return TokenCache::GetToken(m_tenantId, m_clientId, m_authorityHost, scopesStr, [&]() {
+    return m_tokenCredentialImpl->GetToken(context, [&]() {
```
Jinming-Hu, line 84: Correct me if I'm wrong.

The lambda function defined at Line 84 is invoked when token needs to be refreshed, by the time of which, variable `scopesStr` is already destructed because it's a local variable.

But you capture this local variable by reference rather than by value, and use this variable at Line 92, won't this crash the application?

Jinming-Hu: This comment also applies to the code in `managed_identity_source.cpp`

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
```diff
@@ -0,0 +1,88 @@
...
+    };
+
+    /**
+     * @brief Represents immediate cache value (token) and a synchronization primitive to handle its
+     * updates.
+     *
+     */
+    struct CacheValue
+    {
+      std::shared_timed_mutex ElementMutex;
```
Jinming-Hu, line 58: I don't see why we have to use **timed** mutex rather than just `std::shared_mutex`.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
```diff
@@ -0,0 +1,88 @@
...
+     * @brief The cache itself.
+     *
+     */
+    static std::map<CacheKey, std::shared_ptr<CacheValue>> Cache;
+
+    /**
+     * @brief Mutex to access the cache container.
+     *
+     */
+    static std::shared_timed_mutex CacheMutex;
```
Jinming-Hu, line 72: Also, shared_mutex seems just fine to me.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
antkmsft: `std::shared_timed_mutex` is C++14, while `std::shared_mutex` is C++17. We can't go newer than 14.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
antkmsft: `std::shared_mutex` is C++17, unfortunately.

sdk/identity/azure-identity/src/private/token_cache_internals.hpp
Jinming-Hu: ok

sdk/identity/azure-identity/src/client_secret_credential.cpp
antkmsft: No, it shouldn't be dead. The reference to the lambda is not kept anywhere. By the time we're about to return the value, `TokenCache::GetToken()` has already returned the value. It had a chance to call lambda only while it was executing. So by the time we get a return value, we're back in this function, where `scopesStr` is still alive (BTW, should've there been a problem, there would've been the same problem with `m_tokenCredentialImpl` reference).

Simplified, it is this:
```cpp
int f(std::function<int()> g)
{
    return g();
}

int h()
{
    std::string s = "Hello";
    int retval = f([&](){ std::cout << s; return 1; });
    // at this point we returned from f() and from the lambda we passed there.
   // s is stil alive.
   return retval;
}
```

`int retval = ...; return retval` is basically what the compiler generates anyway. Correct, or am I missing something?

Jinming-Hu: Ok, got it. So the lambda function is called synchronously. It's not saved somewhere and get called later. Thanks for the clarification!

Jinming-Hu:
@antkmsft Do you have any idea when this feature will get merged/released?

antkmsft:
@Jinming-Hu, I am thinking of it getting released in 1.4.0-beta.2 on November 8th.

As of merged, I want to collect more feedback/get more reviews. Ideally, review concurrent stuff with @LarryOsterman / @JeffreyRichter, and also get a thumbs up from @schaabs on the correctness of the idea of mapping (TenantId, ClientId, AuthorityHost, Scopes) => Token. I think I will have it merged by EOW.

If you'd suggest OOB/private release for the customer to try it out sooner, let me know, I think we can figure out something.

sdk/identity/azure-identity/src/client_secret_credential.cpp
```diff
@@ -45,7 +51,7 @@ ClientSecretCredential::ClientSecretCredential(
 ClientSecretCredential::ClientSecretCredential(
     std::string tenantId,
     std::string clientId,
-    std::string clientSecret,
+    std::string const& clientSecret,
```
LarryOsterman, line 54: Good catch. Leaking client secrets on the heap is probably a bad thing(tm) :).

sdk/identity/azure-identity/src/managed_identity_source.cpp
```diff
@@ -81,17 +83,26 @@ Azure::Core::Credentials::AccessToken AppServiceManagedIdentitySource::GetToken(
...
+    if (!scopes.empty())
     {
-      auto const& scopes = tokenRequestContext.Scopes;
-      if (!scopes.empty())
+      scopesStr = TokenCredentialImpl::FormatScopes(scopes, true);
+    }
+  }
+
+  return TokenCache::GetToken(std::string(), GetClientId(), GetAuthorityHost(), scopesStr, [&]() {
+    return TokenCredentialImpl::GetToken(context, [&]() {
```
LarryOsterman, line 96: What is being captured with the inner universal byref capture? It doesn't look like there's anything other than scopestr, so it might be better to restrict the capture to just scopestr.

sdk/identity/azure-identity/src/managed_identity_source.cpp
```diff
@@ -141,27 +152,36 @@ Azure::Core::Credentials::AccessToken CloudShellManagedIdentitySource::GetToken(
...
+    auto const& scopes = tokenRequestContext.Scopes;
+    if (!scopes.empty())
     {
-      auto const& scopes = tokenRequestContext.Scopes;
-      if (!scopes.empty())
+      scopesStr = TokenCredentialImpl::FormatScopes(scopes, true);
+    }
+  }
+
+  return TokenCache::GetToken(std::string(), GetClientId(), GetAuthorityHost(), scopesStr, [&]() {
```
LarryOsterman, line 164: Similar comment - every time I see a wildcard lambda capture I get a titch worried. In this case, I understand that the capture is safe, but I'm worried that it's going to sweep up more than anticipated (when the compiler transforms the lambda into an anonymous class, all the captured values are copied into the anonymous class, which can potentially capture more than is anticipated).

sdk/identity/azure-identity/src/managed_identity_source.cpp
```diff
@@ -212,60 +242,61 @@ Azure::Core::Credentials::AccessToken AzureArcManagedIdentitySource::GetToken(
...
-        std::ifstream secretFile(challenge.substr(eq + 1));
-        request->HttpRequest.SetHeader(
-            "Authorization",
-            "Basic "
-                + std::string(
-                    std::istreambuf_iterator<char>(secretFile), std::istreambuf_iterator<char>()));
-
-        return request;
-      });
+  return TokenCache::GetToken(std::string(), GetClientId(), GetAuthorityHost(), scopesStr, [&]() {
```
LarryOsterman, line 255: Why an explicit constructor for `std::string` instead of ""?

sdk/identity/azure-identity/src/token_cache.cpp
```diff
@@ -0,0 +1,148 @@
...
+#endif
+
+namespace {
+bool IsFresh(
+    std::shared_ptr<TokenCache::Internals::CacheValue> const& item,
+    std::chrono::system_clock::time_point now)
+{
+  // If cached token expires in less than MinExpiry from now, it's cached value won't be returned,
+  // newer value will be requested.
+  constexpr auto MinExpiry = std::chrono::minutes(3);
```
LarryOsterman, line 29: Where does 3 minutes come from?

sdk/identity/azure-identity/src/token_cache.cpp
```diff
@@ -0,0 +1,148 @@
...
+  {
+    return found->second;
+  }
+
+  // Clean up cache from expired items (once every N insertions).
+  {
+    auto const cacheSize = TokenCache::Internals::Cache.size();
+
+    // N: cacheSize (before insertion) is >= 32 and is a power of two.
+    if (cacheSize >= 32 && (cacheSize & (cacheSize - 1)) == 0)
```
LarryOsterman, line 69: Where did 32 come from?

You might want to explain what X&(X-1) does, many people don't understand that particular trick (especially since it only works on processors which implement 2s complement arithmetic).

sdk/identity/azure-identity/test/ut/token_cache_test.cpp
```diff
@@ -0,0 +1,339 @@
...
+  });
+
+  EXPECT_EQ(TokenCache::Internals::Cache.size(), 1UL);
+
+  EXPECT_EQ(token.ExpiresOn, Tomorrow);
+  EXPECT_EQ(token.Token, "T1");
+}
+
+TEST(TokenCache, TwoThreadsAttemptToUpdateTheSameToken)
+{
```
LarryOsterman, line 146: In addition to simulating what happens when multiple threads attempt to retreive a token, you might consider a test which actually retrieves a token from a few threads at the same time.

LarryOsterman (approved):

sdk/identity/azure-identity/src/managed_identity_source.cpp
antkmsft: Just a micro-optimization - default constructor would be faster than for the string class to read `const char*` arg to find out it is an empty string, plus `""` is not getting allocated in `.text`.

sdk/identity/azure-identity/src/token_cache.cpp
antkmsft: `BearerTokenAuthenticationPolicy` has the same criteria for 2 minutes, and if it is less, then it asks credential for the new token, which will result in the code in this file being executed.
This is 1 minute more. It does not have to be - can be 5 minutes, can be 2, can be 1, so we're not taking a dependency on what `BearerTokenAuthPolicy` has, but I think it makes things simpler and would produce optimal results, in terms of executing less code.

https://github.com/Azure/azure-sdk-for-cpp/blob/fda88bf7b8389575b9d973821efc604442e10493/sdk/core/azure-core/src/http/bearer_token_authentication_policy.cpp#L22-L25

I could depend on the timeout that BearerTokenAuthPolicy has, but it would probably create an impression that it has to be 1 minute more, which is technically not.

sdk/identity/azure-identity/src/token_cache.cpp
antkmsft: 32 is just a power of two I chose instinctively. It can be any number.
As of the "power of two" explanation, the comment just above `if()` says that:
`"N: cacheSize (before insertion) is >= 32 and is a power of two."`

LarryOsterman: My 2nd comment is actually a *about* the "and is a power of 2" comment. *I* know that X&(X-1) reduces the number of bits in a integer value by 1 (on 2s complement CPUs). But I don't believe that is general knowledge. My comment is that you might want to add a link to something which explains the trick, because to me, it was *NOT* obvious (I once spent *hours* trying to figure out why it worked).

I know it because it was one of the solutions to the "Count the number of 1 bits in a 32bit integer" problem, but otherwise that would be totally mystifying code.

sdk/identity/azure-identity/src/token_cache.cpp
LarryOsterman: Every time I see a magic number (like 32 or 3), I want to know where the number came from. I'm not challenging the choice of number (I think it's fine), I'm asking that the comments explain why that number was chosen.

sdk/identity/azure-identity/src/managed_identity_source.cpp
antkmsft: It is actually `[this, &context, &scopesStr]`. From `this`, we use `m_tokenCredentialImpl` and `m_requestBody`. I can add it.
But in my opinion the only concern here is whether the lambda is going to outlive any of the values referenced. Since it is byref, it is technically possible, unless you know that the lambda is not being saved anywhere. Whether the reader sees `[&]` or `[this, &context, &scopesStr]`, it does not help to answer that question.
Correct me if you have different perception, I certainly can add it.

sdk/identity/azure-identity/src/managed_identity_source.cpp
antkmsft: It is `[this, &context, &scopesStr]`. I can add it, but see my comment below. If you think we still should, let me know and I'll add it.

LarryOsterman: In general, I prefer smaller lambda captures. But it's not a huge deal, especially if you know that the function is synchronous.

sdk/identity/azure-identity/src/client_secret_credential.cpp
antkmsft: The reason I did that is that we don't keep a copy of `clientSecret` so copy is not needed - we partially initialize `m_requestBody` and put it there. For the rest of the arguments, we now do keep a copy, so we can take them byval and `std::move()` them, and if they happen to be rvalues, it would be more efficient. If not - either way a copy needs to be created, be it for the argument or later during the `m_*` initialization.

sdk/identity/azure-identity/src/token_cache.cpp
antkmsft: I'll add a link. He best book for such tricks that I know is "Hacker's Delight" by Hentry S. Warren, but it is not freely available online. I'll find a good online article.
BTW, starting with C++20, two's compliment is required.

sdk/identity/azure-identity/src/token_cache.cpp
antkmsft: Thank you, I already added it to the changes that I am going to commit soon.

sdk/identity/azure-identity/src/managed_identity_source.cpp
antkmsft: BTW, I will also add comment about the lifetime in my next commit. I hope it will be helpful to the reader to give that clarity that Jinming and you probably wanted initially.

sdk/identity/azure-identity/src/client_secret_credential.cpp
LarryOsterman: The thing about client secrets is that they're secrets. If we store them in a std::string (even temporarily) without clearing them after they're used, then we potentially leak those into memory, allowing an attacker to access them (see also the CLR's `SecureString` class).

In practice, it's not a huge deal (and `SecureString` has a host of problems that are nasty enough that Barry Dorrans breaks out in hives whenever anyone mentions the phrase). But marking it as const is a good thing (since it means we don't need to copy the value on the call).

sdk/core/azure-core/src/http/bearer_token_authentication_policy.cpp
```diff
@@ -18,10 +18,13 @@ std::unique_ptr<RawResponse> BearerTokenAuthenticationPolicy::Send(
...
     std::lock_guard<std::mutex> lock(m_accessTokenMutex);
 
-    // Refresh the token in 2 or less minutes before the actual expiration.
-    if (std::chrono::system_clock::now() > (m_accessToken.ExpiresOn - std::chrono::minutes(2)))
+    if (std::chrono::system_clock::now()
+        > (m_accessToken.ExpiresOn - m_tokenRequestContext.MinimumExpiration))
     {
-      m_accessToken = m_credential->GetToken(m_tokenRequestContext, context);
+      auto tokenRequestContext = m_tokenRequestContext;
+      tokenRequestContext.MinimumExpiration += std::chrono::minutes(1);
```
LarryOsterman, line 25: Why decrease the expiration by 1 minute here? Just trying to understand the algorithm.

sdk/identity/azure-identity/src/client_secret_credential.cpp
```diff
@@ -80,27 +80,37 @@ Azure::Core::Credentials::AccessToken ClientSecretCredential::GetToken(
...
+  // TokenCache::GetToken() and m_tokenCredentialImpl->GetToken() can only use the lambda argument
+  // when they are being executed. They are not supposed to keep a reference to lambda argument to
+  // call it later. Therefore, any capture made here will outlive the possible time frame when the
+  // lambda might get called.
+  return TokenCache::GetToken(
+      m_tenantId,
+      m_clientId,
+      m_authorityHost,
+      scopesStr,
+      tokenRequestContext.MinimumExpiration,
```
LarryOsterman, line 92: Would it make more sense to pass the `tokenRequestContext` here to future-proof other uses of context values?

antkmsft: We could, but I rather wouldn't for now - `tokenRequestContext` contains scopes vector. But we get scopes already formatted as string, ready to go to a network request, or to a map key. Although we'd get it by ref, it may get a bit confusing, plus the test code is not going to look as pretty. And it is a code in `_detail` namespace, not exposed to anyone, being used internally only within `Identity` lib, so there's no immediate need to future-proof at this point. If we make it `_internal`, or  - should there come more parameters that need to go into that function - I think that whoever makes that change would go through that thinking process naturally and will likely replace `minimumExpiration` with `TokenRequestContext const& tokenRequestContext`. I'm 51/49 on this, if you feel more strongly towards future-proofing, I can do that.

sdk/core/azure-core/src/http/bearer_token_authentication_policy.cpp
antkmsft: We're saying: "get me a new token, and if possible, get it so that is not expiring for at least next 3 minutes".

Default `MinimumExpiration` is 2 minutes, because it is used on line 22. So it comes to this:
"If token expires in 2 minutes or less" => ("get me a new token, and if possible, get it so that is not expiring for at least next 3 minutes")

antkmsft: You know, thinking more about this, we don't actually need it to be 3 minutes. 2 minutes will work just fine.

antkmsft: Yes, I updated it to 2 minutes. If check within `BearerTokenAuthPolicy` is going to see that token has <= 2 minutes to live, then by the time it gets for the `TokenCache` to decide whether the cached token is still good, the token will certainly have less lifetime when it had when `BearerTokenAuthPolicy` checked it.

PR #4045, opened 2022-10-22
Added the ability to select OpenSSL 1.1.1n if desired.

LarryOsterman:
Added the ability to switch between OpenSSL 3 (the default) and OpenSSL 1.1.1n.

Updated readme file with instructions on how to configure vcpkg and cmake to select OpenSSL 1.1.1n. Added CRL validation which is OpenSSL 1.1.1n compatable.

# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -37,19 +37,35 @@
...
+#if defined(AZ_PLATFORM_POSIX)
 #include <poll.h> // for poll()
 #include <sys/socket.h> // for socket shutdown
 #elif defined(AZ_PLATFORM_WINDOWS)
 #include <winsock2.h> // for WSAPoll();
-#endif
+#endif // AZ_PLATFORM_POSIX/AZ_PLATFORM_WINDOWS
 
 #include <algorithm>
 #include <chrono>
```
Jinming-Hu, line 71: The include header at Line 73 can be removed. It's superfluous.

Jinming-Hu: #include <openssl/ssl.h>

README.md
```diff
@@ -324,6 +324,41 @@ The following SDK library releases are available on [vcpkg](https://github.com/m
...
 > NOTE: In case of getting linker errors when consuming the SDK on Windows, make sure that [vcpkg triplet](https://vcpkg.readthedocs.io/en/latest/users/triplets/) being consumed matches the [CRT link flags](https://docs.microsoft.com/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-160) being set for your app or library build. See also `MSVC_USE_STATIC_CRT` build flag.
 
+## OpenSSL Version
+
+Several packages within the Azure SDK for C++ use the OpenSSL library. By default, the Azure SDK will use whatever the most recent version of OpenSSL is within the VCPKG repository.
+
+If you need to use a specific version of OpenSSL, you can use the vcpkg custom ports feature to specify the version of OpenSSL to use. 
+For example, if you want to use OpenSSL 1.1.1, you should create a folder named `vcpkg-custom-ports` next to to your vcpkg.json file.
+
+Navigate to your clone of the vcpkg vcpkg repo and execute "git checkout 3b3bd424827a1f7f4813216f6b32b6c61e386b2e" - this will reset your repo to the last version of OpenSSL 1.1.1
```
Jinming-Hu, line 334: Does this pin the version to `1.1.1n`? Is there any way to upgrade to newer versions like `1.1.1o` or `1.1.1r`?

Is the FIPS certification the problem or one of problems we're trying to solve in this PR? If so, pinning to an older version doesn't seem to meet the requirement.

LarryOsterman: Yes it does, because that is the most recent FIPS certified version of OpenSSL (and more importantly the last version supported on vcpkg).

There are two problems we're trying to solve with this PR:
1. Customers who are running on OS's which don't support OpenSSL 3 can't use OpenSSL 3.
2. Customers who require FIPS certification cannot use the Azure SDK.

It *is* possible to switch to OpenSSL 1.1.1q (which is the most recent version published on OpenSSL.org) - I did it while testing this PR. It's a bit convoluted and we don't want to commit to long term support for those versions of OpenSSL so I didn't include instructions on how to do this.

Jinming-Hu: In the cryptography security guidance, it says

- Do not hardcode to OpenSSL 1.1.1.
- OpenSSL 1.1.1* is allowed and **must** use a recent release version that includes patches for all known vulnerabilities and **must stay up to date** regarding known vulnerability fixes.

I think technically we hardcode the version in this PR. In addition, how do we upgrade to new versions if there was a new vulnerability found in `1.1.1n` someday?

vcpkg-custom-ports/openssl/portfile.cmake
```diff
@@ -0,0 +1,28 @@
+if(EXISTS "${CURRENT_INSTALLED_DIR}/include/openssl/ssl.h")
+    message(FATAL_ERROR "Can't build openssl if libressl/boringssl is installed. Please remove libressl/boringssl, and try install openssl again if you need it.")
+endif()
+
+set(OPENSSL_VERSION 1.1.1n)
+vcpkg_download_distfile(
+    ARCHIVE
+    URLS "https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz" "https://www.openssl.org/source/old/1.1.1/openssl-${OPENSSL_VERSION}.tar.gz"
+    FILENAME "openssl-${OPENSSL_VERSION}.tar.gz"
+    SHA512 1937796736613dcf4105a54e42ecb61f95a1cea74677156f9459aea0f2c95159359e766089632bf364ee6b0d28d661eb9957bce8fecc9d2436378d8d79e8d0a4
```
RickWinter, line 10: If the openssl version is set to something other than 1.1.1n won't the SHA512 check fail?

LarryOsterman: Yes. The way you deal with this, you set OPENSSL_VERSION=1.1.1q, and set SHA512 to 0. cmake/vcpkg then complains that the SHA512 doesn't match and tells you the SHA512 you should use. You cut and paste that into the portfile.cmake and everything starts working.

Which of course completely defeats the point of the SHA512, but we are going to ignore that.

LarryOsterman:
Ping :).

antkmsft (approved):
👍

PR #4053, opened 2022-10-25
[Perf] Include GA versions of core package

mikeharder:


mikeharder:
@gearama

mikeharder:
> I'm not sure what this actually means (`source` -> `1|source`) but it seems benign.

This string is a regex to select which package versions to run.  Adding `1` includes the `1.X.Y` versions of `azure-core-cpp`.

PR #4054, opened 2022-10-25
Fix to for a SAL annotation

RickWinter:


antkmsft (approved):

antkmsft:
Thank you! It has been annoying me just recently, and I didn't notice that it is in another file, because locally I don't have this change, but on Github, it merges with the latest and was complaining about a very similar place in another file.

It was not failing for me in CI, but it did in non-CI pipelines. I'll run one of them for your PR, if you don't mind.

LarryOsterman (approved):

Issue #3603, opened 2022-04-29
Storage DataMovement Library


Jinming-Hu:
- [ ] Blob Virtual Directory
- [ ] Scheduler
  - [ ] Job Basics
  - [ ] Thread Pool
- [ ] Single Upload Job
- [ ] Single Download Job
- [ ] Directory Upload Job
- [ ] Directory Download Job
- [ ] Job Pause & Resume
  - [ ] Journal file
- [ ] Job Cancellation

Jinming-Hu:
related PR https://github.com/Azure/azure-sdk-for-cpp/pull/3618 https://github.com/Azure/azure-sdk-for-cpp/pull/3594

PR #4067, opened 2022-10-27
DMLib page blob download support

Jinming-Hu:
# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
+++ b/sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
@@ -7,10 +7,12 @@
 #include <vector>
 
 #include <azure/core/azure_assert.hpp>
+#include <azure/core/internal/json/json.hpp>
 
 #include "azure/storage/blobs/blob_options.hpp"
 #include "azure/storage/datamovement/tasks/async_copy_blob_task.hpp"
 #include "azure/storage/datamovement/tasks/download_blob_to_file_task.hpp"
+#include "azure/storage/datamovement/tasks/download_page_blob_to_file_task.hpp"
 #include "azure/storage/datamovement/tasks/upload_blob_from_file_task.hpp"
 #include "azure/storage/datamovement/utilities.hpp"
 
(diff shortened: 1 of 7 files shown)


antkmsft (approved):

sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
```diff
@@ -49,6 +51,108 @@ namespace Azure { namespace Storage { namespace _detail {
...
+          }
+          else
+          {
+            flushSubtask();
+            addRange(range);
+          }
+        }
+      };
+
+      for (auto rangePage = pageBlobClient.GetPageRanges(); rangePage.HasPage();
```
EmmaZhu, line 136: It could take long time or even timeout to GetpageRanges for a large page blob with large number of small ranges. Please change to get page ranges with spans or chunks.

EmmaZhu, line 136: Can has page be used as loop termination condition? Does GetPageRanges only return range with data?

sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
```diff
@@ -49,6 +51,108 @@ namespace Azure { namespace Storage { namespace _detail {
...
+           rangePage.MoveToNextPage())
+      {
+        for (const auto& range : rangePage.PageRanges)
+        {
+          addRange(range);
```
EmmaZhu, line 141: I didn't go through all range handling code. If you have done this, please just ignore this comment.
It might be slow to send one download request for each small range. It'd be faster to merge nearby ranges to a larger one (maybe 1MB  or so) and download a larger chunk with one request.

sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
```diff
@@ -146,14 +250,28 @@ namespace Azure { namespace Storage { namespace _detail {
...
+        task.SourceType = TaskModel::ResourceType::BlockBlob;
+      }
+      else if (blobProperties.BlobType == Blobs::Models::BlobType::AppendBlob)
+      {
+        task.SourceType = TaskModel::ResourceType::AppendBlob;
+      }
+      else if (blobProperties.BlobType == Blobs::Models::BlobType::PageBlob)
+      {
+        task.SourceType = TaskModel::ResourceType::PageBlob;
+        PageBlobTaskDecoration(task, m_model.Source.m_blobClient.Value().AsPageBlobClient());
```
EmmaZhu, line 273: List page range could take pretty long time. Don't put it in the main scheduling thread.

EmmaZhu:
Not sure about current plan. Some benchmark for page blob with different ranges would be good.

sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
Jinming-Hu: I checked the REST API, this API returns paged results since version `2020-10-02`, each page containing at most 10000 items. If this is the case, I don't think there's anything we need to do at client side.

Jinming-Hu: > Can has page be used as loop termination condition?

Yes, that's how pageable API works in C++ clients.

> Does GetPageRanges only return range with data?

Yes

sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
Jinming-Hu: I defined a constant variable at Line 80. The threshold is now 512 bytes, but it can be easily changed to any value.
`constexpr int64_t MergeThreshold = 512;`

`1MB` seems too large. The overhead of issuing a new HTTP request is: http request header size (around 700\~800 bytes), http response header size (around 700\~800 bytes), one extra transaction, one extra round-trip time. With these numbers in mind, I think 512 bytes or 1024 bytes might be a better choice.

sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp
Jinming-Hu: Good point. It's not a good idea to issue network requests in scheduling thread. This can be a bottleneck especially when downloading large number of small files/blobs. Listing page ranges, by comparison, is less of a problem. Because small files can be as small as several bytes, and each file/blob item takes hundreds of bytes in ListBlobs response. However, page range is at least 512 bytes, and a range takes only like ~20 bytes in ListPageRanges response.

Anyway, we already have a [TODO](https://github.com/Azure/azure-sdk-for-cpp/blob/dcf62d20eb42f42c40507b8ef6445a5a8730662e/sdk/storage/azure-storage-datamovement/src/job_engine_task_generation.cpp#L146) item here. We also have GitHub issue #3603 tracking these work items. The purpose of this PR is to make most common scenarios work. We'll make optimizations for other scenarios in separate PRs. I hope this all makes sense to you.

Jinming-Hu:
> Not sure about current plan. Some benchmark for page blob with different ranges would be good.

Perf benchmark is underway. I assume that doesn't block this PR.

EmmaZhu (approved):

PR #4078, opened 2022-10-31
Gtest timeout fix

### LINKED ISSUE #3951: Split test capture step out from build step
Currently the builds run gtest_discover_tests as a final step of the tests.  This step should be split out into its own step to speed up the builds and help isolate points of failure.

gearama:
closes #3951 
Having isolated the point of failure to GTest_Discover_Tests , splitting the step out is pointless and actually lengthens the build process as it being part of the cmake files, the only way to have it executed or not is to run cmake config / build twice, which overall lengthen the build process. 
We cannot have only the discovery test step run , but we can actually increase the timeout of the discovery step from 5 s currently ( which is too short ) to something more appropriate. 
To quote the cmake gtest documentation for discovery timeout "Most test executables will enumerate their tests very quickly, but under some exceptional circumstances, a test may require a longer timeout. The default is 5. " the exceptional circumstances are not elaborated further. 
Why the current implementation of the timeout is not working is due to the fact that it is in a macro nobody calls , protected by an ENV that nobody sets thus it never really executes. 

Properly add timeout (default is 5s, moved to 10 min)  for gtest. The value set in AddGoogleTest  is guarded by a env var that nobody sets, in a macro that nobody calls , thus quite ineffective. 
Updated to the latest version of gtest 1.12.1 from 1.11 , i found mentions that it might fix some of the issues. 
Also needed to increase the timeout for the CMakeGenerate builds as they get too close to 1 HR . actually the build failed a few times in this PR due to timeout in that build. 

# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


LarryOsterman:
Thank you very much George.

LarryOsterman (approved):

cmake-modules/AddGoogleTest.cmake
```diff
@@ -15,7 +15,7 @@ if(CMAKE_VERSION VERSION_LESS 3.11)
     download_project(
         PROJ                googletest
 		GIT_REPOSITORY      https://github.com/google/googletest.git
-		GIT_TAG             release-1.11.0
+		GIT_TAG             release-1.12.1
```
ahsonkhan, line 18: This will require updating our cgmanifest.json sha as well to point to the correct commit:
https://github.com/Azure/azure-sdk-for-cpp/blob/2c4900efaa065b50c296059b070ef69bae316f55/sdk/core/azure-core/cgmanifest.json#L45-L51

https://github.com/google/googletest/commit/58d77fa8070e8cec2dc1ed015d66b454c8d78850

.vscode/cspell.json
```diff
@@ -25,7 +25,8 @@
     "*nlohmann-json*",
     "eng/docs/api/assets/**/*",
     "eng/CredScanSuppression.json",
-    "**/test/ut/recordings/*.json"
+    "**/test/ut/recordings/*.json",
+    "**/CMakeLists.txt"
```
ahsonkhan, line 29: Rather than ignoring the entire CMakeLists.txt file (which is useful to have spell check on), add the incorrectly flagged word into the dictionary of "words" below.

Which false positive word does cspell complain about here?

eng/pipelines/templates/jobs/cmake-generate.tests.yml
```diff
@@ -28,9 +28,13 @@ parameters:
   - name: JobName
     type: string
     default: CMakeGenerate
+  - name: TimeoutInMinutes
+    type: number
+    default: 120
 
 jobs:
 - job: ${{ parameters.JobName }}
+  timeoutInMinutes: ${{ parameters.TimeoutInMinutes }}
```
ahsonkhan, line 37: Is this case-sensitive?

I see `timeoutInMinutes` here, but `TimeoutInMinutes` in the parameter list within the `archetype-sdk-client.yml` file.

eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -73,13 +73,16 @@ parameters:
...
+- name: CMakeGenerationTimeoutInMinutes 
+  type: number
+  default: 120
 
 stages:
   - stage: CMakeGeneration
     jobs:
       - template: /eng/pipelines/templates/jobs/cmake-generate-jobs.yml
         parameters:
+          TimeoutInMinutes: ${{ parameters.CMakeGenerationTimeoutInMinutes }}
```
ahsonkhan, line 85: I am not very familiar with how each of the CI stage is laid out in yaml, so trying to understand this a bit.
In this stage and jobs, the `TimeoutInMinutes` is in the parameters list, but it is using the `cmake-generate-jobs.yml` as a template, which has a `timeoutInMinutes` as a jobs property. Why do we need to have 2 timeout variables set, both `CMakeGenerationTimeoutInMinutes` and `TimeoutInMinutes`?

.vscode/cspell.json
ahsonkhan: Looks like it is just `DISCOVERY_TIMEOUT` that is being added in this PR. Consider adding that to the dictionary.

ahsonkhan (approved):
Other than that, looks good to me.

It would be good for someone on the EngSys team to review the changes in the yaml files regarding the timeout parameters and how it is set.

eng/pipelines/templates/jobs/cmake-generate.tests.yml
```diff
@@ -28,9 +28,13 @@ parameters:
   - name: JobName
     type: string
     default: CMakeGenerate
+  - name: TimeoutInMinutes
+    type: number
+    default: 120
```
ahsonkhan, line 33: Looks like the default value of `timeoutInMinutes` is 60 minutes, do we need to bump this up to 2 hours?

Is the reason for this because the test discovery for each test/CMakeLists.txt is sequential, and since we are bumping the test discovery timeout to 10 minutes (from 5 seconds), it is increasing our CI pipeline times?

https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts

RickWinter (approved):
Please add back the CMakelist file.  We need to scan that file.  Instead add exemptions for the words that are failing

cmake-modules/AddGoogleTest.cmake
gearama: @ahsonkhan  Thank you sir , much appreciated.

eng/pipelines/templates/jobs/cmake-generate.tests.yml
gearama: i had it fail a few times in a row due to 1 hr timeout. so i figured since we bump one timeout we might need to bump this one too , the regular builds are already at 2 hours, i bumped them a while back

.vscode/cspell.json
gearama: it wasn't that , the cspell just hang for no reason while checking the cmake files , no errors reported.

eng/pipelines/templates/jobs/cmake-generate.tests.yml
gearama: TimeoutInMinutes is the incoming param , timputInMinutes is the parameter for the job, being set to the incoming value. we have all incoming  params being pascal case, and the job params are all camel case. Weather case sensitive or not i did it for consistency.

eng/pipelines/templates/stages/archetype-sdk-client.yml
gearama: the CMakeGenerationTimeoutInMinutes is a parameter for the script , with a default value of 120. there is another timeout defined LiveTestTimeout with it's default value. I did not want to reuse the LiveTestTimeout as it has another purpose. TimoutInMinutes is a parameter for the next template down the line which does not define any other timeouts thus i could be the first to use timeout in minutes.

sdk/core/azure-core/test/ut/CMakeLists.txt
```diff
@@ -150,7 +150,8 @@ target_link_libraries(azure-core-global-context-test PRIVATE azure-core gtest_ma
 gtest_discover_tests(azure-core-test
      TEST_PREFIX azure-core.
      NO_PRETTY_TYPES
-     NO_PRETTY_VALUES)
+     NO_PRETTY_VALUES
+     DISCOVERY_TIMEOUT 600 )
```
antkmsft, line 154: Nit: in many cmakefiles, there is an extra space: `600 )`. I suggest to remove space (`600)`), or to move `)` to the next line.

antkmsft (approved):

sdk/core/azure-core/test/ut/CMakeLists.txt
gearama: @antkmsft done

Issue #4131, opened 2022-11-28
CI failed when building azure-core


Jinming-Hu:
```
         base64.cpp
         context.cpp
         md5.cpp
         sha_hash.cpp
         datetime.cpp
         environment.cpp
         environment_log_level_listener.cpp
         etag.cpp
         exception.cpp
         bearer_token_authentication_policy.cpp
         http.cpp
         http_sanitizer.cpp
         log_policy.cpp
         policy.cpp
         raw_response.cpp
         request.cpp
         request_activity_policy.cpp
         retry_policy.cpp
         Running Code Analysis for C/C++...
     3>D:\a\_work\1\s\build\vcpkg_installed\x86-windows-static\include\wil\result_macros.h(1324,9): error C2220: the following warning is treated as an error [D:\a\_work\1\s\build\sdk\core\azure-core\azure-core.vcxproj]
     3>D:\a\_work\1\s\build\vcpkg_installed\x86-windows-static\include\wil\result_macros.h(1314): warning C6387: 'Temp_value_#2184' could be '0':  this does not adhere to the specification for the function 'GetProcAddress'. : Lines: 1302, 1303, 1305, 1306, 1308, 1312, 1314 [D:\a\_work\1\s\build\sdk\core\azure-core\azure-core.vcxproj]
         Generating Code...
         Compiling...
         telemetry_policy.cpp
         transport_policy.cpp
         url.cpp
         user_agent.cpp
         body_stream.cpp
         random_access_file_body_stream.cpp
         logger.cpp
         operation_status.cpp
         strings.cpp
         tracing.cpp
         uuid.cpp
         Running Code Analysis for C/C++...
         Generating Code...
     3>Done Building Project "D:\a\_work\1\s\build\sdk\core\azure-core\azure-core.vcxproj" (default targets) -- FAILED.
     1>Done Building Project "D:\a\_work\1\s\build\sdk\storage\azure-storage-common\azure-storage-common.vcxproj" (default targets) -- FAILED.

Build FAILED.

       "D:\a\_work\1\s\build\sdk\storage\azure-storage-common\azure-storage-common.vcxproj" (default target) (1) ->
       "D:\a\_work\1\s\build\sdk\core\azure-core\azure-core.vcxproj" (default target) (3) ->
       (ClCompile target) -> 
[...]

PR #4132, opened 2022-11-28
Disable warning C6387 coming from wil in win_http_transport due to SAL annotation

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/4131

With the vcpkg baseline update of dependencies like `wil` (from https://github.com/Azure/azure-sdk-for-cpp/pull/4100), it introduced this warning (treated as error), which is causing nightly run pipelines to fail on Windows:
https://learn.microsoft.com/en-us/cpp/code-quality/c6387?view=msvc-170

For example:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2013774&view=logs&j=d6204d4c-df7b-58eb-f6e9-300f4a1ec51a&t=9dd83b26-fa53-5361-4278-635b5975bfe8&l=128

Disabling it, similar to https://github.com/Azure/azure-sdk-for-cpp/pull/4054

### DIFF
--- a/sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
+++ b/sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
@@ -17,7 +17,8 @@
 #include <mutex>
 #pragma warning(push)
 #pragma warning(disable : 6553)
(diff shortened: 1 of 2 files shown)


LarryOsterman (approved):

LarryOsterman:
LGTM. And thank you.

PR #4130, opened 2022-11-27
Correct invalid UTF-8 in comment in share_options.hpp

Ariellink:
/storage/files/shares/share_options.hpp

# Pull Request Checklist

Please leverage this checklist as a reminder to address commonly occurring feedback when submitting a pull request to make sure your PR can be reviewed quickly:

See the detailed list in the [contributing guide](https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#pull-requests).

- [ ] [C++ Guidelines](https://azure.github.io/azure-sdk/cpp_introduction.html)
- [ ] Doxygen docs
- [ ] Unit tests
- [ ] No unwanted commits/changes
- [ ] Descriptive title/description
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [x] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] External references or docs updated
- [ ] Self review of PR done
- [ ] [Any]([url]([url](url))) breaking changes?

图片
图片

PR #4134, opened 2022-11-28
Add CI pipeline step and script that validates all files in the repo contain ASCII-only bytes and no UTF-8 BOM

ahsonkhan:
The goal is to detect issues like this and remain green by making it a part of CI:
https://github.com/Azure/azure-sdk-for-cpp/pull/4130

This is similar to what we do in Embedded C:
https://github.com/Azure/azure-sdk-for-c/blob/cd62e2b78a396c3d5478d3b60c046adb72687274/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L369-L390


antkmsft (approved):
`LibcurlTransportAdapter.md` has the `’` character. If this catches it, we're good.

antkmsft:
It did catch `’` in the `LibcurlTransportAdapter.md`, great!

As for the errors in unit tests like `test/nlohmann-json-test/src/unit-inspection.cpp:301`: `CHECK(json("❤️").dump() == "\"❤️\"");`, if it was in our tests, I'd say we should rewrite string literals using escape characters (for example, `"❤️"` can be written as `"\u2764\uFE0F"` or as `"\xE2\x9D\xA4\xEF\xB8\x8F"`).

But given that it is in 3rd party code, CI probably should be ignoring errors in these files.

ahsonkhan:
Yep, that's the plan. Will filter out test recordings and nlohmann json. The main focus is product/SDK code.

antkmsft:
@ahsonkhan, if this PR is close to achieving its goal, can we get to merge it soon?

LarryOsterman (approved):

PR #4142, opened 2022-11-30
Fixed #4137 - Don't close CURL connection on shutdown

### LINKED ISSUE #4137: Bug: libcurl transport layer doesn't correctly handle `Connection: close` response header
Storage team recently found this when investigating on an RST issue between C++ clients and storage service. We're seeing abnormally high number of RST in TCP connections. After reading the source code, I found libcurl transport layer didn't correctly handle "connection: close" response header.

at Line 872 of `curl.cpp`, the connection is immediately closed the moment we see this response header.

https://github.com/Azure/azure-sdk-for-cpp/blob/121c103d3baf267c12f8943c7cbafcd8c08a3e7d/sdk/core/azure-core/src/http/curl/curl.cpp#L872

But in [RFC2616, section 14.10](https://www.rfc-editor.org/rfc/rfc2616#section-14.10), it says: the connection SHOULD NOT be considered `persistent' **after** the current request/response is complete.

The expected behavior should be: client side reads response body as normal. After that, instead of putting this connection back to the pool, it closes this connection.

LarryOsterman:
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -557,6 +557,12 @@ CURLcode CurlConnection::SendBuffer(
     size_t bufferSize,
     Context const& context)
 {
+  // Once you've shutdown the connection, we can't send any more data (although we can continue to
+  // receive).
+  if (IsShutdown())
+  {
+    return CURLE_SEND_ERROR;
+  }
   for (size_t sentBytesTotal = 0; sentBytesTotal < bufferSize;)
   {
     // check cancelation for each chunk of data.
@@ -1063,16 +1069,6 @@ size_t CurlSession::OnRead(uint8_t* buffer, size_t count, Context const& context
   }
 
   return totalRead;
-}
-
-void CurlConnection::Shutdown()
-{
-#if defined(AZ_PLATFORM_POSIX)
-  ::shutdown(m_curlSocket, SHUT_RDWR);
-#elif defined(AZ_PLATFORM_WINDOWS)
-  ::shutdown(m_curlSocket, SD_BOTH);
-#endif
-  CurlNetworkConnection::Shutdown();
 }
 
 // Read from socket and return the number of bytes taken from socket
--- a/sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
+++ b/sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
@@ -601,6 +601,30 @@ namespace Azure { namespace Core { namespace Test {
       }
     }
 
+    TEST(CurlConnectionPool, forceConnectionClosed)
+    {
+      Azure::Core::Http::Request req(
+          Azure::Core::Http::HttpMethod::Get,
+          Azure::Core::Url(AzureSdkHttpbinServer::Status(101)));
+
+      Azure::Core::Http::CurlTransportOptions options;
+      auto connection
+          = CurlConnectionPool::g_curlConnectionPool.ExtractOrCreateCurlConnection(req, options);
+
+      {
+        // Check we can use the connection to retrieve headers in the response, but the connection is still flagged as shutdown.
+        auto session
+            = std::make_unique<Azure::Core::Http::CurlSession>(req, std::move(connection), options);
+
+        auto r = session->Perform(Azure::Core::Context::ApplicationContext);
+        EXPECT_EQ(CURLE_OK, r);
+        auto response = session->ExtractResponse();
+        EXPECT_EQ(
+            response->GetStatusCode(), Azure::Core::Http::HttpStatusCode::SwitchingProtocols);
+        EXPECT_EQ("close", response->GetHeaders().find("Connection")->second);
+      }
+    }
+
     TEST(CurlConnectionPool, connectionClose)
     {
       /// When getting the header connection: close from an HTTP response, the connection should not
--- a/sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
+++ b/sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
@@ -604,23 +604,22 @@ namespace Azure { namespace Core { namespace Test {
     TEST(CurlConnectionPool, forceConnectionClosed)
     {
       Azure::Core::Http::Request req(
-          Azure::Core::Http::HttpMethod::Get,
-          Azure::Core::Url(AzureSdkHttpbinServer::Status(101)));
+          Azure::Core::Http::HttpMethod::Get, Azure::Core::Url(AzureSdkHttpbinServer::Status(101)));
 
       Azure::Core::Http::CurlTransportOptions options;
       auto connection
           = CurlConnectionPool::g_curlConnectionPool.ExtractOrCreateCurlConnection(req, options);
 
       {
-        // Check we can use the connection to retrieve headers in the response, but the connection is still flagged as shutdown.
+        // Check we can use the connection to retrieve headers in the response, but the connection
+        // is still flagged as shutdown.
         auto session
             = std::make_unique<Azure::Core::Http::CurlSession>(req, std::move(connection), options);
 
         auto r = session->Perform(Azure::Core::Context::ApplicationContext);
         EXPECT_EQ(CURLE_OK, r);
         auto response = session->ExtractResponse();
-        EXPECT_EQ(
-            response->GetStatusCode(), Azure::Core::Http::HttpStatusCode::SwitchingProtocols);
+        EXPECT_EQ(response->GetStatusCode(), Azure::Core::Http::HttpStatusCode::SwitchingProtocols);
         EXPECT_EQ("close", response->GetHeaders().find("Connection")->second);
       }
     }
--- a/sdk/core/azure-core/src/http/curl/curl_connection_private.hpp
+++ b/sdk/core/azure-core/src/http/curl/curl_connection_private.hpp
@@ -230,8 +230,6 @@ namespace Azure { namespace Core {
        */
       CURLcode SendBuffer(uint8_t const* buffer, size_t bufferSize, Context const& context)
           override;
-
-      void Shutdown() override;
     };
   } // namespace Http
 }} // namespace Azure::Core
--- a/sdk/core/azure-core/test/ut/transport_adapter_base_test.hpp
+++ b/sdk/core/azure-core/test/ut/transport_adapter_base_test.hpp
@@ -33,6 +33,10 @@ namespace Azure { namespace Core { namespace Test {
     inline static std::string Delete() { return Schema() + "://" + Host() + "/delete"; }
     inline static std::string Patch() { return Schema() + "://" + Host() + "/patch"; }
     inline static std::string Delay() { return Schema() + "://" + Host() + "/delay"; }
+    inline static std::string Status(int statusCode)
+    {
+      return Schema() + "://" + Host() + "/status/" + std::to_string(statusCode);
+    }
     inline static std::string Host() { return std::string(_detail::AzureSdkHttpbinServer); }
     inline static std::string Schema() { return std::string(_detail::AzureSdkHttpbinServerSchema); }
   };


sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -557,6 +557,12 @@ CURLcode CurlConnection::SendBuffer(
     size_t bufferSize,
     Context const& context)
 {
+  // Once you've shutdown the connection, we can't send any more data (although we can continue to
+  // receive).
+  if (IsShutdown())
```
RickWinter, line 562: Sockets allow you to shutdown the socket for sends while keeping the socket open for receives.  Is there a reason we don't want that behavior here?

https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown

After your change, Where do we shutdown the socket?

Jinming-Hu:
I fixed this issue in a different way. https://github.com/Azure/azure-sdk-for-cpp/issues/4137#issuecomment-1331858548

I'm not quite familiar with the libcurl transport layer implementation. You decide.

Jinming-Hu:
I don't get it. How does the change in this PR prevent the connection being moved back into the pool?

sdk/core/azure-core/src/http/curl/curl.cpp
LarryOsterman: We shut down the socket when the connection is released (via curl_easy_shutdown).

With my change, if a Connection: close is set, we won't put the connection back into the connection pool, it will just be normally freed.

That allows all the normal operations to be performed on the socket, most importantly, if the operation is a "put" operation and the server responds with "Connection: close", if we called shutdown to block sends, then the "put" would fail.

The new behavior is consistent with the RFC mentioned in the original bug report: Connection: close means "don't persist the connection after this HTTP transaction".

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -1065,16 +1071,6 @@ size_t CurlSession::OnRead(uint8_t* buffer, size_t count, Context const& context
...
-void CurlConnection::Shutdown()
-{
-#if defined(AZ_PLATFORM_POSIX)
-  ::shutdown(m_curlSocket, SHUT_RDWR);
-#elif defined(AZ_PLATFORM_WINDOWS)
-  ::shutdown(m_curlSocket, SD_BOTH);
-#endif
-  CurlNetworkConnection::Shutdown();
-}
-
```
vhvb1989, line 1077: This used to be there to test how we react after a connection lost.
There was a test like: https://github.com/Azure/azure-sdk-for-cpp/pull/1721/files#diff-34f6f1c9be490d635445c2e8956dc6aa5f3c6c96f96a33ba9b658fa0f766465cR183

Are we replacing that test to actually make a call to `AzureSdkHttpbinServer::Status(101)` which is expected to shutdown a connection?

All of this logic around `shutdown`  is trying to solve this issue: https://github.com/Azure/azure-sdk-for-cpp/issues/1716

I think, at some point, we started to use the `shutdown` method as a way to avoid a connection to be re-used, which is not good.  In that case, removing the `shutdown` method makes total sense, as that was something for testing only, and not expected to be used by the client implementation.

vhvb1989 (approved):

sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
```diff
@@ -601,6 +601,30 @@ namespace Azure { namespace Core { namespace Test {
       }
     }
 
+    TEST(CurlConnectionPool, forceConnectionClosed)
```
ahsonkhan, line 604: Can we have a test that resembles the real-life scenario that this PR is fixing, a bit more?

I don't follow how the PR is fixing the issue, other than returning a `CURL_SEND_ERROR` in one case, in `SendBuffer`.

LarryOsterman:
> I fixed this issue in a different way. [#4137 (comment)](https://github.com/Azure/azure-sdk-for-cpp/issues/4137#issuecomment-1331858548)
> 
> I'm not quite familiar with the libcurl transport layer implementation. You decide.

Calling Shutdown ensures that the connection is not recycled into the connection pool. If you don't ensure that the connection isn't recycled, then it defeats the purpose of `Connection::close`.

sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
LarryOsterman: I believe that this test should *exactly* reflect the real-life scenario that this PR is fixing. We call into the azuresdkforcpp test server asking for a 102 response, which also includes a Connection: close response. Before my fix, the call to Perform would have failed, after the fix, the call to Perform succeeds and returns headers indicating that the connection was closed.

Could you help me understand why it doesn't reflect the real-life scenario?

sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
```diff
@@ -601,6 +601,30 @@ namespace Azure { namespace Core { namespace Test {
...
+    {
+      Azure::Core::Http::Request req(
+          Azure::Core::Http::HttpMethod::Get,
+          Azure::Core::Url(AzureSdkHttpbinServer::Status(101)));
+
+      Azure::Core::Http::CurlTransportOptions options;
+      auto connection
+          = CurlConnectionPool::g_curlConnectionPool.ExtractOrCreateCurlConnection(req, options);
```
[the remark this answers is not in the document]
LarryOsterman: There is a way we can test the internals of the transport.

We can define "interfaces" (pure virtual base classes) for the underlying socket implementation (ie a CURL interface and a WinHTTP interface) and then mock that interface in our test collateral. We can then call into the transport with the mocked CURL/WinHTTP and have the CURL/WinHTTP behave in whatever fashion we want.

It's a lot of work, but it is completely possible (it will be rather challenging to do this for WinHTTP since we're now using WinHTTP asynchronously, but it is absolutely possible).

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -1065,16 +1071,6 @@ size_t CurlSession::OnRead(uint8_t* buffer, size_t count, Context const& context
   return totalRead;
 }
 
-void CurlConnection::Shutdown()
```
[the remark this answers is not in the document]
LarryOsterman: Yes - that's why I had to remove it (it was called when a Connection: close header was received).

The problem in this bug is that the `CurlConnection::Shutdown` method shut down the underlying socket, which meant that we didn't allow processing any subsequent "recv" calls, which violated the RFC.

According to the bug, the RFC says that the *only* thing we should be doing in receipt of a `Connection: close` call is to prevent the connection from being re-used, we should NOT hang up the connection.

sdk/core/azure-core/src/http/curl/curl.cpp
LarryOsterman: FWIW, I keep that test working (although I'm suspicious of the test because it may cause "put" and "patch" commands to malfunction if they receive a connection: close header.

sdk/core/azure-core/src/http/curl/curl.cpp
LarryOsterman: > I don't get it. How does the change in this PR prevent the connection being moved back into the pool?

When a connection is marked as "Shutdown", it won't be put back into the pool (see `CurlConnectionPool::MoveConnectionBackToPool` which ignores connections marked as `Shutdown`).

What this PR does is to prevent the underlying socket used in the connection from being shut down, which has the effect of allowing the connection to be processed normally (sends and receives continue to work).

Based on the text you quoted from the RFC, all `Connection: close` is supposed to do is to prevent a connection from being re-used, which is what the `CurlConnection::Shutdown` function does.

LarryOsterman:
> I don't get it. How does the change in this PR prevent the connection being moved back into the pool?

When a connection is marked as "Shutdown", it won't be put back into the pool (see `CurlConnectionPool::MoveConnectionBackToPool` which ignores connections marked as `Shutdown`).

What this PR does is to prevent the underlying socket used in the connection from being shut down, which has the effect of allowing the connection to be processed normally (sends and receives continue to work).

Based on the text you quoted from the RFC, all `Connection: close` is supposed to do is to prevent a connection from being re-used, which is what the `CurlConnection::Shutdown` function does.

sdk/core/azure-core/src/http/curl/curl.cpp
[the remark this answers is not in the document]
ahsonkhan: I guess what I am trying to understand is, with this method override gone, what is this call supposed to be doing?
https://github.com/Azure/azure-sdk-for-cpp/blob/b3b2b6774d36bfbe1d0b08174282341152af4a27/sdk/core/azure-core/src/http/curl/curl.cpp#L871-L872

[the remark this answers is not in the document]
LarryOsterman: Without the virtual override, this calls the shutdown method on the base CurlConnection object.

The shutdown method on the base CurlConnection object just sets the "m_isShutdown" member variable, which tells the connection cache to not cache the connection.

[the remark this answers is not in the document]
LarryOsterman: Btw, I misspoke. The virtual CurlNetworkConnection::Shutdown called ::shutdown to force the socket closed. THat was the root of the bug. My fix was to remove that force-close.

Jinming-Hu (approved):

Jinming-Hu:
@LarryOsterman When can we merge this PR? I happen to have an environment that can repro this bug. I want to try it out.

ahsonkhan:
@Jinming-Hu you could try and validate that this PR fixes the issue, without waiting for the PR to be merged. Just checkout Larry's branch:
https://github.com/LarryOsterman/azure-sdk-for-cpp/tree/larryo/Bug4137

Jinming-Hu:
> @Jinming-Hu you could try and validate that this PR fixes the issue, without waiting for the PR to be merged. Just checkout Larry's branch: https://github.com/LarryOsterman/azure-sdk-for-cpp/tree/larryo/Bug4137

Verification of the fix is not an easy job. It may take several hours. Even if I test this PR, I'll still verify after it's merged into `main` branch.

antkmsft:
CI is currently failing on Windows due to #4152.

Jinming-Hu:
I've tested. Now I can confirm #4137 can be fixed with this PR.

RickWinter (approved):

antkmsft:
@LarryOsterman probably won't get to merge this PR himself soon. After consultations within our team, we decided to merge this PR. Everyone approved it, and @Jinming-Hu has tested it. So I am going to resolve all the comment threads so GitHub allows merging, and will merge this PR.

sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
antkmsft: @ahsonkhan, I am going to resolve this comment thread now so that I can merge this PR.
If you'd like more follow up your question, please create a work item so that we can follow up on it later.

sdk/core/azure-core/test/ut/curl_connection_pool_test.cpp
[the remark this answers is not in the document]
antkmsft: @ahsonkhan, I am going to resolve this comment thread now so that I can merge this PR.
If you'd like more follow up your question, please create a work item so that we can follow up on it later.

sdk/core/azure-core/src/http/curl/curl.cpp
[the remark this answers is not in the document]
antkmsft: @ahsonkhan, I am going to resolve this comment thread now so that I can merge this PR.
If you'd like more follow up your question, please create a work item so that we can follow up on it later.

sdk/core/azure-core/src/http/curl/curl.cpp
antkmsft: @vhvb1989, I am going to resolve this comment thread now so that I can merge this PR.
If you'd like more follow up your question, please create a work item so that we can follow up on it later.

sdk/core/azure-core/src/http/curl/curl.cpp
antkmsft: @RickWinter, I am going to resolve this comment thread now so that I can merge this PR.
If you'd like more follow up your question, please create a work item so that we can follow up on it later.

PR #4167, opened 2022-12-07
Use nlohmann::json to parse Identity token

### LINKED ISSUE #4155: Identity: use nlohmann JSON from `Core` to parse token
Feedback from #4146.

antkmsft:
Closes #4155.


sdk/identity/azure-identity/src/token_credential_impl.cpp
```diff
@@ -161,140 +152,52 @@ AccessToken TokenCredentialImpl::ParseToken(
...
+    }
+    else
     {
-      auto c = str[pos];
-      if (c == '\"' || c == '\'')
+      auto expiresOn = parsedJson[expiresOnPropertyName].get<std::string>();
       {
-        break;
+        auto const spacePos = expiresOn.find(' ');
+        if (spacePos != std::string::npos) // LCOV_EXCL_LINE
```
RickWinter, line 187:
```suggestion
        if (spacePos != std::string::npos) 
```

RickWinter (approved):

sdk/identity/azure-identity/test/ut/token_credential_impl_test.cpp
```diff
@@ -340,72 +340,6 @@ TEST(TokenCredentialImpl, NoToken)
       }));
 }
 
-TEST(TokenCredentialImpl, CurrentJsonParserQuirksAndLimitations)
```
ahsonkhan, line 343: Should we keep this test?

What this is testing seems to be useful, regardless of the JSON parser used in the implementation.

It isn't directly testing the helpers like `GetPropertyValueAsInt64`, so I would expect it to pass even when the impl now uses nlohmann.

sdk/identity/azure-identity/test/ut/azure_cli_credential_test.cpp
```diff
@@ -290,3 +290,21 @@ TEST(AzureCliCredential, UnsafeChars)
     EXPECT_THROW(static_cast<void>(azCliCred.GetToken(trc, {})), AuthenticationException);
   }
 }
+
+TEST(AzureCliCredential, StrictIso8601TimeFormat)
```
ahsonkhan, line 294: Good test.

sdk/identity/azure-identity/src/token_credential_impl.cpp
```diff
@@ -161,140 +152,52 @@ AccessToken TokenCredentialImpl::ParseToken(
...
+    }
+    else
     {
-      auto c = str[pos];
-      if (c == '\"' || c == '\'')
+      auto expiresOn = parsedJson[expiresOnPropertyName].get<std::string>();
       {
-        break;
+        auto const spacePos = expiresOn.find(' ');
+        if (spacePos != std::string::npos)
```
ahsonkhan: I wish this custom logic wasn't needed to modify the input JSON before turning it into our DateTime.
Why do we need to do this? Do we need a different parse method?

sdk/identity/azure-identity/src/token_credential_impl.cpp
```diff
@@ -161,140 +152,52 @@ AccessToken TokenCredentialImpl::ParseToken(
...
+
+      accessToken.ExpiresOn
+          = Azure::DateTime::Parse(expiresOn, Azure::DateTime::DateFormat::Rfc3339);
     }
-  }
 
-  return pos;
+    return accessToken;
+  }
+  catch (Azure::Core::Json::_internal::json::parse_error const& ex)
```
ahsonkhan, line 199: Is the whole purpose of this try-catch to avoid leaking the internal nlohmann JSON's parse exception, or is there some other reason?

Is this a pattern we want to follow everywhere we do JSON parsing? I think we have places where the error just bubbles up, without us wrapping the code in a try-catch and our own exception.

sdk/identity/azure-identity/src/token_credential_impl.cpp
```diff
@@ -161,140 +152,52 @@ AccessToken TokenCredentialImpl::ParseToken(
...
-    auto const jsonStringSize = jsonString.size();
-    for (; pos < jsonStringSize; ++pos)
+    else if (expiresOnPropertyName.empty())
     {
-      auto c = jsonString[pos];
-      if (c != ':' && c != ' ' && c != '\"' && c != '\'')
-      {
-        break;
-      }
+      ThrowMissingJsonPropertyError(expiresInPropertyName);
```
ahsonkhan, line 176: With valid response payloads from the service, will these error throw cases ever happen?

ahsonkhan (approved):
Some comments and questions, but looks good overall.

sdk/identity/azure-identity/src/token_credential_impl.cpp
```diff
@@ -161,140 +152,52 @@ AccessToken TokenCredentialImpl::ParseToken(
...
-      if (c < '0' || c > '9')
-      {
-        break;
-      }
-
-      value = (value * 10) + (static_cast<int64_t>(c) - '0');
+      accessToken.ExpiresOn
+          = std::chrono::system_clock::now()
+          + std::chrono::seconds(
+                parsedJson[expiresInPropertyName].get<std::chrono::seconds::duration::rep>());
```
Jinming-Hu, line 172: This field is just an integer representing number of seconds, right? why not just use `.get<int>` or `.get<int64_t>`?

antkmsft: It is implementation-defined, the only thing being said is it should be "signed integer type of at least 35 bits".
This would allow it to work correctly on all possible implementations, and to avoid any possible warnings.
https://en.cppreference.com/w/cpp/chrono/duration

sdk/identity/azure-identity/src/token_credential_impl.cpp
antkmsft: Since we use nlohmann json internally, I think we should be wrapping these exceptions.
Speaking of authentication space, that is even more important, because if something else other than `AuthenticationException` is getting thrown, it becomes hard to distinguish whether the exception came from the request, or from the authentication pipeline.

sdk/identity/azure-identity/src/token_credential_impl.cpp
antkmsft: So far, it is the only known input that has space in it instead of "T". If we have more, we can add (internally first, I think) another parse format support. It is easy to modify the parser to allow space there.
But, I would check with Jeffrey first to ask if he thinks it is safe to update the parser, and whether that still can be called `DateFormat::Rfc3339`.

RFC 3339 itself says:
"_NOTE: ISO 8601 defines date and time separated by "T". Applications using this syntax may choose, for the sake of readability, to specify a full-date and full-time separated by (say) a space character._"
https://www.rfc-editor.org/rfc/rfc3339

antkmsft: #4171 4171

sdk/identity/azure-identity/test/ut/token_credential_impl_test.cpp
antkmsft: Other tests cover parsing ints, and all other things, they just are not being touched in this PR, because they all pass as expected. But they have sufficient coverage.
This test is not needed anymore. Previous implementation would successfully parse a string that starts with double quote and "ends with" a single quote, and these quirks were being tested. Now that the implementation is updated, this test has started to fail as expected, and it is no longer needed.

PR #4180, opened 2022-12-10
Clean up token cache from expired items on Fibonacci cache sizes instead of 2^N

### LINKED ISSUE #4161: Token cache: clean up on Fibonacci numbers instead of powers of two
@LarryOsterman's recommendation was to use Fibonacci numbers instead of powers of two for token cache cleanup:

"_I would suggest a Fibonacci series rather than a power of two – for memory allocations, Fibonacci seems to produce more optimal chunking because powers of two grow too quickly. [...]_

_[M]y experience here is in heap allocation strategies (when you run out of blocks in the current heap chunk, how large do you allocate for the next heap chunk – powers of two grow way too fast and are inefficient (most allocations tend to be in a relatively stable size range and thus a growth algorithm which grows somewhat slowly uses memory more efficiently._

_The size of the cache is somewhat analogous to the growth of memory on the heap._"

IsFibonacci() can be implemented by using a sorted array of values + `std::binary_search()`. An array of values will be rather small, given that the 2^64 range can be covered with just 93 numbers, and we need even less than that.

antkmsft:
Closes #4161.

### DIFF
--- a/sdk/identity/azure-identity/src/token_cache.cpp
+++ b/sdk/identity/azure-identity/src/token_cache.cpp
@@ -3,6 +3,9 @@
 
 #include "azure/identity/detail/token_cache.hpp"
 
+#include <algorithm>
+#include <array>
+#include <limits>
 #include <mutex>
 
 using Azure::Identity::_detail::TokenCache;
@@ -16,6 +19,10 @@ bool TokenCache::IsFresh(
     std::chrono::system_clock::time_point now)
 {
   return item->AccessToken.ExpiresOn > (DateTime(now) + minimumExpiration);
+}
+
+namespace {
+template <typename T> bool ShouldCleanUpCacheFromExpiredItems(T cacheSize);
 }
 
 std::shared_ptr<TokenCache::CacheValue> TokenCache::GetOrCreateValue(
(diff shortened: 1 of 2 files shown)


gearama (approved):

sdk/identity/azure-identity/src/token_cache.cpp
```diff
@@ -120,3 +117,44 @@ AccessToken TokenCache::GetToken(
...
+constexpr auto SortedFibonacciSequence<T, L, R, N, X>::Get()
+{
+  auto result = SortedFibonacciSequence<T, R, R + L, N + 1>::Get();
+  result[N] = L;
+  return result;
+}
+
+template <typename T> bool ShouldCleanUpCacheFromExpiredItems(T cacheSize)
+{
+  static auto const Fibonacci = SortedFibonacciSequence<T, 1, 2>::Get();
```
ahsonkhan, line 156: Why don't we need to specify a `size_T N` counter? What is that template parameter meant for?

antkmsft: It is needed for array - so that when template "stops", we know the size of std::array that we need to create. And on each step we use N to assign the current sum to the corresponding index in the resulting array.

PR #4181, opened 2022-12-10
ClientCertificateCredential: Add AuthorityHost override and Azure Stack support

antkmsft:
Also:
* Use RAII types in `ClientCertificateCredential` implementation
* Replace stringstreams with string concatenations where it makes sense
* Preparation step towards challenge-based/multitenant auth - with what `ClientCredentialCore` does, it will be easier to extend it to do what needs to be done for multi-tenant auth.

### DIFF
--- a/sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
+++ b/sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
@@ -8,10 +8,12 @@
 
 #pragma once
 
+#include "azure/identity/detail/client_credential_core.hpp"
 #include "azure/identity/detail/token_cache.hpp"
 
 #include <azure/core/credentials/credentials.hpp>
 #include <azure/core/credentials/token_credential_options.hpp>
+#include <azure/core/internal/unique_handle.hpp>
 #include <azure/core/url.hpp>
 
 #include <memory>
@@ -20,6 +22,18 @@
 namespace Azure { namespace Identity {
   namespace _detail {
     class TokenCredentialImpl;
+
+    void FreePkeyIfNotNull(void* pkey);
+
+    template <typename> struct UniquePkeyHelper;
+    template <> struct UniquePkeyHelper<void*>
+    {
+      static void FreePkey(void* pkey) { FreePkeyIfNotNull(pkey); }
+
+      using type = Azure::Core::_internal::BasicUniqueHandle<void, FreePkey>;
+    };
+
+    using UniquePkeyHandle = Azure::Core::_internal::UniqueHandle<void*, UniquePkeyHelper>;
   } // namespace _detail
 
   /**
@@ -28,6 +42,15 @@ namespace Azure { namespace Identity {
    */
   struct ClientCertificateCredentialOptions final : public Core::Credentials::TokenCredentialOptions
   {
+    /**
+     * @brief Authentication authority URL.
+     * @note Default value is Azure AD global authority (https://login.microsoftonline.com/).
+     *
+     * @note Example of a \p authority string: "https://login.microsoftonline.us/". See national
+     * clouds' Azure AD authentication endpoints:
+     * https://docs.microsoft.com/azure/active-directory/develop/authentication-national-cloud.
+     */
+    std::string AuthorityHost = _detail::ClientCredentialCore::AadGlobalAuthority;
   };
 
   /**
@@ -38,12 +61,19 @@ namespace Azure { namespace Identity {
   class ClientCertificateCredential final : public Core::Credentials::TokenCredential {
   private:
     _detail::TokenCache m_tokenCache;
+    _detail::ClientCredentialCore m_clientCredentialCore;
     std::unique_ptr<_detail::TokenCredentialImpl> m_tokenCredentialImpl;
-    Core::Url m_requestUrl;
     std::string m_requestBody;
-    std::string m_tokenHeaderEncoded;
     std::string m_tokenPayloadStaticPart;
-    void* m_pkey;
+    std::string m_tokenHeaderEncoded;
+    _detail::UniquePkeyHandle m_pkey;
+
+    explicit ClientCertificateCredential(
+        std::string tenantId,
+        std::string const& clientId,
+        std::string const& clientCertificatePath,
+        std::string const& authorityHost,
+        Core::Credentials::TokenCredentialOptions const& options);
 
   public:
     /**
@@ -55,7 +85,7 @@ namespace Azure { namespace Identity {
      * @param options Options for token retrieval.
      */
     explicit ClientCertificateCredential(
-        std::string const& tenantId,
+        std::string tenantId,
         std::string const& clientId,
         std::string const& clientCertificatePath,
         Core::Credentials::TokenCredentialOptions const& options
@@ -70,7 +100,7 @@ namespace Azure { namespace Identity {
      * @param options Options for token retrieval.
      */
     explicit ClientCertificateCredential(
-        std::string const& tenantId,
+        std::string tenantId,
         std::string const& clientId,
         std::string const& clientCertificatePath,
         ClientCertificateCredentialOptions const& options);
--- a/sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
+++ b/sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
@@ -23,13 +23,12 @@ namespace Azure { namespace Identity {
   namespace _detail {
     class TokenCredentialImpl;
 
-    void FreePkeyIfNotNull(void* pkey);
+    void FreePkeyImpl(void* pkey);
 
     template <typename> struct UniquePkeyHelper;
     template <> struct UniquePkeyHelper<void*>
     {
-      static void FreePkey(void* pkey) { FreePkeyIfNotNull(pkey); }
-
+      static void FreePkey(void* pkey) { FreePkeyImpl(pkey); }
       using type = Azure::Core::_internal::BasicUniqueHandle<void, FreePkey>;
     };
 
@@ -46,7 +45,7 @@ namespace Azure { namespace Identity {
      * @brief Authentication authority URL.
      * @note Default value is Azure AD global authority (https://login.microsoftonline.com/).
      *
-     * @note Example of a \p authority string: "https://login.microsoftonline.us/". See national
+     * @note Example of an authority host string: "https://login.microsoftonline.us/". See national
      * clouds' Azure AD authentication endpoints:
      * https://docs.microsoft.com/azure/active-directory/develop/authentication-national-cloud.
      */
--- a/sdk/identity/azure-identity/inc/azure/identity/client_secret_credential.hpp
+++ b/sdk/identity/azure-identity/inc/azure/identity/client_secret_credential.hpp
@@ -8,8 +8,8 @@
 
 #pragma once
 
+#include "azure/identity/detail/client_credential_core.hpp"
 #include "azure/identity/detail/token_cache.hpp"
-#include "azure/identity/dll_import_export.hpp"
 
 #include <azure/core/credentials/credentials.hpp>
 #include <azure/core/credentials/token_credential_options.hpp>
@@ -21,7 +21,6 @@
 namespace Azure { namespace Identity {
   namespace _detail {
     class TokenCredentialImpl;
-    AZ_IDENTITY_DLLEXPORT extern std::string const g_aadGlobalAuthority;
   } // namespace _detail
 
   /**
@@ -38,7 +37,7 @@ namespace Azure { namespace Identity {
      * clouds' Azure AD authentication endpoints:
      * https://docs.microsoft.com/azure/active-directory/develop/authentication-national-cloud.
      */
-    std::string AuthorityHost = _detail::g_aadGlobalAuthority;
+    std::string AuthorityHost = _detail::ClientCredentialCore::AadGlobalAuthority;
   };
 
   /**
@@ -49,21 +48,15 @@ namespace Azure { namespace Identity {
   class ClientSecretCredential final : public Core::Credentials::TokenCredential {
   private:
     _detail::TokenCache m_tokenCache;
+    _detail::ClientCredentialCore m_clientCredentialCore;
     std::unique_ptr<_detail::TokenCredentialImpl> m_tokenCredentialImpl;
-    Core::Url m_requestUrl;
     std::string m_requestBody;
-
-    std::string m_tenantId;
-    std::string m_clientId;
-    std::string m_authorityHost;
-
-    bool m_isAdfs;
 
     ClientSecretCredential(
         std::string tenantId,
-        std::string clientId,
+        std::string const& clientId,
         std::string const& clientSecret,
-        std::string authorityHost,
+        std::string const& authorityHost,
         Core::Credentials::TokenCredentialOptions const& options);
 
   public:
@@ -77,7 +70,7 @@ namespace Azure { namespace Identity {
      */
     explicit ClientSecretCredential(
         std::string tenantId,
-        std::string clientId,
+        std::string const& clientId,
         std::string const& clientSecret,
         ClientSecretCredentialOptions const& options);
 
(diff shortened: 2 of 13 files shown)


sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
```diff
@@ -28,6 +42,15 @@ namespace Azure { namespace Identity {
    */
   struct ClientCertificateCredentialOptions final : public Core::Credentials::TokenCredentialOptions
   {
+    /**
+     * @brief Authentication authority URL.
+     * @note Default value is Azure AD global authority (https://login.microsoftonline.com/).
+     *
+     * @note Example of a \p authority string: "https://login.microsoftonline.us/". See national
```
ahsonkhan, line 49: Isn't \p meant for parameter? Do we need it for `authority`?

sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
```diff
@@ -55,7 +85,7 @@ namespace Azure { namespace Identity {
      * @param options Options for token retrieval.
      */
     explicit ClientCertificateCredential(
-        std::string const& tenantId,
+        std::string tenantId,
```
ahsonkhan, line 88: Why is this change needed?

Going from const to non-const is non-breaking, so we are allowed to do so.

sdk/identity/azure-identity/inc/azure/identity/client_secret_credential.hpp
```diff
@@ -77,7 +70,7 @@ namespace Azure { namespace Identity {
      */
     explicit ClientSecretCredential(
         std::string tenantId,
-        std::string clientId,
+        std::string const& clientId,
```
ahsonkhan, line 73: Isn't this breaking?

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <azure/core/credentials/credentials.hpp>
+#include <azure/core/url.hpp>
+
+#include <string>
+
+namespace Azure { namespace Identity { namespace _detail {
+  class ClientCredentialCore final {
+    Core::Url m_authorityHost;
+    std::string m_tenantId;
+    bool m_isAdfs;
```
ahsonkhan, line 17: Curious: What's Adfs?

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+  public:
+    AZ_IDENTITY_DLLEXPORT static std::string const AadGlobalAuthority;
+
+    explicit ClientCredentialCore(std::string tenantId, std::string const& authorityHost);
+
+    Core::Url GetRequestUrl() const;
+
+    std::string GetScopesString(decltype(Core::Credentials::TokenRequestContext::Scopes)
```
ahsonkhan, line 26: Can we  use a regular function signature declaration here rather than `decltype`?

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include "azure/identity/dll_import_export.hpp"
+
+#include <azure/core/credentials/credentials.hpp>
+#include <azure/core/url.hpp>
+
+#include <string>
+
+namespace Azure { namespace Identity { namespace _detail {
+  class ClientCredentialCore final {
```
ahsonkhan, line 14: Why do you think it's valuable to have this private type rather than the fields be part of `ClientCredential` as they were previously?

What does "Core" imply here?

sdk/identity/azure-identity/src/environment_credential.cpp
```diff
@@ -53,18 +56,28 @@ EnvironmentCredential::EnvironmentCredential(
     // }
     else if (!clientCertificatePath.empty())
     {
-      m_credentialImpl.reset(
-          new ClientCertificateCredential(tenantId, clientId, clientCertificatePath, options));
+      if (!authority.empty())
+      {
+        ClientCertificateCredentialOptions clientCertificateCredentialOptions;
+        static_cast<TokenCredentialOptions&>(clientCertificateCredentialOptions) = options;
```
ahsonkhan, line 62: What does `static_cast` on the L.H.S. mean? How should I be reading this line and is it useful to avoid the cast here?

sdk/identity/azure-identity/test/ut/client_certificate_credential_test.cpp
```diff
@@ -165,6 +165,269 @@ TEST(ClientCertificateCredential, Regular)
   EXPECT_LE(response1.AccessToken.ExpiresOn, response1.LatestExpiration + 7200s);
 }
 
+TEST(ClientCertificateCredential, AzureStack)
```
ahsonkhan, line 168: What aspects of this test is related to Azure Stack? It isn't clear to me, so trying to understand better.

sdk/identity/azure-identity/src/client_certificate_credential.cpp
```diff
@@ -36,214 +47,217 @@ template <typename T> std::vector<uint8_t> ToUInt8Vector(T const& in)
 
   return outVec;
 }
+
+template <typename> struct UniqueHandleHelper;
```
ahsonkhan, line 51: Do we need such helper types when using the RAII pattern with `BasicUniqueHandle`?

sdk/identity/azure-identity/src/client_certificate_credential.cpp
```diff
@@ -36,214 +47,217 @@ template <typename T> std::vector<uint8_t> ToUInt8Vector(T const& in)
...
     }
 
-    using Azure::Core::Url;
+    // Get thumbprint as hex string:
     {
-
-      m_requestUrl = Url("https://login.microsoftonline.com/");
-      m_requestUrl.AppendPath(tenantId);
-      m_requestUrl.AppendPath("oauth2/v2.0/token");
+      std::ostringstream thumbprintStream;
```
ahsonkhan, line 174: Can we avoid using string stream here as well?

sdk/identity/azure-identity/src/client_certificate_credential.cpp
```diff
@@ -36,214 +47,217 @@ template <typename T> std::vector<uint8_t> ToUInt8Vector(T const& in)
...
-            std::ostringstream payloadStream;
-
-            const Azure::DateTime now = std::chrono::system_clock::now();
-            const Azure::DateTime exp = now + std::chrono::minutes(10);
-
-            payloadStream << m_tokenPayloadStaticPart << Uuid::CreateUuid().ToString()
-                          << "\",\"nbf\":" << PosixTimeConverter::DateTimeToPosixTime(now)
-                          << ",\"exp\":" << PosixTimeConverter::DateTimeToPosixTime(exp) << "}";
+            DateTime const now = std::chrono::system_clock::now();
+            DateTime const exp = now + std::chrono::minutes(10);
```
ahsonkhan, line 255: Where does this 10 minutes come from? Add a comment.

ahsonkhan (approved):
I don't know why Credential Core is useful to have.

Other than the potential breaking change with changing `const` params, looks reasonable to me.

That said, it might be helpful to also get someone else on the team (maybe from the Identity v-team) to take a look.

sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
antkmsft: It would be a breaking change to replace "`std::string const& fnParam`" with "`std::string& fnParam`", but replacing it with "`std::string fnParam`" is non-breaking and is unnoticeable for the user. Replacing "`std::string fnParam`" with "`std::string const& fnParam`" is also non-breaking and users won't notice it. Everything that was compiling before, will keep compiling.

Why we need this: previously, we would keep no copy of `tenantId` - so there was no need to copy it as a function arg - reference was enough: we'd read it and return. Now we do copy `tenantId` and store it in `ClientCredentialCore`. It is possible for `ClientCertificateCredential` to still have `tenantId` arg as const ref, but then when we copy, we'd invoke copy constructor. But if we have argument as by-value, we can `std::move` it.

sdk/identity/azure-identity/inc/azure/identity/client_secret_credential.hpp
antkmsft: `std::string const& fnParam` <=> `std::string fnParam` is not breaking (either way, vice versa replacement is also not breaking).

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
antkmsft: Active Directory Federated Services, a tenant that is used in Azure Stack.

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
antkmsft: We always can put a real type, but I like decltypes - if type gets changed in the `TokenRequestContext`, everything keeps working, no need to update the type here. Decltype allows me to say "Whatever the type `TokenRequestContext::Scopes` is". It can make sense to avoid it in user-facing APIs for simplicity, but for the internal code, quite the opposite.

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
antkmsft: This class does common parts of the job for both `ClientSecretCredential` and `ClientCertificateCredential`. It's functionality will be expanded even more with Multi-tenant/Challenge-based auth support. Normally it probably would've been an abstract base class havinf some methods implemented, but we can achieve the same by aggregating it, and using the functionality without inheritance (it does not override any virtual functions anyway). Plus, it keeps inheritance structure, that is visible to the SDK users, clean.

sdk/identity/azure-identity/src/environment_credential.cpp
antkmsft:
```cpp
#include <string>
#include <iostream>

struct Base { int i; };

struct Derived : public Base { std::string s; };

void Print(Base const& b)
{
    std::cout << "i == " <<  b.i << "\n";
}

void Print(Derived const& d)
{
    std::cout << "s == " << d.s << ", i == " << d.i << "\n";
}


int main()
{
    Base base1;
    Base base2;
    Derived derived1;
    Derived derived2;
    
    base1.i = 1;
    base2.i = 2;
    
    derived1.i = 3;
    derived1.s = "Foo";
    
    derived2.i = 4;
    derived2.s = "Bar";
    
    // Will compile because Base has operator=(Base const&)
    base1 = base2;
    Print(base1); // i == 2
    
    // Will compile because Derived has operator=(Derived const&) 
    derived1 = derived2;
    Print(derived1); // s == "Bar", i == 4
    
    // The line below won't compile because Derived does not have operator=(Base const&).
    // But Base has it. It is not gone, it is hidden, for safety
    // (you don't want implicit slicing).
    // derived1 = base2; <-- won't compile
    
    // But here, we do know what we are doing, so we are invoking
    // Base::operator=(Base const&) on Derived.
    // But to do that, we need to cast Derived to Base first, and then invoke.
    static_cast<Base&>(derived1) = base2;
    Print(derived1); // s == "Bar", i == 2
}
```

sdk/identity/azure-identity/test/ut/client_certificate_credential_test.cpp
antkmsft: In Azure Stack, TenantId == "adfs" instead of UUID, and URL path for the auth endpoint is slightly different.

sdk/identity/azure-identity/src/client_certificate_credential.cpp
antkmsft: Unlike using `std::ostringstream` simply for concatenating to strings, or for doing `std::to_string()`, here, with all these  `std::uppercase`, `std::hex`, `std::setfill('0')`, `std::setw(2)` , usage of `std::ostringstream` is justified, it handles all that formatting for us. There is no simple `to_string()` that would do all that for us using the same amount of lines, so probably we'll need to extract it to a helper function. And it is better to have the code that you don't even write, because it is less chance for bugs. This is certainly not a perf bottleneck.
We could've left stringstreams in other places either, they are not perf bottlenecks too, but if it's string concat or `to_string()`, replacement is straightforward and easy. This one - not so much.

sdk/identity/azure-identity/src/client_certificate_credential.cpp
antkmsft: You are right, they can be simplified now (I was using more verbose form for debugging). Plus I found out that nullptr check prior to free is not necessary - unique_ptr does the check.

sdk/identity/azure-identity/inc/azure/identity/client_secret_credential.hpp
ahsonkhan: Really? I thought adding a `const &` to a param is considered breaking. Or does that only apply to adding `const`?

Wouldn't the caller now need to pass in a  `const` local to this function, when previously both const and non-const would have been fine?

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
ahsonkhan: nit: Adding a doc comment would be helpful for readability.

sdk/identity/azure-identity/inc/azure/identity/detail/client_credential_core.hpp
ahsonkhan: Though every dev has preferences for implementation details and empowered to choose what C++ patterns to follow, I think it is valuable to keep the entire code base consistent for maintenance (similar to, but not necessarily to the same extent as public API surface).

I see the value of `decltype` in some dynamically typed scenarios, but **to me**, it makes the code more complex to read and understand. Therefore, I would try to minimize its use within the SDK.

`TokenRequestContext` is a public type and won't ever change (as that's breaking). I don't think it makes sense to use `decltype` here:
https://github.com/Azure/azure-sdk-for-cpp/blob/4de2423934cc5526bcca9439d5e643052d3201db/sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp#L44

antkmsft: I respectfully disagree, decltype() is the best tool for the job, especially when all the function does is passes it to another function. When you read this function, in no place you need to know its actual type, that it is actually a vector. It gets passed further. Using decltype() where it was meant to be does actually mean easier maintenance, not harder. I don't find decltype() hard to read or understand - it is a language feature, and we use it as it was meant to be used, plus we use inside the implementation, and unlike the public API service, users who don't know about this language feature can't expect to read implementation sources and understand everything without further reading.

sdk/identity/azure-identity/inc/azure/identity/client_secret_credential.hpp
antkmsft: Adding `const&` to a by-value function param is not breaking. Users can still invoke with the same semantics and pass everything that was been passed before:
```cpp
#include <string>
#include <utility>

void by_cref(std::string const&) {}
void by_val(std::string) {}

static std::string g_s;
std::string& get_ref() { return g_s; }
std::string const& get_cref() { return g_s; }
std::string&& get_rvalref() { return std::move(std::string()); }

int main()
{
    // All the semantics keeps working:
    std::string s;
    by_cref(s);
    by_val(s);

    by_cref({});
    by_val({});

    by_cref("hello");
    by_val("hello");

    by_cref(std::string("hello"));
    by_val(std::string("hello"));

    by_cref(get_ref());
    by_val(get_ref());

    by_cref(get_cref());
    by_val(get_cref());

    by_cref(get_rvalref());
    by_val(get_rvalref());
}
```

PR #4184, opened 2022-12-12
Add support for space character instead of 'T' when parsing RFC3339 DateTimes

### LINKED ISSUE #4171: Can we update DateTime::Parse(string, DateFormat::Rfc3339) to allow space in place of T?
So far, it is the only known input that has space in it instead of "T". If we have more, we can add (internally first, I think) another parse format support. It is easy to modify the parser to allow space there.
But, I would check with Jeffrey first to ask if he thinks it is safe to update the parser, and whether that still can be called `DateFormat::Rfc3339`.

RFC 3339 itself says:
"_NOTE: ISO 8601 defines date and time separated by "T". Applications using this syntax may choose, for the sake of readability, to specify a full-date and full-time separated by (say) a space character._"
https://www.rfc-editor.org/rfc/rfc3339

_Originally posted by @antkmsft in https://github.com/Azure/azure-sdk-for-cpp/pull/4167#discussion_r1042897351_
      
FWIW, .NET's DateTime.Parse() would parse `"2021-02-03 12:34:56"`, ours would not.

antkmsft:
Closes #4171

### DIFF
--- a/sdk/core/azure-core/test/ut/datetime_test.cpp
+++ b/sdk/core/azure-core/test/ut/datetime_test.cpp
@@ -895,3 +895,10 @@ TEST(DateTime, LeapYear)
   EXPECT_NO_THROW(static_cast<void>(DateTime(2021, 2, 28)));
   EXPECT_THROW(static_cast<void>(DateTime(2021, 2, 29)), std::invalid_argument);
 }
+
+TEST(DateTime, Rfc3339Space)
+{
+  auto const datetime
+      = DateTime::Parse("2022-08-24 00:43:08.0004308Z", DateTime::DateFormat::Rfc3339);
+  EXPECT_EQ(datetime.ToString(DateTime::DateFormat::Rfc3339), "2022-08-24T00:43:08.0004308Z");
+}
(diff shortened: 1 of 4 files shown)


JeffreyRichter (approved):

sdk/core/azure-core/test/ut/datetime_test.cpp
```diff
@@ -895,3 +895,10 @@ TEST(DateTime, LeapYear)
   EXPECT_NO_THROW(static_cast<void>(DateTime(2021, 2, 28)));
   EXPECT_THROW(static_cast<void>(DateTime(2021, 2, 29)), std::invalid_argument);
 }
+
+TEST(DateTime, Rfc3339Space)
```
ahsonkhan, line 899: Good test!

sdk/identity/azure-identity/src/token_credential_impl.cpp
```diff
@@ -181,17 +181,9 @@ AccessToken TokenCredentialImpl::ParseToken(
...
-        auto const spacePos = expiresOn.find(' ');
-        if (spacePos != std::string::npos)
-        {
-          expiresOn = expiresOn.replace(spacePos, 1, 1, 'T');
-        }
-      }
-
-      accessToken.ExpiresOn
-          = Azure::DateTime::Parse(expiresOn, Azure::DateTime::DateFormat::Rfc3339);
+      accessToken.ExpiresOn = Azure::DateTime::Parse(
```
ahsonkhan, line 184: 👍

ahsonkhan (commented):
Looks great!

PR #4189, opened 2022-12-14
exclude template from test proxy

gearama:
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -178,7 +178,7 @@ jobs:
       - pwsh: |
           get-content test-proxy.log
         displayName: TestProxy Log
-        condition: and(succeededOrFailed(), contains(variables.CmakeArgs, 'BUILD_TESTING=ON'))
+        condition: and(succeededOrFailed(), contains(variables.CmakeArgs, 'BUILD_TESTING=ON'),ne('${{parameters.ServiceDirectory}}', 'template'))
         workingDirectory: $(Build.SourcesDirectory)
         continueOnError: true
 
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -152,15 +152,15 @@ jobs:
           runProxy: true
           rootFolder: '$(Build.SourcesDirectory)/sdk/${{parameters.ServiceDirectory}}'
           templateFolder: '$(Build.SourcesDirectory)/sdk/${{parameters.ServiceDirectory}}'
-          condition: and(succeeded(), contains(variables.CmakeArgs, 'BUILD_TESTING=ON'))
+          condition: and(succeeded(), contains(variables.CmakeArgs, 'BUILD_TESTING=ON'),ne('${{parameters.ServiceDirectory}}', 'template'))
```
ahsonkhan, line 155: I would prefer to have someone from EngSys look at this change, before merging.

ahsonkhan: Why do we need to exclude the template package from test proxy? And if we really need to, is this the cleanest way to do it?

We are adding exception in multiple places, which seems a bit fragile. Are these the only places that need to opt-out?

benbp: @gearama @ahsonkhan we also depend on the template pipeline to validate changes, often ones we make that affect all repos. I would prefer not to special case template so that we can catch any possible common proxy setup issues that might affect c++.

PR #4195, opened 2022-12-15
Move Request ctor impl to the request.cpp file instead of http.cpp

ahsonkhan:
We already have a `request.cpp` file, so all method implementations for the `Request` object belong there, rather than in the common `http.cpp`.

These constructors were moved from the `http.hpp` header file to source here: https://github.com/Azure/azure-sdk-for-cpp/pull/1741

### DIFF
--- a/sdk/core/azure-core/src/http/http.cpp
+++ b/sdk/core/azure-core/src/http/http.cpp
@@ -9,7 +9,6 @@
 
 using namespace Azure::Core;
 using namespace Azure::Core::Http;
-using namespace Azure::Core::IO::_internal;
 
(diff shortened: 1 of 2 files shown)


antkmsft (approved):

LarryOsterman:
My only comment here is that this change slightly changes code generation - when you move code out of a header, it means that you depend on LTCG to inline constructor implementations rather than letting the optimizer do the work.

That may make a difference, especially for simple constructors.

ahsonkhan:
This PR is moving the implementation to the correct cpp file. It was already in a cpp file before (not hpp), just the wrong one.

LarryOsterman:
Doh - didn't realize it was http.cpp. Sigh.

PR #4196, opened 2022-12-16
Add new Request ctors that accept content-type and obsolete previous ones that only took a request body without a content type

ahsonkhan:


sdk/storage/azure-storage-blobs/src/rest_client.cpp
```diff
@@ -1215,13 +1215,10 @@ namespace Azure { namespace Storage { namespace Blobs {
         const SubmitServiceBatchOptions& options,
         const Core::Context& context)
     {
-      auto request = Core::Http::Request(Core::Http::HttpMethod::Post, url, &requestBody, false);
+      auto request = Core::Http::Request(
```
Jinming-Hu, line 1218: Have you updated codeGen?

Jinming-Hu:
What is the motivation of this change?

Issue #3674, opened 2022-05-24
Audit and fix-up places where methods are marked as noexcept too liberally


ahsonkhan:
https://github.com/Azure/azure-sdk-for-cpp/search?l=C%2B%2B&q=noexcept

The string extension methods shouldn't be marked noexcept. Audit other instances, like the ones in DateTime.

Use of it in places like `Nullable<T>` is expected and OK. Ignore instances in 3rd party libraries like nlohmann json.

ahsonkhan:
The relevant instances that need another look are in core.

cc @antkmsft, @LarryOsterman

antkmsft:
When 4 of the linked PRs above, we should no longer have places where `noexcept` is used too liberally.

PR #4197, opened 2022-12-16
Audit string comparisons

antkmsft:
* Ensure that locale-independent character functions are used: i.e. `std::isalnum(n)` => `std::isalnum(c, std::locale::classic)`
* Do not `ToUpper()`/`ToLower()` if the only goal is `LocaleInvariantCaseInsensitiveEqual()`
* That also includes the case when one'd `ToUpper()`/`ToLower()` before searching in a `CaseInsensitiveMap`.
* Speaking of which, `map<string, string, CaseInsensitiveComparator>` are replaced with `CaseInsensitiveMap`, and there is a corresponding codegen update: https://github.com/Azure/autorest.cpp/pull/170
* `StringExtensions`: remove `noexcept` where it cannot be guaranteed (string versions of `ToUpper()`\`ToLower()`) - partially covers #3674
  * Simplify some stuff in `StringExtensions` (or rewrite so that `noexcept` can be guaranteed), everything can be moved to a header now.
* Other related code - simplify:
  * Header name validation
  * `GetHeader()` / `SetHeader()`
  * URL parsing, encode, decode
  * Drop some unused headers


antkmsft:
I have run pipelines for Core, Identity, Storage, KeyVault, and Attestation.

Jinming-Hu (changes requested):

sdk/storage/azure-storage-common/src/shared_key_policy.cpp
```diff
@@ -32,7 +31,7 @@ namespace Azure { namespace Storage { namespace _internal {
           "If-Unmodified-Since",
           "Range"})
     {
-      auto ite = headers.find(Azure::Core::_internal::StringExtensions::ToLower(headerName));
+      auto ite = headers.find(headerName);
```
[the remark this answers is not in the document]
antkmsft: @Jinming-Hu, headers are `CaseInsensitiveMap`, `find()` will work the same for either case.

Jinming-Hu (approved):
Changes to storage LGTM

sdk/storage/azure-storage-common/src/shared_key_policy.cpp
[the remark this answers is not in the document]
antkmsft: @Jinming-Hu, to reassure you, : https://github.com/Azure/azure-sdk-for-cpp/blob/c73ad854c21e2aee7e1aefd041efb4294e0f1624/sdk/core/azure-core/test/ut/case_insensitive_containers_test.cpp#L12-L32

sdk/core/azure-core/src/http/http.cpp
```diff
@@ -25,153 +28,30 @@ const HttpMethod HttpMethod::Put("PUT");
 const HttpMethod HttpMethod::Delete("DELETE");
 const HttpMethod HttpMethod::Patch("PATCH");
 
+namespace {
+bool IsInvalidHeaderNameChar(char c)
```
ahsonkhan, line 32: Wouldn't this be slower than a single lookup in the `validChars` array, we had before?

sdk/core/azure-core/src/http/user_agent.cpp
```diff
@@ -133,10 +133,10 @@ std::string GetOSVersion()
 
 std::string TrimString(std::string s)
 {
-  auto const isSpace = [](int c) { return !std::isspace(c); };
+  auto const isNotSpace = [](char c) { return !std::isspace(c, std::locale::classic()); };
 
-  s.erase(s.begin(), std::find_if(s.begin(), s.end(), isSpace));
-  s.erase(std::find_if(s.rbegin(), s.rend(), isSpace).base(), s.end());
+  s.erase(s.begin(), std::find_if(s.begin(), s.end(), isNotSpace));
+  s.erase(std::find_if(s.rbegin(), s.rend(), isNotSpace).base(), s.end());
```
ahsonkhan, line 139: The double negative in `isNotSpace` is a bit confusing, especially combined with what `find_if` does, which is:
> An iterator to the first element in the range for which pred does not return false.

Would using an `isSpace` function be more clear, logically?

sdk/core/azure-core/inc/azure/core/internal/strings.hpp
```diff
@@ -19,24 +20,61 @@ namespace Azure { namespace Core { namespace _internal {
...
+    {
+      std::transform(src.begin(), src.end(), src.begin(), [](auto c) { return ToLower(c); });
+      return src;
+    }
+
+    static std::string ToUpper(std::string src)
+    {
+      std::transform(src.begin(), src.end(), src.begin(), [](auto c) { return ToUpper(c); });
+      return src;
+    }
```
ahsonkhan, line 77: Why did we decide to move the implementation detail from a cpp file, into a header-only file?

sdk/core/azure-core/test/ut/string_test.cpp
```diff
@@ -23,18 +25,20 @@ TEST(String, invariantCompare)
 TEST(String, toLowerC)
 {
   using Azure::Core::_internal::StringExtensions;
-  for (unsigned char ch = 0; ch < 255; ch += 1)
+  for (unsigned i = 0; i <= 255; ++i)
```
ahsonkhan, line 28: What's the type of `i`? When not specified, is it just a 32-bit unsigned `int`?

LarryOsterman: Yes - in C (and C++), the "unsigned" type is a shorthand for "unsigned int".

sdk/core/azure-core/src/http/user_agent.cpp
antkmsft: To be clear, I did not touch the logic, but I updated the name to actually correspond to what the function does, which before was the exact opposite of the name :)

I agree, using "actual" `isSpace` could have been better, however for std::erase we couldn't put it with a negation operator, i.e. unfortunately we can't write `find_if(..., ..., !isSpace)`, so "isNotSpace" is needed.

Look at it this way, it is not unreadable, and it is not a double negation, although normally functions like "isNotSomething()" are not common.
It is readable as this: "erase from front of the string to the first character from beginning that is NOT space;" then next line: "erase from the back of the string (rbegin) to the first character from the end that is NOT space". In speech it sounds more normal, as it is more normal for us to operate with wording like "first non-space character", as we don't seem to have a better word for it except calling such characters "non-space".

sdk/core/azure-core/inc/azure/core/internal/strings.hpp
antkmsft: After static variable removal, no hard reason left to keep any implementation in a .cpp file, allowing better inlining to be used for the functions, including the functions that get called for each individual character, or from a case insensitive comparator in maps, plus I think logistically it is easier to support smaller amount of code in less files, especially if a function is short, and no .cpp file exists already.

sdk/core/azure-core/src/http/http.cpp
antkmsft: It is likely slightly slower, but I think it is way easier to support (less lines) and therefor has a higher chance to have less bugs. It is unlikely to be a perf bottleneck, especially in the SDK that talks to the network, and I'd say until we get data that it is a perf bottleneck, we should prefer ~5 line version to 256+ lines version. The version we had before did rely on comments, on 0 not being in the wrong spot, took at least 256 bytes, and it would have been harder to ensure it works correctly by reading it, than reading an isalnum and ~10 extra allowed character that fit on the screen and comparison is straighforward and not index-based.

sdk/core/azure-core/inc/azure/core/internal/strings.hpp
LarryOsterman: In general I'm a massive fan of header-only implementations of types, but Casey corrected me about implementation of header-only vs cpp - it's actually not that much better if LTCG (link time code generation) is enabled.

But as I said, I generally prefer header-only implementations when possible.

antkmsft: I'd actually move some inline functions we have in headers to .cpp files - (longer ones, the ones with asciiz literals), but not these ones.

PR #4199, opened 2022-12-16
More accurate argument type for the ExtendableEnumeration == and !=

antkmsft:
Can be considered as part of the #3674 effort

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
+++ b/sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
@@ -63,7 +63,7 @@ namespace Azure { namespace Core { namespace _internal {
      *
      * @param other Another extendable enumeration to be compared.
      */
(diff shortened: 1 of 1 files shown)


gearama (approved):

sdk/core/azure-core/inc/azure/core/internal/extendable_enumeration.hpp
```diff
@@ -63,7 +63,7 @@ namespace Azure { namespace Core { namespace _internal {
      *
      * @param other Another extendable enumeration to be compared.
      */
-    bool operator==(T const& other) const noexcept
+    bool operator==(ExtendableEnumeration<T> const& other) const noexcept
```
ahsonkhan, line 66: @antkmsft can you please help me understand why this change was necessary and the implications?

Even though this type is internal, is this considered a breaking change (where a previous call to `==` with different types of "arguments" now fail to compile?

PR #204, opened 2020-06-23
Nullable type

gilbertw:

PR #4200, opened 2022-12-16
noexcept adjustments

antkmsft:
Part of work for the #3674 effort

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/nullable.hpp
+++ b/sdk/core/azure-core/inc/azure/core/nullable.hpp
@@ -98,7 +98,7 @@ public:
    * @brief Destructs the contained value, if there is one.
    *
    */
-  void Reset() noexcept /* enforces termination */
+  void Reset() noexcept(std::is_nothrow_destructible<T>::value) /* enforces termination */
   {
     if (m_hasValue)
     {
--- a/sdk/core/azure-core/inc/azure/core/operation.hpp
+++ b/sdk/core/azure-core/inc/azure/core/operation.hpp
@@ -139,7 +139,7 @@ namespace Azure { namespace Core {
      * @brief Gets the current #Azure::Core::OperationStatus of the long-running operation.
      *
      */
-    OperationStatus Status() const noexcept { return m_status; }
+    OperationStatus const& Status() const noexcept { return m_status; }
 
     /**
      * @brief Checks if the long-running operation is completed.
--- a/sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
+++ b/sdk/core/azure-core/src/http/winhttp/win_http_request.hpp
@@ -55,7 +55,7 @@ namespace Azure { namespace Core { namespace Http { namespace _detail {
         DWORD_PTR dwContext,
         DWORD dwInternetStatus,
         LPVOID lpvStatusInformation,
-        DWORD dwStatusInformationLength) noexcept;
+        DWORD dwStatusInformationLength);
 
     /*
      * Callback from WinHTTP called after the TLS certificates are received when the caller sets
--- a/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
+++ b/sdk/core/azure-core/src/http/winhttp/win_http_transport.cpp
@@ -517,7 +517,7 @@ namespace Azure { namespace Core { namespace Http { namespace _detail {
       DWORD_PTR dwContext,
       DWORD internetStatus,
       LPVOID statusInformation,
-      DWORD statusInformationLength) noexcept
+      DWORD statusInformationLength)
   {
     // If we're called before our context has been set (on Open and Close callbacks), ignore the
     // status callback.


gearama (approved):

sdk/core/azure-core/inc/azure/core/operation.hpp
```diff
@@ -139,7 +139,7 @@ namespace Azure { namespace Core {
      * @brief Gets the current #Azure::Core::OperationStatus of the long-running operation.
      *
      */
-    OperationStatus Status() const noexcept { return m_status; }
```
LarryOsterman, line 142: This is a fairly significant change because it silently ties the lifetime of the returned `OperationStatus` to the lifetime of the `Operation`. If the `Operation` is deleted, the returned `OperationStatus` is a dangling reference to uninitialized memory.

It's also a breaking change since this is a change to the API signature of a public API.

sdk/core/azure-core/inc/azure/core/nullable.hpp
```diff
@@ -98,7 +98,7 @@ template <class T> class Nullable final {
    * @brief Destructs the contained value, if there is one.
    *
    */
-  void Reset() noexcept /* enforces termination */
+  void Reset() noexcept(std::is_nothrow_destructible<T>::value) /* enforces termination */
```
ahsonkhan, line 101: @CaseyCarter would adding this type of constraint on T on a method on `Nullable<T>` be considered a breaking change, or was it not possible to satisfy this constraint previously, based on the existing `is_nothrow_...` within the ctors above?

CaseyCarter: Formally, relaxing exception guarantees is a breaking change since callers may be relying on those exception guarantees for correctness. In practice, this is only a change for abominable types with throwing destructors, which should be very rare because they are impossible to reason about. Case in point: trying to support throwing destructors has introduced a bug in this function. Line 105 and 106 need to be reordered.

ahsonkhan: Thanks, that makes a lot of sense.

Aside from when is relaxing exception guarantees is considered breaking (generally and technically), and aside from the fact that, in practice, we shouldn't have throwing destructors anyway, I am wondering specifically about this type, and whether adding this `is_nothrow...` now on the method is just making it explicit what was already the type of constraint, implicitly, with no observable impact, OR if it has some theoretical change of behavior?

For example, the ctor already has this in the `noexcept` clause:
https://github.com/Azure/azure-sdk-for-cpp/blob/aec83d672b0c0863c64bf73b786b4f2b1a939412/sdk/core/azure-core/inc/azure/core/nullable.hpp#L51

Is there some intersecting relationship between T constrainted to `is_nothrow_move_constructible` versus `is_nothrow_destructible` or are they mutually exclusive?

ahsonkhan: > Line 105 and 106 need to be reordered.

Interesting, really?! What's the bug? Should we not wait to return from the Value's destructor, before setting `m_hasValue` to false?

This was probably inspired by `std::optional`, when added here:
https://github.com/Azure/azure-sdk-for-cpp/pull/204
https://github.com/microsoft/STL/blob/8ddf4da23939b5c65587ed05f783ff39b8801e0f/stl/inc/optional#L130-L135

sdk/core/azure-core/inc/azure/core/operation.hpp
antkmsft: No one should call `operation.Status()` after it has been destructed, or while it is destructing. And no one can expect to have a reference without controlling its lifetime and expect it won't go bad. But in order to do that, you have to be very persistent:
```cpp
OperationStatus s1 = operation.Status(); // creates a copy
auto s2 = operation.Status(); // creates a copy - auto does not declare references even if the type is reference.

OperationStatus const& s3 = operation.Status(); // would break*, but why'd you do that in the past?
decltype(operation.Status()) s4 = operation.Status(); // would break*, but why'd you do decltype when auto is sufficient?

// * Examples where it would break, are only possible if the code does not control the lifetime of an operation and it gets destroyed while the code still has a reference to status.
```

sdk/core/azure-core/inc/azure/core/nullable.hpp
antkmsft: Thank you @CaseyCarter, good catch with the ordering.

Is removing `noexcept` practically breaking though? In the past, if code ends up throwing, terminate() gets called, without much chance for the user code to interrupt. So all the programs that were relying on nothrow behavior were basically crashing. Now you are going to get an exception and survive past the point in the past, where `terminate()` was called. Can it be considered as an actually "extended" behavior and not the "narrowing" one, that is normally an attribute of a breaking change? That's of course, as long as we don't start throwing exceptions left and right when previously we would avoid doing so. So, given that the code remains the same and does the best effort to not throw, is `noexcept` removal practically a breaking change, or in spirit a breaking change?

sdk/core/azure-core/inc/azure/core/operation.hpp
antkmsft: Regarding lifetime, 5 operations are statically allocated: https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/src/operation_status.cpp#L8-L12

So, in order to get in trouble with the old code after upgrading to new Core version, you should be getting the value as `s3` or `s4` from example above, not control the lifetime of Operation (not have it as copy), and the value should be not the one of the pre-defined 5 operations.

sdk/core/azure-core/inc/azure/core/nullable.hpp
antkmsft: #4232

sdk/core/azure-core/inc/azure/core/operation.hpp
LarryOsterman: You're absolutely right that you need to be persistent. But our customers are VERY good at doing silly things.

What value does making `Status()` return a reference bring to the table, especially since the reference is const (which means that it's value can never change)?

antkmsft: It can remain `noexcept`. If we return a copy, constructor may throw: https://en.cppreference.com/w/cpp/string/basic_string/basic_string

So, we should either remove `noexcept` or make it to return const ref.

sdk/core/azure-core/inc/azure/core/nullable.hpp
CaseyCarter: > Interesting, really?! What's the bug? Should we not wait to return from the Value's destructor, before setting `m_hasValue` to false?

The object is destroyed whether the destructor throws or not. (This is the fundamental reason why throwing destructors make little sense: you can't "abort" destruction.) We need to ensure that `m_hasValue` is set to `false` _before_ calling the destructor which may throw so we don't destroy the contained object again in `Nullable`'s destructor.

> In the past, if code ends up throwing, terminate() gets called

Right. So this change has no effect for programs that don't use types with throwing destructors in `Nullable`, but programs that do can now get into a state that wasn't possible/expected/planned for/tested before.

PR #4219, opened 2023-01-04
Core January Release

gearama:
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 1.8.0-beta.3 (Unreleased)
+## 1.8.0-beta.3 (2023-01-05)
 
 ### Features Added
 
@@ -10,10 +10,6 @@
 
 - Bearer token authentication will not work for endpoint URL protocol schemes other than `"https"`. This ensures token security and is consistent with the Azure SDKs for other languages.
 - Removed `noexcept` specification from `Azure::DateTime::clock::now()`.
-
-### Bugs Fixed
-
-### Other Changes
 
 ## 1.8.0-beta.2 (2022-11-03)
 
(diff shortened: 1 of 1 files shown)


antkmsft (approved):

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -9,11 +9,7 @@
 ### Breaking Changes
 
 - Bearer token authentication will not work for endpoint URL protocol schemes other than `"https"`. This ensures token security and is consistent with the Azure SDKs for other languages.
-- Removed `noexcept` specification from `Azure::DateTime::clock::now()`.
-
-### Bugs Fixed
-
-### Other Changes
+- Removed `noexcept` specification from `Azure::DateTime::clock::now()`.
```
ahsonkhan, line 12: I don't think this is worth calling out as a breaking change. I would leave this under 'Other changes', or maybe 'Bug fix' since the API now reflects what the Implementation was actually doing.

@antkmsft which one makes more sense to you, if you had to pick between 'bug fix' and 'other changes'?

ahsonkhan (approved):
Other than where to put the noexcept change in the changelog, looks good.

sdk/core/azure-core/CHANGELOG.md
antkmsft: I see arguments for both - technically, when you remove `noexcept`, the app will continue working if the exception is thrown, and user will have a chance to handle the exception, while previously the app would call `terminate()` and the game is over. So, I am not sure if it is a "breaking" change. Therefore, it can be marked as "Other", I agree.
Left to human judgement, I think it would've been a breaking change, if it was `noexcept` before, and now we were explicitly throwing an exception.
But in this case, it could throw only theoretically, under extremely rare set of conditions, yet we removed it to be accurate. So it this case, this can be categorized as "Other Changes", I agree.

gearama: it is breaking as now it would throw exceptions potentially, thus need to try/catch , could potentially blow up the customer code if it throws.

PR #4230, opened 2023-01-09
Use non-template-template parameter in UniqueHandle

antkmsft:
If needed, UniqueHandle's template-template argument can be rewritten as regular typed argument. There could be subjective pros and cons to each, but in the end same things are possible using either of two semantics.

We can change it while both changes are in betas (Identity and Core, Identity depends on Core).
But the Core Beta has already shipped, while Identity Beta ships tomorrow, so don't merge this until tomorrow after the Identity release.

Regular template parameters are supported by current ApiView parser implementation, while for template-template parameters, ApiView needs to be updated.

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/internal/unique_handle.hpp
+++ b/sdk/core/azure-core/inc/azure/core/internal/unique_handle.hpp
@@ -62,6 +62,5 @@ namespace Azure { namespace Core { namespace _internal {
   template <typename T> struct UniqueHandleHelper;
 
   // *** Now users can say UniqueHandle<T> if they want:
-  template <typename T, template <typename> class U = UniqueHandleHelper>
-  using UniqueHandle = typename U<T>::type;
+  template <typename T, typename U = UniqueHandleHelper<T>> using UniqueHandle = typename U::type;
 }}} // namespace Azure::Core::_internal
--- a/sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
+++ b/sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
@@ -25,8 +25,7 @@ namespace Azure { namespace Identity {
 
     void FreePkeyImpl(void* pkey);
 
-    template <typename> struct UniquePkeyHelper;
-    template <> struct UniquePkeyHelper<void*>
+    struct UniquePkeyHelper
     {
       static void FreePkey(void* pkey) { FreePkeyImpl(pkey); }
       using type = Azure::Core::_internal::BasicUniqueHandle<void, FreePkey>;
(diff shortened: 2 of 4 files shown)


antkmsft:
I'd leave the decision mostly up to @LarryOsterman, but any team member feedback is appreciated. As for me, I'm fine either way. There are cases when you can't do without template-template parameters, this one is not one of them.

sdk/identity/azure-identity/inc/azure/identity/client_certificate_credential.hpp
```diff
@@ -25,8 +25,7 @@ namespace Azure { namespace Identity {
 
     void FreePkeyImpl(void* pkey);
 
-    template <typename> struct UniquePkeyHelper;
-    template <> struct UniquePkeyHelper<void*>
+    struct UniquePkeyHelper
```
ahsonkhan, line 28: Is such a change non-breaking?

antkmsft: This one specific is not - it is in Identity's `_detail` namespace.
The broader one, in `unique_handle.hpp` is breaking. It is in `_internal` namespace. If we go with this change, the next Core will break the Identity Beta that we release tomorrow, but it is Beta (both are Betas actually) so it is ok to break.

ahsonkhan: Missed the `_detail` namespace. Makes sense.

sdk/core/azure-core/inc/azure/core/internal/unique_handle.hpp
```diff
@@ -62,6 +62,5 @@ namespace Azure { namespace Core { namespace _internal {
   template <typename T> struct UniqueHandleHelper;
 
   // *** Now users can say UniqueHandle<T> if they want:
-  template <typename T, template <typename> class U = UniqueHandleHelper>
-  using UniqueHandle = typename U<T>::type;
+  template <typename T, typename U = UniqueHandleHelper<T>> using UniqueHandle = typename U::type;
```
ahsonkhan, line 65: Should the `using ...` be on a new line?

antkmsft: It is how clang-format formatted it. But if we really want the newline, we could put an empty line comment (`//`) after `>` and before `using`.

ahsonkhan: I don't have an opinion one way or another.

This new approach is more readable/understandable than how it was written previously (as long as they are identical).

That said, both ways highlight a complexity in C++ templates that I have trouble parsing the syntax of and understanding.
Unrelated to this PR: For me, some source comments explaining how I should be reading this would be helpful.

PR #4232, opened 2023-01-10
Nullable::Reset() hardening

antkmsft:
Thanks to @CaseyCarter, per https://github.com/Azure/azure-sdk-for-cpp/pull/4200#discussion_r1065142310.


CaseyCarter (approved):

RickWinter (dismissed):

sdk/core/azure-core/inc/azure/core/nullable.hpp
```diff
@@ -102,8 +102,8 @@ template <class T> class Nullable final {
   {
     if (m_hasValue)
     {
-      m_value.~T();
       m_hasValue = false;
+      m_value.~T();
```
RickWinter, line 106: This seems to be a race.  If we switch this order then m_value could be reassigned to another value of T by another thread doing an assignment.

@CaseyCarter This order is different than the order in std::optional.  Why do we want inverted behavior in this function?

CaseyCarter: > This seems to be a race. If we switch this order then m_value could be reassigned to another value of T by another thread doing an assignment.

If another thread calls member functions of this object while this function is modifying it that's a data race leading to undefined behavior - and that has nothing to do with the order in which this function performs its various tasks.
 
> @CaseyCarter This order is different than the order in std::optional. Why do we want inverted behavior in this function?

Because this function is trying to support contained objects with throwing destructors, which `std::optional` does not do.

antkmsft:
@RickWinter, are you ok with this change going in?

sdk/core/azure-core/inc/azure/core/nullable.hpp
RickWinter: How doe std::optional prevent the type from having a throwing destructor.  It seems we should put that same constraint in place.

RickWinter (approved):

sdk/core/azure-core/inc/azure/core/nullable.hpp
ahsonkhan, line 106: It would be helpful for future maintainers if we add a source comment here to explain why the line order is intentionally this way since it matters, and to not reverse it.

Issue #4213, opened 2022-12-27
Bug: `Host` header is not correctly set for non-default port


Jinming-Hu:
It's specified in RFC2616 that

> A "host" without any trailing port information implies the default
>   port for the service requested (e.g., "80" for an HTTP URL).

so for HTTP or HTTPS that are not running on 80 or 443, we need to include port in `Host` header.

This is a bug in our code:

https://github.com/Azure/azure-sdk-for-cpp/blob/f872841ffc9d6339fcd7303af20c6763b95d5b2f/sdk/core/azure-core/src/http/curl/curl.cpp#L414

Jinming-Hu:
this issue can be fixed with below path

```
diff --git a/sdk/core/azure-core/src/http/curl/curl.cpp b/sdk/core/azure-core/src/http/curl/curl.cpp
index f8467e0c..30dd4b00 100644
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -411,7 +411,13 @@ CURLcode CurlSession::Perform(Context const& context)
     if (hostHeader == headers.end())
     {
       Log::Write(Logger::Level::Verbose, LogMsgPrefix + "No Host in request headers. Adding it");
-      this->m_request.SetHeader("Host", this->m_request.GetUrl().GetHost());
+      std::string headerValue = this->m_request.GetUrl().GetHost();
+      auto port = this->m_request.GetUrl().GetPort();
+      if (port != 0)
+      {
+        headerValue += ":" + std::to_string(port);
+      }
+      this->m_request.SetHeader("Host", headerValue);
     }
     auto isContentLengthHeaderInRequest = headers.find("content-length");
     if (isContentLengthHeaderInRequest == headers.end())
```

ahsonkhan:
> so for HTTP or HTTPS that are not running on 80 or 443, we need to include port in `Host` header.

In what scenario would that happen? How did you run into this issue?

Jinming-Hu:
> In what scenario would that happen? How did you run into this issue?

When we send requests to non 80 or 443 ports.

I ran into this issue when running test-proxy (localhost:5001) through Fiddler.

Issue #4147, opened 2022-12-01
LIVEONLY tests in Storage 


gearama:
There are some issues with test proxy behavior and some of the storage tests in a few cases there are errors around json deserialization "expected string received null", which suggest that the deserailization of the incoming json might not handle nulls properly. 
The followin g tests i swiched to _LIVEONLY_ 
         
 
	 37 - azure-storage-blobs.BlobContainerClientTest.Lease (Failed)
	 
	114 - azure-storage-blobs.PageBlobClientTest.Lease (Failed)
	
	420 - azure-storage-files-shares.FileShareClientTest.Lease (Failed)
	421 - azure-storage-files-shares.FileShareClientTest.SnapshotLease (Failed)
	
	447 - azure-storage-files-shares.FileShareFileClientTest.LeaseRelated (Failed)

TEST_F(BlobBatchClientTest, DISABLED_SubmitDeleteBatch) 
 TEST_F(BlobBatchClientTest, DISABLED_SnapshotVersion
 TEST_F(BlobBatchClientTest, DISABLED_TokenAuthorization)
TEST_F(QueueClientTest, PeekMessage_LIVEONLY_)

[testlogs.zip](https://github.com/Azure/azure-sdk-for-cpp/files/10135304/testlogs.zip)
[AppendBlobClientTest.ContentHash.zip](https://github.com/Azure/azure-sdk-for-cpp/files/10157180/AppendBlobClientTest.ContentHash.zip)

gearama:
https://github.com/Azure/azure-sdk-tools/issues/4853

Jinming-Hu:
@gearama 

Do you want storage team to look into this or you want to assign this to yourself just for tracking?

Jinming-Hu:
Hi @gearama , 

I found some secrets were not properly redacted when fetching tokens. 
In the below recorded request, tenant ID, client ID, client secret were in plaintext. Not sure if the cookie should be redacted.

![image](https://user-images.githubusercontent.com/52266153/211984065-a1cb0309-f94d-4d19-8d63-82421ff8d25d.png)

Jinming-Hu:
Hi @gearama , have you got a chance to look at the above issue?

What is the correct way to record/playback token-fetching/refreshing requests anyway? Is it by-design to now record such requests?

PR #4118, opened 2022-11-15
Test proxy

gearama:
Replaces the homegrown test-proxy with the commn one used by other languages, created byt the ENG team . (https://github.com/Azure/azure-sdk-tools/tree/main/tools/test-proxy )
One bebefit is that the test recordings are kept in a separate repo ( https://github.com/Azure/azure-sdk-assets) from where the tests are restored at playback time. 
All recordinggs have been moved to the new location. A few misbehaving tests(<10) have been switched to LIVEONLY , a bug has been filed for those to keep track
To install the proxy locally and run it read : https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md#installation 

or execute "\eng\scripts\Start-TestProxy.ps1 -AssetsPath <sdk/service> "

# Pull Request Checklist

Please leverage this checklist as a reminder to address commonly occurring feedback when submitting a pull request to make sure your PR can be reviewed quickly:

See the detailed list in the [contributing guide](https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#pull-requests).

- [x] [C++ Guidelines](https://azure.github.io/azure-sdk/cpp_introduction.html)
- [x] Doxygen docs
- [x] Unit tests
- [x] No unwanted commits/changes
- [x] Descriptive title/description
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] External references or docs updated
- [x] Self review of PR done
- [x] Any breaking changes?

PR #4241, opened 2023-01-12
Test proxy & storage tests improvements

Jinming-Hu:
In this PR, we:

1. Removed unused definitions in `sdk/core/azure-core-test/inc/azure/core/test/network_models.hpp`.
2. Removed unused functions `TestBase::ValidateSkippingTest()` and `TestBase::IsValidTime()`.
3. Added an overload of function `TestBase::InitClientOptions()`.
4. Removed unused functions `TestProxyManager::SetStopRecordMode()`, `TestProxyManager::SetStopPlaybackMode()`.
5. Rewrote proxy matcher setting logic.
    1. Use json object rather than hard-coded json string, for better readability and maintenance.
    2. Exclude `Expect` and `Connection` headers so that recording made by libcurl can be used by WinHTTP and also the other way around
    3. Removed some unnecessary ignored headers
    4. Added some ignored query parameters
6. Fixed an issue in libcurl transport layer. Closes https://github.com/Azure/azure-sdk-for-cpp/issues/4213
7. Fixed an issue where libcurl transport layer adds `Content-Length` header for HEAD/GET/DELETE requests, which caused a behavior misalignment between libcurl and WinHTTP, making the requests recorded by one of the transport layer unusable by the other.
    1. Also removed `Content-Length` header for GET requests from existing recordings.
8. Fixed a bug where test proxy sometimes discarded request body.
9. Fixed a bug where requests were not properly redirected if `Host` was set in the original request *
10. Enabled some storage test cases that were disabled in https://github.com/Azure/azure-sdk-for-cpp/pull/4118. Closes https://github.com/Azure/azure-sdk-for-cpp/issues/4147
11. Removed `CHECK_SKIP_TEST` macro. We check if test cases is live-only in `SetUpTestBase()` function. With this change, the only thing we need to do to mark a test case live-only is to add `_LIVEONLY_` suffix in case name. There's no need to call `CHECK_SKIP_TEST` macro anymore.

\* A little more explanation of this bug: 
Transport layer accepts a request with a URL and request headers in it. If `Host` header is not in request headers, libcurl transport will extract host from URL and set `Host` header. If `Host` is already in request, libcurl transport does nothing.
Test-proxy policy modifies host in URL to `localhost:5001`, but it doesn't modify `Host` header (I think it's a negligence).
For most of the requests we don't set `Host` header. Test-proxy modifies host in the URL to `localhost:5001`, libcurl extracts `localhost:5001` from URL and set `Host` header. Everything works perfectly.
For very few requests, `Host` header is already set. Test-proxy only modifies host in the URL, libcurl doesn't set `Host` header because there's one already. Now the host in `Host` header and URL don't match. These requests cannot be recorded & played back.

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
+++ b/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
@@ -22,13 +22,6 @@
 #include <regex>
 #include <thread>
 
-#define CHECK_SKIP_TEST() \
-  std::string const readTestNameAndUpdateTestContext = GetTestName(); \
-  if (shouldSkipTest()) \
-  { \
-    GTEST_SKIP(); \
-  }
-
 using namespace std::chrono_literals;
 
 namespace Azure { namespace Core { namespace Test {
@@ -374,6 +367,12 @@ namespace Azure { namespace Core { namespace Test {
       if (!m_wasSkipped && !m_testContext.IsLiveMode())
       {
         m_testProxy = std::make_unique<Azure::Core::Test::TestProxyManager>(m_testContext);
+      }
+
+      std::string const readTestNameAndUpdateTestContext = GetTestName();
+      if (shouldSkipTest())
+      {
+        GTEST_SKIP();
       }
     }
 
--- a/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
+++ b/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
@@ -134,6 +134,15 @@ namespace Azure { namespace Core { namespace Test {
 
     bool shouldSkipTest() { return m_wasSkipped; }
 
+    void SkipTest()
+    {
+      if (!m_wasSkipped)
+      {
+        m_wasSkipped = true;
+        GTEST_SKIP();
+      }
+    }
+
     // Reads the current test instance name.
     // Name gets also sanitized (special chars are removed) to avoid issues when recording or
     // creating
--- a/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
+++ b/sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
@@ -134,37 +134,6 @@ namespace Azure { namespace Core { namespace Test {
 
     bool shouldSkipTest() { return m_wasSkipped; }
 
-    void SkipTest()
-    {
-      if (!m_wasSkipped)
-      {
-        m_wasSkipped = true;
-        GTEST_SKIP();
-      }
-    }
-
-    inline void ValidateSkippingTest()
-    {
-      if (m_wasSkipped)
-      {
-        GTEST_SKIP();
-      }
-    }
-
-    bool IsValidTime(const Azure::DateTime& datetime)
-    {
-      // Playback won't check dates
-      if (m_testContext.IsPlaybackMode())
-      {
-        return true;
-      }
-
-      // We assume datetime within a week is valid.
-      const auto minTime = std::chrono::system_clock::now() - std::chrono::hours(24 * 7);
-      const auto maxTime = std::chrono::system_clock::now() + std::chrono::hours(24 * 7);
-      return datetime > minTime && datetime < maxTime;
-    }
-
     // Reads the current test instance name.
     // Name gets also sanitized (special chars are removed) to avoid issues when recording or
     // creating
@@ -236,11 +205,13 @@ namespace Azure { namespace Core { namespace Test {
       return std::make_unique<T>(url, credential, options);
     }
 
+    template <class T> void InitClientOptions(T& options) { PrepareOptions(options); }
+
     template <class T> T InitClientOptions()
     {
       // Run instrumentation before creating the client
       T options;
-      PrepareOptions(options);
+      InitClientOptions(options);
       return options;
     }
 
--- a/sdk/core/azure-core-test/src/test_proxy_manager.cpp
+++ b/sdk/core/azure-core-test/src/test_proxy_manager.cpp
@@ -95,7 +95,6 @@ void TestProxyManager::StartPlaybackRecord(TestMode testMode)
   if (IsPlaybackMode() || IsRecordMode())
   {
     std::string mode = (IsPlaybackMode() ? "playback" : "record");
-    std::cout << "TestProxy already in " + mode + " mode.";
     return;
   }
 
@@ -225,12 +224,50 @@ void TestProxyManager::SetProxySanitizer()
   Azure::Core::Url matcherRequest(m_proxy);
   matcherRequest.AppendPath("Admin");
   matcherRequest.AppendPath("SetMatcher");
-  const std::string matcherBody
-      = "{\"compareBodies\": false ,\"ignoreQueryOrdering\": true , \"ignoredHeaders\": "
-        "\"x-ms-copy-source,x-ms-proposed-lease-id,x-ms-lease-id,x-ms-file-change-"
-        "time,x-ms-file-creation-time,x-ms-file-last-write-time,x-ms-destination-"
-        "lease-id,x-ms-source-lease-id,x-ms-source-content-crc64,x-ms-content-crc64,x-ms-source-"
-        "content-md5,x-ms-content-md5,Content-MD5\"}";
+  std::string matcherBody;
+  {
+    auto jsonRoot = Json::_internal::json::object();
+    jsonRoot["compareBodies"] = false;
+    jsonRoot["ignoreQueryOrdering"] = true;
+    const std::vector<std::string> excludedHeaders = {
+        "Expect",
+        "Connection",
+    };
+    jsonRoot["excludedHeaders"] = std::accumulate(
+        excludedHeaders.begin(),
+        excludedHeaders.end(),
+        std::string(),
+        [](const std::string& lhs, const std::string& rhs) {
+          return lhs + (lhs.empty() ? "" : ",") + rhs;
+        });
+    const std::vector<std::string> ignoredHeaders = {
+        "x-ms-copy-source",
+        "x-ms-file-change-time",
+        "x-ms-file-creation-time",
+        "x-ms-file-last-write-time",
+        "x-ms-rename-source",
+    };
+    const std::vector<std::string> ignoreQueryParameters = {
+        "st",
+        "se",
+        "sig",
+    };
+    jsonRoot["ignoredHeaders"] = std::accumulate(
+        ignoredHeaders.begin(),
+        ignoredHeaders.end(),
+        std::string(),
+        [](const std::string& lhs, const std::string& rhs) {
+          return lhs + (lhs.empty() ? "" : ",") + rhs;
+        });
+    jsonRoot["ignoredQueryParameters"] = std::accumulate(
+        ignoreQueryParameters.begin(),
+        ignoreQueryParameters.end(),
+        std::string(),
+        [](const std::string& lhs, const std::string& rhs) {
+          return lhs + (lhs.empty() ? "" : ",") + rhs;
+        });
+    matcherBody = jsonRoot.dump();
+  }
 
   {
     Azure::Core::IO::MemoryBodyStream payloadStream(
--- a/sdk/core/azure-core-test/src/test_proxy_manager.cpp
+++ b/sdk/core/azure-core-test/src/test_proxy_manager.cpp
@@ -94,7 +94,6 @@ void TestProxyManager::StartPlaybackRecord(TestMode testMode)
 {
   if (IsPlaybackMode() || IsRecordMode())
   {
-    std::string mode = (IsPlaybackMode() ? "playback" : "record");
     return;
   }
 
--- a/sdk/core/azure-core-test/src/test_proxy_policy.cpp
+++ b/sdk/core/azure-core-test/src/test_proxy_policy.cpp
@@ -53,11 +53,10 @@ std::unique_ptr<RawResponse> TestProxyPolicy::Send(
   // Copy all headers
   for (auto& header : request.GetHeaders())
   {
-    if (header.first == "host")
+    if (header.first != "host")
     {
-      continue;
+      redirectRequest.SetHeader(header.first, header.second);
     }
-    redirectRequest.SetHeader(header.first, header.second);
   }
   // QP
   for (auto const& qp : request.GetUrl().GetQueryParameters())
--- a/sdk/core/azure-core-test/src/test_proxy_policy.cpp
+++ b/sdk/core/azure-core-test/src/test_proxy_policy.cpp
@@ -42,7 +42,10 @@ std::unique_ptr<RawResponse> TestProxyPolicy::Send(
   {
     // This is a download with keep connection open. Let's switch the request
     redirectRequest = Azure::Core::Http::Request(
-        request.GetMethod(), Azure::Core::Url(m_testProxy->GetTestProxy()), false);
+        request.GetMethod(),
+        Azure::Core::Url(m_testProxy->GetTestProxy()),
+        request.GetBodyStream(),
+        false);
   }
 
   redirectRequest.GetUrl().SetPath(request.GetUrl().GetPath());
--- a/sdk/core/azure-core-test/src/test_proxy_policy.cpp
+++ b/sdk/core/azure-core-test/src/test_proxy_policy.cpp
@@ -53,6 +53,10 @@ std::unique_ptr<RawResponse> TestProxyPolicy::Send(
   // Copy all headers
   for (auto& header : request.GetHeaders())
   {
+    if (header.first == "host")
+    {
+      continue;
+    }
     redirectRequest.SetHeader(header.first, header.second);
   }
   // QP
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -413,10 +413,18 @@ CURLcode CurlSession::Perform(Context const& context)
     if (hostHeader == headers.end())
     {
       Log::Write(Logger::Level::Verbose, LogMsgPrefix + "No Host in request headers. Adding it");
-      this->m_request.SetHeader("Host", this->m_request.GetUrl().GetHost());
+      std::string headerValue = this->m_request.GetUrl().GetHost();
+      auto port = this->m_request.GetUrl().GetPort();
+      if (port != 0)
+      {
+        headerValue += ":" + std::to_string(port);
+      }
+      this->m_request.SetHeader("Host", headerValue);
     }
-    auto isContentLengthHeaderInRequest = headers.find("content-length");
-    if (isContentLengthHeaderInRequest == headers.end())
+    if (this->m_request.GetMethod() != HttpMethod::Get
+        && this->m_request.GetMethod() != HttpMethod::Head
+        && this->m_request.GetMethod() != HttpMethod::Delete
+        && headers.find("content-length") == headers.end())
     {
       Log::Write(Logger::Level::Verbose, LogMsgPrefix + "No content-length in headers. Adding it");
       this->m_request.SetHeader(
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -413,13 +413,13 @@ CURLcode CurlSession::Perform(Context const& context)
     if (hostHeader == headers.end())
     {
       Log::Write(Logger::Level::Verbose, LogMsgPrefix + "No Host in request headers. Adding it");
-      std::string headerValue = this->m_request.GetUrl().GetHost();
+      std::string hostName = this->m_request.GetUrl().GetHost();
       auto port = this->m_request.GetUrl().GetPort();
       if (port != 0)
       {
-        headerValue += ":" + std::to_string(port);
+        hostName += ":" + std::to_string(port);
       }
-      this->m_request.SetHeader("Host", headerValue);
+      this->m_request.SetHeader("Host", hostName);
     }
     if (this->m_request.GetMethod() != HttpMethod::Get
         && this->m_request.GetMethod() != HttpMethod::Head
--- a/sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
+++ b/sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
@@ -7,32 +7,47 @@
 #include <azure/storage/blobs/blob_lease_client.hpp>
 #include <azure/storage/common/crypt.hpp>
 
+namespace Azure { namespace Storage { namespace Blobs { namespace Models {
+
+  bool operator==(const BlobHttpHeaders& lhs, const BlobHttpHeaders& rhs);
+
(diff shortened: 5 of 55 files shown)


antkmsft (approved):
You probably want George's review, but as for me, changes look good.
You are touching a lot of tests and as a result, code coverage may improve. I suggest to grab code coverage numbers from a successful run closer to the end of code review, and update the numbers in `ci.yml`s.

Jinming-Hu:
Thanks @antkmsft !

@gearama Can you please help review this PR?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -225,12 +224,50 @@ void TestProxyManager::SetProxySanitizer()
   Azure::Core::Url matcherRequest(m_proxy);
   matcherRequest.AppendPath("Admin");
   matcherRequest.AppendPath("SetMatcher");
-  const std::string matcherBody
-      = "{\"compareBodies\": false ,\"ignoreQueryOrdering\": true , \"ignoredHeaders\": "
-        "\"x-ms-copy-source,x-ms-proposed-lease-id,x-ms-lease-id,x-ms-file-change-"
-        "time,x-ms-file-creation-time,x-ms-file-last-write-time,x-ms-destination-"
-        "lease-id,x-ms-source-lease-id,x-ms-source-content-crc64,x-ms-content-crc64,x-ms-source-"
-        "content-md5,x-ms-content-md5,Content-MD5\"}";
+  std::string matcherBody;
```
gearama, line 227: you're missing a lot of the headers, some of those are not only for storage but for attestation, keyvault etc.

sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
```diff
@@ -7,103 +7,101 @@
...
+    properties = blobClient.GetProperties().Value;
+    EXPECT_EQ(properties.BlobSize, 2 * originalLength);
 
-    auto deleteResponse = appendBlobClient.Delete();
+    auto deleteResponse = blobClient.Delete();
     EXPECT_TRUE(deleteResponse.Value.Deleted);
-    EXPECT_THROW(appendBlobClient.Delete(), StorageException);
-  }
-
-  TEST_F(AppendBlobClientTest, CreateWithTags)
```
gearama, line 84: why are you removing this test ?

gearama:
need to make sure all the other piplines run , make a change in the read me in the folder of the services to kick in the test runs
image

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
```diff
@@ -236,11 +214,13 @@ namespace Azure { namespace Core { namespace Test {
       return std::make_unique<T>(url, credential, options);
     }
 
+    template <class T> void InitClientOptions(T& options) { PrepareOptions(options); }
```
gearama, line 217: what's the point ? toghether with the next change ?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -95,7 +95,6 @@ void TestProxyManager::StartPlaybackRecord(TestMode testMode)
   if (IsPlaybackMode() || IsRecordMode())
   {
     std::string mode = (IsPlaybackMode() ? "playback" : "record");
-    std::cout << "TestProxy already in " + mode + " mode.";
```
gearama, line 98: why remove this log line ?

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -412,10 +412,18 @@ CURLcode CurlSession::Perform(Context const& context)
...
-      this->m_request.SetHeader("Host", this->m_request.GetUrl().GetHost());
+      std::string headerValue = this->m_request.GetUrl().GetHost();
+      auto port = this->m_request.GetUrl().GetPort();
+      if (port != 0)
+      {
+        headerValue += ":" + std::to_string(port);
+      }
+      this->m_request.SetHeader("Host", headerValue);
     }
-    auto isContentLengthHeaderInRequest = headers.find("content-length");
```
LarryOsterman, line 417: I didn't see an explanation of the change to remove the content-length header for HEAD and DELETE.

Why was this change made? Is it a protocol correctness issue? something else?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: Why do we need it anyway?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: Oh, I thought the headers I removed were for storage.

Do you know which of those are necessary for attestation and keyvault?

Jinming-Hu: @gearama @LarryOsterman

sdk/storage/azure-storage-blobs/test/ut/append_blob_client_test.cpp
Jinming-Hu: This is already covered in test case `AppendBlobClientTest.CreateAppendDelete`

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
Jinming-Hu: This is useful when we already have an options, and we just want to append the record-playback policy into the pipeline.

sdk/core/azure-core/src/http/curl/curl.cpp
Jinming-Hu: Good catch! I added some explanation in the PR description.

Jinming-Hu:
> need to make sure all the other piplines run , make a change in the read me in the folder of the services to kick in the test runs
> image

@gearama  What change do I need to make to enable the tests?

antkmsft:
@Jinming-Hu, I think @gearama meant that you can touch some files in each of the library, so the CI would trigger a run of `cpp - (library) - ci`. The other way to do this is probably by manually kicking off these pipelines.

antkmsft:
@Jinming-Hu, you can trigger CI pipelines manually, as I just did for identity pipeline, but the runs won't be visible here.

image

Maybe it is better to just add a space into each library's `inc/library.hpp` header (undoing it after the PR is done).

I wonder if running non-CI pipelines would give us the same level of verification (although taking longer). These ones you can run via `/azp run cpp - (library)`.

Next week I'll bring this up, I think we should allow the `/azp` command to trigger runs for CI pipelines, even if none of the files from that library was modified in the PR. Because 1. a responsible human triggers the command knowing best what he wants; 2. This PR demonstrates it could be useful/needed sometimes.

Jinming-Hu:
@antkmsft Really? There's no way to trigger the pipeline without actually making changes? Is that how pipeline works? It's kind of inconvenient.

So we'll need to add an extra empty line or something in the `README.md` to trigger the pipeline. If all the tests pass, we revert the changes. It that correct?

Jinming-Hu:
@antkmsft thanks for the tips. There are  other issues to be resolved before this PR can be merged. I'll trigger other pipelines when I think it's ready for merging.

antkmsft:
@Jinming-Hu, apparently, you can trigger the non-CI pipelines, which are heavier, and do also run Live tests, but not the CI pipelines, for some reason. Maybe as a slight side effect of some config, but almost no one noticed until now (I've noticed before, but I didn't have that much to run so I worked around it. In your case, it shows that it is better we have that ability).

I would also ask for "run multiple pipelines via single comment" support. Every time I need to run 5 pipelines, I need to leave 5 commants, receiving 5 bot replies back. This at least clutters the comment space in a PR.

antkmsft:
FWIW, your current changes did succeed in the identity CI pipeline: https://dev.azure.com/azure-sdk/public/_build/results?buildId=2114159&view=results

Jinming-Hu:
> FWIW, your current changes did succeed in the identity CI pipeline: https://dev.azure.com/azure-sdk/public/_build/results?buildId=2114159&view=results

I noticed WinHTTP transport layer is not covered in the CI. Is this a known issue? Why isn't it tested in CI?

antkmsft:
Hmm, is it not? Can you open an issue and bring more details?

antkmsft:
Could it be just naming? The pipelines that build on Windows + libcurl, have it in their name, while the ones using WinHTTP don't have `_winhttp` in their name?

Jinming-Hu:
> Hmm, is it not? Can you open an issue and bring more details?

Opened an issue https://github.com/Azure/azure-sdk-for-cpp/issues/4245

sdk/core/azure-core-test/src/test_proxy_manager.cpp
gearama: not of the top of my head, just run the pipelines with your changes, if they pass then you are correct and you can ignore my comment :)

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
gearama: PrepareOptions is getting wrapped in InitClientOptions method that only calls the PrepareOptions. i don't see the benefit of this construct , what does the indirection provide vs calling PrepareOptions alone?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
gearama: it's pretty

ahsonkhan:
> @Jinming-Hu, apparently, you can trigger the non-CI pipelines, which are heavier, and do also run Live tests, but not the CI pipelines, for some reason. Maybe as a slight side effect of some config, but almost no one noticed until now (I've noticed before, but I didn't have that much to run so I worked around it. In your case, it shows that it is better we have that ability).
> 
> I would also ask for "run multiple pipelines via single comment" support. Every time I need to run 5 pipelines, I need to leave 5 commants, receiving 5 bot replies back. This at least clutters the comment space in a PR.

cc @danieljurek any suggestions? Is there an `azp run` command we could use to run difference service pipelines on a PR?

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
Jinming-Hu: `PrepareOptions` is a private method, thus cannot be called in inherited classes.

danieljurek:
> > @Jinming-Hu, apparently, you can trigger the non-CI pipelines, which are heavier, and do also run Live tests, but not the CI pipelines, for some reason. Maybe as a slight side effect of some config, but almost no one noticed until now (I've noticed before, but I didn't have that much to run so I worked around it. In your case, it shows that it is better we have that ability).
> > I would also ask for "run multiple pipelines via single comment" support. Every time I need to run 5 pipelines, I need to leave 5 commants, receiving 5 bot replies back. This at least clutters the comment space in a PR.
> 
> cc @danieljurek any suggestions? Is there an `azp run` command we could use to run difference service pipelines on a PR?
> 
> @antkmsft and @Jinming-Hu did you try something like `/azp run cpp - attestation - ci`? I assumed that would work to trigger a CI run of any pipeline (using its name), without having to manually make file changes.
> 
> > run:
> > Run all pipelines or specific pipelines for this repository using a comment. Use this command by itself to trigger all related pipelines, or specify specific pipelines to run.
> > Example: "run" or "run pipeline_name, pipeline_name, pipeline_name"

You _can_ trigger builds against PRs but not using `/azp` commands in the GitHub PR. This is because the `/azp` command only works on builds whose trigger conditions include the PR against which you're testing. We've requested that this feature be changed to support the scenario described.

To do this go to the project (public) where you want to trigger the pipeline and select the pipeline you want to trigger. 

Click on the blue `Run Pipeline` button: 
image

In the `Branch/tag` field enter `refs/pull/<PR-Number>/merge` (in this case `refs/pull/4241/merge`): 
image

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
gearama: then you could move it to protected

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -225,12 +224,50 @@ void TestProxyManager::SetProxySanitizer()
...
+  std::string matcherBody;
+  {
+    auto jsonRoot = Json::_internal::json::object();
+    jsonRoot["compareBodies"] = false;
+    jsonRoot["ignoreQueryOrdering"] = true;
+    const std::vector<std::string> excludedHeaders = {
+        "Expect",
+        "Connection",
+    };
+    jsonRoot["excludedHeaders"] = std::accumulate(
```
LarryOsterman, line 236: Why use std::vector and std::accumulate instead of using initializer syntax?

In other words, something like:
```c++
jsonRoot["excludedHeaders"] = { "Expect", "Connection" };
```

sdk/core/azure-core-test/src/test_proxy_policy.cpp
```diff
@@ -42,14 +42,21 @@ std::unique_ptr<RawResponse> TestProxyPolicy::Send(
...
-        request.GetMethod(), Azure::Core::Url(m_testProxy->GetTestProxy()), false);
+        request.GetMethod(),
+        Azure::Core::Url(m_testProxy->GetTestProxy()),
+        request.GetBodyStream(),
+        false);
   }
 
   redirectRequest.GetUrl().SetPath(request.GetUrl().GetPath());
 
   // Copy all headers
```
LarryOsterman, line 53: Update comment to explain why you're redacting "host"?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: @LarryOsterman Your code will emit string:

```
{"excludedHeaders":["Expect","Connection"]}
```

What we want is:

```
{"excludedHeaders": "Expect,Connection"}
```

This is defined by the test-proxy, we have to follow this grammer.

sdk/core/azure-core-test/src/test_proxy_policy.cpp
Jinming-Hu: @LarryOsterman I'm not redacting `Host`. It's to fix a bug. I added explanation in the description of the PR.

sdk/core/azure-core-test/src/test_proxy_manager.cpp
LarryOsterman: @#$@#$ - in other words, test-proxy's schema is broken JSON (there's no way that parameter should be a string instead of an array). Sigh. 

So be it.

LarryOsterman:
FWIW, I think I may have a short-term fix for the attestation tests; I'll include a better fix, but it's going to collide with the changes that Jinming has made (the ignored headers list appears to be highly storage specific, we need to include a bunch of headers to the ignored headers list). 

I think we also need to reengineer the list a bit to allow individual SDKs to add service specific ignored headers. that would allow the core list be the set of headers that apply to all SDKs and allow individual SDKs to add service specific headers.

Jinming-Hu:
Test  pipeline:

Storage: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2178053&view=results
Identity: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2178050&view=results
Keyvault: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2178052&view=results
Attestation: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2178047&view=results

sdk/core/azure-core-test/inc/azure/core/test/test_base.hpp
Jinming-Hu: `PrepareOptions` is used in multiple places. It's mainly for internal implementation. I don't feel comfortable exposing it to derived classes.

Jinming-Hu:
@LarryOsterman @antkmsft @gearama

I think this PR is now ready for merging. I rebased on top of the latest main branch. CI pipelines all passed. Can you review the PR again?

Jinming-Hu:
ping @LarryOsterman @gearama @antkmsft

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -95,7 +95,6 @@ void TestProxyManager::StartPlaybackRecord(TestMode testMode)
   if (IsPlaybackMode() || IsRecordMode())
   {
     std::string mode = (IsPlaybackMode() ? "playback" : "record");
```
LarryOsterman, line 97: What is the purpose of this line? It appears to initialize a local variable named `mode` and then immediately return without using the local variable.

In the old code, it was used for a printf, but the new code seems to simply ignore the variable.

sdk/core/azure-core-test/src/test_proxy_manager.cpp
LarryOsterman: Attestation doesn't have any specific headers, beyond the standard ones like `x-ms-client-request-id`, etc.

I can't speak to keyvault.

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -225,12 +224,50 @@ void TestProxyManager::SetProxySanitizer()
...
-        "\"x-ms-copy-source,x-ms-proposed-lease-id,x-ms-lease-id,x-ms-file-change-"
-        "time,x-ms-file-creation-time,x-ms-file-last-write-time,x-ms-destination-"
-        "lease-id,x-ms-source-lease-id,x-ms-source-content-crc64,x-ms-content-crc64,x-ms-source-"
-        "content-md5,x-ms-content-md5,Content-MD5\"}";
+  std::string matcherBody;
+  {
+    auto jsonRoot = Json::_internal::json::object();
+    jsonRoot["compareBodies"] = false;
+    jsonRoot["ignoreQueryOrdering"] = true;
+    const std::vector<std::string> excludedHeaders = {
```
LarryOsterman, line 232: Btw, I *love* this change, I'm not a fan of raw JSON.

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -225,12 +224,50 @@ void TestProxyManager::SetProxySanitizer()
...
+    };
+    jsonRoot["excludedHeaders"] = std::accumulate(
+        excludedHeaders.begin(),
+        excludedHeaders.end(),
+        std::string(),
+        [](const std::string& lhs, const std::string& rhs) {
+          return lhs + (lhs.empty() ? "" : ",") + rhs;
+        });
+    const std::vector<std::string> ignoredHeaders = {
+        "x-ms-copy-source",
```
LarryOsterman, line 244: These headers appear to be specific to storage, would it make sense to have a mechanism which allows each service test to add its own service-specific headers to ignore in recordings?

Otherwise each service which has its own custom headers is going to need to modify the global test base.

sdk/core/azure-core-test/src/test_proxy_manager.cpp
```diff
@@ -225,12 +224,50 @@ void TestProxyManager::SetProxySanitizer()
...
+        "x-ms-file-change-time",
+        "x-ms-file-creation-time",
+        "x-ms-file-last-write-time",
+        "x-ms-rename-source",
+    };
+    const std::vector<std::string> ignoreQueryParameters = {
+        "st",
+        "se",
```
LarryOsterman, line 252: Same comment as above - I don't think these are common to all services and appear to be storage specific.

sdk/core/azure-core-test/src/test_proxy_policy.cpp
```diff
@@ -42,14 +42,21 @@ std::unique_ptr<RawResponse> TestProxyPolicy::Send(
...
   for (auto& header : request.GetHeaders())
   {
+    if (header.first == "host")
```
LarryOsterman, line 56: Minor style nitpicking: In general, I would prefer to see something like:

```cpp
if (header.first != "host")
{
  redirectRequest.SetHeader(header.first, header.second);
}
```
simply because (to me) it makes the intent of the change clearer.
```

File: sdk/core/azure-core/src/http/curl/curl.cpp [hunk 13]
```diff
@@ -413,10 +413,18 @@ CURLcode CurlSession::Perform(Context const& context)
     if (hostHeader == headers.end())
     {
       Log::Write(Logger::Level::Verbose, LogMsgPrefix + "No Host in request headers. Adding it");
-      this->m_request.SetHeader("Host", this->m_request.GetUrl().GetHost());
+      std::string headerValue = this->m_request.GetUrl().GetHost();
```
LarryOsterman, line 416: I think the `headerValue` variable is probably better named `hostName` since that's closer to its value.

LarryOsterman (changes requested):

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: > would it make sense to have a mechanism which allows each service test to add its own service-specific headers to ignore in recordings?

@LarryOsterman That would be the best. Such mechanism is not the purpose of this PR. (It's not a change that can be done with a few lines of code change. It should be a separate PR.)

I'm not sure how often we will modify these headers or query parameters. So I don't know if the benefit is worth the effort. How about we just leave them here for a while and revisit later?

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: Yeah, maybe we need some mechanism so that we can set these headers and query parameters at per-service leve.

sdk/core/azure-core-test/src/test_proxy_policy.cpp
Jinming-Hu: Ok

sdk/core/azure-core/src/http/curl/curl.cpp
Jinming-Hu: Sure

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: removed

sdk/core/azure-core-test/src/test_proxy_manager.cpp
Jinming-Hu: Tracking in https://github.com/Azure/azure-sdk-for-cpp/issues/4372

LarryOsterman (approved):
It looks good to me, don't forget to handle #4372 at some point :).

gearama (approved):

PR #4251, opened 2023-01-17
Removed unused APIs `SetScheme()` and `SetPort()` from `Core::Url`.

ahsonkhan:
These APIs are only used in testing atm.

After construction from a url string, these attributes don't need to be changed. We want to make more things immutable, unless absolutely necessary. For examples, aspects of the Url like query parameters need to remain mutable.

This came out of the discussion on client parameter validation, and how much validation we should do to keep core types well-formed after construction.

If we want to keep these, we need to decide on appropriate validation. For example, a scheme cannot allow any arbitrary characters:
https://www.rfc-editor.org/rfc/rfc1738#section-2.1
> Scheme names consist of a sequence of characters. The lower case
   letters "a"--"z", digits, and the characters plus ("+"), period
   ("."), and hyphen ("-") are allowed. For resiliency, programs
   interpreting URLs should treat upper case letters as equivalent to
   lower case in scheme names (e.g., allow "HTTP" as well as "http").

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Removed unused APIs `SetScheme()` and `SetPort()` from `Azure::Core::Url`. After construction from a url string, these attributes don't need to be changed.
+
 ### Bugs Fixed
 
 ### Other Changes
@@ -19,7 +21,7 @@
 ### Breaking Changes
 
 - Bearer token authentication will not work for endpoint URL protocol schemes other than `"https"`. This ensures token security and is consistent with the Azure SDKs for other languages.
-- Removed `noexcept` specification from `Azure::DateTime::clock::now()`. 
+- Removed `noexcept` specification from `Azure::DateTime::clock::now()`.
 
 ## 1.8.0-beta.2 (2022-11-03)
 
--- a/sdk/core/azure-core/inc/azure/core/url.hpp
+++ b/sdk/core/azure-core/inc/azure/core/url.hpp
@@ -15,6 +15,13 @@
 #include <memory>
 #include <string>
 
+#if defined(TESTING_BUILD)
+// Define the class used from tests to validate http scheme
+namespace Azure { namespace Core { namespace Test {
+  class TransportAdapterOptions_ProxyWithPasswordHttp_Test;
+}}} // namespace Azure::Core::Test
+#endif
+
 namespace Azure { namespace Core {
   namespace _detail {
     inline std::string FormatEncodedUrlQueryParameters(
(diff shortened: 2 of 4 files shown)


Jinming-Hu:
Are you going to bump up the major version?

ahsonkhan:
> Are you going to bump up the major version?

TBD. That is up for discussion with the team, based on the specifics of this particular change, assuming we are going forward with it.

antkmsft:
I like the change, but it might be too late for a breaking change, and not worth doing it.
Customers could be already using it.
Good news, it is harmless - someone sets scheme, so nothing bad happens.

Validation on these methods would be technically breaking. We could add it, but I don't see big problem there either - you have no chance to succeed with invalid characters anyway.

gearama:
The old saying once it's out there you can't take it back.

LarryOsterman:
> it's a public API, you can potentially break customers. What is the policy around this? do we need to bump the major version ? Do we need to bump the deps versions along with it?

What evidence do we have of customers using this type? From what I see, it's not a part of the public API surface, which means it probably should have been tagged as _internal from the get-go. And if a customer is using it, the aren't using it to interact with the SDK.

More importantly, the Url class should not and can not be considered a C++ equivalent to .Net's System.Uri because it doesn't have the same correctness guarantees that System.Uri has.

We could take on the work to ensure that Azure::Core::Uri has the same level of correctness, but since it's not a part of our API surface, I'm not sure it's worth the effort.

antkmsft:
@LarryOsterman, but it has been already GA'd, it is not in `_internal` namespace (we couldn't make the whole Url class `_internal` if that's what you mean + being public is one of the reasons for this class to exist at all - so clients can have Url constructor parameter). So it is in the `Azure::Core` namespace, GA since `v1.0.0` then all it methods are de facto part of the SDK's public API surface until `azure-core` v2.0.

ahsonkhan:
The `Url` class isn't intended to be used in client ctors (we explicitly decided to keep those as strings), and isn't used in any public surface area of **service** SDKs. Like I said, it was largely meant to be an internal type, used as implementation detail.

It isn't in question that the Url class happens to be public, and removing it or it's methods would be "technically" breaking, but the discussion is more nuanced here and we are considering the intended usage and whether any customer would actually use the type given our SDK shape (not discuss whether theoretically they could).

The only reason this type ended up being public is because it is used in the `Request` ctor (which was also meant to be internal). But, `Request` is in the `Send` method for `HttpPipelinePolicy`. If a customer wants to create their own policy, they would have to implement that method. The Url type just happens to be carried along the Request object, but isn't an exchange type in any other scenario.

ahsonkhan:
Looks like Core::Url's `SetScheme()` **is used** within the websockets feature branch:

https://github.com/Azure/azure-sdk-for-cpp/blob/7c529242885e40f5815599896b103b81482e553b/sdk/core/azure-core/src/http/curl/curl_websockets.cpp#L43-L56

It isn't decided whether websockets will be in its own package or be part of azure-core. Rather than removing these APIs (or make them private and rely on `friend`), we have decided to keep them as public, and add appropriate input validation.

Therefore, closing this PR.

ahsonkhan:
Will address as part of https://github.com/Azure/azure-sdk-for-cpp/issues/4279

Issue #4250, opened 2023-01-17
Warning C4100 for unreferenced formal parameter in the OpenTelemetry installed dependency, when building the tests in the repo via command line.


ahsonkhan:
I get this build warning coming from the installed vcpkg dependency, when building the repo from the command line (on Windows):
```text
mkdir build
cd build
cmake ..
cmake --build .
```

```text 
opentelemetry.cpp
  azure-core-tracing-opentelemetry.vcxproj -> E:\GitHub\Fork\azure-sdk-for-cpp\build\sdk\core\azure-core-tracing-opente
  lemetry\Debug\azure-core-tracing-opentelemetry.lib
  azure_core_otel_test.cpp
  service_support_test.cpp
E:\GitHub\Fork\azure-sdk-for-cpp\build\vcpkg_installed\x64-windows-static\include\opentelemetry/sdk/trace/simple_proces
sor.h(44,57): error C2220: the following warning is treated as an error [E:\GitHub\Fork\azure-sdk-for-cpp\build\sdk\cor
e\azure-core-tracing-opentelemetry\test\ut\azure-core-tracing-opentelemetry-test.vcxproj]
E:\GitHub\Fork\azure-sdk-for-cpp\build\vcpkg_installed\x64-windows-static\include\opentelemetry/sdk/trace/simple_proces
sor.h(44,57): warning C4100: 'parent_context': unreferenced formal parameter [E:\GitHub\Fork\azure-sdk-for-cpp\build\sd
k\core\azure-core-tracing-opentelemetry\test\ut\azure-core-tracing-opentelemetry-test.vcxproj]
E:\GitHub\Fork\azure-sdk-for-cpp\build\vcpkg_installed\x64-windows-static\include\opentelemetry/sdk/trace/simple_proces
sor.h(43,28): warning C4100: 'span': unreferenced formal parameter [E:\GitHub\Fork\azure-sdk-for-cpp\build\sdk\core\azu
re-core-tracing-opentelemetry\test\ut\azure-core-tracing-opentelemetry-test.vcxproj]
E:\GitHub\Fork\azure-sdk-for-cpp\build\vcpkg_installed\x64-windows-static\include\opentelemetry/sdk/trace/simple_proces
sor.h(59,33): warning C4100: 'timeout': unreferenced formal parameter [E:\GitHub\Fork\azure-sdk-for-cpp\build\sdk\core\
[...]

PR #4252, opened 2023-01-17
Ignore warning C4100 on Windows for unused parameter in OpenTelemetry headers.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-cpp/issues/4250

Should we disable 4244 and 6323 as well, like in other files?


antkmsft (approved):

antkmsft:
> Should we disable 4244 and 6323 as well, like in other files?

Let's disable all warnings from 3rd party libraries, there's nothing we can do, so they are unactionable but they clutter the warning log.

LarryOsterman:
> > Should we disable 4244 and 6323 as well, like in other files?
> 
> Let's disable all warnings from 3rd party libraries, there's nothing we can do, so they are unactionable but they clutter the warning log.

I agree with Anton. When we encounter warnings like this in 3rd party libraries we should disable them. But IMHO we only do it when we encounter them.

And we should probably file issues against those libraries (when possible) to get the issues resolved.

gearama (approved):
shoudl we hide the warnings or look into why the warnings?

LarryOsterman:
> shoudl we hide the warnings or look into why the warnings?

When I've seen this before, it's because 3rd party code has bugs in it. Our nightly CI pipeline runs the VC analytics which find these problems, but the actual root bug is in 3rd party code.

As far as I know, the best we can do is to suppress the warnings and file issues against the 3rd pary component to get them to fix their bugs.

ahsonkhan:
@LarryOsterman I don't see any recent pipeline runs failing with this warning treated as error. This error seems to only show up locally for me. You mentioned some of our open telemetry nightly runs are failing with the warning that I am fixing with this PR. Can you point me to one, please? Is anyone else hitting the same error?
https://dev.azure.com/azure-sdk/internal/_build?definitionId=1615&_a=summary

ahsonkhan:
The build error is unrelated, due to a link verification error from https://github.com/Azure/azure-sdk-for-cpp/pull/4269
Resolved in https://github.com/Azure/azure-sdk-for-cpp/pull/4273/

PR #4259, opened 2023-01-18
Update changelog for the Azurite fix

antkmsft:
Update changelog for the issue #4206 that was fixed in PR #4208. Thank you @ahojnnes!

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -8,6 +8,14 @@
 
 ### Bugs Fixed
 
+- Fixed Azurite parallel connectivity problem (#4206).
+
+### Acknowledgments
+
+Thank you to our developer community members who helped to make Azure Core better with their contributions to this release:
+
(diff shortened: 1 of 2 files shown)


ahsonkhan (approved):
Thanks!

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
 
 ### Bugs Fixed
 
+- [[#4206]](https://github.com/Azure/azure-sdk-for-cpp/issues/4206) Azurite parallel connectivity problem.
```
LarryOsterman, line 11: From what I understand, this is not an issue with Azurite, but instead a general problem which occurs whenever a server terminates the TCP connection during the HTTP exchange.

LarryOsterman: Maybe "Fixed connectivity issues which can occur if a TCP connection is dropped prematurely"?

antkmsft: @LarryOsterman, #4262

PR #4262, opened 2023-01-19
Update changelog wording

antkmsft:
Per https://github.com/Azure/azure-sdk-for-cpp/pull/4259#discussion_r1081725580

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugs Fixed
 
-- [[#4206]](https://github.com/Azure/azure-sdk-for-cpp/issues/4206) Fixed connectivity issues which can occur if a TCP connection is dropped prematurely.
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 
 ### Bugs Fixed
 
-- [[#4206]](https://github.com/Azure/azure-sdk-for-cpp/issues/4206) Azurite parallel connectivity problem.
+- [[#4206]](https://github.com/Azure/azure-sdk-for-cpp/issues/4206) Connectivity issues which can occur if a TCP connection is dropped prematurely.
```
ahsonkhan, line 11: Nit: Add the word "Fixed"?
```suggestion
- [[#4206]](https://github.com/Azure/azure-sdk-for-cpp/issues/4206) Fixed connectivity issues which can occur if a TCP connection is dropped prematurely.
```

ahsonkhan (approved):

sdk/core/azure-core/CHANGELOG.md
antkmsft: Yes, thank you - I thought we don't add "Fixed X" for bugs, but just put "X", but I just looked at our logs and there is no real consistency - we use both forms. I like "Fixed ..." more in this situation, my first instinct was also to put it.

PR #4270, opened 2023-01-24
[Perf] Refactor tests.yml into one file per package per language

mikeharder:
- Depends on Azure/azure-sdk-tools#5083


gearama (approved):

sdk/storage/azure-storage-blobs/perf-tests.yml
```diff
@@ -0,0 +1,35 @@
+Service: storage-blob
+
+Project: azure-storage-blobs-perf
+
+PrimaryPackage: azure-storage-blobs-cpp
+
+PackageVersions:
+- azure-storage-blobs-cpp: 12.6.2
```
Jinming-Hu, line 8: What is this version? Should this always point to the latest GA version, or can it be beta?

Do we need to update this version every time we ship a new release?

mikeharder: This version should usually point to the latest GA version (or a few versions older).  It is updated manually every month or so, currently by the EngSys team, after we have validated there are no regressions in main.

I am considering adding a special version like "latest-ga" that always points to the latest GA package on the package manager (https://github.com/Azure/azure-sdk-tools/issues/3553), but am hesitant because we might automatically upgrade and miss a regression.

We can test multiple versions (not just two), so yes we can add more versions, including betas.  If you would like to add more versions, please reach out to me and we can discuss how to add them and how it impacts the perf pipelines.

PR #4303, opened 2023-01-30
Policies do not overwrite HTTP headers if they are already set

### LINKED ISSUE #4297: TelemetryPolicy should not be setting User-Agent if it is set already
Autorest-test has a testcase that makes the codegen to set `User-Agent` header, and expects to receive a specific value. `TelemetryPolicy` overwrites that value.

.NET SDK supports it: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/AutoRest.TestServer.Tests/header.cs#L22-L27

Java SDK supports it: https://github.com/Azure/autorest.java/blob/0a11263d5a3e28f042fdfd8c535c61d0335ffd26/protocol-tests/src/samples/java/fixtures/header/generated/HeaderParamExistingKey.java

Python as well: https://github.com/Azure/autorest.python/blob/c1dd57d1ffd45199744d8c9883447f05f61f2286/packages/autorest.python/test/vanilla/version-tolerant/AcceptanceTests/test_header.py#L162-L166

Go SDK has it commented out, because they have the same problem: https://github.com/Azure/autorest.go/blob/main/test/autorest/headergroup/headergroup_test.go#L118-L127

### LINKED ISSUE #4301: RequestIdPolicy should not override the value of `x-ms-client-request-id` header if it is set already
Same situation and circumstances as for #4297, only this time, Go SDK does not have the same problem as we do.

antkmsft:
Closes #4297
Closes #4301


sdk/core/azure-core/src/http/telemetry_policy.cpp
```diff
@@ -13,6 +13,10 @@ std::unique_ptr<RawResponse> Azure::Core::Http::Policies::_internal::TelemetryPo
     NextHttpPolicy nextPolicy,
     Context const& context) const
 {
-  request.SetHeader("User-Agent", m_telemetryId);
```
LarryOsterman, line 16: I would argue that having a separate policy which sets User-Agent is a coding error. The Azure SDK has fairly strict requirements on user-agent, if the telemetry_policy is in the pipeline, it means that developers expect that the pipeline will implement the Azure SDK guidelines.

If we have tests which depend on the user-agent policy not being in the pipeline, we should change the tests.

antkmsft: This is about bringing codegen up to spec.

Autorest.testserver has a test (https://github.com/Azure/autorest.testserver/blob/main/swagger/header.json#L13-L31) that specifically check that if the swagger says that `User-Agent` should have some specific value, the generated code will send that and nothing else. Not even appending. Java, .NET, and Python do pass that test. They have logic to intentionally enable this scenario.

There is another similar test, for `x-ms-client-request-id`, C++ fails it, but Go, .NET, and Python do pass.

The test specifies these headers intentionally, there is no other way to read it. For example, for User-Agent, the value should be `"overwrite"`, i.e. "overwrite the user-agent". As almost it expects that naive policy implementation would simply overwrite the values, and wants to prevent that.

LarryOsterman: Why can't the test configure the pipeline with the desired set of policies, excluding the TelemetryPolicy and RequestActivityPolicy?

antkmsft: It is not the SDK test, it is codegen test. Codegen generates client with the default set of options. It should not be generating new pipeline without a policy just for one method that happens to use `User-Agent` or `x-ms-client-request-id`.

antkmsft: For example, for that user-agent test, the following code gets generated:

```
Response<ParamHeaderExistingKeyResult> HeaderClient::ParamExistingKey(
      const ParamHeaderExistingKeyOptions& options,
      const Core::Context& context)
  {
    auto url = m_url;
    url.AppendPath("header/param/existingkey");
    auto request = Core::Http::Request(Core::Http::HttpMethod::Post, url);
    if (!options.UserAgent.empty())
    {
      request.SetHeader("User-Agent", options.UserAgent);
    }
    auto pRawResponse = m_pipeline->Send(request, context);
    auto httpStatusCode = pRawResponse->GetStatusCode();
    if (httpStatusCode != Core::Http::HttpStatusCode::Ok)
    {
      throw Core::RequestFailedException(pRawResponse);
    }
    ParamHeaderExistingKeyResult response;
    return Response<ParamHeaderExistingKeyResult>(std::move(response), std::move(pRawResponse));
  }
```

LarryOsterman: In the sample above, where did m_pipeline come from? Presumably it was created by test code, which we control.

Btw, I checked the .Net implementation for both TelemetryPolicy and RequestActivityPolicy - TelemetryPolicy has a similar "if not set use the default" behavior you describe. RequestActivityPolicy is somewhat different - it depends on the ClientRequestId property in the Request object, which is defaulted in the constructor to be a Guid.NewGuid.

antkmsft: It is not created by the test code, it is created by codegen for clients. Codegen does not know whether it generates client for tests or for an Azure service.
It gets generated as such:

Header -
```cpp
  class HeaderClient final {
  private:
    std::shared_ptr<Azure::Core::Http::_internal::HttpPipeline> m_pipeline;
    Core::Url m_url;

  public:
    HeaderClient();

    Response<ParamHeaderExistingKeyResult> ParamExistingKey(
        const ParamHeaderExistingKeyOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderExistingKeyResult> ResponseExistingKey(
        const Core::Context& context = {});

    Response<ParamHeaderProtectedKeyResult> ParamProtectedKey(
        const ParamHeaderProtectedKeyOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderProtectedKeyResult> ResponseProtectedKey(
        const Core::Context& context = {});

    Response<ParamHeaderIntegerResult> ParamInteger(
        const ParamHeaderIntegerOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderIntegerResult> ResponseInteger(
        const ResponseHeaderIntegerOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderLongResult> ParamLong(
        const ParamHeaderLongOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderLongResult> ResponseLong(
        const ResponseHeaderLongOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderFloatResult> ParamFloat(
        const ParamHeaderFloatOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderFloatResult> ResponseFloat(
        const ResponseHeaderFloatOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderDoubleResult> ParamDouble(
        const ParamHeaderDoubleOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderDoubleResult> ResponseDouble(
        const ResponseHeaderDoubleOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderBoolResult> ParamBool(
        const ParamHeaderBoolOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderBoolResult> ResponseBool(
        const ResponseHeaderBoolOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderStringResult> ParamString(
        const ParamHeaderStringOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderStringResult> ResponseString(
        const ResponseHeaderStringOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderDateResult> ParamDate(
        const ParamHeaderDateOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderDateResult> ResponseDate(
        const ResponseHeaderDateOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderDatetimeResult> ParamDatetime(
        const ParamHeaderDatetimeOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderDatetimeResult> ResponseDatetime(
        const ResponseHeaderDatetimeOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderDatetimeRfc1123Result> ParamDatetimeRfc1123(
        const ParamHeaderDatetimeRfc1123Options& options,
        const Core::Context& context = {});

    Response<ResponseHeaderDatetimeRfc1123Result> ResponseDatetimeRfc1123(
        const ResponseHeaderDatetimeRfc1123Options& options,
        const Core::Context& context = {});

    Response<ParamHeaderDurationResult> ParamDuration(
        const ParamHeaderDurationOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderDurationResult> ResponseDuration(
        const ResponseHeaderDurationOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderByteResult> ParamByte(
        const ParamHeaderByteOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderByteResult> ResponseByte(
        const ResponseHeaderByteOptions& options,
        const Core::Context& context = {});

    Response<ParamHeaderEnumResult> ParamEnum(
        const ParamHeaderEnumOptions& options,
        const Core::Context& context = {});

    Response<ResponseHeaderEnumResult> ResponseEnum(
        const ResponseHeaderEnumOptions& options,
        const Core::Context& context = {});

    Response<CustomHeaderRequestIdResult> CustomRequestId(const Core::Context& context = {});
  };
```

.cpp file -
```cpp
  HeaderClient::HeaderClient()
      : m_pipeline(
          new Core::Http::_internal::HttpPipeline({}, "azure-test-autorest-header", "0", {}, {})),
        m_url("http://localhost:3000")
  {
  }
```

LarryOsterman: In my response above, I thought I described the behavior for the .Net SDK: It initializes the ClientRequestId to a GUID in the Request constructor,  and then blindly overwrites any existing x-ms-client-request-id header. That's why you see the "preservation" behavior (which isn't *quite* preservation since it will overwrite any headers set in the headers map).  And the TelemetryPolicy uses the UserAgent header set by the client if it's set.

FWIW, I usually try to track down the implementation of the functionality rather than the tests because often times the tests have hidden assumptions that may not be backed up in live code (for instance dealing with mocked HTTP pipelines or transports).

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
```diff
@@ -430,9 +430,12 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
           NextHttpPolicy nextPolicy,
           Context const& context) const override
       {
-        auto uuid = Uuid::CreateUuid().ToString();
+        if (!request.GetHeader(RequestIdHeader).HasValue())
```
ahsonkhan, line 433: Add a changelog entry for these changes please, since it is an observable behavioral change. Maybe it's a bug fix?

ahsonkhan (approved):
Should this overwrite behavior apply to all request headers or just the two specific ones you are changing?

Before merging, please make sure (with consistency across languages), this is precisely the behavior we want. The change looks good otherwise.

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
antkmsft: I don't think it is visible to the user: policies are internal, HTTP requests are being created by the Clients.
The SDK will continue to beave exactly as it was before.

antkmsft:
@ahsonkhan, there is rather a consistency, see details in the linked issues.

I am not aware of any other headers that we're setting.

sdk/core/azure-core/src/http/telemetry_policy.cpp
antkmsft: Here is `x-ms-client-request-id` preservation logic in the Azure SDK for Java:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RequestIdPolicy.java#L40-L43

PR #4305, opened 2023-01-30
Add date-only ToString() for DateTime

### LINKED ISSUE #4300: `DateTime` should have `ToString()` that gives back just the date
In order to be able to interoperate with JSON dates (not datetimes), there needs to be `ToString()` which does not print the time portion in case the time is `"00:00:00T0000000Z"`. It can be `TimeFractionFormat::Date`.

Alternatively, there can be `ToDateString()` which only gives back the date regardless of time (Or, `TimeFractionFormat::Date` could do that).

It would allow us to send values like `"2010-01-01"` (https://github.com/Azure/autorest.testserver/blob/main/swagger/header.json#L554)

It will be basically implementation of support for the date (not datetime) type.

antkmsft:
Closes #4300

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/datetime.hpp
+++ b/sdk/core/azure-core/inc/azure/core/datetime.hpp
@@ -152,7 +152,10 @@ public:
     AllDigits,
 
     /// Drop all the fractional time digits.
-    Truncate
+    Truncate,
+
+    /// Include only the date portion, skipping time and time zone.
+    Date,
   };
 
   /**
--- a/sdk/core/azure-core/test/ut/datetime_test.cpp
+++ b/sdk/core/azure-core/test/ut/datetime_test.cpp
@@ -902,3 +902,14 @@ TEST(DateTime, Rfc3339Space)
       = DateTime::Parse("2022-08-24 00:43:08.0004308Z", DateTime::DateFormat::Rfc3339);
   EXPECT_EQ(datetime.ToString(DateTime::DateFormat::Rfc3339), "2022-08-24T00:43:08.0004308Z");
 }
+
+TEST(DateTime, DateOnly)
+{
+  {
+    auto dt = DateTime::Parse("1975-04-04T01:23:45.6789Z", DateTime::DateFormat::Rfc3339);
+
+    EXPECT_EQ(
+        dt.ToString(DateTime::DateFormat::Rfc3339, DateTime::TimeFractionFormat::Date),
+        "1975-04-04");
+  }
+}
--- a/sdk/core/azure-core/src/datetime.cpp
+++ b/sdk/core/azure-core/src/datetime.cpp
@@ -886,34 +886,38 @@ std::string DateTime::ToString(DateFormat format, TimeFractionFormat fractionFor
   std::ostringstream dateString;
 
   dateString << std::setfill('0') << std::setw(4) << static_cast<int>(year) << '-' << std::setw(2)
-             << static_cast<int>(month) << '-' << std::setw(2) << static_cast<int>(day) << 'T'
-             << std::setw(2) << static_cast<int>(hour) << ':' << std::setw(2)
-             << static_cast<int>(minute) << ':' << std::setw(2) << static_cast<int>(second);
+             << static_cast<int>(month) << '-' << std::setw(2) << static_cast<int>(day);
 
-  if (fractionFormat == TimeFractionFormat::AllDigits)
+  if (fractionFormat != TimeFractionFormat::Date)
   {
-    dateString << '.' << std::setw(7) << static_cast<int>(fracSec);
-  }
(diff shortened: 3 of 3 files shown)


LarryOsterman:
Silly question: Why do we need this?

antkmsft:
@LarryOsterman, it is described in the bug that's linked in the description, #4300. We need this to interchange with the swaggers that require date and that do not accept datetime.

LarryOsterman:
> @LarryOsterman, it is described in the bug that's linked in the description, #4300. We need this to interchange with the swaggers that require date and that do not accept datetime.

FWIW, I feel fairly strongly that pull requests shouldn't force reviewer chase down the linked bug, especially if it's straightforward. The pull request description should contain at least a base level of information that the reviewers need. That is especially true for straightforward pull requests like this one.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -152,7 +152,10 @@ class DateTime final : public _detail::Clock::time_point {
     AllDigits,
 
     /// Drop all the fractional time digits.
-    Truncate
+    Truncate,
+
+    /// Include only the date portion, skipping time and time zone.
+    Date,
```
LarryOsterman, line 158: I think this should be `DateOnly`, which correctly reflects the behavior.

Note that if you specify `TimeFractionFormat::Date`, you are no longer generating valid Rfc3339 dates, but the documentation doesn't reflect that specifying `TimeFractionFormat::Date` breaks the `DateFormat` contract.

 As such, you should consider instead of using `TimeFractionFormat` as the mechanism to represent these, you might want to use `DateFormat::Rfc3339FullDate`.

sdk/core/azure-core/inc/azure/core/datetime.hpp
```diff
@@ -152,7 +152,10 @@ class DateTime final : public _detail::Clock::time_point {
...
   };
 
   /**
```
LarryOsterman, line 161: The comment for `Rfc3339` should express which construct in `Rfc3339` the DateFormat conforms to (I believe it's `date-time`, based on the code, but the comment should be specific).

LarryOsterman (changes requested):

Jinming-Hu:
As far as I know, Date and DateTime are usually implemented as two separate types in other languages, like .Net [DateOnly](https://learn.microsoft.com/en-us/dotnet/api/system.dateonly?view=net-7.0), in Java [LocalDate](https://www.w3schools.com/java/java_date.asp), in Python [datetime.date](https://docs.python.org/3/library/datetime.html#datetime.date). Is there a strong reason you want to reuse the same type for Date in our C++ SDK?

LarryOsterman:
> As far as I know, Date and DateTime are usually implemented as two separate types in other languages, like .Net [DateOnly](https://learn.microsoft.com/en-us/dotnet/api/system.dateonly?view=net-7.0), in Java [LocalDate](https://www.w3schools.com/java/java_date.asp), in Python [datetime.date](https://docs.python.org/3/library/datetime.html#datetime.date). Is there a strong reason you want to reuse the same type for Date in our C++ SDK?

Remember that Azure::DateTime is a wrapper around the std::chrono::time_point type, as such it encapsulates both time and dates (std::chrono doesn't distinguish between "time" and "date"). If we were to bifurcate the DateTime into Date and Time, they would still be wrappers around std::chrono based types. 

I'm not sure that this use case is sufficiently unique to justify adding a brand-new time type (with all the inherent complexities of implementation (we'd need to add conversion functions between Date and DateTime, DateTime and Date, Time and DateTime, etc). In this case, we only need is a way of representing a time_point with a print representation which is just the `full-date` representation in RFC 2339.

sdk/core/azure-core/test/ut/datetime_test.cpp
```diff
@@ -902,3 +902,14 @@ TEST(DateTime, Rfc3339Space)
...
 }
+
+TEST(DateTime, DateOnly)
+{
+  {
+    auto dt = DateTime::Parse("1975-04-04T01:23:45.6789Z", DateTime::DateFormat::Rfc3339);
+
+    EXPECT_EQ(
+        dt.ToString(DateTime::DateFormat::Rfc3339, DateTime::TimeFractionFormat::Date),
+        "1975-04-04");
```
ahsonkhan, line 913: Would this be considered data loss when round-tripping @LarryOsterman? Should we only ever return `DateOnly` if the time is exactly 0?

LarryOsterman: I don't believe it's possible to round-trip this. The date-only format is an invalid Rfc3339 date-time string and we parse date-time strings, not date-only strings.

antkmsft:
Closing this PR for now, I'll get to it later and then I'll probably reopen it instead of submitting a new one. Thank you for your feedback, I'll address it.

PR #4307, opened 2023-01-31
move cleanup tool.

### LINKED ISSUE #4290: The `keyvault/cleanup` folder at the top level with a sample cpp file needs to be moved
Introduced in https://github.com/Azure/azure-sdk-for-cpp/pull/4222

The SDK directory structure and guideline is to have `/src`, `/test`, `/samples` within a `/sdk/<service>/<package>` directory (see https://azure.github.io/azure-sdk/policies_repostructure.html#sdk-directory-layout).

That doesn't match what we are doing here:
https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/keyvault/cleanup

The `keyvault/cleanup` folder seems to contain a single sample file, given the comment:
https://github.com/Azure/azure-sdk-for-cpp/blob/a3e9ad1d2c2509608ba6aaf78d0a400587e9b4fe/sdk/keyvault/cleanup/cleanup.cpp#L5-L6

This needs to be moved to a samples directory, or deleted, as appropriate.
One viable home could be:
https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/keyvault/azure-security-keyvault-secrets/samples

gearama:
closes #4290 
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


LarryOsterman (approved):

antkmsft (approved):

tools/cleanup/README.md
```diff
@@ -0,0 +1,6 @@
+# Get Environment Variable for Samples
```
ahsonkhan, line 1: I don't think this belongs in the general purpose `tools` folder. It isn't relevant to many SDKs or services. In fact, it only contains a copy of a cmake module for customers to use, which we might remove, all together.

We should put this in the keyvault-secrets/samples folder, instead. Here:
https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/keyvault/azure-security-keyvault-secrets/samples

ahsonkhan (changes requested):
Please move this to `keyvault-secrets` samples as it is keyvault specific and not at the root `tools` folder of the repo.

tools/cleanup/README.md
gearama: it's not specific to secrets, , it handles keys and certificates , it is a tool to cleanup a keyvault.

antkmsft: @gearama, can this go under `/keyvault/tools/cleanup` then?

ahsonkhan: Gotcha @gearama. In that case, please put it under `keyvault/samples`.

We should keep code and files specific to a particular service under it's parent folder.
Also, sample fits better than tool in this case, using our SDK APIs.

I have seen precedence of this in another language repo:
https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/keyvault/samples

gearama:
it's a tool thus goes under tools , it's not intended to be a sample, it's useful when diagnosing tests that attempt to create a new resource, not just a new version of the resource, since our tests do not do any cleanup, relying on the fact that new resources are deployed each time the pipeline runs and are removed up once the tests are done.

tools/cleanup/README.md
gearama: it's a tool thus goes under tools , it's not intended to be a sample, it's useful when diagnosing tests that attempt to create a new resource, not just a new version of the resource, since our tests do not do any cleanup, relying on the fact that new resources are deployed each time the pipeline runs and are removed up once the tests are done.

ahsonkhan:
In that case, consider moving it under the test directory.

ahsonkhan (approved):
Tests directory might be a better home, but tools under keyvault is better than at root.

ahsonkhan:
> it's not specific to secrets, , it handles keys and certificates

> it's a tool thus goes under tools , it's not intended to be a sample

@gearama The comment in the file is confusing. Please update the doc comment on this file to highlight what the purpose of this tool is, if it isn't a sample.

https://github.com/Azure/azure-sdk-for-cpp/blob/5ac3f00d589eb003162ed89adcc062b50dfd225c/sdk/keyvault/tools/cleanup/src/cleanup.cpp#L5-L6

PR #4317, opened 2023-02-01
Identity February Release

antkmsft:


gearama (approved):

LarryOsterman (approved):

sdk/identity/azure-identity/CMakeLists.txt
```diff
@@ -38,9 +38,9 @@ if(FETCH_SOURCE_DEPS)
         add_subdirectory(${oneFolder} EXCLUDE_FROM_ALL)
     endforeach()
 elseif(NOT AZ_ALL_LIBRARIES)
-  find_package(azure-core-cpp "1.3.1" CONFIG QUIET)
+  find_package(azure-core-cpp "1.8.0" CONFIG QUIET)
```
ahsonkhan, line 41: Is this the **minimum** version of core that identity now depends on? Will anything older cause it to break?

sdk/identity/azure-identity/CHANGELOG.md
```diff
@@ -1,14 +1,14 @@
 # Release History
 
-## 1.4.0-beta.4 (Unreleased)
+## 1.4.0 (2023-02-07)
 
 ### Features Added
```
ahsonkhan, line 5: I like we are emphasizing the key features added for the GA release.

ahsonkhan (approved):

ahsonkhan:
Presumably the CI error is expected until after core 1.8.0 ships.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2156207&view=logs&j=5b40fe45-3be6-55b8-8b6c-da8150e7e0d7&t=11993ae9-0d33-503c-0a4e-62d7d35853c8&l=80
```text
CMake Error at build/_deps/vcpkg-src/scripts/buildsystems/vcpkg.cmake:834 (_find_package):
  Could not find a configuration file for package "azure-core-cpp" that is
  compatible with requested version "1.8.0".

  The following configuration files were considered but not accepted:

    /Users/runner/work/1/s/sdk/identity/azure-identity/build/vcpkg_installed/x64-osx/share/azure-core-cpp/azure-core-cppConfig.cmake, version: 1.7.2

Call Stack (most recent call first):
  CMakeLists.txt:43 (find_package)

```

sdk/identity/azure-identity/CMakeLists.txt
antkmsft: Yes, it needs the new Core version, In Core's changelog for "1.8.0", there is an entry under "Other Changes": "Added support for Identity token caching, and for configuring token refresh offset in `BearerTokenAuthenticationPolicy`."

sdk/identity/azure-identity/CHANGELOG.md
antkmsft: Thank you, I also prefer this over the "Changes from Beta are now generally available", which would be most of the time the GA changelog, given that we do release as Beta first most of the time.

Yet this is not the full verbatim details concatenated from beta releases - some we do summarize in shorter sentences, or not mention at all (i.e. bug fixes for a bug introcuced in one beta, fixed in subsequent beta).

antkmsft:
@ahsonkhan, ah, yes, now that we build each package individually, plus after the manifest mode, that may not be working. Let me try to fix. The ideal case is we can update the `vcpkg.json` for core (the vcpkg.json that is used for manifest mode, not the one that gets used for releases), worst case, we drop the minimum version in CMakeLists - the changes in the `/vcpkg` directory should be sufficient.

PR #4327, opened 2023-02-02
Fix vcpkg install

antkmsft:
There is `ApiViewSetting.json` in each `inc` directory.
We install the contents of `inc` directory.
So when user installs multiple azure vcpkgs, there is a collision, because the file already exists, and build fails.

This fix is to install `inc/azure` directory into `{vcpkg_install_directory}/azure`, instead of simply `inc/*` into `{vcpkg_install_directory}`.

Better than that, there will no longer be `{vcpkg_install_directory}/ApiViewSettings.json` installed on user machines, which is not needed, especially in the include directory, especially at the root include level (i.e. before the fix they'd type `#include <`, and `ApiViewSettings.json` might have shown up)

### DIFF
--- a/cmake-modules/AzureVcpkg.cmake
+++ b/cmake-modules/AzureVcpkg.cmake
@@ -93,7 +93,7 @@ macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
   include(GNUInstallDirs)
 
   # When installing, copy our "inc" directory (headers) to "include" directory at the install location.
-  install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/inc/" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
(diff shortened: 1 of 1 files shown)


LarryOsterman (approved):

cmake-modules/AzureVcpkg.cmake
```diff
@@ -93,7 +93,7 @@ macro(az_vcpkg_export targetName macroNamePart dllImportExportHeaderPath)
   include(GNUInstallDirs)
 
   # When installing, copy our "inc" directory (headers) to "include" directory at the install location.
-  install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/inc/" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+  install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/inc/azure/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/azure")
```
ahsonkhan, line 96: Will this affect the header include paths of the SDK for customers?

Also, CL entry?

antkmsft: The change is invisible to our customers, everything will work exactly as before.

PR #4320, opened 2023-02-01
Prepare OpenTelemetry for beta release

LarryOsterman:
# Prepare OpenTelemetry and Core for February SDK release.

Also updates the DistributedTracing.md file to reflect the recent changes to distributed tracing.

## Pull Request Checklist

Please leverage this checklist as a reminder to address commonly occurring feedback when submitting a pull request to make sure your PR can be reviewed quickly:

See the detailed list in the [contributing guide](https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#pull-requests).

- [ ] [C++ Guidelines](https://azure.github.io/azure-sdk/cpp_introduction.html)
- [ ] Doxygen docs
- [ ] Unit tests
- [ ] No unwanted commits/changes
- [ ] Descriptive title/description
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] External references or docs updated
- [ ] Self review of PR done
- [ ] Any breaking changes?

PR #4331, opened 2023-02-02
Remove mention of testing related changes from telemetry changelog

ahsonkhan:
Follow-up from https://github.com/Azure/azure-sdk-for-cpp/pull/4320#discussion_r1094984178 

The test changes don't impact customers and trying to reduce noise to signal.

### DIFF
--- a/sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
+++ b/sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Aligned OpenTelemetry tracing infrastructure with OpenTelemetry 1.17.0 conventions for use with Azure Monitor.
 
 ### Other Changes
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core-tracing-opentelemetry/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
 - Aligned OpenTelemetry tracing infrastructure with OpenTelemetry 1.17.0 conventions for use with Azure Monitor.
 
 ### Other Changes
-- Updated test collateral to reflect Azure Core tracing changes.
+
```
antkmsft, line 10: Should we decide against that item removal, at least this newline between "Other Changes" should still be checked in.

antkmsft (approved):
I agree with the change, but I'd leave the final decision up to @LarryOsterman

PR #4343, opened 2023-02-07
Trigger`keyvault` on proxy changes

scbedd:
So I don't need to manually queue this anymore. I prefer to only merge updates to `eng/common/testproxy/target_version.txt` after a completed CI run from every major language that uses the proxy.

This PR just ensures that I don't need to remember to queue it manually anymore.

### DIFF
--- a/sdk/keyvault/ci.yml
+++ b/sdk/keyvault/ci.yml
@@ -21,6 +21,7 @@ pr:
   paths:
     include:
       - sdk/keyvault
+      - eng/common/testproxy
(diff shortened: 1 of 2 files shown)


gearama (commented):
i would make the same change to storage also

gearama:
@scbedd

scbedd:
> i would make the same change to storage also

Finally got a chance to get back to this. Added!

gearama (approved):

PR #4344, opened 2023-02-08
reworked docker file

gearama:
PR for @benbp 
reworked so that docker image builds the executable it needs instead fo having it fed from outside
# Pull Request Checklist

- [ ] Doxygen docs
- [ ] Unit tests
  - [ ] PR is single purpose
  - [ ] Related issue listed
- [ ] Comments in source
- [ ] No typos
- [ ] Update changelog
- [ ] Not work-in-progress
- [ ] Self review of PR done
- [ ] Any breaking changes?


sdk/core/azure-core/test/libcurl-stress-test/Dockerfile
```diff
@@ -3,11 +3,21 @@
...
+RUN su
 # install the mem check tool 
 RUN apt-get update -y
-RUN apt-get install valgrind -y
+RUN apt-get install gcc cmake make g++ git zip unzip build-essential pkg-config wget curl valgrind -y
+RUN wget -O vcpkg.tar.gz https://github.com/microsoft/vcpkg/archive/master.tar.gz
+RUN mkdir /opt/vcpkg
+RUN tar xf vcpkg.tar.gz --strip-components=1 -C /opt/vcpkg
+RUN /opt/vcpkg/bootstrap-vcpkg.sh
+RUN ln -s /opt/vcpkg/vcpkg /usr/local/bin/vcpkg
```
LarryOsterman, line 14: Forgive me, but this is somewht terrifying... We're downloading arbitrary code from the internet and running it without even checking hashes to verify that we're getting the binary we expected.

Then we download the azure SDK and rebuild the libcurl stress test, which we will run, again without verifying what we downloaded.

IMHO you should be pulling a binary from an official build (probably using universal artifacts) similar to how we get the APIView tool from our official builds.

Downloading unknown, arbitrary code from the internet and then compiling and running it is not a good idea.

LarryOsterman (dismissed):

sdk/core/azure-core/test/libcurl-stress-test/Dockerfile
gearama: How is this random code? 
What part of github  microsoft or vcpkg seem random? 
i know my web cam / mic might not be the best but do i look like a script kiddie that has no idea what they're doing? nothing is random there, every step has a purpose, 
if you would take five minutes to read the commands, they would make perfect sense. 

Allow me to elaborate : 
L9 -> install missing packages required for vcpkg and subsequent operations 
L10 -> wget the latest release from GITHUB under MICROSOFT organization VCPKG project 
L11 -> create folder under /opt to host vcpkg 
L12 -> extract the downloaded tar to it's destination and remove the first part of the name structure ( strip component)
L13 -> run bootstrap-vcpkg ( see vcpkg instructions ) 
L14 -> create a symbolic link in  usr/local/bin to  vcpkg under opt

Explain to me what is random here, what did i not understand, what did i put there for no reason whatsoever. 

Also how is that diffrent from  https://github.com/microsoft/vcpkg#quick-start-unix  ?
It's the same thing, wget is the equivalent of git clone , my way i get to control where things go more cleanly than git clone. 

And what is wrong with building the code?
That's what stress tests in go do https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azservicebus/internal/stress/Dockerfile . 
That's what i was requested to do, i did not do it because i had nothing better to do with my life. 

Since you bring up that security point please show me where in the CI or LIVE or any other CPP pipeline we verify the hash of ANYTHING. such as the code, vcpkg, cahed components ? Let me save you some time , WE DO NOT. 
We choose to trust that github, and vcpkg are valid sources, that other tools we use are not malicious. 
Is this good or bad ? That's another discussion that if you want you can bring up with EngSys.  

Another point to bring up is you assume stress runs against a specific build that i can extract the binary from. That assumption is incorrect, it runs on it's own schedule, it runs on a weekly schedule. Thus i need to grab and build the code as needed, not based on some glorious build.  

I understand you are busy and don't have time to look and google all the commands, that's fine, i don't expect everyone to know everything, but assuming that i'm an imbecile who has no idea what they're doing i do find insulting.

sdk/core/azure-core/test/libcurl-stress-test/Dockerfile
```diff
@@ -3,11 +3,21 @@
 
 FROM mcr.microsoft.com/mirror/docker/library/ubuntu:22.04
 
-# copy the tagrget binary
-COPY ./bin/libcurl-stress-test/azure-core-libcurl-stress-test ./azure-core-libcurl-stress-test
-RUN chmod +x ./azure-core-libcurl-stress-test
-
+RUN su
```
benbp, line 6: Containers run as root by default so this shouldn't be necessary.

gearama: done

gearama:
@benbp done

sdk/core/azure-core/test/libcurl-stress-test/Dockerfile
```diff
@@ -3,11 +3,20 @@
...
+RUN mkdir /opt/vcpkg
+RUN tar xf vcpkg.tar.gz --strip-components=1 -C /opt/vcpkg
+RUN /opt/vcpkg/bootstrap-vcpkg.sh
+RUN ln -s /opt/vcpkg/vcpkg /usr/local/bin/vcpkg
 
+RUN git clone https://github.com/Azure/azure-sdk-for-cpp.git 
+WORKDIR /azure-sdk-for-cpp/build
+RUN cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DBUILD_TRANSPORT_CURL=ON ..
+RUN cmake --build . --target azure-core-libcurl-stress-test 
+
```
richardpark-msft, line 19: I think we can split this Dockerfile here and have the first part be a build container and the second part be the actual final container. This lets you create a runtime image that doesn't have all the dev tools and source code.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azservicebus/internal/stress/Dockerfile

This is the start of the build container:
https://github.com/Azure/azure-sdk-for-go/blob/552351111cd8a8ec9a7619e0eacf644e1dbe5905/sdk/messaging/azservicebus/internal/stress/Dockerfile#L1

# this is the runtime container
https://github.com/Azure/azure-sdk-for-go/blob/552351111cd8a8ec9a7619e0eacf644e1dbe5905/sdk/messaging/azservicebus/internal/stress/Dockerfile#L15

benbp: @gearama here is a working dockerfile that will build from whatever version the checked out repo is on (working both locally and in CI). However it takes ~5 mins to build on my CPU hungry machine, is this expected?

NOTE: in order for this to work, the `imageBuildDir` field in the scenarios-matrix must be updated:

```
imageBuildDir: "../../../../../"
```

Dockerfile:

```
# Copyright (c) Microsoft Corporation. All rights reserved.

FROM mcr.microsoft.com/mirror/docker/library/ubuntu:22.04 as build
# FROM mcr.microsoft.com/mirror/docker/library/ubuntu:22.04

# install the mem check tool along side the other deps
RUN apt-get update -y
RUN apt-get install -y gcc cmake make g++ git zip unzip build-essential pkg-config wget curl valgrind
RUN wget -O vcpkg.tar.gz https://github.com/microsoft/vcpkg/archive/master.tar.gz
RUN mkdir /opt/vcpkg
RUN tar xf vcpkg.tar.gz --strip-components=1 -C /opt/vcpkg
RUN /opt/vcpkg/bootstrap-vcpkg.sh
RUN ln -s /opt/vcpkg/vcpkg /usr/local/bin/vcpkg

ADD . /src
WORKDIR /build
RUN cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DBUILD_TRANSPORT_CURL=ON /src
RUN cmake --build . --target azure-core-libcurl-stress-test

FROM mcr.microsoft.com/mirror/docker/library/ubuntu:22.04

RUN apt-get update -y
RUN apt-get install -y valgrind
WORKDIR /

# copy the target binary
COPY --from=build ./build/sdk/core/azure-core/test/libcurl-stress-test/azure-core-libcurl-stress-test ./azure-core-libcurl-stress-test
RUN chmod +x ./azure-core-libcurl-stress-test

CMD valgrind --tool=memcheck -s ./azure-core-libcurl-stress-test
```

gearama: i trust your code is correct , i got it things working local , only that when trying to use deploy.ps1 to deploy in the test cluster i keep getting event for the pod : 


 Warning   FailedScheduling   pod/cpp-constantdetach-libcurl-stress-test-9-v4sb6   0/8 nodes are available: 2 node(s) had untolerated taint {CriticalAddonsOnly: true}, 3 node(s) didn't match Pod's node affinity/selector, 6 Insufficient cpu. preemption: 0/8 nodes are available: 3 Preemption is not helpful for scheduling, 5 No preemption victims found for incoming pod.

benbp: @gearama you might have been hitting noisy neighbors on the cluster with node usage. Looks like the cluster scaled up node count in response and was able to schedule your test. I bumped up the node count baseline as well.

sdk/core/azure-core/test/libcurl-stress-test/Dockerfile
```diff
@@ -1,22 +1,33 @@
...
-RUN cp /azure-sdk-for-cpp/build/sdk/core/azure-core/test/libcurl-stress-test/azure-core-libcurl-stress-test ./azure-core-libcurl-stress-test
+RUN apt-get update -y
+RUN apt-get install -y valgrind
+WORKDIR /
+
+# copy the target binary
+COPY --from=build ./build/sdk/core/azure-core/test/libcurl-stress-test/azure-core-libcurl-stress-test ./azure-core-libcurl-stress-test
 RUN chmod +x ./azure-core-libcurl-stress-test
+
+#CMD valgrind --tool=memcheck -s ./azure-core-libcurl-stress-test
```
benbp, line 32: I added a `CMD` directive just so it would be easy to check if the dockerfile/binary built was working when running `docker run` locally. Even if it's not included it's easy enough to just run `docker run -it <container name> bash` and test out the command from the shell.

```suggestion
CMD ./azure-core-libcurl-stress-test
```

benbp (approved):

gearama:
since @LarryOsterman did not review for more than a week, we need this checked in, i will have to dismiss his review in order t unblock this PR

PR #4352, opened 2023-02-13
Fix compilation on Visual Studio 2017

jorgen:
This is needed to be able to compile on Visual Studio 2017.

### DIFF
--- a/sdk/core/azure-core/src/http/request.cpp
+++ b/sdk/core/azure-core/src/http/request.cpp
@@ -9,6 +9,7 @@
 #include <string>
 #include <vector>
 
+using namespace Azure::Core;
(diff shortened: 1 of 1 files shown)


antkmsft (approved):

RickWinter (approved):

antkmsft:
Thank you @jorgen! We are happy to take your fix.
Note that we do not have CI for VS 2017, and at this point we support VS 2019 and 2022. I.e. we can't guarantee you that we won't break VS 2017 build in future commits in some other place, as we don't have tools to verify that and we don't have plans to add them. But we will likely accept fixes like this in the future as well (I hope that it won't be needed often).

PR #4361, opened 2023-02-15
Added support to ignore invalid cert common name

Jinming-Hu:
We need this feature when using storage library in Azure Storage server-side codebase. TLS certificate used for testing is self-signed. Service needs to be accessed via multiple hostnames including localhost, so we need to be able to ignore CN mismatch errors.

# Pull Request Checklist

- [x] Doxygen docs
- [x] Unit tests
  - [x] PR is single purpose
  - [x] Related issue listed
- [x] Comments in source
- [x] No typos
- [x] Update changelog
- [x] Not work-in-progress
- [x] Self review of PR done
- [x] Any breaking changes?

### DIFF
--- a/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
+++ b/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
@@ -174,6 +174,14 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
     bool EnableCertificateRevocationListCheck{false};
 
     /**
+     * @brief Verify peer's SSL certificate.
+     *
+     * @remark This field is only used if the customer has not specified a default transport
+     * adapter. If the customer has set a Transport adapter, this option is ignored.
+     */
+    bool SslVerifyPeer{true};
+
+    /**
      * @brief Base64 encoded DER representation of an X.509 certificate expected in the certificate
      * chain used in TLS connections.
      *
--- a/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
+++ b/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
@@ -174,12 +174,17 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
     bool EnableCertificateRevocationListCheck{false};
 
     /**
-     * @brief Verify peer's SSL certificate.
+     * @brief Disable SSL/TLS certificate verification. This option allows transport layer to
+     * perform insecure SSL/TLS connections and skip SSL/TLS certificate checks while still having
+     * SSL/TLS-encrypted communications.
+     *
+     * @remark Disabling TLS security is generally a bad idea because it allows malicious actors to
+     * spoof the target server and should never be enabled in production code.
      *
      * @remark This field is only used if the customer has not specified a default transport
      * adapter. If the customer has set a Transport adapter, this option is ignored.
      */
-    bool SslVerifyPeer{true};
+    bool DisableTlsCertificateValidation{false};
 
     /**
      * @brief Base64 encoded DER representation of an X.509 certificate expected in the certificate
--- a/sdk/core/azure-core/src/http/curl/curl.cpp
+++ b/sdk/core/azure-core/src/http/curl/curl.cpp
@@ -311,7 +311,7 @@ Azure::Core::Http::CurlTransportOptions CurlTransportOptionsFromTransportOptions
     curlOptions.SslOptions.PemEncodedExpectedRootCertificates
         = PemEncodeFromBase64(transportOptions.ExpectedTlsRootCertificate, "CERTIFICATE");
   }
-  curlOptions.SslVerifyPeer = transportOptions.SslVerifyPeer;
+  curlOptions.SslVerifyPeer = !transportOptions.DisableTlsCertificateValidation;
   return curlOptions;
 }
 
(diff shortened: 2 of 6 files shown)


LarryOsterman:
QQ: Where is the libcurl version of this fix? Or does Libcurl already ignore invalid CN names?

Jinming-Hu:
> QQ: Where is the libcurl version of this fix? Or does Libcurl already ignore invalid CN names?

It turns out libcurl will ignore all these kinds of errors when [verify peer](https://github.com/Azure/azure-sdk-for-cpp/blob/afaefd3d05d5a43f20cd1894a52f6a32f2bce5a0/sdk/core/azure-core/inc/azure/core/http/curl_transport.hpp#L143) is disabled.

WinHTTP, however, has more granular control. https://learn.microsoft.com/en-us/windows/win32/winhttp/option-flags#winhttp_option_security_flags

LarryOsterman:
> 



> > QQ: Where is the libcurl version of this fix? Or does Libcurl already ignore invalid CN names?
> 
> It turns out libcurl will ignore all these kinds of errors when [verify peer](https://github.com/Azure/azure-sdk-for-cpp/blob/afaefd3d05d5a43f20cd1894a52f6a32f2bce5a0/sdk/core/azure-core/inc/azure/core/http/curl_transport.hpp#L143) is disabled.
> 
> WinHTTP, however, has more granular control. https://learn.microsoft.com/en-us/windows/win32/winhttp/option-flags#winhttp_option_security_flags

There are global options now for disabling peer validation, so if you're ok with disabling peer validation on curl, why not also disable it globally?

In general, I'm *really* not a fan of per-transport options, because enabling them forces the customer to manually create their own transport object for the pipeline (which then forces them to make curl vs winhttp decisions, etc). So I would much prefer to see new transport features be available for all of them.

Jinming-Hu:
> There are global options now for disabling peer validation,

Are you referring to [TransportOptions](https://github.com/Azure/azure-sdk-for-cpp/blob/30afd1eb89f586a69f448495965311278927de55/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp#L137)? There's no option to disable peer validation in there. In other words, if customer wants to disable peer validation, he has to specify a transport (curl or winhttp). 

That's what I always think. Am I wrong?

LarryOsterman:
> > There are global options now for disabling peer validation,
> 
> Are you referring to [TransportOptions](https://github.com/Azure/azure-sdk-for-cpp/blob/30afd1eb89f586a69f448495965311278927de55/sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp#L137)? There's no option to disable peer validation in there. In other words, if customer wants to disable peer validation, he has to specify a transport (curl or winhttp).
> 
> That's what I always think. Am I wrong?

@#$  - for some reason, I thought I'd added them with the websocket work.

Jinming-Hu:
> IMHO, this would be better set as a global option to disable peer validation that would then be modified to the appropriate HTTP transport (so for WInHTTP, it would set the existing IgnoreUnknownCertificateAuthority and the new IgnoreInvalidCertificateCommonName field, for curl it would set SslVerifyPeer to false.

Done!

Jinming-Hu:
@LarryOsterman Can you take a look at the updated changes, to ensure you're happy with the it, especially the variable name?

sdk/core/azure-core/inc/azure/core/http/policies/policy.hpp
```diff
@@ -173,6 +173,14 @@ namespace Azure { namespace Core { namespace Http { namespace Policies {
      */
     bool EnableCertificateRevocationListCheck{false};
 
+    /**
+     * @brief Verify peer's SSL certificate.
+     *
+     * @remark This field is only used if the customer has not specified a default transport
+     * adapter. If the customer has set a Transport adapter, this option is ignored.
+     */
+    bool SslVerifyPeer{true};
```
LarryOsterman, line 182: In general, all of our options "enable" a behavior. So we have "EnableSystemDefaultProxy", "EnableCertificateRevocationListCheck", etc. This option is backwards because the default is true.

In general, I prefer options to be defaulted to false and when set to "true" it defines a new behavior.

So instead of `SslVerifyPeer`, I think something like `DisableTlsPeerValidation` or `DisableTlsCertificateValidation` is probably a better name for this option (the encryption technology previously (in 2000 or so) called SSL (Secure Sockets Layer) is now called "TLS" (Transport Layer Security)).

Also if you could add a comment indicating that disabling TLS security is generally a bad idea because it allows malicious actors to spoof the target server and should never be enabled in production code.

LarryOsterman (changes requested):
My only concern is that the semantics of `SslVerifyPeer` is backwards from the other options (and yes, I know that the CURL option is `SslVerifyPeer` - that makes sense in the CURL context since this is exactly the CURL option name. But this setting applies to all transports.

Jinming-Hu:
@LarryOsterman Can you review this PR again?

LarryOsterman (approved):
Looks great to me. Thank you for your patience.

sdk/core/azure-core/src/http/curl/curl.cpp
```diff
@@ -311,6 +311,7 @@ Azure::Core::Http::CurlTransportOptions CurlTransportOptionsFromTransportOptions
     curlOptions.SslOptions.PemEncodedExpectedRootCertificates
         = PemEncodeFromBase64(transportOptions.ExpectedTlsRootCertificate, "CERTIFICATE");
   }
+  curlOptions.SslVerifyPeer = !transportOptions.DisableTlsCertificateValidation;
```
ahsonkhan, line 314: Just making sure (because it seems to be ok), but @LarryOsterman would we need to update the connection key logic that we use for the connection pooling/caching? It is meant to be unique for each set of options (based on `SslVerifyPeer` being set).
https://github.com/Azure/azure-sdk-for-cpp/blob/e2a1b99a0625d6309e22474332fb3c342b4a2dc6/sdk/core/azure-core/src/http/curl/curl.cpp#L1293

LarryOsterman: Good question: Based on my reading, the SslVerifyPeer option is already included in the connection key logic - see line 1284 of `GetConnectionKey`.

